PR #1, opened 2020-09-08
Initial bundle

nmadruga:
# Documentation:
- Basic structure
- Webpack setup
- Apollo client setup

## Tips to review:
Remove from the Review files with extensions (just imported no code changes):
`.eot`
`.less`
`.overrides`
`.variables`
`.ttf`
`.woff`
`.woff2`

### App basic setup using:
* Typescript
* React
* React hooks
* Apollo client 3
* Webpack 4
* Semantic UI React
* Eslint


src/graphql/mutations/updateApplicationById.mutation.ts
```diff
@@ -0,0 +1,12 @@
+import { gql } from '@apollo/client'
+
+export default gql`
+    mutation updateApplicationById($id: Int!, $applicationPatch: ApplicationPatch!) {
```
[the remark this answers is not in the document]
nmadruga: Done. Thanks for the reminder 👍

tsconfig.json
```diff
@@ -0,0 +1,17 @@
+{
+    "compilerOptions": {
+      "jsx": "react",
+      "module": "commonjs",
+      "noImplicitAny": true,
```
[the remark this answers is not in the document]
nmadruga: Done! I also copied the description of each one in use from the `tsconfig.json` file in the Back-end.

CarlosNZ (approved):

webpack.config.js
```diff
@@ -0,0 +1,37 @@
...
+
+  return addons.map((addonName) => require(`./build-utils/addons/webpack.${addonName}.js`))
+}
+
+// 'env' will contain the environment variable from 'scripts'
+// section in 'package.json'.
+// console.log(env); => { env: 'dev' }
+module.exports = (env) => {
+  // We use 'buildValidations' to check for the 'env' flag
+  if (!env) {
```
andreievg, line 21: this is not really checking that env.dev exists and if appropriate webpack file exists, maybe just:

`module.exports = ({ env = 'dev' }) => {`
and
` const envConfig = require(`./build-utils/webpack.${env}.js`)`
that way don't really need the error check

build-utils/webpack.common.js
```diff
@@ -0,0 +1,66 @@
...
+        use: 'url-loader',
+      },
+      {
+        // Load other files, images etc
+        test: /\.(png|j?g|gif|ico)?$/,
+        use: 'url-loader',
+      }
+    ]
+  },
+  optimization: {
```
andreievg, line 40: Was wondering what this would do ?

build-utils/webpack.prod.js
```diff
@@ -0,0 +1,38 @@
...
+  },
+  devtool: 'source-map',
+  module: {
+    rules: [
+      {
+        test: /\.less$/,
+          use: [
+            {
+              // We configure 'MiniCssExtractPlugin'              
+              loader: MiniCssExtractPlugin.loader,
```
andreievg, line 18: if we add this a seperate one here would it work (and move the other style loaders to common:
i.e.

does this work ?

```JSON
 "rules": [
  ...
    {
    "test": /\.css$/,
    "use": [MiniCssExtractPlugin.loader, 'css-loader']
    }
   {
    "test": /\.less$/,
    "use": [
     {
      "options": {
       "importLoaders": 2
      }
     },
     "postcss-loader",
     "less-loader"
    ]
   }

```

Also was wondering why we need postcss ?

webpack.config.js
andreievg: also forgot to mention, personal preference for me, is to use full word 'development', 'production'

nmadruga: Yes, good point about the possibility of not `env` passed. 
And for me the naming in this case is not a big deal. Most times I agree to never user shortage version of names. But in this case is like using `i` for iteration. I'm changing to `development` and `production` anyway.

build-utils/webpack.common.js
nmadruga: This is called [Code splitting](https://webpack.js.org/guides/code-splitting/) and it's used to separate the `semantic-ui-react` vendor in another chunk. Not sure exactly what the one for styles is doing. It would be something similar for css files using the `semantic-ui-less` in a separated chunk. I got this instructions from this [tutorial](https://medium.com/free-code-camp/learn-webpack-for-react-a36d4cac5060#4e46). 
The results when I run `yarn build` with using the code splitting:
Screen Shot 2020-09-17 at 10 23 05 AM
And without using the code splitting:
Screen Shot 2020-09-17 at 10 30 15 AM

build-utils/webpack.prod.js
nmadruga: I was able to build like that:
```
rules: [
      {
        test: /\.css$/,
        use: [MiniCssExtractPlugin.loader, 'css-loader']
      },
      {
        test: /\.less$/,
        use: [
          {
            loader: 'css-loader',
            options: {
              importLoaders: 2,
            }
          },
          'postcss-loader',
          'less-loader'
        ]
      }
```
But now the bundled `app.js` is huge and no `.css` was created:
Screen Shot 2020-09-17 at 10 38 48 AM

I'll make a quick research to answer this question:
> Also was wondering why we need postcss ?

nmadruga: Ok. I have removed the `postcss` from the Webpack. It seems useless for now. We can change this if we need later on. This is the bundle result that I'm getting after changing the code to this:
```
{
        test: /\.less$/,
        use: [
          {
            loader: MiniCssExtractPlugin.loader
          },
          'css-loader',
          'less-loader'
        ]
      }
```
Screen Shot 2020-09-17 at 11 20 11 AM

webpack.config.js
nmadruga: Oh, I reverted the change on the file names because in VS Code it stopped showing as `webpack` types if the different name, so I think this is what is expected to be used for naming these files.

PR #10, opened 2020-09-17
#9 context user area states

### LINKED ISSUE #9: Example of User area with useContext and useReducer hooks

nmadruga:
Fixes #9.
Branched from PR #1. Needs to wait until this one is finished to Review/Merge.

Basically setup the `UserState` using the useContext and useReducer hooks in a structured way.
Adds folder: 
* src/containers
 
Moves the `App.tsx` file to containers folder and new folder `User`.

Inside `src/containers/User` there are:
* UserArea - container that would have nested components (accessing the same context)
* UserState - where the user states are stored and the dispatch function to be used by components

In the `App.tsx` the `UserArea` is wraped by the `UserContextProvider` to make the magic.


src/containers/User/UserArea.tsx
```diff
@@ -0,0 +1,36 @@
+import React from 'react'
+import { Button, Container, Label, Segment } from 'semantic-ui-react';
+import { useUserState } from './UserState' 
+
+const UserArea: React.FC = () => {
+    const {
```
CarlosNZ, line 6: I notice the indentation is 4 spaces instead of the 2 specified in `.prettierrc`. So when I save this, it changes back to the more narrow indentation. Perhaps you haven't got your editor set to auto-format on Save? Or something else? 🤷‍♂️

CarlosNZ (changes requested):
Thanks, Nicole, this is a really good demonstration of the state management/context.

I'm keen to see how this plays with database requests. I see you've already made an issue to make a mutation query. I'd probably extend that to include a fetch query as well (since the `UseQuery` used here is separate from the `UserArea` component.

I'd suggest we test that out and merge that PR into here before merging this one to the `master`. Does that seem reasonable? Just that, if there's issues with it working nicely, then this might not be the best way to go.

Can I make a quick suggestion as to what to implement on top of what you've done here?
- Since there's 4 users in the Database already, maybe make the app fetch the names of the users, and make those names the buttons that appear in the User Area. 
- Add a input field and when it is submitted, it updates the name of the current user to the new value (and writes it to the database).

Up to you if you've got a better idea to demonstrate similar concepts.

If that all works nicely, then bingo, I think this will be awesome. :) 

Also, just check the formatting with Prettier, as I think it should be 2 spaces.

src/containers/User/UserArea.tsx
nmadruga: Yes. I think mine is still not configured to use. I'll do this now.

nmadruga:
> I see you've already made an issue to make a mutation query.

Yes, I'm working on this now. 

> I'd probably extend that to include a fetch query as well

No problem

> I'd suggest we test that out and merge that PR into here before merging this one to the `master`.

Yes. I can do that. Although I have also merged in the PR #11 so I can use some features. But I think that's alright since the code for routing should be merge to the master today.

> Since there's 4 users in the Database already, maybe make the app fetch the names of the users, and make those names the buttons that appear in the User Area.

Great suggestion. I'll do that.

> Add a input field and when it is submitted, it updates the name of the current user to the new value (and writes it to the database).

Hmm I think I prefer doing the following: 

On the (existing) register page I'll add a new user - pretending we don't need to fill a registration for that. And when this new user is added the buttons should display the new one.

nmadruga:
Closing to add as part of #13 PR

PR #11, opened 2020-09-18
#2 Basic Routing and query parameter handling

### LINKED ISSUE #2: Create all basic routes and management for filter in routes

CarlosNZ:
Fixes #2.

Have implemented basic routes in `App.tsx` and made "dummy" components to resolve all the endpoints. Also added some functionality to `ApplicationList` (and others) to demonstrate passing query string parameters to the component.

Just for explanation, there are two types of URL parameters that can be received by a component:

1. `URL parameters` are the "variable" parts of the main URL, defined with a colon `:` in the ReactRouter route path, e.g.  
`/application/:appId`  
These are accessed by the called component with ReactRouter's built-in `useParams` hook. (Can also be acquired from default props using the `props.match` object, but `useParams` seems cleaner.)
2. `Query parameters` are the key-value pairs after the `?` in a URL, as in:
`/applications?status=draft&stage=assessment`  
These are exposed by the `useLocation` hook, in the `.search` field, but require parsing. The browser API `URLSearchParams` extracts the key-value pairs, but I've written an additional function called `useQueryParameters` to capture these in a simple object with `{ key: value }` 

Linking (i.e. updating the URL and updating the above parameter values) can be done in (at least) two ways:

1. The ReactRouter `<Link to="/applications">` component for simple HTML links.
2. The `useHistory` hook -- it's basically a stack of previous URLs, with the last element being the current. So pushing a URL onto this list updates the current URL. Use for buttons or other event handlers that need to update the URL.

### [applications](http://localhost:3000/applications)
- Application list. Have added a couple of links to the page so you can how the URL can be updated and the parameters captured for filtering or display. (e.g. [/applications?key=value&anotherkey=anothervalue](http://localhost:3000/applications?key=value&anotherkey=anothervalue))

### [/applications/new](http://localhost:3000/applications/new)

New Application start page.

### [/applications/new?type=drug-registration](http://localhost:3000/applications/new?type=drug-registration)

Note the page is aware of the type that is specified, which can be used for default selection in drop-down menu.

### [/applications/1234](http://localhost:3000/applications/1234)

An application home page.

### [/applications/1234/SectionA/page1](http://localhost:3000/applications/1234/SectionA/page1)

Specific Section/Page of an application. Note the `Next page` button to increment the pages.

### [/applications/1234/summary](http://localhost:3000/applications/1234/summary)

Summary page for an application

### [/applications/1234?mode=assignment](http://localhost:3000/applications/1234?mode=assignment)

"Assignment" mode for Application. The `?mode=assignment` query can be appended to any of the (above) application pages. (See also [`?mode=consolidation`](http://localhost:3000/applications/9876?mode=consolidation))

### [/applications/1234/approval](http://localhost:3000/applications/1234/approval)

Application approval page.

### Admin pages (No dynamic parameters)

- [/admin](http://localhost:3000/admin)
- [/admin/users](http://localhost:3000/admin/users)
- [/admin/permissions](http://localhost:3000/admin/permissions)
- [/admin/config](http://localhost:3000/admin/config)

### Template Builder

- [/admin/templates](http://localhost:3000/admin/templates) (Template list)
- [/admin/templates/new](http://localhost:3000/admin/templates/new) (New Template)
- [/admin/templates/abc789/triggers-and-actions](http://localhost:3000/admin/templates/abc789/triggers-and-actions) (The Triggers/Actions step of a template -- change ID and Step to whatever)

### [/account](http://localhost:3000/account)

User account page (no dynamic parameters)

### Organisation Management

- [/organisations/CompanyA](http://localhost:3000/organisations/CompanyA) (Org home page)
- [/organisations/CompanyA/members](http://localhost:3000/organisations/CompanyA/members) (Edit members of Org)

### Notification Centre

- [/notifications](http://localhost:3000/notifications) (Notification Inbox -- some clickable links provided to demonstrate changing filter parameters, and a single notification link)
- [/notifications/567](http://localhost:3000/notifications/567) (Single notification message)

### Products

- [/products](http://localhost:3000/products) (Product list -- some clickable links provided to demonstrate changing filter parameters, and a single product link)
- [/products/abc123](http://localhost:3000/products/abc123) (Single product page)


documentation/internal/Overview-of-Navigation-&-URL-structure.md
```diff
@@ -85,9 +83,9 @@ Example: [https://balsamiq.cloud/scs7giw/ponj59g/r1D2B](https://balsamiq.cloud/s
 
 ---
 
-## `/applications/[type]/[guid]/[section]/[page]`
+## `/application/[appId]/[sectionName]/p[page]`
```
nmadruga, line 86:
```suggestion
## `/application/[appId]/[sectionName]/[page]`
```

documentation/internal/Overview-of-Navigation-&-URL-structure.md
```diff
@@ -65,15 +63,15 @@ Example: [https://balsamiq.cloud/scs7giw/ponj59g/r4C62](https://balsamiq.cloud/s
 
 ---
 
-## `/applications/new`
+## `/application/new`
```
nmadruga, line 66: I'm confused, in the PR description it says `applications/new`. Anyway when I try `application/new` the page is not found. :(

nmadruga: Oh. My bad. I hasn't changed the branch correctly.

documentation/internal/Overview-of-Navigation-&-URL-structure.md
nmadruga: Oh, now I see this was intended. Do you think it's better than just the page number? I think I prefer just the number.

src/components/App.tsx
```diff
@@ -1,14 +1,34 @@
...
+import Template from './Template'
+import Account from './Account'
+import { Organisation, OrgMemberEdit } from './Organisation'
+import { AdminUsers, AdminPermissions, Config } from './AdminOther'
+import { NotificationsList, Notification } from './Notification'
 import NoMatch from './NoMatch'
+import { ProductList, Product } from './Product'
+
+// queryParams is an object that gets the URL query params as key-value pairs
+// This object should be used for filtering the getApplication query
```
nmadruga, line 25: I think it's good to have a header for this function, but can this be improved to use a pattern of function header? I'll find one tomorrow if you haven't already. But just to make it easier to find what if you are saying is regarding the result, parameters of the function logic. Also could you change to use ES6 arrow functions instead?

src/components/App.tsx
```diff
@@ -27,9 +47,72 @@ const App: React.FC = () => {
           <Grid.Column stretched width={12}>
             <Segment>
               <Switch>
-                <Route path="/" exact component={Home} />
-                <Route path="/form" component={Register} />
-                <Route path="/applications" component={ApplicationsList} />
+                <Route exact path="/">
+                  <Home />
```
nmadruga, line 51: Just by curiosity: Why not use the `component={Home}` pattern?

src/components/App.tsx
```diff
@@ -27,9 +47,72 @@ const App: React.FC = () => {
...
+                <Route exact path="/admin/config">
+                  <Config />
+                </Route>
+                <Route exact path="/account">
+                  <Account />
+                </Route>
+                <Route exact path="/organisation/:orgName">
+                  <Organisation />
+                </Route>
+                <Route exact path="/organisation/:orgName/member-edit">
```
nmadruga, line 101: I think this path is odd. I would expect something as `/organisation/:orgName/members` and the action to edit member would be just a option in there. Maybe something that the user could edit directly in there and then save it. Not sure if 1. we need a separated page for this and 2. If we would allow adding/removing then this path would be with a different meaning, or we would have one per action. For now, I think it could just be called members.

src/components/Application.tsx
```diff
@@ -0,0 +1,57 @@
+import React from 'react'
+import { useQueryParameters } from './App'
+import { useHistory, useLocation, useParams } from 'react-router-dom'
+
+type TParams = { appId: string; sectionName?: string; page?: string }
```
nmadruga, line 5: Nice! Just curios what is the different between using `?` and not using it?

src/components/Application.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+export interface AppProps {
+  summary?: boolean
+}
+
+const Application: React.FC<AppProps> = (props) => {
+  const { appId, sectionName, page }: TParams = useParams()
+
+  console.log(useLocation().pathname)
```
nmadruga, line 14: console here. Please remove.

src/components/Application.tsx
```diff
@@ -0,0 +1,57 @@
...
+import { useHistory, useLocation, useParams } from 'react-router-dom'
+
+type TParams = { appId: string; sectionName?: string; page?: string }
+
+export interface AppProps {
+  summary?: boolean
+}
+
+const Application: React.FC<AppProps> = (props) => {
+  const { appId, sectionName, page }: TParams = useParams()
```
nmadruga, line 12: This is great!!! I'm so happy about this solution using hooks : D

src/components/Application.tsx
```diff
@@ -0,0 +1,57 @@
...
+export default Application
+
+type ButtonProps = { sectionName?: string; page?: string }
+
+const NextPageButton: React.FC<ButtonProps> = (props) => {
+  const history = useHistory()
+  const handleClick = () => {
+    history.push('p' + (Number(props.page) + 1))
+  }
+  console.log(props.sectionName)
```
nmadruga, line 48: Another console here

src/components/Notification.tsx
```diff
@@ -0,0 +1,56 @@
...
+  return (
+    <div>
+      <h1>Title of Message</h1>
+      <p>Dear user, this message has the id: {notificationId}.</p>
+      <p>
+        <Link to="./">Back to Notification Center</Link>
+      </p>
+    </div>
+  )
+}
```
nmadruga, line 56: I know this is just an example, but we should really break down into separated files. For the future. Do you agree?

documentation/internal/Overview-of-Navigation-&-URL-structure.md
```diff
@@ -359,10 +357,16 @@ Example: [https://balsamiq.cloud/scs7giw/ponj59g/r6C38](https://balsamiq.cloud/s
 
 ---
 
-## `/products/[product-id]`
+## `/products/[productId]`
```
nmadruga, line 360:
```suggestion
## `/product/[productId]`
```

To keep consistent with `application`  and `template`

nmadruga (commented):
Cool! This is looking great. Very well done! I really don't have anything to add.

Have just a few comments there, see what you think. But overall is looking great. I really liked.

## Tests:
- [x] applications 
- [x] /application/new
- [x] /application/new?type=drug-registration
- [x] /application/1234
- [x] /application/1234/SectionA/p1
- [x] /application/1234/summary
- [x] /application/1234?mode=assignment
- [x] /application/1234/approval
### Admin pages (No dynamic parameters)
- [x] /admin
- [x] /admin/users
- [x] /admin/permissions
- [x] /admin/config
### Template Builder
- [x] /admin/templates (Template list)
- [x] /admin/template/new (New Template)
- [x] /admin/template/abc789/triggers-and-actions (The Triggers/Actions step of a template -- change ID and Step to whatever)
- [x] /account
### Organisation Management
- [x] /organisation/CompanyA (Org home page)
- [x] /organisation/CompanyA/member-edit (Edit members of Org)
### Notification Centre
- [x] /notifications (Notification Inbox -- some clickable links provided to demonstrate changing filter parameters, and a single notification link)
- [x] /notifications/567 (Single notification message)

Oh! you forgot to mention the product section:
- [x] /products
- [x] /products/XYZ276

CarlosNZ:
> Oh! you forgot to mention the product section:
> 
Whereabouts did I forget to mention it? It's in the docs and the notes for this PR. 🤷‍♂️

documentation/internal/Overview-of-Navigation-&-URL-structure.md
CarlosNZ: I thought it was bit nicer since the `[sectionName]` shows a meaningful name, so it would make it quite clear to the user that what section and page we're on: `/applications/1234/product-details/p2`

But I'm not too fussed, it's only a minor aesthetic thing. Shall we see if Andrei has a preference?

src/components/App.tsx
CarlosNZ: Good question! I noticed the [ReactRouter docs](https://reactrouter.com/web/example/basic) used this style throughout, so I asked on SSNZDevs what the difference was (if any). There isn't really any, but both Josh and Will said this way is the preferred way these days, as it's more "declaritive" which is in keeping with general React style. 

Also, the other way gets really ugly once you start having to pass props to the components. ;)

src/components/App.tsx
CarlosNZ: Yup, good call. Have changed it to just `/members` for now.

src/components/Application.tsx
CarlosNZ: The `?` means its an optional field. I thought this was necessary since `sectionName` and `page` aren't always available from the URL (e.g. on the application home page). 

I see now that it doesn't actually matter in this case, since the values are only ever read from `useParams` and not explicitly defined in the code. But without the `?`, if you tried to put `const myParams = {appId: "1234"}` Typescript complains about missing fields.

src/components/Application.tsx
CarlosNZ: 😊

documentation/internal/Overview-of-Navigation-&-URL-structure.md
CarlosNZ: Fair call -- have also done the same to `notifications`.

src/components/App.tsx
CarlosNZ: Changed to arrow function. Not sure quite what you mean by "function header"? Can you give me an example?

src/components/Notification.tsx
CarlosNZ: Not sure. I only put these into one file because I was getting tired of making so many small files and having a massive `import` list. I agree that it's good to keep things well contained, but I think when you have a few small(ish) components that all work together, then it's fine to keep them in one file. Especially if they share common functions or types, then it saves having to keep importing things all over the place. 

Seems to me the more isolated you make each component, the more messy and convoluted it becomes to keep track of all the imports.

Happy to go with whatever we decide though.

documentation/internal/Overview-of-Navigation-&-URL-structure.md
nmadruga: I think instead of `p1` we should use either just the number of something more declarative like: `Page-1`

src/components/App.tsx
nmadruga: The comments before your function. isn't that called a function header? Example of what I would prefer to use:
```
/**
 * Represents a book.
 * @param {string} title - The title of the book.
 * @param {string} author - The author of the book.
 * @returns { (promt:string) => void } - the returned function
 */
```
It seems this is using the JsDocs pattern? I'm not sure what that is, but I thought it would be good to have a pattern for commenting functions. What do you think?

src/components/Application.tsx
nmadruga: Interesting! Thanks.

src/components/App.tsx
CarlosNZ: Okay, have added something now. Not 100% sure I expressed the return value correctly. Let me know what you think.

nmadruga:
@CarlosNZ I'll have a look again on this one later on today. And will reply to your messages.
In the meanwhile could you do the update for the path on `application/id` as discussed in the meeting please?

CarlosNZ:
> In the meanwhile could you do the update for the path on `application/id` as discussed in the meeting please?

Done! 446347a

- Changed `p<page>` to `page<page>`
- All URL sections kept in plural form (ie. `/application/1234` is now `/applications/1234`

Also updated docs and links in original post

src/components/Notification.tsx
nmadruga: I don't think this should be defined (or changed in this PR) what I was suggesting is that we create the habit of creating new files for every new component. Since they will grow in size most definitely in the future. What is shared between then should be passed as props or added to the context area, this way there wouldn't be much else to share. For now I understand that was just not worth, but let's work with that in mind for the next issues?

documentation/internal/Overview-of-Navigation-&-URL-structure.md
```diff
@@ -313,7 +311,7 @@ Example: [https://balsamiq.cloud/scs7giw/ponj59g/r650F](https://balsamiq.cloud/s
 
 ---
 
-## `/organisation/[name-of-org]/member-edit`
+## `/organisation/[name-of-org]/members`
```
nmadruga, line 314:
```suggestion
## `/organisations/[name-of-org]/members`
```

src/components/App.tsx
```diff
@@ -27,9 +52,72 @@ const App: React.FC = () => {
...
+                <Route exact path="/admin/permissions">
+                  <AdminPermissions />
+                </Route>
+                <Route exact path="/admin/config">
+                  <Config />
+                </Route>
+                <Route exact path="/account">
+                  <Account />
+                </Route>
+                <Route exact path="/organisation/:orgName">
```
nmadruga, line 103:
```suggestion
                <Route exact path="/organisations/:orgName">
```

src/components/App.tsx
```diff
@@ -27,9 +52,72 @@ const App: React.FC = () => {
...
+                  <Organisation />
+                </Route>
+                <Route exact path="/organisation/:orgName/members">
```
nmadruga, line 106:
```suggestion
                <Route exact path="/organisations/:orgName/members">
```

src/components/App.tsx
```diff
@@ -27,9 +52,72 @@ const App: React.FC = () => {
...
+                <Route exact path="/notifications/:notificationId">
+                  <Notification />
+                </Route>
+                <Route exact path="/products">
+                  <ProductList />
+                </Route>
+                <Route exact path="/products/:productId">
+                  <Product />
+                </Route>
                 <Route component={NoMatch} />
```
nmadruga, line 121:
```suggestion
                <Route>
                  <NoMatch />
                </Route>
```

Just to keep consistent with other routes

src/components/Organisation.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+  return (
+    <div>
+      <h1>Organisation Home page</h1>
+      <p>
+        Home page for organisation: <strong>{orgName}</strong>.
+      </p>
+      <p>List of Members, summaries of applications (past and present), etc.</p>
+      <p>
+        <Link to={'/organisation/' + orgName + '/members'}>Edit this org's members</Link>
```
nmadruga, line 17:
```suggestion
        <Link to={'/organisations/' + orgName + '/members'}>Edit this org's members</Link>
```

nmadruga (changes requested):
Pretty good. Just some tweaks to do on the documentation and I suggested using `organisations` to keep consistent. Although we don't list the organisations, we know that is part of a group. And maybe we could list the organisation that the user is part of in the `/organisations` page. Not for now.

## Tests 2:
- [x] applications 
- [x] /applications/new
- [x] /applications/new?type=drug-registration
- [x] /applications/1234
- [x] /applications/1234/SectionA/Page1
- [x] /applications/1234/summary
- [x] /applications/1234?mode=assignment
- [x] /applications/1234/approval
### Admin pages (No dynamic parameters)
- [x] /admin
- [x] /admin/users
- [x] /admin/permissions
- [x] /admin/config
### Template Builder
- [x] /admin/templates (Template list)
- [x] /admin/templates/new (New Template)
- [x] /admin/templates/abc789/triggers-and-actions (The Triggers/Actions step of a template -- change ID and Step to whatever)
- [x] /account
### Organisation Management
- [ ] /organisations/CompanyA (Org home page) **[FAIL]**
- [ ] /organisations/CompanyA/members (Edit members of Org) **[FAIL]**
### Notification Centre
- [x] /notifications (Notification Inbox -- some clickable links provided to demonstrate changing filter parameters, and a single notification link)
- [x] /notifications/567 (Single notification message)
### Products
- [x] /products
- [x] /products/XYZ276

CarlosNZ:
Changed `/organisation/...` to `/organisations/...` to keep consistent with others, and updated docs accordingly.

Think that addresses everything. :)

nmadruga (approved):

PR #13, opened 2020-09-25
#12 mutation example register

### LINKED ISSUE #9: Example of User area with useContext and useReducer hooks

### LINKED ISSUE #12: Example of User area change state after useMutation

nmadruga:
Fixes #9 & #12 

Add folders: 
* src/containers
* src/containers/User

* Moves the `App.tsx` file to containers folder and new folder `User`.

## (Issue 9) Changes: `useQuery`
* Basic setup the `UserState` using the useContext and useReducer hooks in a structured way.

* Add files in `src/containers/User`:
* **UserArea** - container that would have nested components (accessing the same context)
* **UserState** - where the user states are stored and the dispatch function to be used by components
* In the `App.tsx` the `UserArea` is wraped by the `UserProvider` to make the magic.

## (Issue 12) Changes: `useMutation`
* Add example of mutation on component `Register`.
* Add cache management to update the current local fetched list of user after the mutation runs.

## Usage & tests
* Run `yarn dev`
* See the list of existing users on the top (`User Area` component).
* Go to the Register page, add first name, last name, email and select a role.
* Click the submit button to see the submission message (snackbar).
* Next the new added user will be displayed on the top `User Area`.


src/containers/User/UserArea.tsx
```diff
@@ -0,0 +1,48 @@
...
+const UserArea: React.FC = () => {
+  const {
+    userState: { currentUser, users },
+    setUserState,
+  } = useUserState()
+  const { data, loading, error } = useGetUsersQuery()
+
+  useEffect(() => {
+    if (data) {
+      if (data.users && data.users.nodes) {
```
CarlosNZ, line 16: Does this need to be nested?
Why not just `if (data && data.users && data.users.nodes)`?

src/components/Register.tsx
```diff
@@ -32,32 +67,124 @@ const Register: React.FC = () => {
...
+      setTimeout(() => {
+        removeSnackbar()
+      }, 2000)
+    },
+  })
+
+  const updateUser = async () => {
+    try {
+      const { username, password, email, role } = newUser
+      const updatedUser = await createUserMutation({
```
CarlosNZ, line 113: `updatedUser` is not used anywhere. Do you need to declare it at all?

Why not just `await createUserMutation({...` ?

CarlosNZ (approved):
Nice work, Nicole. Just a couple of very minor nits I've noted.

I don't fully understand how this works, so I'll get you to talk me through some aspects of it next time I'm in the office.

How is the `graphql.tsx` file generated?

src/components/Register.tsx
nmadruga: Yes. I think I was just doing something else and left it there. Thanks.

PR #17, opened 2020-10-02
Small refactoring - share location using 

### LINKED ISSUE #16: Update AppMenu and navigation states

nmadruga:
Fixes #16 

## Changes

- Create new container folder: **Main** where the **NavigationState** is located. 
  - Perhaps **AppMenu** should be moved from component to container as well.
- Add folder in the components folder `Application` that stores all the application related components.
- Keep the `queryParameters` and `pathname` on the **NavigationState** which is used in many components. Why not use the useLocation directly? - the reducer on this context ares is called after every change on the path, which can be tracked by the DevTools. Also - the utility method that transform the query parameters into an object happens automatically after every URL update.


nmadruga:
This PR has been included in #23. Closing this without merge.

PR #23, opened 2020-10-12
#14 Display of new application (User Registration)

### LINKED ISSUE #14: Build components for Basic user registration applications
UI layout should be coming from designer next week (12/10/2020). Start with the most basic layout that display all elements/sections from what is fetched from GraphQL (saving to cache). On submit should trigger the action to save the new user.

nmadruga:
Fixes #14 

## Required:
* `yarn install`
* Back-end server running master [HEAD](https://github.com/openmsupply/application-manager-server/commit/6dfcf263e7ccea2e83bc02469e080403c86e9257)

Initially this PR would be covering not only, create the application based on the template (code=UserRego1), but also display the application with a few question elements using a local version of the components. After working for a while on the workflow to show an application after its created on the back-end which added many changes to this PR I decide to create the PR and continue with the creation of elements in another PR.

The local state is mainly done by the Apollo cache now, which seems more reasonable. For now the only state that is kept on the context is `serialNumber` which is created in the front-end (temporarily) after the **Create Application** button is clicked. The main component which is responsible for creating the application will use this state to make the mutation. 

## Changes:
- Created folder `utils` and moved folders `generated` and `graphql` (with queries, mutations and fragments) in there.
- Renamed component `Application.tsx` to `ApplicationHeader.tx`
- Added folders `component/Application` to keep all related components of an application in there.
- Added component `ApplicationStart` that shows the button to create an application (that has information of the template).
- Added component `ApplicationStep` to display the content of one step of the new/edited application.
- Added component `ApplicationSummary` to display the summary as the final step of the application creation.
- Added component `FilterList` to store the list of filters of the applications. Extra change, not required for this PR.
- Added new folder `container/Application` for elements that deal with **Graphql queries/mutations**. I'm not sure if we need this differentiation. Open to discussion, but for now that's how it's implemented:
  * `ApplicationCreate` -> **Query** template to display options (passed as props to `Application Start`). **Query** application and call another container to display the first page of an application (`ApplicationPage`).
  * `ApplicationNew` -> **Mutation** to create application and section after the button on `Application Start` is pressed. Or show the `ApplicationSelectType` in case no template was passed in the URL.
  * `ApplicationPage` -> **Query** the application to display. This page should get the content to be displayed in each page on component `ApplicationStep`.
- Added some new queries and mutations (some not in use yet)
- Replace hooks used for navigation with new `utils/hooks/useRouter` that keeps all navigation state discoverable in the components.
- New file `utils/types.ts` to keep shared types between components.


CarlosNZ:
It's running, but I'm getting this error when starting up:
```
ERROR in [at-loader] ./src/utils/hooks/useRouter.tsx:2:25 
    TS7016: Could not find a declaration file for module 'query-string'. '/Users/carl/Github/openmsupply-application-manager/application-manager-web-app/node_modules/query-string/index.js' implicitly has an 'any' type.
  Try `npm install @types/query-string` if it exists or add a new declaration (.d.ts) file containing `declare module 'query-string';`
```

Is this expected?

nmadruga:
@CarlosNZ , sorry I forgot to say, You need to run `yarn install` because there is a new package called `query-string`.
I'll add to the issue description too.

src/utils/graphql/queries/getElements.query.ts
```diff
@@ -0,0 +1,17 @@
+import { gql } from '@apollo/client'
+
+export default gql`
+  query getElements($sectionId: Int!) {
+    templateElements(filter: { sectionId: { equalTo: $sectionId } }) {
```
CarlosNZ, line 5: Just wondering why you used `filter` here rather than `condition`. (Filter is from advanced filtering plugin, condition is built-in). I guess it doesn't really matter, but there is a difference when using live queries (not really an issue here). 

Do you think we should stick to using `condition` where we can and only extend to the advanced `filter` when we really need it?

src/utils/graphql/queries/getTemplate.query.ts
```diff
@@ -0,0 +1,38 @@
+import { gql } from '@apollo/client'
+
+export default gql`
+  query getTemplate($code: String!, $status: TemplateStatus = AVAILABLE) {
+    templates(condition: { code: $code, status: $status }) {
```
CarlosNZ, line 5: You used `condition` here, so keeping it all the same except where we need advanced filtering makes it slightly more consistent.

CarlosNZ (commented):
This is great, Nicole. A lot of work here. I haven't gone through it in huge details as there's so much (!) and I don't fully understand it all. It seems like a really good backbone to start building the rest of the application on, so it'll probably be helpful to give us (well me at least ;) ) a talk through of the main concepts, particularly how you're managing state with the Apollo cache -- I don't really understand it yet.

Just a couple of tiny tweaks/considerations. It'd be good to get this into `master` and start working Andrei's plugin elements into it.

nmadruga:
>  it'll probably be helpful to give us (well me at least ;) ) a talk through of the main concepts, particularly how you're managing state with the Apollo cache

I'll prepare something to present to you today. : ) 
It will be also good to discuss some things that I'm not sure how we should manage about cache and bad latency.

src/components/Application/ApplicationHeader.tsx
```diff
@@ -0,0 +1,22 @@
+import React from 'react'
+import { Header, Segment } from 'semantic-ui-react'
+
+export interface AppHederProps {
```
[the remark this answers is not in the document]
nmadruga: Thanks. Will fix this now.

src/utils/graphql/queries/getElements.query.ts
nmadruga: Yes, keeping consistency will be be better. I haven't seen the difference and it works the same with `condition`. I'll change this now.

nmadruga:
After having a meeting where the workflow of the Frontend was discussed some points were raised that will be treated in different PR (which can be merged to this PR so we can keep track of these changes).

List of topics:

1. - [ ] Move the logic to hooks to manipulate the queries/mutations calls with the useEffect that done inside each container:
      - [ ] Hook of `useGetTemplateQuery` from `ApplicationCreate`
      - [ ] Hook of `useGetApplicationQuery` from `ApplicationCreate`
      - [ ] Hook of `useGetApplicationQuery` from `ApplicationPage`
      - [ ] Hook of `useCreateApplicationMutation` and `useCreateSectionMutation` (should also add creation of `Responses`) from `ApplicationNew`
     - [ ] Hook of `useGetSectionElementsQuery` from `ApplicationStep`
2. - [ ] Fix wrong use of `replace` to use `push` of router in `ApplicationCreate`
3. - [ ] Remove creation of local state for pages - After creation just call `firstSection/Page1`. The `ApplicationPage` would get the current Section and figure out what to display in the first page - ideally the query would be preset to receive only the elements to be displayed.
4. - [ ] Create local **context state** (like applicationState) to store all question and responses in a simple object with { `key` = questionCode, `value` = questionValue }
5. - [ ] `ApplicationQuestion` should be watching the new **context state** and call the evaluate expression. It would only re-render if some of the dynamic values (evaluated by evaluateExpression) has changed. e.g.: `isEditable` has changed.
6. - [ ] Replace use of `refetchQuery` with `readQuery` and `writeQuery` for better latency feedback.
7. - [ ] Research how to send cached local changes to the server in one call. Example: update cache `onChange` of responses and then send a mutation to the server when click the `Next button`.
8 - [ ] Research how to define a query should check the server or just use cache - What we have though so far is: 
     - Subscription to always get `live` data from the server. 
     - Use the query/mutation to fetch once and then use cached data - (will have to update cached data after changes)
     - `readQuery` and `writeQuery` to use only cache

PR #25, opened 2020-10-14
Use expression-evaluator in plugins -- 1st example

CarlosNZ:
Starts to address #24 -- this version shouldn't be merged, will implement properly using a Hook (probably)

Just a quick demonstration of how the evaluator might work in question plugins, for visibility condition. Branched off #21 with the following changes:

- each question/element handles its own visibility evaluation in `ApplicationViewWrapper` using useState/useEffect localised to each components own scope.
- instead of `isVisible` being passed into ApplicationViewWrapper, it just passes the unevaluated expression as `isVisibleExpression`
- each ApplicationViewWrapper has its own state object `isVisible`, with its visibility and error status

Currently only handles simple operators, no objects or database connections are passed in. If you put an unresolvable query in the Visibility Condition input, the UI element should appear in red with the error message.

This is just a proof of concept. Once we all agree with this general approach, need to do:

- pass in object(s) containing relevant local state -- i.e. responses indexed by code, user info
- pass in GraphQL connection and fetch object (for API calls)
- generalise to other conditions, including validation

Hot tip: evaluator can handle literal values, so you can just write `true` rather than `{"value": true}` and it will work fine.


src/elementPlugins/ApplicationViewWrapper.tsx
```diff
@@ -1,15 +1,31 @@
...
 const ApplicationViewWrapper = (props: ApplicatioViewProps) => {
   const {
     templateElement: { elementTypePluginCode: pluginCode },
-    isVisible,
+    isVisibleExpression,
   } = props
 
+const [isVisible, setIsVisible]:any[] = useState({visible: evaluateExpression(isVisibleExpression), error: null})
+
+useEffect(() => {
```
andreievg, line 15: @CarlosNZ, when would this be called ? i.e. would there be a prop to 'applicationState' on this component, and whenever anything changes in 'applicationState', React would call ApplicationViewWrapper which would in turn call useEffect ? What if nothing in applicationState changes this ApplicationViewWrapper (and it doesn't need to re render, would useEffect be called) ? Would be great to have an example of relying on some application state.

I think the reason i had it in parent component, was to use onUpdate method from any of the wrappers, to re run the evaluationExpression for visibility of all components. If it can be done within the wrapper, it's probably better

CarlosNZ: Yeah, in this case it's only going to update when the actual _expression_ changes, which happens in the template builder, but not in an actual application.

In the real one, we'd also pass in the `{question_code: response}` object containing all the questions (plus the `user` object and any other  useful bits of local state), and monitor those objects with `useEffect`. That should work, shouldn't it?

andreievg: I wasn't sure exactly when useEffect would be called, I think it should be called any time it re-renders though 👍

nmadruga (changes requested):
Just testing something here

nmadruga (commented):
## Test
**how does this look?**

nmadruga (approved):

andreievg:
Don't think we need this here anymore

Issue #33, opened 2020-10-15
Move the logic of getting the section and elements to ApplicationPage 


nmadruga:
Remove `pages` from ApplicationState. And create the page url (on ApplicationCreate) with simple call firstSection/Page1. The ApplicationPage would get the current Section and figure out what to display in the first page - ideally the query would be preset to receive only the elements to be displayed.

---

This issue is part of a list of issues created after a meeting to discuss how we would like to structure the cache and calls to back-end. See list [here](https://github.com/openmsupply/application-manager-web-app/pull/23#issuecomment-708878476)

Please, `base` and `merge` to [PR #14](https://github.com/openmsupply/application-manager-web-app/pull/23).

We will merge everything to the master when most of this issues are completed.

PR #42, opened 2020-10-18
#28 Hook to load application (with all elements)

### LINKED ISSUE #28: Hook of useGetApplicationQuery from ApplicationCreate

nmadruga:
Fix #28. Merging to #14 (will be merged back to master as bulk).

## Changes
- Removed use of `pages` from ApplicationState - since the logic to get the pages in each section will be moved (see #33)
- Created hook to do `getApplicationQuery`. It makes the query using another Apollo-client hook that returns `data`, `loading` and `error`.  The goal is to have all the elements from one application loaded in cache before the application can be filled by the user - all the next queries for each section will be using the data stored in the cache. 
For now what is returned by this hook is the currentSection `code` to be used in the URL. This will change depending on the application be a draft - it should show the application on the same step the applicant was filling - which is not been done yet!

### DIFF
--- a/src/containers/Application/ApplicationCreate.tsx
+++ b/src/containers/Application/ApplicationCreate.tsx
@@ -1,18 +1,15 @@
 import React, { useEffect, useState } from 'react'
 import {
-  Application,
-  ApplicationSection,
   Template,
-  TemplateElement,
   TemplateSection,
-  useGetApplicationQuery,
   useGetTemplateQuery,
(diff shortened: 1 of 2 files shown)


nmadruga:
After chatting with Andrei this PR seems not required. Also - next PR (https://github.com/openmsupply/application-manager-web-app/pull/50) has the changes from this one, so closing it without merging.

Issue #35, opened 2020-10-15
Hook to query the full application, sections and elements and create flatten object with all questions and responses 


nmadruga:
ApplicationQuestion should be watching the new context state and call the evaluate expression. It would only re-render if some of the dynamic values (evaluated by evaluateExpression) has changed. e.g.: isEditable has changed.

Either in the ApplicationQuestion or in the ApplicationWrapper - which should be doing things that are common to every question plugin.

----

This issue is part of a list of issues created after a meeting to discuss how we would like to structure the cache and calls to back-end. See list [here](https://github.com/openmsupply/application-manager-web-app/pull/23#issuecomment-708878476)

Please, `base` and `merge` to [PR #14](https://github.com/openmsupply/application-manager-web-app/pull/23).

We will merge everything to the master when most of this issues are completed.

PR #49, opened 2020-10-19
#35 Response object hook

### LINKED ISSUE #34: Create local context state (like applicationState) to store all question and responses in a simple object with { key = questionCode, value = questionValue }

CarlosNZ:
Fixes #34. Merge back into #28 (which in turn goes into #14)

Created the hook (`useGetAllResponses`) and confirmed that it works by calling it in `ApplicationPage`. 

Updated the `useGetApplicationQuery` to also return all responses for this application.

We'll probably want to add this hook to the main ApplicationState context at some point, but leaving that out for now to avoid conflicting with @nmadruga 's changes.

### DIFF
--- a/src/utils/hooks/useGetAllResponses.tsx
+++ b/src/utils/hooks/useGetAllResponses.tsx
@@ -1,6 +1,11 @@
 import { useState, useEffect } from 'react'
-import { Application, ApplicationSection, useGetApplicationQuery } from '../generated/graphql'
-import { AllResponses } from '../types'
+import {
+  Application,
+  ApplicationResponse,
+  ApplicationSection,
+  useGetApplicationQuery,
+} from '../generated/graphql'
+import { ResponsesByCode } from '../types'
 
 interface useLoadApplicationProps {
   serialNumber: number
@@ -8,7 +13,7 @@ interface useLoadApplicationProps {
 
 const useGetAllResponses = (props: useLoadApplicationProps) => {
   const { serialNumber } = props
-  const [allResponses, setAllResponses] = useState({})
+  const [responsesByCode, setResponsesByCode] = useState({})
   const { data, loading, error } = useGetApplicationQuery({
     variables: { serial: serialNumber },
   })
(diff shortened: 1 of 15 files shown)


src/utils/hooks/useGetAllResponses.tsx
```diff
@@ -0,0 +1,42 @@
...
+    if (data?.applications) {
+      if (data.applications.nodes.length === 0) return
+      if (data.applications.nodes.length > 1)
+        console.log('More than one application returned. Only one expected!')
+
+      const applicationResponses = data.applications.nodes[0]?.applicationResponses.nodes
+
+      const currentResponses = {} as AllResponses
+
+      applicationResponses?.forEach((response) => {
```
nmadruga, line 26: Not a good idea to be using `forEach` on a variable that can be undefined. Better to do a check before this I think.

src/utils/hooks/useGetAllResponses.tsx
```diff
@@ -0,0 +1,42 @@
...
+        const code = response?.templateElement?.code
+        if (code) currentResponses[code] = response?.value?.text || response?.value
```
nmadruga, line 28: In this case if value is undefined or null it would add to the `currentResponses` object right?

nmadruga (changes requested):
It seems good. 
I wonder why the difference with prettier config. I need to check that with you, mine Vs code can be missing some configuration that yours have.
Just some changes requested - checking possible `undefined` variable and need to update the use of `serial` to be a string.

src/utils/hooks/useGetAllResponses.tsx
CarlosNZ: Yes :)

src/utils/hooks/useGetResponsesByCode.tsx
```diff
@@ -0,0 +1,48 @@
...
+  }, [data, error])
+
+  return {
+    error,
+    loading,
+    responsesByCode,
+  }
+}
+
+export default useGetResponsesByCode
```
nmadruga, line 48: I would recommend using a similar code to what was suggested by Andrei on branch [27-suggested-improvements](https://github.com/openmsupply/application-manager-web-app/blob/%2327-suggested-improvements/src/utils/hooks/useLoadTemplate.tsx). 
Basically renaming the variables `error` to `apolloError` and `loading`to `apolloLoading` (see my previous suggestions). Then keeping track of each unexpected problems found (more than one application found) to the **new** `error` variable. Which can be done in a separated function (check Andrei suggestion). Also the loading should have another one that is handled by us and is set to `true` only after the object is filled or if an error occurred.

src/utils/hooks/useGetResponsesByCode.tsx
```diff
@@ -0,0 +1,48 @@
...
+import { ResponsesByCode } from '../types'
+
+interface useLoadApplicationProps {
+  serialNumber: string
+}
+
+const useGetResponsesByCode = (props: useLoadApplicationProps) => {
+  const { serialNumber } = props
+  const [responsesByCode, setResponsesByCode] = useState({})
+  const { data, loading, error } = useGetApplicationQuery({
```
nmadruga, line 17:
```suggestion
  const { data, loading: apolloLoading, error: errorLoading } = useGetApplicationQuery({
```

nmadruga (changes requested):
Also, could you merge the current #23 into here, I think some of the styling changes will be removed from here after that. If you need any help with conflicts just let me know.

src/utils/hooks/useGetResponsesByCode.tsx
CarlosNZ: Okay, think I've covered it. Please let me know if it's correct.

CarlosNZ:
Closing this without merging as I'll create a new branch based on #54 and move my changes to that.

Issue #52, opened 2020-10-20
Troubleshoot why Questions/Elements are not being rendered correctly.


CarlosNZ:
Based on @nmadruga's recent work, the question/elements framework that @andrei has provided should show up in Application page view, but they are not.

PR #53, opened 2020-10-20
#52 Element rendering fix

CarlosNZ:
This is just a basic fix for #52. Main problem was the `useGetSectionElementsQuery` query was not fetching `elementTypePluginCode`. 

Also, removed `ApplicationQuestion` component, since it was doing the same thing as the ApplicationViewWrapper.

Have hard-coded an initial `pluginManifest` in `pluginProvider` as per @andreievg's example.

This is not meant to be a complete implementation for the rendering of elements, some values are still hard-coded (`isVisible` etc.)

Will use this as basic for using the evaluator next. Let me know if you think I should do any additional changes in this issue.


nmadruga (approved):
Perfect. It is actually a small PR, but because all of the style changes it looks bigger.
I'm happy to merge this one now, hopefully we won't get too many conflicts with the PR
#54 :octocat:

CarlosNZ:
Going to re-do this one to base on #54, so don't review this just yet.

nmadruga (changes requested):
As per previous comment. Flagging this PR should wait to be merged.

CarlosNZ:
Closing -- replaced by #57

PR #54, opened 2020-10-21
#29 #31 & #33 Refactoring: ApplicationCreate and ApplicationPage using hooks

nmadruga:
Fixes #29 
Fixes #31
Fixes #33. 

Branched from [PR #28](https://github.com/openmsupply/application-manager-web-app/pull/42)!
Merging to #14 (will be merged back to master as bulk).

## Changes
- Move all the code logic to get elements of current `section` and `page` to new hook `useGetElementsInPage`.
- Removed old container `ApplicationNew` - now combined with `ApplicationCreate`! Yey.
- Took use of hook `useLoadApplication` off the `ApplicationCreate`, it just uses template to call the `ApplicationPage`.
- Added logic on the hook `useLoadApplication` to find `start` and `total` pages per section.
- Updated `useCreateApplication` to return object with all responses loaded. It will only call `ApplicationPage` when all is done.
- Pass functions to component `ApplicationStep` to control **next** and **previous** buttons:
  - The **previous** simply uses `goBack()` (utility method from `useRouter`). Available from page 2 & beyond.
  - The **next** required a object per section that defines for each section what is the `start` and `total` number of pages. This is used to determine if next page is in another section.

PR #56, opened 2020-10-21
35(2) Response object hook

### LINKED ISSUE #35: Hook to query the full application, sections and elements and create flatten object with all questions and responses

CarlosNZ:
Fixes #35.

This PR is a replacement for #49 as it has been refactored to be based on #54.

### DIFF
--- a/src/utils/hooks/useGetResponsesByCode.tsx
+++ b/src/utils/hooks/useGetResponsesByCode.tsx
@@ -28,23 +28,17 @@ const useGetResponsesByCode = (props: useLoadApplicationProps) => {
       return
     }
 
-    if (data?.applications) {
-      if (data.applications.nodes.length === 0) return
-      if (data.applications.nodes.length > 1)
-        console.log('More than one application returned. Only one expected!')
+    const applicationResponses = data?.applications?.nodes[0]?.applicationResponses
+      .nodes as ApplicationResponse[]
 
-      const applicationResponses = data.applications.nodes[0]?.applicationResponses
-        .nodes as ApplicationResponse[]
+    const currentResponses = {} as ResponsesByCode
 
-      const currentResponses = {} as ResponsesByCode
+    applicationResponses.forEach((response) => {
+      const code = response?.templateElement?.code
+      if (code) currentResponses[code] = response?.value?.text || response?.value
+    })
 
-      applicationResponses.forEach((response) => {
-        const code = response?.templateElement?.code
-        if (code) currentResponses[code] = response?.value?.text || response?.value
-      })
-
-      setResponsesByCode(currentResponses)
-    }
+    setResponsesByCode(currentResponses)
   }, [data, apolloError])
 
   return {
--- a/src/utils/hooks/useGetResponsesByCode.tsx
+++ b/src/utils/hooks/useGetResponsesByCode.tsx
@@ -14,6 +14,7 @@ const useGetResponsesByCode = (props: useLoadApplicationProps) => {
   const { serialNumber } = props
   const [responsesByCode, setResponsesByCode] = useState({})
   const [error, setError] = useState('')
+  const [loading, setLoading] = useState(true)
   const { data, loading: apolloLoading, error: apolloError } = useGetApplicationQuery({
     variables: { serial: serialNumber },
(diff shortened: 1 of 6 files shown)


src/utils/hooks/useGetResponsesByCode.tsx
```diff
@@ -0,0 +1,60 @@
+import { useState, useEffect } from 'react'
+import {
+  Application,
+  ApplicationResponse,
+  ApplicationSection,
+  useGetApplicationQuery,
+  GetApplicationQuery,
+} from '../generated/graphql'
```
nmadruga, line 8:
```suggestion
import {
  ApplicationResponse,
  useGetApplicationQuery,
  GetApplicationQuery,
} from '../generated/graphql'
```
Unused types

src/utils/hooks/useGetResponsesByCode.tsx
```diff
@@ -0,0 +1,60 @@
...
+      setError(error)
+      return
+    }
+
+    const applicationResponses = data?.applications?.nodes[0]?.applicationResponses
+      .nodes as ApplicationResponse[]
+
+    const currentResponses = {} as ResponsesByCode
+
+    applicationResponses.forEach((response) => {
```
nmadruga, line 36: This line fails if I try accessing the address of one existing application page with the **server not running**: e.g: ` http://localhost:3000/applications/1236/SectionA/page1`. 
I started doing that to make sure we are covering cases of bad responses from the server. I think this is another case to be checked on the `checkForApplicationErrors` or perhaps just returning the `apolloError` might fix that. See what I've done on `ApplicationCreate`.

src/utils/hooks/useGetResponsesByCode.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+    if (error) {
+      setError(error)
+      return
+    }
+
+    const applicationResponses = data?.applications?.nodes[0]?.applicationResponses
+      .nodes as ApplicationResponse[]
+
+    const currentResponses = {} as ResponsesByCode
```
nmadruga, line 34: Oh, that's interesting. I usually initialise the variable like this: `const currentResponses: ResponsesByCode = {}`, but I liked the way this is done it's more declarative : )

src/utils/hooks/useGetResponsesByCode.tsx
```diff
@@ -0,0 +1,60 @@
...
+      const code = response?.templateElement?.code
+      if (code) currentResponses[code] = response?.value?.text || response?.value
+    })
+
+    setResponsesByCode(currentResponses)
+  }, [data, apolloError])
+
+  return {
+    apolloError,
+    apolloLoading,
```
nmadruga, line 46: It would be nice to also have the loading done here. Similar to what was done in `useLoadTemplate`, because the `apolloLoading` will be set to false as soon as the query is fetch, but we would like to watch a `loading` that is only set to be false after this hook has finished doing the manipulation of data.

nmadruga (changes requested):
Summary of requests:
- Create a local variable to control `loading` of our custom hook
- Fix problem when server is not reachable.

src/utils/hooks/useGetResponsesByCode.tsx
CarlosNZ: Ha ha, so do I actually -- I guess I'd just got used to seeing your `as` style a lot recently. 😄

src/utils/hooks/useGetResponsesByCode.tsx
nmadruga: Created a separated issue for that: #60

nmadruga (approved):
It seems good to me! I tried to check the ApolloError and is also returning undefined to me. So a new issue on `Issues Backlog` to just deal with that - when we can start doing a proper global apollo error handling too.

PR #57, opened 2020-10-21
#52(2) -- Fix element rendering problem

### LINKED ISSUE #52: Troubleshoot why Questions/Elements are not being rendered correctly.

CarlosNZ:
Replaces PR #53 (which fixes #52).

Copied from that PR:

This is just a basic fix for #52. Main problem was the `useGetSectionElementsQuery` query was not fetching `elementTypePluginCode`. 

Also, removed `ApplicationQuestion` component, since it was doing the same thing as the ApplicationViewWrapper.

Have hard-coded an initial `pluginManifest` in `pluginProvider` as per @andreievg's example.

This is not meant to be a complete implementation for the rendering of elements, some values are still hard-coded (`isVisible` etc.)

Will use this as basis for using the evaluator next. Let me know if you think I should do any additional changes in this issue.

Also did:

- fixed misspelling of `ApplicatioViewProps` type.
- renamed plugin `basicText` to `shortText` to match our [spec](https://github.com/openmsupply/application-manager-server/issues/61)


nmadruga (approved):
LGTM. Thanks for doing that!

Issue #58, opened 2020-10-21
Update front-end page logic to use indexed ordering for Sections and Elements


CarlosNZ:
Complementary back-end issue: https://github.com/openmsupply/application-manager-server/issues/82

PR #59, opened 2020-10-22
#36 replace refetch query with write query

nmadruga:
2 examples:

- After `Section` mutation -> To update the cache with the new section.
- After `User` mutation (example) -> To update the list of users.

## Unexpected behaviour
Might need to wait for issue to persist cache #46 to find out how we need to configure our cache. It seems that updating the cache is **not** updating the connections. So, the Application still gets the sections, but probably it's getting it from the server. To be confirmed!


CarlosNZ:
Hey, when I click "User Registration" button, I can see the cache is created correctly in the dev tools, but no next page with elements is loaded -- it just goes blank.

nmadruga:
> Hey, when I click "User Registration" button, I can see the cache is created correctly in the dev tools, but no next page with elements is loaded -- it just goes blank.

The reason might be because I wasn't looking at that code in this PR - it is something that was fixed (and changed entirely in the next PRs - I think it's fixed on the next PR #58 or, definitely fixed on PR #69).

nmadruga:
@andreievg @CarlosNZ Can I merge this one now? Or would you like to have another look on this?
I'm planning to add some documentation for the current Front-end development flux which might be helpful to review the next PRs. Hopefully what was done in this PR is simple to review, but let me know if there is anything that is not clear.

CarlosNZ:
> @CarlosNZ Can I merge this one now?

Fine with me.

Issue #51, opened 2020-10-19
Problem controlling state when page is refresh or link followed


CarlosNZ:
As originally noted in #30:

> When I start the application form, so the URL is `http://localhost:3000/applications/6944/S1/page1`, and I refresh the page, it reloads with "Some problem" where the list of elements should have been.

> Also -- after I start an application, if I click "Register" in the side-bar again, it stays at the current page

Suspect that the Apollo cache needs to be cleared when doing either of these actions.

CarlosNZ:
Based on current master, going directly to a page or refreshing a page just loads a blank screen.

CarlosNZ:
Seems to be a problem with the `useGetResponsesByCode` hook. When the following line is commented out in ApplicationPage, the page loads okay:

```http://localhost:3000/applications/3549/S1/page1``` (Line 20)

CarlosNZ:
The problem is that when `useGetResponsesByCode` calls `useGetApplicationQuery`, the Data returned from the query is `undefined` when started from a specific page, but correct when launched from starting a new application.

PR #61, opened 2020-10-23
getResponseByCode moved and skeleton for update state

CarlosNZ:
I've moved the functionality from the `useGetResponsesByCode` hook into `useLoadApplication`, so it only needs to call `useGetApplicationQuery` once, which seems to help somewhat with #51.

However, this only runs once, so what do we use to update the `responsesByCode` object each time a form element is updated?

Also, placed a skeleton for a function `updateState` which runs whenever a form element in `shortText` loses focus. Need to write the actual query to mutate the cache, or local state --whatever we're using.

Maybe some discussion is required on both these points before we can merge this?


nmadruga (commented):
Maybe delete the `src/utils/hooks/useGetResponsesByCode.tsx`

src/utils/hooks/useLoadApplication.tsx
```diff
@@ -99,6 +59,19 @@ const useLoadApplication = (props: useLoadApplicationProps) => {
         }
       })
       setSections(mapSectionsDetails)
+
+      // Build application responsesByCode object -- TO:DO add to Context applicationState
```
nmadruga, line 63: > -- TO:DO add to Context applicationState

I think that's done.

nmadruga:
Closing this one for now - will have to be changed to be used in a different place.

PR #62, opened 2020-10-23
#58 update logic of elements order by index

### LINKED ISSUE #58: Update front-end page logic to use indexed ordering for Sections and Elements

nmadruga:
Fixes #58.

**Important**: Run using back-end on https://github.com/openmsupply/application-manager-server/pull/83

## Changes:
- Change from using `goBack` for previous page to having a handler for the previous button.
- Elements to be displayed on the `ApplicationPage` are selected by index. 
- Next button logic also changed to consider:
  - Each section starts from `page1`
  - Go to the next section based on `section.index`
- The `useLoadTemplate` hook currently is finding the number of pages per section which is stored in the **ApplicationState**
- The `useLoadApplication` logic to find each section pages is removed.

### DIFF
--- a/src/cache.ts
+++ b/src/cache.ts
@@ -1,0 +1,20 @@
+import { InMemoryCache } from '@apollo/client'
+import { TemplateElementsConnection } from './utils/generated/graphql'
+
+const cache = new InMemoryCache({
+    typePolicies: {
+      TemplateSection: {
+        fields: {
+          pageCount: {
+              read(_, { readField }) {
+                // TODO: Move code to find number of pages per section (in useLoadTemplete) to here
+                  const elementsConnection = readField('templateElementsBySectionId') as TemplateElementsConnection
+                  return 0
+              }
+          }
+        }
+      }
+    }
+})
+
+export default cache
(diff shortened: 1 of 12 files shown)


CarlosNZ (changes requested):
Nice! Working great, other than the page refresh still not working.

Since you mentioned you think you know what that is, maybe try fixing that here now?

nmadruga:
> Since you mentioned you think you know what that is, maybe try fixing that here now?

I have fixed the problem with reloading the page now - although I have commented the line calling the `useGetResponsesByCode` hook on the `ApplicationPage`. There a problem with calling sequential hooks (that depends on the previous one), since the calls are made asynchronously the App gets lost. What I have noticed is that the best use of hooks is basically one query per controller. So we might need to define exactly where to call the `useGetResponsesByCode` to be passed to each wrapper controller : )

CarlosNZ (approved):
Sweet. 👍

(Haven't scrutinised the code in great detail, but functionality does what it should, and the refresh page is working nicely now.)

CarlosNZ:
Can you merge [#83 in the back-end](https://github.com/openmsupply/application-manager-server/pull/83) at the same time?

nmadruga:
> Can you merge #83 in the back-end at the same time?

Sure. I'll start doing this now

PR #69, opened 2020-10-27
#47 basic application summary and submit

### LINKED ISSUE #47: Basic application submission logic + summary page
Also add response passed as props to ApplicationQuestion and mutation for when the value is changes by the user
Just a simple json of application with submit button.
Don't need to have the summary version of plugins yet!

nmadruga:
Fixes #47 

Basically to get this one working I had to edit some of the **query** and **hooks**, keep only one **hook** per controller and add the response **mutation** and application **mutation**. Now each page shows the existing responses and that's been send to the server on Input **change** - not ideal. Issue #46 should take care of just keeping changes on Cache and sending to the server either when **Next** button is pressed or only when **Submit** button is pressed.

## Changes NOT done yet - should be done in other issues:
- Plugin to do ‘onUpdate’ get triggered onLooseFocus instead of onChange
- On Summary validate each response to show error (and disallow submission)

## Changes
1. Rename component `ApplicationStep` to `ElementsArea` (that is now a container since it's also dealing with logic).
2. Add the response value to each element on the `ApplicationPage`.
3. Add **mutation** that updates each response in the application with props passed to `ApplicationViewWrapper`.
4. Edit `useLoadElements` hook  to either retrieve elements in one page or for the whole section.
5. Create `useLoadSummary` hook  that calls the `useLoadElements` for each section in the application.
6. Edit `useLoadApplication` hook to also return the sections of the current application.
7. Create controller `ApplicationOverview` that is called when url: `/applications/serialNumber/summary` and first thing is to call the `useLoadApplication` hook to allow the summary page to be reloaded. Then it will call the `ApplicationSummary` component passing some required props.
8. Display all sections and responses on the `ApplicationSummary` page with the Submit button.
9. Add **mutation** that updates the application `trigger` field to `onApplicationSubmitted`.


nmadruga:
The problem with reloading the page, I think we might need another intermediate page that is called before the  `ApplicationSummary`. And this page/container is responsible for checking weather the `id` and `sections` have been set, in case they haven't it would call the existing **hooks** and update the `applicationState` before calling the page to display the summary. What you guys think? I'll create a separated issue for that (since this one is so big already).

nmadruga:
OK. After changing to **not store sections** in the `applicationState` on the branch that this PR is based of, I got conflicts here. Now, in order to solve these conflicts, I will have to do the fix to allow reloading the `ApplicationSummary`. Which is basically adding a call of the `useLoadApplication` hook in the workflow of displaying the summary page.

Should be ready to review in 1 hour.

src/containers/Application/ElementsArea.tsx
```diff
@@ -0,0 +1,111 @@
...
+import { Button, Container, Grid, Header, Label, Segment } from 'semantic-ui-react'
+import { ApplicationViewWrapper } from '../../elementPlugins'
+import useLoadElements from '../../utils/hooks/useLoadElements'
+import { Loading } from '../../components'
+import { useUpdateResponseMutation } from '../../utils/generated/graphql'
+
+interface ElementsAreaProps {
+  applicationId: number
+  sectionTitle: string
+  sectionTempId: number
```
CarlosNZ, line 11: Maybe should be called `sectionTemplateId` to avoid confusion?

CarlosNZ (approved):
Really nice, works well. There's a few problems with the format of the response data (should be JSON object), and a few other things I noticed, but it's all outside the scope of this issue, so no need to deal with it now.

Would recommend changing that one variable name though (`sectionTempId`) for consistency ;)

nmadruga:
Thank you Carl. Sorry it took me so long to reply. I've done the improvement you suggested. 

Going to merge this one now.

Issue #18, opened 2020-10-04
Use the evaluateExpression on each question


nmadruga:
using evaluationExpression on UI example: Select list of companies based on toggle for Applicant category: Manufacturer/Supplier

andreievg:
@CarlosNZ 

I thought about this a bit. And I think we would need to re-run evaluator on isVisible and isRequired on ALL questions (not just the current page), this would be required for progress bar and for summary view.

Thus I think we should try to minimise the number of times evaluator should run, suggestions:

**Basic**
Identify an evaluation express as 
* static -> needs to run only once, evaluation outcome does not change by state of application
* dynamic -> can change based on application outcome

**Advanced**
For dynamics ->
* specify which fields in the application impact evaluation outcome
* compute an object with field response value difference (in the parent container), and only run evaluationExpression for elements that specify reliance on those fields

Also something to keep in mind, if element becomes not visible, we should change it's value to null. (not sure how to best do it, probably from within the wrapper ?)

CarlosNZ:
> Also something to keep in mind, if element becomes not visible, we should change it's value to null. (not sure how to best do it, probably from within the wrapper ?)

Do we really, though? I was thinking about this, and I don't really see a downside to just leaving it as is. The `isVisible` field will dictate that it should be ignored in all subsequent processing, and in the occasional case that the applicant switches a conditional answer back to what it used to be, it'd be nice for the previous response to pop back into the form as it was.(i.e. less work for more features!)

PR #83, opened 2020-11-04
#34 Local state responses and evaluation handling

CarlosNZ:
Addresses #34 and #18.

This PR adds an additional hook to `ApplicationPageWrapper` (renamed from "ApplicationPage") called `useGetResponsesAndElementState`. This creates three state objects at the PageWrapper level:
- `responsesByCode` - all current user responses in the form `{ <questionCode> : <responseValue>... }` )
- `responsesFullByCode` - same as previous but contains the full response object ( `{ text, optionIndex, reference }` )
- `elementsState` - contains evaluated values for `isVisible`, `isEditable`, and `isRequired` for each element, indexed by `elementCode`. This object is recomputed whenever `responsesByCode` changes.

Note, I have not yet implemented a mutation to actually change the responseValues when the user fills in the form, will be done in a separate PR.

In order to see "Visibility condition" in action, change one of the `visibility_condition` fields in `template_element` to `{"value": false}` (or another query that evaluates to false), and notice that question no longer appears in the form.


documentation/internal/Element-Type-Specs.md
```diff
@@ -0,0 +1,129 @@
...
+- **is_editable**: `JSON` -- dynamic query determining whether can be edited (Would only be false in rare circumstances)
+  - default: `{"value": true}`
+- **parameters**: `JSON` -- the parameters specific to each question/element type. See individual plugins below for parameter breakdown
+- **validation**: `JSON` -- a dynamic expression for checking if the user's response is a valid input.
+  - default: `{"value": true}`
+- **validation_message**: `string` -- the message that shows in the UI when validation fails.
+
+## Question/Element types
+
+**Note**: all parameter fields can also have a dynamic query object instead of a primitive. The `evaluateExpression` function will return literal strings (or numbers, booleans) as is. The types described for the parameters below are the type that is expected to be _returned_ from a query expression.
```
nmadruga, line 24: Maybe add a link to the description of evaluateExpression (on back-end) here: https://github.com/openmsupply/application-manager-server/wiki/Query-Syntax

documentation/internal/Element-Type-Specs.md
```diff
@@ -0,0 +1,129 @@
...
+
+### Short Text Input _(not yet implemented -- highest priority)_
```
nmadruga, line 26: This one is implemented - maybe not entirely but there is already one in the master branch, and it is showing on the User Registration. I'll create issues for the next ones.

documentation/internal/Element-Type-Specs.md
```diff
@@ -0,0 +1,129 @@
...
+
+- **type/code**: `shortText`
+- **category**: `Question`
+
+_Free-form, single-line text input element_
+
+#### Input parameters (in the `parameters` JSON)
+
+- **label**: `string` -- Text that shows in the HTML "label" attribute of the form element (suggestion: make this interpret Markdown)
```
nmadruga, line 35: > **label** - [...] suggestion: make this interpret Markdown

I have added a issue regarding this suggestion #86

documentation/internal/Element-Type-Specs.md
```diff
@@ -0,0 +1,129 @@
...
+
+#### Response type
+
+_This describes the expected object that will be stored in the `application_response` table `value` field. from the user's response_
+`{ text: <string> }`
+
+---
+
+### Text Information _(not yet implemented)_
+
```
nmadruga, line 49: I think this one is already implemented - just missing the `SummaryView` I'll add an issue about that.

documentation/internal/Element-Type-Specs.md
```diff
@@ -0,0 +1,129 @@
...
+- **hasOther**: `boolean` -- if `true`, an additional text-entry field is provided so the user can add their own alternative option _(may not implement in first iteration but good to have the option in future)_
+
+#### Response type
+
+```
+{
+  optionIndex: <integer> (index from the options array)
+  text: <string> (actual text from options array)
+  reference?: To-do: some way to link the response to an entity in database (e.g. an organisation)
+  -- @nmadruga maybe you have an idea of what we'd need here
```
nmadruga, line 85: I think to be able to have this as part of the response the **Input** would also need to receive that information... In my thinking the options would receive not only a string per row but what some reference to what it is linked to. But maybe using a different element plugin - e.g. Radio Buttons Link ?

documentation/internal/Element-Type-Specs.md
nmadruga: I've added a issue to finish this one considering anything that could be missing: #87

documentation/internal/Element-Type-Specs.md
nmadruga: #88

documentation/internal/Element-Type-Specs.md
```diff
@@ -0,0 +1,129 @@
...
+#### Input parameters
+
+- **title**: `string` -- Heading text to display [Optional]
+- **text**: `string` -- body text to display
+  _(Maybe have some formatting options, but not initially, although `text` should support Markdown)_
+
+---
+
+### Radio Buttons _(not yet implemented)_
+
```
nmadruga, line 64: #89

documentation/internal/Element-Type-Specs.md
```diff
@@ -0,0 +1,129 @@
...
+  reference?: To-do: some way to link the response to an entity in database (e.g. an organisation)
+  -- @nmadruga maybe you have an idea of what we'd need here
+}
+
+```
+
+---
+
+### Drop-down Selector _(not yet implemented)_
+
```
nmadruga, line 93: #90

documentation/internal/Element-Type-Specs.md
```diff
@@ -0,0 +1,129 @@
...
+  text: <string> (actual text from options array)
+  reference?: To-do -- as above
+}
+
+```
+
+---
+
+### Page Break _(not yet implemented)_
+
```
nmadruga, line 120: #91

nmadruga (commented):
Overall comment - Why the file `Overview-of-Navigation-&-URL-structure.md` was renamed to copy?

src/containers/Application/ElementsArea.tsx
```diff
@@ -1,9 +1,10 @@
 import React from 'react'
 import { Button, Container, Grid, Header, Label, Segment } from 'semantic-ui-react'
 import { ApplicationViewWrapper } from '../../elementPlugins'
-import useLoadElements from '../../utils/hooks/useLoadElements'
+import useLoadElements from '../../utils/hooks/useLoadElementsOfSection'
```
nmadruga, line 4:
```suggestion
import useLoadElementsOfSection from '../../utils/hooks/useLoadElementsOfSection'
```

And rename where is in use. It makes it simpler to use the same name as the file.

src/containers/Application/ApplicationPageWrapper.tsx
```diff
@@ -17,14 +18,27 @@ const ApplicationPage: React.FC = () => {
...
+    responsesFullByCode,
+    elementsState,
+  } = useGetResponsesAndElementState({
+    serialNumber: serialNumber as string,
+  })
+
+  console.log('Responses BY Code', responsesByCode)
+  console.log('Responses Full by Code', responsesFullByCode)
+
+  console.log('Element State', elementsState)
```
nmadruga, line 34: Please remove logs

src/containers/Application/ApplicationPageWrapper.tsx
```diff
@@ -1,14 +1,15 @@
-import React, { useEffect } from 'react'
+import React, { useState, useEffect } from 'react'
```
nmadruga, line 1:
```suggestion
import React, { useEffect } from 'react'
```

CarlosNZ:
> Overall comment - Why the file `Overview-of-Navigation-&-URL-structure.md` was renamed to copy?

🤦‍♂️ My bad -- I duplicated it as a basis for a new doc file, and I guess I modified the original rather than the copy. Good spotting, fixed now.

src/utils/hooks/useGetResponsesAndElementState.tsx
```diff
@@ -0,0 +1,127 @@
...
+  const { serialNumber } = props
+  const [responsesByCode, setResponsesByCode] = useState({})
+  const [responsesFullByCode, setResponsesFullByCode] = useState({})
+  const [elementsExpressions, setElementsExpressions] = useState([{}])
+  const [elementsState, setElementsState] = useState({})
+  const [error, setError] = useState('')
+  const [loading, setLoading] = useState(true)
+  const { data, loading: apolloLoading, error: apolloError } = useGetElementsAndResponsesQuery({
+    variables: { serial: serialNumber },
+  })
```
nmadruga, line 21: There is a problem with that, I think. For the cached data to be used it would check by id, and since we aren't passing the id here it is just making another call to the server. I think we can make a different issue to solve that, but just flagging this here. That's why I was storing the Application `id` in the context area, and we would need to use the skip to not call this hook until we have the application loaded. Otherwise it's just making 2 calls to the server.

src/utils/hooks/useGetResponsesAndElementState.tsx
```diff
@@ -0,0 +1,127 @@
...
+    error,
+    loading,
+    responsesByCode,
+    responsesFullByCode,
+    elementsState,
+  }
+}
+
+function checkForApplicationErrors(data: GetElementsAndResponsesQuery | undefined) {
+  if (data?.applicationBySerial) {
```
nmadruga, line 121: Check if the result `data` isn't undefined and if `data.applicationBySerial` isn't undefined it's better, so you don't need to be checking inside the scope of the useEffect - since that's called after doing this checks

nmadruga: Oh and `data.applicationBySerial.applicationResponses` just in case

src/utils/hooks/useGetResponsesAndElementState.tsx
```diff
@@ -0,0 +1,127 @@
...
+
+    if (error) {
+      setError(error)
+      setLoading(false)
+      return
+    }
+
+    if (apolloError) return
+
+    const applicationResponses = data?.applicationBySerial?.applicationResponses?.nodes
```
nmadruga, line 34:
```suggestion
    const applicationResponses = data?.applicationBySerial?.applicationResponses.nodes as ApplicationResponse[]
```

And add the import from `graphql`.

src/utils/hooks/useGetResponsesAndElementState.tsx
```diff
@@ -0,0 +1,127 @@
...
+
+    if (applicationResponses) {
+      applicationResponses.forEach((response) => {
+        const code = response?.templateElement?.code
+        if (code) {
+          currentResponses[code] = response?.value?.text
+          currentFullResponses[code] = response?.value
+        }
+      })
+    }
```
nmadruga, line 57:
```suggestion
    applicationResponses.forEach((response) => {
      const code = response.templateElement?.code
      if (code) {
        currentResponses[code] = response.value?.text
        currentFullResponses[code] = response.value
      }
    })
```

Since you already casted to `ApplicationResponse[]` at the top (and the checks will make sure you have received a valid object response).

src/utils/hooks/useGetResponsesAndElementState.tsx
nmadruga: Oops needs to also check for correct responses for: `template` and `template?.templateSections`

src/utils/hooks/useGetResponsesAndElementState.tsx
```diff
@@ -0,0 +1,127 @@
...
+      setError(error)
+      setLoading(false)
+      return
+    }
+
+    if (apolloError) return
+
+    const applicationResponses = data?.applicationBySerial?.applicationResponses?.nodes
+
+    const templateSections = data?.applicationBySerial?.template?.templateSections?.nodes
```
nmadruga, line 36:
```suggestion
    const templateSections = data?.applicationBySerial?.template?.templateSections
      .nodes as TemplateSection[]
```

src/utils/hooks/useGetResponsesAndElementState.tsx
```diff
@@ -0,0 +1,127 @@
...
+
+    const templateElements: TemplateElementState[] = []
+
+    templateSections?.forEach((sectionNode) => {
+      sectionNode?.templateElementsBySectionId?.nodes.forEach((element) => {
+        if (element) templateElements.push(element as TemplateElementState) // No idea why ...[spread] doesn't work here.
```
nmadruga, line 42:
```suggestion
templateSections.forEach((sectionNode) => {
      const elementsInSection = sectionNode.templateElementsBySectionId?.nodes as TemplateElement[]
      elementsInSection.forEach((element) => {
         templateElements.push(element as TemplateElementState) // No idea why ...[spread] doesn't work here.
```

nmadruga: > No idea why ...[spread] doesn't work here.

I think the problem is that the types aren't the same. The structure of `TemplateElementState` isn't considering `undefined` which is a possible return from the GraphQL results for each `TemplateElement`.

src/utils/hooks/useGetResponsesAndElementState.tsx
```diff
@@ -0,0 +1,127 @@
...
+      }
+    })
+    return elementsState
+  }
+
+  async function evaluateSingleElement(element: any): Promise<EvaluatedElement> {
+    const evaluationParameters = {
+      objects: [responsesByCode, responsesFullByCode], // TO-DO: Also send user/org objects etc.
+      // graphQLConnection: TO-DO
+    }
```
nmadruga, line 93: This is not in use?

nmadruga (changes requested):
I'm still trying to understand how this works : P

Just a heads up on my first run reviewing this. 
I'll probably use this branch as base for the next issue that I'll be working on - Keep the responses in cache until we send to the server in a set.

I like the idea of passing this objects to `ElementsArea` I'm still not sure why we need the `ElementsFullByCode` but, the elementsState is great. I still see a problem with how we are querying things from the server - not using what is in cache, but like I added to a comment I think this can be treated in a separated issue. I'll add a not in `Issues Backlog`.

documentation/internal/Element-Type-Specs.md
CarlosNZ: Thanks.

documentation/internal/Element-Type-Specs.md
CarlosNZ: Yeah, although it's not implemented "completely" yet, in that it's not handling changes etc.

documentation/internal/Element-Type-Specs.md
CarlosNZ: Let's discuss soon, but will probably be important when we do #90, etc.

src/utils/graphql/queries/getApplication.query.ts
```diff
@@ -31,8 +31,15 @@ export default gql`
         applicationResponses {
           nodes {
             value
+            id
             templateElement {
               code
```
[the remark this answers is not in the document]
CarlosNZ: Done.

src/containers/Application/ElementsArea.tsx
CarlosNZ: Yes, good catch. This is what happens when you use the "Rename Symbol" feature in VSCode. Annoying.

src/utils/hooks/useGetResponsesAndElementState.tsx
CarlosNZ: Yes, I was hoping to use the id, but I knew it wouldn't have finished loading yet. @andreievg  has a couple of ideas of how to make one hook wait for the preceding one (he wrote a comment on it somewhere, but can't find right now).

We should optimise this once we've got some of the other things solidly in place.

src/utils/hooks/useGetResponsesAndElementState.tsx
CarlosNZ: OK

src/utils/hooks/useGetResponsesAndElementState.tsx
CarlosNZ: Thanks.

src/utils/hooks/useGetResponsesAndElementState.tsx
CarlosNZ: What do you mean? What is not in use?

Do you mean the graphQLConnection? No, there is no expressions that use it yet, but we can pass this object in here quite easily when we need it.

src/utils/hooks/useGetResponsesAndElementState.tsx
CarlosNZ: This suggestion doesn't work. `sectionNode` is underlined, with the error `Object is possibly 'null'`.  🤷‍♂️

src/utils/hooks/useGetResponsesAndElementState.tsx
CarlosNZ: Done.

src/utils/hooks/useGetResponsesAndElementState.tsx
CarlosNZ: Nevermind, it works after I apply the suggestion further up. 👍 thx.

nmadruga:
I'm noticing another problem. Not sure if it's related to the changes on this PR.
Steps:
- Go to the Register page
- Click on `User Registration`
- Go to the next back
- Click `Go back` on the browser
- Now the page is not reloading anymore.

The error that is showing in the console:
Screen Shot 2020-11-05 at 4 10 28 PM

src/utils/hooks/useGetResponsesAndElementState.tsx
CarlosNZ: Okay, hope what I've done is okay now:
```
function checkForApplicationErrors(data: GetElementsAndResponsesQuery | undefined) {
  if (
    !data?.applicationBySerial?.applicationResponses?.nodes ||
    !data?.applicationBySerial?.template?.templateSections
  )
    return 'Data undefined'
  if (data?.applicationBySerial?.applicationResponses?.nodes.length === 0)
    return 'No responses found'
  return null
}
```

src/utils/hooks/useGetResponsesAndElementState.tsx
nmadruga: I think this is what you are looking for: https://github.com/openmsupply/application-manager-web-app/compare/example-skip-on-useQuery

By the way - I'm thinking about changing the cache for `Application` to use the serial as the identifier for applications stored in the cache. I've done this already in the branch that I'm currently working - but I'm still seeing calls to the graphql server when this runs so I'll continue to investigate. 🕵️‍♀️

src/utils/hooks/useGetResponsesAndElementState.tsx
nmadruga: Sorry, the `evaluationParameters`.

src/utils/hooks/useGetResponsesAndElementState.tsx
```diff
@@ -0,0 +1,137 @@
...
+
+function checkForApplicationErrors(data: GetElementsAndResponsesQuery | undefined) {
+  if (
+    !data?.applicationBySerial?.applicationResponses?.nodes ||
+    !data?.applicationBySerial?.template?.templateSections
+  )
+    return 'Data undefined'
+  if (data?.applicationBySerial?.applicationResponses?.nodes.length === 0)
+    return 'No responses found'
+  return null
```
nmadruga, line 135:
```suggestion
if (!data?.applicationBySerial) return 'Data undefined'
  const application = data?.applicationBySerial as Application
  if (!application.applicationResponses || !application.template?.templateSections)
    return 'Application missing parameters'
  if (application.template?.templateSections.nodes.length === 0) return 'No sections found'
  if (application.applicationResponses.nodes.length === 0) return 'No responses found'
  application.template?.templateSections.nodes.forEach((section) => {
    const missingElements = section?.templateElementsBySectionId.nodes.some((element) => !element)
    if (missingElements) return 'Application missing elements'
  })
  return null
```

src/utils/hooks/useGetResponsesAndElementState.tsx
```diff
@@ -0,0 +1,137 @@
...
+      if (code) {
+        currentResponses[code] = response?.value?.text
+        currentFullResponses[code] = response?.value
+      }
+    })
+
+    setResponsesByCode(currentResponses)
+    setResponsesFullByCode(currentFullResponses)
+    setElementsExpressions(templateElements)
+    setLoading(false)
```
nmadruga, line 69: I've found what was causing the problem to refresh the page! This `setLoading` should be moved to the next useEffort. I'll add one suggestion next.

src/utils/hooks/useGetResponsesAndElementState.tsx
```diff
@@ -0,0 +1,137 @@
...
+  }, [data, apolloError])
+
+  useEffect(() => {
+    evaluateElementExpressions().then((result) => setElementsState(result))
+  }, [responsesByCode])
```
nmadruga, line 74:
```suggestion
useEffect(() => {
    if (responsesByCode)
      evaluateElementExpressions().then((result) => {
        setElementsState(result)
        setLoading(false)
      })
  }, [responsesByCode])
```

nmadruga (changes requested):
Ok. Second run on reviewing this PR.
I think the error checking should include more cases, suggested a replacement. See what do you think. And also another suggestion to fix the problem with refreshing the page.

src/utils/hooks/useGetResponsesAndElementState.tsx
CarlosNZ: Great! Thanks for picking up on this. :)

src/utils/hooks/useGetResponsesAndElementState.tsx
CarlosNZ: Ah, right, yes -- good catch, they should be passed in to the `evaluateExpression` function.

CarlosNZ:
Have made changes to make sure the page loads properly on refresh.

I've also made small change to the response format so it reads and writes `{value: text}` rather than literal value string. Still need to handle the other response fields, but that'll will come later when we get to plugins that actually use them.

For now, you can see the conditional `isVisible` in action by setting one of the fields visibility condition to something like:

```
{
  "children": [
    {
      "children": [
        {
          "value": {
            "property": "Q1"
          }
        }
      ],
      "operator": "objectProperties"
    },
    {
      "value": "Carl"
    }
  ],
  "operator": "="
}
``` 
(recommending compacting that.)
Now that field will only appear once Q1 has been answered with "Carl".

src/utils/hooks/useGetResponsesAndElementState.tsx
```diff
@@ -0,0 +1,144 @@
...
+        isEditable: element.isEditable,
+        isRequired: element.isRequired,
+        isVisible: element.isVisible,
+        // isValid: element.isValid,
+      }
+    })
+    return elementsState
+  }
+
+  async function evaluateSingleElement(element: any): Promise<EvaluatedElement> {
```
nmadruga, line 100:
```suggestion
  async function evaluateSingleElement(element: TemplateElementState): Promise<EvaluatedElement> {
```

src/utils/hooks/useGetResponsesAndElementState.tsx
nmadruga: Yes, it's sorted now on the PR to-be-created. I fixed by using the `serial` as the primary key on the GraphQL queries (no changes on the back-end). And also changing the order and fields queried in each query.

src/utils/hooks/useGetResponsesAndElementState.tsx
nmadruga: I'm also thinking how we could combine these three types: `TemplateElementState`, `EvaluatedElement` and (a new one I have created to be used as following (for other PRs):
```
interface ElementState {
  id: number
  title: string
  elementTypePluginCode: string
  section: number
  category: TemplateElementCategory
  isRequired: boolean
  isVisible: boolean
  isValid: boolean
  // validationMessage: string | null
  isEditable: boolean
}

interface ApplicationElementState {
  [key: string]: ElementState
}
```

Since these 3 have so much in common it would be good to have some intersection between types. Can you think on something? I hasn't spent to much time on it, just think it was kind of annoying to have to add fields in these 3 while I was extending this to use somewhere else that required more fields...

nmadruga (commented):
Hey, I have had another go on this one. Hopefully we can finish this one up tomorrow, my last ones are just about the types. If we can merge them somehow. Or have on use to other, or create some in between are that both use... Just because it was annoying to have to enter fields in many `interface` while I was trying to expand this work in another branch.

src/utils/hooks/useGetResponsesAndElementState.tsx
CarlosNZ: Okay, have done the Typing on the suggestion.

I'm not sure how to combine these, maybe I'll wait to see your PR and then we'll consider all three of them in its own issue?

nmadruga: Yes, all right. I think it would be better to merge this one already. And then we will tidy up these types altogether in the other PR. I was trying to avoid to many changes to the other one, but it will be alright to wait until then to change this.

nmadruga (approved):

PR #84, opened 2020-11-04
#46 persist cache

### LINKED ISSUE #46: Persist cache

nmadruga:
#Fixes #46 

Requires to run `yarn`
Then `yarn dev` to run :)

### Changes
- Rename `App.tsx` to `AppWrapper.tsx` and moves to `containers/Main` folder.
- Create `index.ts` in `containers/Main` folder to export all files in this folder. 
- Includes new package `apollo3-cache-persist`
- Start the local storage to keep cache on window.localStorage.

### Local storage checking
[Check this guide](https://developers.google.com/web/tools/chrome-devtools/storage/localstorage) on how see your local storage. 
And I'm also using the [Apollo client dev tools on Chrome](https://chrome.google.com/webstore/detail/apollo-client-developer-t/jdkknkkbebbapilgoeccciglkfbmbnfm)

See how the cached data is stored in local storage:
* Use Screen Shot 2020-11-04 at 5 11 40 PM to clear the local storage that will remove the apollo-cache-persist object. => Although it will get the users again since the App will query that!

Screen Shot 2020-11-04 at 4 58 35 PM

And here is how it looks on the Apollo dev tools
Screen Shot 2020-11-04 at 4 58 52 PM

Then reloading the page just calls the GraphQL server to get these 2 things: Introspective query and OPTIONS:
Screen Shot 2020-11-04 at 5 00 13 PM
Screen Shot 2020-11-04 at 5 00 24 PM


src/index.tsx
```diff
@@ -1,20 +1,52 @@
...
+    persistCache({
+      cache,
+      storage: window.localStorage,
+      /**
+       * Storage options.
+       */
+      // Maximum size of cache to persist (in bytes).
+      // Defaults to 1048576 (1 MB). For unlimited cache size, provide false.
+      // If exceeded, persistence will pause and app will start up cold on next launch.
+      //maxSize: false,
```
andreievg, line 28: hmm, maybe we should change this to false for now with comment to change to appropriate size. Just to avoid confusion if app cold restarts during dev

andreievg (approved):

src/index.tsx
nmadruga: Yes, aggreed.

Issue #98, opened 2020-11-06
Epic #16 - List and Filter applications


nmadruga:
## Description

Any logged user can see some applications.
There are options on the left nav bar to select the Application Type to see. After the user clicks on the Application type a list (or cards) is displayed.
There is an option for the user to change from list to cards. After the user picks one option this is persisted on the user's settings.
The list of applications  is associated to:
Current user's permissions (JWT)
Toggled view mode (Apply/Review)
Current Application Type selected
Current filters

### Documentation:
- [x] #225 Documenting List applications headers and functionality - Effort 2
- [ ] TODO: Add as part of [Epic #38](338) Public docs & tutorials 

### Sprint 8:
- [x] #66 Show status/stage of application on the applications list: basically draft and submitted - Effort 1
- [x] [Back-end#99](https://github.com/openmsupply/application-manager-server/issues/99) Show status and stage as "Pending" when trigger is "Processing" - (AVG Effort 2)
- [x] #104 Display loader for application with  application trigger is processing - Effort 3

### Sprint 16:
- [x] #131 Header columns for Application list - Effort 2
- [x] #207 List applications with specific cells on each column  - Effort 3

### Sprint 17:
- [x] #251 Filter application using URL parsed filters - Effort 4
- [x] [Back-end#185](https://github.com/openmsupply/application-manager-server/issues/185) Application VIEW with all relevant columns for list - Effort 3
- [x] #226 Update URL for lists to handle redirects and documentation - Effort 1

### Sprint 20:
- [x] #283 Bug: Too many render calls to ListWrapper when changing the template types - Effort 2
- [x] #210 Use pagination to list applications - Effort 3
- [x] #208 Search applications - Effort 2
- [X] #209 Sort columns - Effort 2

### Sprint 21:
- [x] #329 Application List -- reload only table contents when changing filters #329 - Effort 2

### Sprint 22:
- [x] #211 Expand application sections in list - Effort 2

### Backlog:
- [ ] #363 Show sections of Application in expanded area on the list - Effort 1 - **Might not get implemented**
- ~~#212 Progress logic on Application/Section in the list~~ Duplicated

Issue #67, opened 2020-10-27
Go to some page in application when click on an application with Draft status


nmadruga:

CarlosNZ:
Just to clarify -- shouldn't ALL applications link to their appropriate application page, regardless of status? The page itself will determine whether its editable (status = "Draft", go to Page 1), or not (status = "Completed", etc., go to Summary page)

PR #107, opened 2020-11-09
#66 list applications stage and status

### LINKED ISSUE #66: Show status/stage of application on the applications list: basically draft and submitted

nmadruga:
Fixes #66 

## Changes
- Create 2 fragments `Template` and `Application` that can be used as part of others queries
- Edit `getApplications.query` to use the new fragments
- Create new hook `useListApplications` to flatten properties to be display in the application list
- Removes the logic (and call to hook) from the `ApplicationSummary` component


src/utils/hooks/useListApplications.tsx
```diff
@@ -0,0 +1,62 @@
...
+  const [applications, setApplications] = useState<ApplicationDetails[] | undefined>()
+  const [loading, setLoading] = useState(true)
+  const [error, setError] = useState('')
+
+  const { data, loading: apolloLoading, error: apolloError } = useGetApplicationsQuery()
+
+  useEffect(() => {
+    if (apolloError) return
+    if (apolloLoading) return
+    // Check that only one tempalte matched
```
CarlosNZ, line 20:
```suggestion
   // Check that only one template matched
```
Who cares about spelling in comments, right? ;)

src/containers/Application/ApplicationList.tsx
```diff
@@ -0,0 +1,49 @@
...
+          <List.Item key={`app_selected_parameter_${value}`} content={key + ' : ' + value} />
+        ))}
+      </List>
+      {applications && applications.length > 0 && (
+        // TODO: Implement sortable columns -> Requires reducer
+        <Table sortable stackable selectable>
+          <Table.Header>
+            <Table.Row>
+              {Object.entries(applications[0]).map(([key]) => (
+                <Table.HeaderCell key={`app_header_${key}`}>{key}</Table.HeaderCell>
```
CarlosNZ, line 30: I'm getting this error in the console:
```
Warning: Encountered two children with the same key, `app_2_null`. 
```
Is there a problem getting `null` values for some of these keys?

CarlosNZ: I think I know what's happening. The default data has no stage or status for the "Company Registration", so these cells are getting `null` as part of their key.

Maybe we should assume that any application with no stage or status is erroneous, and therefore we shouldn't show it in this list?

CarlosNZ (changes requested):
Cool. So you're not implementing the Filters (by URL queries) in this PR? Do we have an issue for that yet? Probably should as basic filtering seems like a fairly high priority.

CarlosNZ:
One another quick comment -- it's still using the cache even after a full page reload. We talked about adding a 'refetchQueries' to this page every reload, but was that meant to be in this issue, or another one?

src/containers/Application/ApplicationList.tsx
```diff
@@ -0,0 +1,49 @@
...
+              ))}
+            </Table.Row>
+          </Table.Header>
+          <Table.Body>
+            {applications.map((application, index) => (
+              <Table.Row key={application.serial}>
+                {Object.values(application).map((value) => (
+                  <Table.Cell key={`app_${index}_${value}`}>{value}</Table.Cell>
```
CarlosNZ, line 38: Small thing: maybe the key should be `app_${index}_${Object.keys(application)[index]}` so it'll be `app_0_status` rather than `app_0_DRAFT` -- seems a bit more descriptive and it'll solve the above problem of having `null` values.

I know it's a bit ugly having an additional `Object.keys`, but I'm about to use it here in #67 anyway, so why not assign the current key to a variable at the start of the per-cell loop, and then we can use it freely wherever within the row?

src/containers/Application/ApplicationList.tsx
```diff
@@ -0,0 +1,49 @@
...
+                ))}
+              </Table.Row>
```
CarlosNZ, line 40:
```suggestion
              <Table.Row key={application.serial}>
                {Object.entries(application).map(([key, value]) => (
                  <Table.Cell key={`app_${index}_${key}`}>{value}</Table.Cell>
                ))}
              </Table.Row>
```
I think this would be better, as we need to use the `key` for #67 anyway.

nmadruga:
> So you're not implementing the Filters (by URL queries) in this PR? 

No, I don't think it is since this is just a `Effort 1` issue.

> Do we have an issue for that yet? Probably should as basic filtering seems like a fairly high priority.

Oh! I though I had listed as one issue to be added but hasn't before. Now I added in here so we can decide on the next meeting, give an effort and work on the following sprint: https://github.com/openmsupply/application-manager-web-app/issues/98

src/utils/graphql/fragments/application.fragment.ts
```diff
@@ -0,0 +1,11 @@
+import { gql } from '@apollo/client'
+
+export default gql`
+  fragment Application on Application {
+    serial
```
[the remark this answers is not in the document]
nmadruga: No need - now we are using the `serial` as the primary key for GraphQL queries. Unless we need to include the id to be send in some other mutation (I'm not sure if we would need on the creation of the application). For now it's not needed.

src/utils/hooks/useListApplications.tsx
nmadruga: : ) I do! Everytime someone reads it would have been annoying - and you end up adding as part of another PR very unrelated. Thanks!

src/containers/Application/ApplicationList.tsx
nmadruga: Ah. Yep, that's a good solution. I'll work on it now.

nmadruga:
> We talked about adding a 'refetchQueries' to this page every reload, but was that meant to be in this issue, or another one?

It should be on this issue. Well reminded. I was working on the mSupply side in the morning - so I'll work on this now and add this as part of this PR.

src/containers/Application/ApplicationList.tsx
CarlosNZ: Hmm, sorry to walk back on this now, but I think the [solution below](https://github.com/openmsupply/application-manager-web-app/pull/107#discussion_r520173067) is probably better. That way we can still list applications with no stage or status, without causing `null` keys.

src/containers/Application/ApplicationList.tsx
CarlosNZ: Ignore this suggestion, [this one](https://github.com/openmsupply/application-manager-web-app/pull/107#discussion_r520173067) is better, sorry.

nmadruga: Yes. Indeed. Thanks for this suggestion!

nmadruga: Well, I think I'll be combining with your next suggestion to make it not have to do this Object.keys inside the call

andreievg (approved):

andreievg:
Approved, pending changes request by Carl

andreievg:
@CarlosNZ 
```
One another quick comment -- it's still using the cache even after a full page reload. We talked about adding a 'refetchQueries' to this page every reload, but was that meant to be in this issue, or another one?
```
I would have thought that network only policy for the query should reload it on page reload

nmadruga:
> I would have thought that network only policy for the query should reload it on page reload

Issue #104, opened 2020-11-06
Display loader for application with  application trigger is processing


nmadruga:
Redirect to the correct page when ready - either on view or create.

* TODO: link issue In back-end if a computed column could show that the application is **pending** the front-end could simply call one query to check if the application has any trigger with some process in progress aka trigger is **processing**.

Issue #103, opened 2020-11-06
Redirection to the summary in case the application is submitted


nmadruga:

CarlosNZ:
Another re-direction this needs is for the application base URL:
i.e. if we try and go to `/applications/689/`:
 - If "Draft", redirect to `/applications/689/S1/page1` (or maybe the last page edited)
 - If "Completed" (etc.), redirect to `/applications/689/summary`

CarlosNZ:
This seems like a duplicate of #68, isn't it, @nmadruga ? Shall we close?

Issue #96, opened 2020-11-06
Epic #17 - Create organisation


nmadruga:
## Description

Create template for users to be able to create a new organisation in on the system using the Application management workflow. After the template goes through the process of being approved by FDA authorities it gets created in the database and the user who applied is granted owner permission in the organisation to be able to allow other user to join or not the organisation. After the application is approved the use should now have the option to login associated with this organisation and apply for application (Drug registration) as the company as long receive new application to join this organisation.

## Documentation
- [ ] TODO: Add as part of [Epic #38](338) Public docs & tutorials

### Sprint 10:
- [x] [Back-end#97](https://github.com/openmsupply/application-manager-server/issues/97) Create template action to change status of 'Create company' when submit application - Effort 1

### Sprint 18:
- [x] [Back-end#114](https://github.com/openmsupply/application-manager-server/issues/114) Create Company Action 
- [x] #297 updateApplication query should include user and org id - Effort 1
- [x] [Back-end#113](https://github.com/openmsupply/application-manager-server/issues/113) Create action to join username to company when application to 'Create company' is approved - Effort 2
- [x] [Back-end#104](https://github.com/openmsupply/application-manager-server/issues/104) Extend action to grant permission for user link to a company - Effort 2

### Remaining issues to get this all up to MVP:
- [x] "ApplicationData" for front-end evaluator -- put into FullStructure: Issue #458, PR #557
- [x] LongText element plugin: Issue #479 PR #555
- [x] DisplayImage element plugin: Issue #549 PR #556
- [x] "Other" option for RadioButton: Issue #488 PR #554

### Sprint 28:
- [x] Final tidy-ups for Create and Join company #583 **PR #602**
- [x] Rename permission and template names to use org instead of company #308

### Backlog:
- [ ] Template element -- Address autocomplete #552

Issue #68, opened 2020-10-27
Display application View page when click an application with Submitted status


nmadruga:
Not to do the collapsable sections yet!
Just show all the sections. Use summary page and hide buttons and edit options.

CarlosNZ:
One thing to probably squeeze into this issue (or make a new one) -- the Summary page should also respect the `isVisible` status. I notice it's currently showing ALL questions, even the ones that were hidden on the actual form.

nmadruga:
Still missing a few things on this one to be completed:
- Summary page (after application is submitted) shouldn't show the `Submit` button not the `edit` sections links.
- One thing to probably squeeze into this issue (or make a new one) -- the Summary page should also respect the isVisible status. I notice it's currently showing ALL questions, even the ones that were hidden on the actual form.

PR #108, opened 2020-11-10
#67 Links from Application list and Re-directs

CarlosNZ:
This should address #67 and #103 (and maybe #68?), although there might be more to do if I've misunderstood the issue requirements.

Please test with back-end branch `#96-more-data-for-trigger-payload` (in PR #98), as this includes a fix for the stage/status computed columns,  which you'll need to see updated statuses in the Application list.

**Changes**:

- Each application in Application List is click-able (on Title field) and goes to `/applications/{serialNo}`
- `useLoadApplication` query also returns current stage and status
- Implemented `isReady` boolean to wait for `useLoadApplication` (and skip that query subsequently) and perform some re-direct logic after application is loaded. (We discussed this way of handling waiting for initial Hooks -- hopefully it's implmented as envisaged)
- Some simple re-direct logic, with the idea that we can add more complexity in this method (`processRedirect`) as we go along. Currently it does:
   - If application status is NOT "DRAFT", re-direct to `/applications/{serialNo}/summary`
   - If no sectionCode or Page (i.e. URL is just `/applications/{serialNo}` with no query params), re-direct to Section 1, Page 1

**Note:**
 - there is some errors when loading the summary page of the pre-loaded applications but newly created ones work fine. I assume this is due to missing/inconsistent data in the dummy applications -- will investigate at some point.
 - because #104 is not done yet, newly created applications might show up as "Draft" in the list and when loading the application, as the back-end has changed status to "Completed" but front-end is using cached data. In order to make the application page know that status has changed, I had to manually delete my Local Storage cache and refresh. This should be addressed in a later issue

### DIFF
--- a/src/containers/Application/ApplicationPageWrapper.tsx
+++ b/src/containers/Application/ApplicationPageWrapper.tsx
@@ -20,8 +20,7 @@ const ApplicationPageWrapper: React.FC = () => {
   })
 
   if (isReady) {
-    console.log(appStatus)
-    processRedirect({ ...appStatus, serialNumber, sectionCode, page, push })
+    processRedirect({ ...appStatus, serialNumber, sectionCode, page, templateSections, push })
   }
 
   const {
@@ -190,9 +189,24 @@ function nextPageButtonHandler({
 
 function processRedirect(appState: any): void {
   // All logic for re-directing/configuring page based on application state, permissions, roles, etc. should go here.
-  console.log('State', appState)
-  const { stage, status, outcome, serialNumber, sectionCode, page, push } = appState
-  if (status !== 'DRAFT') push('summary')
+  const {
+    stage,
+    status,
+    outcome,
+    serialNumber,
+    sectionCode,
+    page,
+    templateSections,
+    push,
+  } = appState
+  if (status !== 'DRAFT') {
+    push(`/applications/${serialNumber}/summary`)
+    return
+  }
+  if (!sectionCode || !page) {
+    const firstSection = templateSections[0].code
+    push(`/applications/${serialNumber}/${firstSection}/page1`)
+  }
 }
 
 export default ApplicationPageWrapper
--- a/src/containers/Application/ApplicationPageWrapper.tsx
+++ b/src/containers/Application/ApplicationPageWrapper.tsx
@@ -11,7 +11,7 @@ import { useApplicationState } from '../../contexts/ApplicationState'
 const ApplicationPageWrapper: React.FC = () => {
   const { setApplicationState } = useApplicationState()
   const [isReady, setIsReady] = useState(false)
-  const { query, push } = useRouter()
+  const { query, push, replace } = useRouter()
   const { mode, serialNumber, sectionCode, page } = query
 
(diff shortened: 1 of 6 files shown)


src/containers/Application/ApplicationPageWrapper.tsx
```diff
@@ -179,4 +187,26 @@ function nextPageButtonHandler({
...
+    status,
+    outcome,
+    serialNumber,
+    sectionCode,
+    page,
+    templateSections,
+    push,
+  } = appState
+  if (status !== 'DRAFT') {
+    push(`/applications/${serialNumber}/summary`)
```
andreievg, line 203: should this be 'replace' ?

src/utils/hooks/useLoadApplication.tsx
```diff
@@ -33,6 +36,12 @@ const useLoadApplication = (props: useLoadApplicationProps) => {
 
       const sections = getApplicationSections(application.applicationSections)
       setSections(sections)
+
+      setAppStatus({
+        stage: application?.stage,
+        status: application?.status,
+        outcome: application?.outcome,
+      })
```
andreievg, line 44: I was originally thinking that isLoading (isReady), would be a hook inside useLoadApplication. (rather then in ApplicationPage). What do you think ?

andreievg (changes requested):
A couple of minor comments

src/containers/Application/ApplicationPageWrapper.tsx
CarlosNZ: Yeah, good call.

src/utils/hooks/useLoadApplication.tsx
CarlosNZ: So the same logic that I have here is in `useLoadApplication`, and ApplicationPage just gets its state from the returned `loading` value? Is that what you mean?

CarlosNZ: Okay, have moved `isReady` into `useLoadApplication` (so it's only responsible for setting `skip` in that hook) and moved the `processRedirect` inside the `useEffect` in the ApplicationPage.

0d3b95b

Is that what you had in mind?

andreievg: Yp, but I was also thinking that it will be returned from useLoadApplication

src/containers/Application/ApplicationPageWrapper.tsx
andreievg: looks good, I think there might be some artefact `push` being passed along and imported

src/utils/hooks/useLoadApplication.tsx
CarlosNZ: Was going to do that, but figured it could be determined by `application` and `loading`, no?

andreievg: yep, that's good. I was just wondering for use cases further down the track. I guess can always determine if application has been loaded by checking loading or application. No need to change.

andreievg (approved):

PR #109, opened 2020-11-11
#104 Application Loading -- wait for trigger

### LINKED ISSUE #104: Display loader for application with  application trigger is processing

CarlosNZ:
Fixes #104 -- waits for all triggers associated with current application to finish before loading the application.

Requires back-end branch from [this PR](https://github.com/openmsupply/application-manager-server/pull/100) in order to work.

Basically, created a new hook `useTriggerProcessing`, which performs the query `getTriggers` to get required trigger field associated with this application. The hook returns `true` until it has confirmed the requested trigger returns `null` (and then it returns `false`, i.e. trigger is NOT currently processing)

The `useLoadApplication` hook calls this hook, and won't be able to execute the `useGetApplicationQuery` (via `skip`) until `triggerProcessing` returns false.

Have combined the `loading` and `error` values for **both** hooks into one `loading` and `error` value each, so the calling component ("ApplicationPageWrapper") has a single `loading` value regardless of which hook it's waiting for.

Testing is a bit tricky, because the triggers get set back to `null` almost instantaneously, so you can't really see the effect. However, as an initial test, you can start a new application (register) and it should load and continue just fine, albeit with a slightly longer loading time than before.

To test the waiting-for-trigger properly, just quit the Node server, but keep the PostGraphile server running (`yarn pg`) -- that way the triggers won't be set back to `null` immediately.

Then load an application page, note the serial id, then in PostGres, set the `trigger` field for this application to a valid trigger (e.g. `onApplicationCreate`). The trigger value will now be stuck in the state "Processing".

Re-load the application page in the browser, and note the "Loading" screen remains up indefinitely. While still "loading", set the trigger back to `null` in Postgres, and the application should then load properly. _Voila!_


src/utils/graphql/queries/getTriggers.query.ts
```diff
@@ -0,0 +1,15 @@
+import { gql } from '@apollo/client'
+
+export default gql`
+  query getTriggers($serial: String!) {
+    applicationTriggerStates(condition: { serial: $serial }) {
```
nmadruga, line 5: Maybe use the ApplicationBySerial here and just return the required fields.

src/utils/hooks/useTriggerProcessing.tsx
```diff
@@ -0,0 +1,28 @@
...
+    variables: {
+      serial: serialNumber,
+    },
+    pollInterval: 500,
+    skip: !isProcessing,
+    fetchPolicy: 'no-cache',
+  })
+
+  useEffect(() => {
+    if (data?.applicationTriggerStates?.nodes[0]) {
```
nmadruga, line 20: Probably worth checking for unexpected data before doing this!

nmadruga (changes requested):
Just 2 comments. Nice and simple : )

I'll be testing this now and see if it helps with problems we found on PR #110

src/utils/graphql/queries/getTriggers.query.ts
CarlosNZ: Not sure what you mean -- this is querying a seperate View table that collates ALL trigger fields.

nmadruga (approved):

nmadruga:
Approving this after our discussion. Just needs to update from master before this can be merged - since there are conflicts!

src/utils/hooks/useTriggerProcessing.tsx
CarlosNZ: Okay, have added an additional `triggerError` field, which is combined into the apollo error field on return. 

You can test it in action by temporarily removing the `serial` field from the `getTriggers.query.ts`. :)

Issue #45, opened 2020-10-18
Mutate application from cache.


nmadruga:
Need to decide when

nmadruga:
Based on what I was researching on issue #37 storing the responses in the cache to be sent to the server only on a particular event isn't a common design pattern while using Apollo client. It seems better to continue sending the responses to the server as soon as the response is changed - and do other things to manage how the UI performances better for low latency users. The way to do this (from what I have found so far) is using:
- [Optimistic UI](https://www.apollographql.com/docs/react/performance/optimistic-ui/) to render a pseudo result from the server immediately after sending it to the server - than waiting for the server response to render the component.
- [HTTP Batch Link](https://www.apollographql.com/docs/react/api/link/apollo-link-batch-http/) to batch queries and mutations to be send to the server in a batch (in internals) - this is done for the entire App from what I understood. So, we can postpone adding this to make sure that's what we want - and it would work well for other queries (ex: load application - we don't want to wait for to long to get this result from the server) and decide if that's what we want.

andreievg:
I am ok with just doing mutation for now. 

With Optimistic UI, it’s not guaranteed that server receives the data, so we would need to send the whole application on submit, is that right ?

HTTP Batch Link can be the way to go, but not sure if we can have full control of the flow (you've mentioned that already)

I was soo convinced that this workflow of storing to cache and then updating from cache was in built to Apollo, as it seems like the main use case for such a tool.

Btw graphile does allow the following:

```
mutation MyMutation {
  mutation1: updateApplicationResponse(input: {patch: {value: "yow"}, id: 1}) {
    clientMutationId
  }
 mutation2: updateApplicationResponse(input: {patch: {value: "sup"}, id: 2}) {
    clientMutationId
  
  }
}
```

andreievg:
I've mentioned a couple of time that we shouldn't do dynamic queries, now I think it might be the best option (also would make creating application logic nicer)

nmadruga:
> With Optimistic UI, it’s not guaranteed that server receives the data, so we would need to send the whole application on submit, is that right ?

That's a good idea. I'll work on that in this PR. I'm not sure yet how Optimistic UI would work, but I think it might give some error if the response wasn't send to the server. Anyway I like the idea of sending the full application at the end.

nmadruga:
> I've mentioned a couple of time that we shouldn't do dynamic queries, now I think it might be the best option (also would make creating application logic nicer)

[...]

PR #110, opened 2020-11-11
Use Serial as identifier for Application + Tidying up

nmadruga:
While I was working on issue #45 and merging many PRs into it, I found out a few improvements to be added. So break down in this PR with more tidying up changes - and will base this one for the #45 related changes. : )

## Changes
- Use `watch` flag on script `yarn generate` to make it continue checking for changes on graphql queries/mutation/fragments to update the generated ` graphql file.
- Update `cache.ts` file using the `serial` as the identifier key in the Application table (for cache management).
- Add fragments for each entity and use in existing queries/mutations (Attention: This may cause different behaiviour in exisiting functionalities. Please revisit all existing ones and let me know if any problem is happening).
- Update `getApplication` and `updateApplication` query/mutation to use the `serial` to identify the entity.
- Update `ApplicationSummary` to receive from `ApplicationOverview` elements using new hook `getResponsesAndElements` instead of specific one `useLoadSummary` - which is now removed.

## Existing problems:
- When creating the application the elements aren't loaded - I think it is because we haven't got the `skip` on the query added to master yet. Will need to investigate further.
- Responses aren't showing correctly on the application:
Screen Shot 2020-11-12 at 11 01 45 AM

- ~~The progress bar is not showing the pages here. Will need to investigate too.~~ 
  - Problem found - missing `index` in fragment.


src/cache.ts
```diff
@@ -3,6 +3,9 @@ import { TemplateElementsConnection } from './utils/generated/graphql'
 
 const cache = new InMemoryCache({
     typePolicies: {
+      Application: {
+        keyFields: ['serial']
```
andreievg, line 7: who does this need to be done ?

nmadruga: This was done while I was trying to fix what was causing the cache management to not work properly. The problem happening was that after an application was fetched (onLoadApplication), regardless being present in the cache, the App was trying to fetch from the server again for the same application because we are using the `serial` to find it. The cache doesn't have a way to check for other fields in an entity that is already in the cache just by its identifier (aka `keyField`). Since we are using the `serial` to query for applications at all places it makes more sense to use it as the identifier in the cache. And now the Apollo cache is correctly getting the application from cache the 2nd time.

andreievg:
@nmadruga 
> This was done while I was trying to fix what was causing the cache management to not work properly. The problem happening was that after an application was fetched (onLoadApplication), regardless being present in the cache, the App was trying to fetch from the server again for the same application because we are using the serial to find it. The cache doesn't have a way to check for other fields in an entity that is already in the cache just by its identifier (aka keyField). Since we are using the serial to query for applications at all places it makes more sense to use it as the identifier in the cache. And now the Apollo cache is correctly getting the application from cache the 2nd time.

I saw that ID was added to application fragment as one of the later commits, could it have been that application wasn't stored and retrieved from cache because ID was missing ? I would have thought that Apollo Client would allow queries from cache by other fields (rather then just the one that was queried originally)

src/components/Application/ApplicationSummary.tsx
```diff
@@ -79,4 +73,19 @@ const ApplicationSummary: React.FC<ApplicationSummaryProps> = ({
...
+  return question.category === TemplateElementCategory.Question ? (
+    <Form.Field>
+      <Header as="h3" content={question.title} />
+      <Input disabled error={!response}>
+        {response?.text}
+      </Input>
+    </Form.Field>
+  ) : question.elementTypePluginCode !== 'pageBreak' ? (
+    <Form.Field>
+      <Header as="h5" content={question.title} />
```
andreievg, line 86: was wondering what this type of element would be ?

src/components/Application/ApplicationSummary.tsx
```diff
@@ -1,57 +1,49 @@
...
-      </Form.Field>
-    ) : null
-  }
+  const sectionsAndElements = Object.entries(elementsState).reduce(
+    (sectionGroups: SectionGroups, [code, item]) => {
+      const group = sectionGroups[item.section] || {}
+      group[code] = item
+      sectionGroups[item.section] = group
+      return sectionGroups
+    },
```
andreievg, line 35: I think if we are doing preliminary reshaping like this, it would be great to put all of the required info in the new object (so we do the minimum amount of logic in jsx)

Also it maybe easier to deal with arrays (rather than objects) in jsx.

i.e. if sectionsAndElements was:

```
[
 {
  section: {full section info}
  elements: [ 
      {
          element: {code etc}
          value: {  }
       }
   ]
 }
]
```

It's also easier to sort arrays, by the way are the elements sorted by index within the section ?

src/utils/hooks/useGetResponsesAndElementState.tsx
```diff
@@ -108,8 +102,11 @@ const useGetResponsesAndElementState = (props: { serialNumber: string }) => {
     // const isValid = evaluateExpression(element.validation, evaluationParameters)
     const results = await Promise.all([isEditable, isRequired, isVisible])
     const evaluatedElement = {
-      code: element.code,
       id: element.id,
+      code: element.code,
+      title: element.title,
+      elementTypePluginCode: element.elementTypePluginCode,
+      section: element.section as number,
```
andreievg, line 109: its not really part of this PR, but was wondering if we should put the response value here too, and was wondering why we would need 3 types  of return rather then just elementsState, in this hook.

andreievg (changes requested):
Mainly it's looking pretty good, a few comments re application summary page

src/containers/Application/ApplicationOverview.tsx
```diff
@@ -10,23 +11,37 @@ const ApplicationOverview: React.FC = () => {
...
+    serialNumber: serialNumber as string,
+  })
+
+  const {
+    error: responsesError,
+    loading: responsesLoading,
+    responsesByCode,
+    responsesFullByCode,
+    elementsState,
+  } = useGetResponsesAndElementState({
```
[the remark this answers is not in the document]
CarlosNZ: I already did this in my one: #112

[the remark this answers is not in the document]
nmadruga: Cool. So should be sorted when we merge those together. I might merge #112 to this branch to sort out conflicts.

src/components/Application/ApplicationSummary.tsx
nmadruga: Not sure if I understood your question. Are you asking about the `elementTypePluginCode`?

src/containers/Application/ApplicationOverview.tsx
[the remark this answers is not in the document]
CarlosNZ: Yeah. #112 should be ready to merge if you're happy -- have addressed your review comments.

src/components/Application/ApplicationSummary.tsx
nmadruga: Doing this now!

src/containers/Application/ApplicationOverview.tsx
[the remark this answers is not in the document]
nmadruga: Ok! I have merged #112 with this PR. I noticed a problem with using this `isReady` now, but I'll put in another issue so we can merge this issues soon!

src/components/Application/ApplicationSummary.tsx
nmadruga: Have added this suggestion. Please have a look see if you like. I have moved to use arrays as you suggested. And I'm using a sort on the sections by index to keep in the correct order.

src/utils/hooks/useGetResponsesAndElementState.tsx
nmadruga: Not sure Andrei, I think we need to add this improvement on another issue really. I'm also not sure if we need all 3 objects, but so far they have been in use, so let's try to tidy up more as we go.

src/components/Application/ApplicationSummary.tsx
andreievg: The template element that's not a question or page break is just displayed as h5 header, was wondering if it needs to be there

src/utils/hooks/useGetResponsesAndElementState.tsx
andreievg: cools

src/components/Application/ApplicationSummary.tsx
andreievg: looks great thanks !

src/containers/Application/ApplicationOverview.tsx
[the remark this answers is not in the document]
andreievg: cool, I realised that we can use loading response from LoadApplication for this (but still there seems to be an improvement to be done in useLoadApplications to only load it once)

andreievg (approved):
Looks good, thanks Nicole.

Still there is a comment regarding keyFields 'serial' that i've made, be good if you can check it out sometime
 https://github.com/openmsupply/application-manager-web-app/pull/110#issuecomment-727645790

src/components/Application/ApplicationSummary.tsx
nmadruga: Oh, that was just to replace what will be the `SumaryView` of other elements types in the future. While its not implemented in the plugins its showing like that.

PR #111, opened 2020-11-12
#45 mutate single block responses and application

### LINKED ISSUE #45: Mutate application from cache.

nmadruga:
Fixes #45 

Based on PR #110, where some tidying up was break down from this PR. 

Finally! The original idea was to only send the response when the application was submitted. But, since Apollo client allow us to write to both server and cache `onUpdate` of each field, we decided it would be better to send the response straight to the server, but if the server isn't reachable - it would be stored in cache and then send to the server when the user submits the application.

## Changes
-  New hook `useUpdateApplication` to update the application - it first get all responses related to application in cache and add to the payload for the updateApplication mutation.
- New query `getResponsesInApplication` to gather the responses from one application by its `serial` field
- Update `ApplicationOverview` to show a modal (not so pretty tough) when the application is being submitted.


CarlosNZ:
How should we test this? New applications are just going straight to the Summary page, or crashing. How are you using it?

Maybe rebase from #114?

nmadruga:
Hi @CarlosNZ I'll do that. And let you now when it's ready.

nmadruga:
Done! I have merged PR #110 which has latest changes from the branch you suggested. It seems to be working fine for me. Please @CarlosNZ have another look and let me know if you keep finding problems.

src/utils/graphql/mutations/updateApplication.mutation.ts
```diff
@@ -1,8 +1,19 @@
 import { gql } from '@apollo/client'
 
 export default gql`
-  mutation updateApplication($serial: String!, $applicationTrigger: Trigger!) {
-    updateApplicationBySerial(input: { serial: $serial, patch: { trigger: $applicationTrigger } }) {
+  mutation updateApplication($serial: String!, $applicationTrigger: Trigger = ON_APPLICATION_SUBMIT,
+  $responses: [ApplicationResponseOnApplicationResponseForApplicationResponseApplicationIdFkeyUsingApplicationResponsePkeyUpdate!]) {
```
CarlosNZ, line 5: Whoa! 😆

CarlosNZ (changes requested):
Looks good, but there is a nasty bug when filling in the form.

Whenever I type a keystroke in any field, the page reloads, and the field loses focus.

Somehow the page is reloading on every keystroke.

I think something has broken when my bug-fixes branch and your branch were merged. For example, in the latest version of `bug-fixes-12-11`, I had this line in useLoadApplication:

`skip: triggerProcessing || isApplicationLoaded,`

However, this branch currently has:

`skip: triggerProcessing,`, which was how my branch was a few days ago before the latest fixes, and a few other later fixes I did. Any chance you might have merged an older version of my branch into your other PR?

I think it can be fixed by re-merging the _latest_ commit of `bug-fixes-12-11` into this again. Hopefully. :)

CarlosNZ:
Okay, have tested it locally and I think this will sort it:

- Merge the *latest* version of `bug-fixes-12-11` into this branch
- In ApplicationOverview.tsx, change `isReady` to `isApplicationLoaded`
- Now there are a couple of type errors left in ApplicationOverview that you'll need to fix

That seems to get my bug fixes back in working order with the latest changes.

Can we please figure out how this happened so we can avoid it in the future? 😕

nmadruga:
Hi @CarlosNZ I'm so sorry this have happened. 
I think the problem was during conflict resolution, because we were merging one branch into the other and then first one got new changes into the master giving so many conflicts to be solved (in the branches that got a latest version merged). Very difficult to solve, I think the best thing is to only merge one other branch into the next one - which didn't happen when we started merging the "bug fixes" branch into the others... If you are working with too many branches that intersect - fixes got done in another before the first one is merged and then we need to get the second one in the first one and so on. It's hard. So, the best solution to me is - merge branch as soon as possible and then add new PRs to fix the bugs, that's easier to follow.

CarlosNZ:
> Very difficult to solve, I think the best thing is to only merge one other branch into the next one - which didn't happen when we started merging the "bug fixes" branch into the others...

Yes, we should have merged bug-fixes to master before merging into other branches.

src/utils/graphql/mutations/createApplication.mutation.ts
```diff
@@ -1,16 +1,17 @@
 import { gql } from '@apollo/client'
 
 export default gql`
-  mutation createApplication($name: String!, $serial: String!, $templateId: Int!) {
+  mutation createApplication($name: String!, $serial: String!, $templateId: Int!,
+  $outcome: ApplicationOutcome = PENDING, $trigger: Trigger = ON_APPLICATION_CREATE ) {
```
CarlosNZ, line 5: Are these default values? That's cool, good to know we can do this.

src/utils/graphql/queries/getResponsesInApplication.query.ts
```diff
@@ -0,0 +1,15 @@
+import { gql } from '@apollo/client'
```
CarlosNZ, line 1: Is is necessary to make a new query for this? It seems very similar to `getElementsAndResponses`. Can you not just use that?

src/utils/hooks/useUpdateApplication.tsx
```diff
@@ -0,0 +1,78 @@
...
+  applicationTrigger?: Trigger
+}
+
+const useUpdateApplication = ({ applicationSerial }: useUpdateApplicationProps) => {
+  const [submitted, setSubmitted] = useState(false)
+  const [processing, setProcessing] = useState(false)
+  const [error, setError] = useState<ApolloError | undefined>()
+
+  // Hook to get existing responses in cache - triggered when user submits application
+  const { data, error: responsesError } = useGetResponsesInApplicationQuery({
```
CarlosNZ, line 21: Related to the earlier comment -- all the response values have already been loaded by `useGetResponsesAndElementState`. Can you not just pass them into the `useUpdateApplication` hook rather than re-query for them?

CarlosNZ (commented):
Hey, just a comment about the additional query (see inline comment). You probably have a good reason for doing it like this, just keen to know why.

CarlosNZ:
Also -- when you resolve the conflicts -- I'm pretty sure this is the version you want:
```
  const [responsesByCode, setResponsesByCode] = useState<ResponsesByCode>({})
  const [responsesFullByCode, setResponsesFullByCode] = useState<ResponsesFullByCode>({})
```

src/utils/graphql/mutations/createApplication.mutation.ts
nmadruga: Yep!

src/utils/graphql/queries/getResponsesInApplication.query.ts
nmadruga: Thanks! Done: 93d4720c8b79c37c181ce8b2d466825030d3db46

src/utils/hooks/useUpdateApplication.tsx
nmadruga: Like chatted in person - What Apollo client does is that - if the elements are already in cache this query will only get from the cache, so no need to pass every element down the track.

nmadruga:
No, I checked and its all good to have it as I resolved. No `{}` so is just an undefined variable.

CarlosNZ (approved):

PR #121, opened 2020-11-13
#113 Change create application to create sections and responses on single-block

### LINKED ISSUE #113: Create application mutation in single-block

nmadruga:
Fixes #113 

Based on issue #45 - Should be merged before this one.

### Changes
- Removed files `createSection.mutation` and `createResponse.mutation` not in use anymore.
- Retrieve elements ids from the template on hook `useLoadTemplate` - to be used on the CreateApplication for responses.
- Simplify a lot the hook `useCreateApplication` to create one application and its sections and responses together.
- Update container `ApplicationCreate` to make the magic! It calls the mutation passing along the info gathered (and organised as the input is expected) to the ApplicationMutation. 

### Minor remaining to do
- I left a comment `// TODO: Update this query to use fragments!` which I will return to it at some point during this Sprint 9.


src/containers/Application/ApplicationCreate.tsx
```diff
@@ -3,42 +3,47 @@ import { TemplateTypePayload } from '../../utils/types'
...
-      serialNumber &&
-      templateSections &&
-      templateSections.length > 0
-    ) {
+    if (!processing && serialNumber && templateSections && templateSections.length > 0) {
       // Call Application page on first section
       const firstSection = templateSections[0].code
       // The pageNumber starts in 1 when is a new application.
       const pageNumber = 1
       push(`${serialNumber}/${firstSection}/page${pageNumber}`)
```
andreievg, line 37: I think this push navigation should happen in the callback of create mutation

src/containers/Application/ApplicationCreate.tsx
```diff
@@ -47,33 +52,53 @@ const ApplicationCreate: React.FC = () => {
...
+interface handleCreateProps {
+  create: (props: createApplicationProps) => void
+  setApplicationState: any
+  templateName: string
+  templateId: number
+  templateSections: { templateSectionId: number }[]
+  templateResponses: { templateElementId: number }[]
+}
+
+function handleCreate({
```
andreievg, line 83: At this point, I am not sure what is the benefit f moving handleCreate method outside of AppicationCreate component.

If it's a scope method inside ApplicationCreate, I think it would be easier to digest the code (and less lines).

Also it would expose things like this logic:

```
setApplicationState({ type: 'reset'})
setApplicationState({ type: 'setSerialNumber', serialNumber })
```

andreievg (changes requested):
A couple of preliminary comments, base don a quick look

src/containers/Application/ApplicationCreate.tsx
nmadruga: Oh, nice! That's a good idea I'll see if I can change to that now.

src/containers/Application/ApplicationCreate.tsx
nmadruga: Yes... I was trying to follow your original idea to keep everything out of the main component. So it would only have the return block with some variables and hook defined at the top. Bu I agree that it would be easier to read if this were inside. I'll move back in the component.

src/containers/Application/ApplicationCreate.tsx
nmadruga: Done

nmadruga:
@andreievg is this one good to go?

andreievg (approved):
Looks good, thanks for changes! 

> Yes... I was trying to follow your original idea to keep everything out of the main component. So it would only have the return block with some variables and hook defined at the top. Bu I agree that it would be easier to read if this were inside. I'll move back in the component.

True, i thought it was the case, I think once there are too many scoped variables needing to be accessed by an outside function, it becomes cumbersome. Although I don't really know the best practice for this (kind of like the idea of classes in this case, where you can define seperate render method at the top and all the helper methods below, not sure why react want to be so functional).

Another couple of comments, not needed changes just opinions, for future dev, if you like them.

`useCreateApplicationMutation` is pretty empty

Similar comment to what i've made in progress bard logic about nested conditional operators in jsx, although it looks pretty readable here. I kind of like the idea of doing early return of jsx if there are errors and then actual non erroneous jsx return at the end.

Issue #64, opened 2020-10-27
Front-end query server to get template types and actionables (based on template type) that the current user is allowed and display on App navigation


nmadruga:

nmadruga:
The UserArea - that is currently listing all users in the system - can be used to set the `current` user. This can be an easy way to check if we are correctly setting the list of available options based on the user permission.

andreievg:
Relocating this one to epic 11

andreievg:
Thought to break this issue up to just get the user permissions. Actually displaying allowable applications can be done in the layout & menue epic.

Also instead of name, it would be good if template in permissions is keyed by code

PR #123, opened 2020-11-16
#64 use user permission end point

andreievg:
Fixes: #64 

Requires: https://github.com/openmsupply/application-manager-server/pull/115

Note: New placeholder issue created in epic 10, (to actually display applications that are allowed)

#### Changes

* Config file now has url for both GraphQL and REST
* Added hook to get permissions via userPermission
* Added permission fields (and user fields) to user context
* Local storage for JWT (in chrome, dev tools, to go Application tab, and persistJWT, you will see key and value pair once page was loaded, can replace token with the one from login page to see it take affect on next reload)
* Added example of how templatePermissions can be accessed in app menu

In below screenshot, current user comes from userPermission end point and JSON at the bottom of menu is also from there

Screen Shot 2020-11-16 at 5 31 59 PM


CarlosNZ (approved):
Okay, cool. I have a few questions, but they can wait. 

Looks like you have a small conflict in types.ts, but otherwise good to go :)

Issue #118, opened 2020-11-13
JWT is used for GraphQL for existing queries


nmadruga:
Should be done in parallel with issue [Back-end#109](https://github.com/openmsupply/application-manager-server/issues/109) JWT is compiled for graphile to restrict row level permissions

Issue #105, opened 2020-11-06
Pass through evaluatorExpression to plugin and make validation in plugin


nmadruga:

CarlosNZ:
Also -- while I think of it -- the plugin also needs to evaluate the templateElement parameters, too.
(Maybe there can be an initial check and it only evalutes them if the `type` for that parameter is not a JSON/object.)

PR #125, opened 2020-11-17
#71 Plugin validation

### LINKED ISSUE #71: Validation before update the response in cache/server
- Update plugin (ShortText) to do validation and update on loose focus event
- Change the plugins (ShortText) to make the validation

CarlosNZ:
Fixes #71 and #105.

Requires back-end PR/branch #118 (https://github.com/openmsupply/application-manager-server/pull/118)
Please review both PRs together.

Please **reset database** (`database_init`), restart server, and completely wipe front-end local storage/cache before testing, otherwise you'll have problems.

Uses the expression-evaluator to validate input as the user types, and displays a message when invalid:
- on first input, only shows error when leaving input field, shows on every keystroke after that.

Please check the template definition in `insertData.js` to see the validation queries that should be active:
- Q2 / Last name -- requires Q1 to not be empty
- Q3 / username -- requires at least 3 chars from alphanumeric set
- Q4 / Email -- must be valid email
- Q5 / Password -- must be at least 8 characters, no spaces

Note that the object passed to the evaluator is a new object called "responses", which contains `allResponses` (all questions responses by code) and a new field called `thisResponse`. The `allResponses` object is only updated after entries are sent to database, whereas `thisResponse` is constantly updated with every keystroke, which is what we need to validate. Also, I figured that having a "this" reference would make evaluator queries simpler, as most validation will be checking against its own current value.

### Issues:
- I wanted to use the `check-unique` endpoint to validate username and email. However, there seems to be a problem with the evaluator using the API operator. I suspect it's something to do with CORS/localhost restrictions, but not sure right now -- will need to be addressed in [another issue](https://github.com/orgs/openmsupply/projects/6#card-49496498) Any suggestions welcome :) 
- I'm thinking it's probably a good idea to allow multiple validation criteria, each with their own distinct validation message. For example, with "email", we want to check that it conforms to appropriate REGEX, and also do an "isUnique" lookup. It's fine to write a query that can do both, but the error message wouldn't be very useful:  
"Email is either not unique or not a valid email address".  
I think we can make the `validation` and `validationMessage` fields contain arrays of queries and messages, and then we could show the user a much more specific validation message. e.g. "Not a valid email address", "That email address in already in use". What do you think?
- The red border around the input box is not working, and it looks like it'll need manual CSS editing, so left that for now.


src/containers/Application/ElementsBox.tsx
```diff
@@ -49,6 +50,7 @@ const ElementsBox: React.FC<ElementsBoxProps> = ({
           isEditable={elementsState[question.code].isEditable}
           isRequired={elementsState[question.code].isRequired}
           allResponses={responsesByCode}
+          evaluator={evaluateExpression}
```
andreievg, line 53: I think a better place for import of this one is in ApplicationViewWrapper (not a required change though)

src/elementPlugins/ApplicationViewWrapper.tsx
```diff
@@ -6,15 +6,14 @@ const ApplicationViewWrapper: React.FC<ApplicationViewProps> = (props) => {
...
-    isEditable,
-    isRequired,
-    allResponses,
   } = props
 
   if (!pluginCode || !isVisible) return null
 
   const { ApplicationView }: PluginComponents = pluginProvider.getPluginElement(pluginCode)
 
+  // console.log('pluginCode', pluginCode)
```
andreievg, line 15: stray console log ?

src/elementPlugins/ApplicationViewWrapper.tsx
```diff
@@ -6,15 +6,14 @@ const ApplicationViewWrapper: React.FC<ApplicationViewProps> = (props) => {
   const {
     templateElement: { elementTypePluginCode: pluginCode },
     isVisible,
-    isEditable,
-    isRequired,
```
andreievg, line 10: was wondering if isRequired would be something that ApplicationViewWrapper would display (generic for all plugins)

src/elementPlugins/shortText/src/ApplicationView.tsx
```diff
@@ -9,43 +9,70 @@ const ApplicationView: React.FC<ApplicationViewProps> = ({
   isEditable,
   isRequired,
   allResponses,
+  evaluator,
 }) => {
-  const [validationMessage, setValidationMessage] = useState('')
+  const [validationMessageDisplay, setValidationMessageDisplay] = useState('')
   const [value, setValue] = useState(initialValue?.text)
+  const [isValid, setIsValid] = useState<boolean>()
```
andreievg, line 16: I've just realised, for clarity sakes, we might want three state variables ('unknown', 'valid', 'invalid').

Not required change atm

src/elementPlugins/shortText/src/ApplicationView.tsx
```diff
@@ -9,43 +9,70 @@ const ApplicationView: React.FC<ApplicationViewProps> = ({
...
-    } else {
-      updateValidationMessage('')
-      onUpdate({ value: value, isValid: true })
-    }
+  function handleChange(e: any) {
+    setValue(e.target.value)
   }
+
+  function handleLoseFocus(e: any) {
+    onUpdate({ value: { text: e.target.value }, isValid })
```
andreievg, line 51: shouldn't this just be ` {value, isValid} ` ?

src/elementPlugins/shortText/src/ApplicationView.tsx
```diff
@@ -9,43 +9,70 @@ const ApplicationView: React.FC<ApplicationViewProps> = ({
...
+        fluid
+        label={templateElement.title}
+        placeholder={placeholder}
+        onChange={handleChange}
+        onBlur={handleLoseFocus}
+        value={value}
+        disabled={!isEditable}
+        type={maskedInput ? 'password' : undefined}
+        error={
+          !isValid && allResponses[templateElement.code]
```
andreievg, line 66: && value ? vs allResponses[templateElement.code]

src/elementPlugins/shortText/src/ApplicationView.tsx
```diff
@@ -9,43 +9,70 @@ const ApplicationView: React.FC<ApplicationViewProps> = ({
...
   isRequired,
   allResponses,
+  evaluator,
 }) => {
-  const [validationMessage, setValidationMessage] = useState('')
+  const [validationMessageDisplay, setValidationMessageDisplay] = useState('')
   const [value, setValue] = useState(initialValue?.text)
+  const [isValid, setIsValid] = useState<boolean>()
+  const [responses, setResponses] = useState({
+    thisResponse: undefined,
```
andreievg, line 18: `thisResponse: initialValue?.text` ?

Hmm, i am wondering if this is a tad too much boilerplate for a plugin. I have a couple of ideas. Should be fine for now thought

src/containers/Application/ElementsBox.tsx
```diff
@@ -57,10 +59,9 @@ const ElementsBox: React.FC<ElementsBoxProps> = ({
              * Also considering send to server on 'Next' or adding a Save button.
              **/
             // TODO: Only send mutation on loose focus event.
-            if (isValid)
-              responseMutation({
-                variables: { id: response?.id as number, value: { text: value } },
-              })
+            responseMutation({
+              variables: { id: response?.id as number, value, isValid },
```
andreievg, line 63: I've made a comment on an earlier PR from Nicole. I think this mutation should happen inside ApplicationViewWrapper (original plan for this callback was based on assumption that are now invalidated)

andreievg (changes requested):
Looks goods, a couple of minor changes please

src/containers/Application/ElementsBox.tsx
CarlosNZ: Done.

src/elementPlugins/shortText/src/ApplicationView.tsx
CarlosNZ: Yeah, I think `undefined`, `true`, `false` is working okay for now, but might run into issues down the line...

src/elementPlugins/shortText/src/ApplicationView.tsx
CarlosNZ: No, because we need to send the updated Value in the correct Response object back to onUpdate, which is:
```
{
  text: <actual value string>
  option?: int
  ...others?:any
}
```

Previously I was just sending the raw value up and constructing this object in the onUpdate call, but it's better to do it here since the plugin itself will need to provide this additional information. It doesn't make any difference in this plugin, but for selectors, etc., this will be necessary.

src/elementPlugins/shortText/src/ApplicationView.tsx
CarlosNZ: This is my sneaky way of making the validation error notification not show up until you actually *leave* the input box, but only on first visit. After that allResponses[templateElement.code] will have a value in it, so it will show error pop-up on every keystroke. I like it, you may have a different view. Did you actually try it in the UI?

src/elementPlugins/shortText/src/ApplicationView.tsx
CarlosNZ: I'm open to ideas, we just need to make sure we meet the following two criteria:
- we need the evaluator to have access to an object that will change on every keystroke (the main `allResponses` object only gets updated onLoseFocus)
- evaluator expressions shouldn't have to know the code of their own question when defining queries (i.e. you should be able to copy and paste an evaluator validation query from one question to another without having to change anything)

Maybe we can abstract more of this logic away to the ApplicationViewWrapper and keep the plugins themselves quite limited?

src/containers/Application/ElementsBox.tsx
CarlosNZ: Yup, done 852bbf1

src/elementPlugins/ApplicationViewWrapper.tsx
CarlosNZ: Yeah, fair enough. Have put them back in here. 762aa63

CarlosNZ:
FYI -- I've got the API check working the back-end now (thanks to @andreievg 's CORS fix), so fetch the latest version of `#71F-validation-improvements` and you'll see live uniqueness check on the "Username" field :)

src/elementPlugins/shortText/src/ApplicationView.tsx
andreievg: cool, I guess I was thinking to still use the value from state (but still send it back in object)

src/elementPlugins/shortText/src/ApplicationView.tsx
andreievg: I haven't tried it yet, will do before final approval.

I am still not sure what is the difference in validating on keystroke during second entry vs first entry

src/elementPlugins/shortText/src/ApplicationView.tsx
CarlosNZ: Oh, right, you mean just:
```
 function handleLoseFocus(e: any) {
    onUpdate({ value: { text: value }, isValid })
```

Done 7d2261c9

andreievg (approved):
Looks good, thanks @CarlosNZ 

Remaining bits:

* Password hashing logic
* Maybe second entry validation on keystroke could be optional via parameter ?
* When `field A` shows message for validation of other fields, error will persist unless something is entered in `filed A` (I think this will be dealt with here: https://github.com/openmsupply/application-manager-web-app/issues/126)

CarlosNZ:
> Remaining bits:
> 
>     * Password hashing logic
> 
>     * Maybe second entry validation on keystroke could be optional via parameter ?
> 
>     * When `field A` shows message for validation of other fields, error will persist unless something is entered in `filed A` (I think this will be dealt with here: #126)

Great. Have added those suggestions to #126. 👍

PR #82, opened 2020-11-03
#39 application progress bar

nmadruga:
Fixes #39 

Create new progress bar to be used by `ApplicationPage` on each page and `ApplicationOverview` for the summary.

### Changes
- Move `PageButton` to a separated component in the `components/Application` folder.
- Create new `ProgressBar` component in `components/Application` folder.
- Add component `ProgressBar` in `ApplicationOverview` and `ApplicationPage` containers.
- Rename `ElementsArea` to `ElementsBox` container to display all elements of one application.
- Create new `ControlBox` (moving code from `ApplicationPage` to here) to display the Previous and Next buttons.
- Move `AppMenu` to the components folder in `components/Main` folder.
- Move `Footer` to the components folder in `components/Main` folder.
- In the `getApplicationQuery` also receive `index` for each section (to fix problem with reloading the page) 

### Progress bar
- Show all sections + 'Review and Submit' sections with numbers (as defined in UI Design)
- Show pages of section when that's the current section.
- Using Semantic-ui `Accordion` to expand only the current section - the current one.
- Temporary showing **console.log** on click of one Section - needs to add logic to decide if user can go to that section or not (visited) and allow the `current section` to be update to the clicked one. Shouldn't have children pages.
- Temporary code on `Review and Submit` section not ideal.  
- Attempt to link a page to its URL - **not working** properly.
- Set the current section accordingly with URL

Issue #48, opened 2020-10-18
Check validation of progress structure before getting to Summary


nmadruga:
Should not allow getting to the summary page before all fields are valid.

CarlosNZ:
Also need to run validation again for all questions on Summary page load (to catch any earlier questions that might be invalidated by later answers). Not sure if this can be done as part of this issue -- if not, can whoever is doing this issue please create a new issue for that?

nmadruga:
@CarlosNZ 
> Also need to run validation again for all questions on Summary page load (to catch any earlier questions that might be invalidated by later answers). Not sure if this can be done as part of this issue -- if not, can whoever is doing this issue please create a new issue for that?

After we have chatted about the difference with this new idea of only allowing the user to go to the summary page when all required elements have valid responses, this should be fine. We have 2 issues now to deal with it (before we can do this one): 
#194 - To recalculate all responses validity when one is changed
#193 - To redirect user to the invalid response when they try accessing the summary page 

> I'm thinking that they probably should be done together, as the result of the validation check (on load) can be used to inform the submission check (so it doesn't have to recompute validations on submit, it just checks the result of the on-load validation check).

Yes. I think that's what will be done now. When we are loading the Summary, the code will check if all elements are valid, and otherwise send the user to the page with the invalid element, if not will load. Then the user can Submit.

PR #127, opened 2020-11-18
#81 progress bar logic

### LINKED ISSUE #81: Progress Bar 'state' logic
## Description
**From what was discussed in the team meeting on 17/11/2020**
**EDITED on 24/11/2020**: Considering what was discussed in the comments: [1](https://github.com/openmsupply/application-manager-web-app/issues/81#issuecomment-728671725), [2](https://github.com/openmsupply/application-manager-web-app/issues/81#issuecomment-728679792)

## Summary
* Template should have a flag to indicate if application should follow a `linear` progress.
* Indication on the Progress bar for sections/pages (and section icon has the combination of pages' statuses):
  - **neutral** => number for sections, empty space for pages
  - **valid** => Screen Shot 2020-11-17 at 1 05 54 PM
  - **invalid** => Screen Shot 2020-11-17 at 1 05 50 PM

## Terminology
When validation a page, will do either:
  - **regular** validation: Only check non-empty elements in the page
  - **strict** validation: Check for all elements in the page

#### Loading page 
- Runs **strict** validation for previous pages.
- Progress bar should only show **valid** icon for previous pages. And **neutral** icon for current page.
- **[TODO #135]** Redirects to last page that has invalid or unanswered elements - if there is one.

#### Navigation box
- Can only go to the some next page when all responses in current page are valid. 
- **[TODO #136]** When click "Next" => validates the current page
   - Show problems inside page for ALL elements (even unanswered)

#### Progress bar
- Next pages and sections are disabled
- After clicking Next, before moving to the next page:
   - If current page is valid (all required responses answered & no invalid elements) - show **valid** icon
   - Show **neutral** icon on the new current page
- Previous pages are enabled so user can click to go back
  - When changing some response on a previous page that makes it invalid -- it doesn't let you go forward until its corrected (even though the later page might still show "Valid")

#### Summary page
- can't get to it unless all pages are valid (i.e. same logic as all pages)

#### Loading page
  - Checks All pages that have some elements that have been responded to (i.e. NOT undefined)
  - Show indication in progress bar for pages that have elements answered (OR all elements are valid - in case no element is required in the page):
    - If page is valid (all required responses answered & no invalid elements) - show **valid** icon
    - If page is invalid (at least one invalid element) - show **invalid** icon
    - All other pages show the **neutral** icon

#### Navigation box
- Can go to next pages using "Next" or clicking on another page in the Progress bar.
- When click "Next" (or in another page) => **strict** validation of the current page
- Go to the new current page regardless of the validation status of previous page.

#### Progress bar
- Current page/section always show the **neutral** icon
- When clicking on other page => Run **strict** validation of the current page
- Go to the new current page regardless of the validation status of previous page.
- Show indication for previous pages:
  - If page is valid (all required responses answered & no invalid elements) - show **valid** icon
  - If page is invalid (all required responses answered & at least one invalid element) - show **invalid** icon
  - All other pages show the **neutral** icon

#### Summary page
* **[TODO #48]**
- Only show problems in each section for invalid answered elements at the start 
- Required elements that haven't been answered - would only show empty and the * to identify it's required.
- After clicking on the **Submit** button:
  - If all required elements are answered with valid responses & no other invalid elements - Show message to confirm the submission indicating that after submitted the application can't be changed.
  - If any required responses is empty or any invalid elements - Show message saying there are problem in the application
    - Run validation over ALL elements and show problems (even for required unanswered elements)
    - Going back to a section should show all problems now
    - The progress bar will show icons for all pages:
      - If all required elements have valid responses & no other invalid elements - show **valid** icon.
      - If at least one element has an invalid response - show **invalid** icon.
   - NOTE: this "submittedClicked" status is for the current session only -- if page is refreshed or visited later, it will only show problems as described above.

--------
## Previous description

**Progress bar need to indicate:**
* Pages that can be visited (for 'linear view')
  - Show visited pages enabled
* Valid/Invalid pages - Icons to replace the Page number: 
  - Invalid `exclamation circle`
  - Valid `check circle`
* Set page is visited on Next button & Previous buttons

**Pages that can be visited:**

`Linear`

* Page where first isRequired and isVisible response has no value. (this should allow for gap pages, gap page is when a page becomes visible in between valid pages)
* Navigation to summary can only be done if all pages are valid

`Non-Linear`

Any page

**Valid/Invalid pages**

What is invalid page ?

For a page, all isVisible elements, that are also isRequired have a value (plugin must make sure to only send value in onUpdate event when it's valid)

`Linear`

Should be easy for this case, need to validate this page before moving to next page.

(what happens when we navigate back to page and change things that may cause validation or visibility changes in other pages ? I think rules in `Deduce by logic` should be appropriate)

`Non-Linear Problem`

We don't want to show whole form with all of the pages having invalid status form the start. So need some sort of mechanism for knowing that a page is eligible for validation indicator.

`isVisited` flag

One solutions is to use isVisited tag on a page, in local store rather than DB. Basically when user navigated to a page, and then changes page thereafter, it would be eligible for validation indication.

We would need to persist this store in order to work with refreshes, we would also need to make sure to run this check on submission (or on summary page load). When persisting this store we would need to know which record it is related to.

`Deduce by logic` (kind of a hybrid of isVisited, but without persist)

Eligible for validation, is a flag set on a page (non persistent)
* A page has at least one response
* When navigation to summary page (or on submission), we mark all pages as eligible for validation
* When navigating away from a page (on next button, on select another page in the future parts of the application)

nmadruga:
Fixes #81 

~~Needs to run using [Back-end#120](https://github.com/openmsupply/application-manager-server/pull/120) and [Back-end#121](https://github.com/openmsupply/application-manager-server/pull/121) - using the Back-end#121 will have both changes.~~ Back-end issue already merged in.

Use [Back-end-PR#127](https://github.com/openmsupply/application-manager-server/pull/127) to see the Company registration with more pages per section. This will make the test of the Progress bar logic for sections/pages easier.

A brief of the logic for the Progress bar is described in the issue related to this PR, if further discussion regarding the logic it would be good to add directly in there (in the comments) so we can keep it up-to-date.

I wasn't sure how to structure this object that is passed to the ProgressBar component, so any improvements ideas will be very welcome. For now it doesn't show the progress after you move from the Summary back to the sections... But to be able to do something like that, the structure for the ProgressBar needs to be a shared state instead of a local one in the `ApplicationPageWrapper`.

### Changes
- Create 3 types `ProgressInApplication`, `ProgressInPage`, `ProgressInSection` used in hierarchical order to keep the current progress in Application (stored in local state in the `ApplicationPageWrapper`.
- The `ProgressBar` and `NavigationBox` will receive the `validateCurrentPage` function in the props to run the validation on the current page, store in the `progressInApplication` state on the ApplicationPage and return if can proceed to next page.
- Some logic was removed from the `ProgressBar` that now will just present the icons, disable/enable sections depending on what is passed in the `progressInApplication` structure.

### Remaining todo
- Use the validation of fields in the `progressInApplication` structure (is only using `isRequired` and `isVisible` for now)
- On linear Application need to disable next sections (that haven't been visited) on the progress bar.
- Display validation on Progress bar after coming back from the Summary page.

### Screenshots
The progress bar on the Application: Company Registration
* To create one company registration application go to `http://localhost:3000/applications/new?type=CompRego1` and start a new one. It will have 3 sections with one page in each, so when moving to the next pages the whole section is validated.
Screen Shot 2020-11-19 at 12 36 14 PM
 
* **Designed in Miro**:
Screen Shot 2020-11-19 at 2 01 40 PM


CarlosNZ:
Now that I see it in action, my initial comment would be that the Progress bar _should_ show up on the Summary page (at least before it's submitted, not after). Otherwise, you never actually see the "All done" icon for the last page/all the pages at once, which is kind of unsatisfying from a user POV. ;)

CarlosNZ (commented):
Just going to make functionality observations first, haven't really looked at the code yet:

- When going back using the "Previous" button, the progress bar is fine, but when going back using Browser back (or direct to a URL, or clicking in the Application list), the progress bar no longer has the correct/warning icons.
- It seems weird not having the progress bar on the summary page, IMHO.
- Wouldn't it be easier to make the "neutral" icon a generic one rather than numbers? I think we talked about not bothering to number the Sections as you can just put a number in the section name if you want it.

andreievg (changes requested):
Hi @nmdruga,

I've reviewed `ProgressBars.tsx`, and then moved on to `ApplicationWrapper`, had a quick look at the logic there. I think it's sound, but it's a bit hard to follow, what all of the set, update and validate methods do and how they behave together.

My original thought for the process of validating pages and displaying them was as follow:

(in ApplicationWrapper)

`useGetResponseAndElementState` <- happens when new value (and now isValid) is stored in cache, this one calculates isVisibility etc, return one object with all info for a response/element 

`useEffect on when elementsState has change` <- here we can calculate, base on return from `useGetResponseAndElementState` validity and return an create an object, something like this:

```
[
    {
        sectionName: 'Section 1',
        status: 'valid'|'inValid'|'undefined',
        canNavigate: true|false,
        isActive: true|false,
        pages: [
            {
                pageName: 'Page 1',
                status: 'valid'|'inValid'|'undefined',
                canNavigate: true|false,
                isActive: true|false,
            }
        ]

    }
]
```

The process above kind of automatically deals with all of the use case (i.e. reloading the application etc..).

What do you think ?

nmadruga:
Thanks for the feedbacks @carl and @andrei much appreciated!

I'll just comment briefly on the overall reviews - and will be working on some changes for the suggestions you gave next.

nmadruga:
> When going back using the "Previous" button, the progress bar is fine, but when going back using Browser back (or direct to a URL, or clicking in the Application list), the progress bar no longer has the "complete"/"warning" icons.

Good point! That's a problem with the current way I have implemented this, with the local state on the `ApplicationWrapper` then anytime it refreshes is going to the initial state. Not good -> I'll be changing this structure to be returned by the hook `useGetResponseAndElementState` as suggested.

> It seems weird not having the progress bar on the summary page, IMHO.

Yes, I was also a bit annoyed by that, hmmm maybe we could have it sort of hidden? I'll try to illustrate this idea on Monday.

> Wouldn't it be easier to make the "neutral" icon a generic one rather than numbers? I think we talked about not bothering to number the Sections as you can just put a number in the section name if you want it.

I think we should go with what has been designed otherwise we aren't considering the inputs from the designer.

nmadruga:
> My original thought for the process of validating pages and displaying them was as follow:

> (in ApplicationWrapper)

> useGetResponseAndElementState <- happens when new value (and now isValid) is stored in cache, this one calculates isVisibility etc, return one object with all info for a response/element

>useEffect on when elementsState has change <- here we can calculate, base on return from useGetResponseAndElementState validity and return an create an object, something like this:
> ...

I agree its confusing and I totally agree it needs some changes (move to a hook **or** use a context to store the shared state **or** else), I'm not 100% sure that using a `useEffect` to build this object is the best solution. But I'll be giving a try on this solution, let's see if it can make this simpler. : )

I'm already using the result from `useGetResponseAndElementState` to generate the structure that is sent to the ProgressBar component, but the problem with it storing this object here is that the object is reset when the page refreshes, which might be fixed with this solution since it will be generating the object everytime. Let's see!

src/components/Application/ProgressBar.tsx
```diff
@@ -9,34 +9,55 @@ interface SectionPage {
...
+            onClick={() =>
+              attemptChangeToPage({
+                serialNumber,
+                sectionCode,
+                page: Number(number),
+                push,
+                validateCurrentPage,
+              })
+            }
+            active={currentPage.visited} // TODO: Change to only show active when visited in linear applications
```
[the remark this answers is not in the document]
nmadruga: Yes, you are right. I'll change this. Ty!

src/components/Application/ProgressBar.tsx
```diff
@@ -9,34 +9,55 @@ interface SectionPage {
...
+          >
+            {currentPage.pageStatus == true ? (
```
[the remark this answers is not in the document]
nmadruga: Yes, that's what I was aiming with this `pageStatus` and `sectionStatus`, meaning:
`true` => All (required) responses are valid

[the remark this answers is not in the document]
nmadruga: I'm changing this property to `status` as you suggested on the previous comment. And considering creating a type to store the 3 cases for each page validation: `valid`, `invalid` or `incomplete` => Which would be the status when only a part of the **required responses** have been answered regardless of valid or invalid. What do you think?

src/components/Application/ProgressBar.tsx
```diff
@@ -45,20 +66,35 @@ const ProgressBar: React.FC<ProgressBarProps> = ({
           children: (
             <Grid>
               <Grid.Column width={4} textAlign="right" verticalAlign="middle">
-                {getStepNumber(stepNumber)}
+                {section.sectionStatus !== undefined ? (
```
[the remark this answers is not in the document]
nmadruga: I'll will be working on this one now.

nmadruga:
Ok, I think this is ready for a re-review... @CarlosNZ @andreievg 

I could set the progress of sections not visited (after the current) to not display the statue on the linear application:
Screen Shot 2020-11-23 at 3 21 12 PM

The non-linear, as discussed on the planning meeting on **23/11/2020**,  not having a shared local state to determine what sections have been visited and which didn't makes the validation show over the ones which have been visited from the start, for example if there is **no required field** in a page, and it hasn't been filled -> the page is valid.

CarlosNZ:
Just noticed:
Screen Shot 2020-11-23 at 4 21 37 PM
-- Sections appearing out of order?

Also -- isn't there meant to be valid/invalid icons for the sections once we've filled them in?
Screen Shot 2020-11-23 at 4 24 16 PM

CarlosNZ:
Screen Shot 2020-11-23 at 4 29 34 PM
Also -- it's odd that Page 1 has an active link when I'm on page 1, and page 2 doesn't (and I've been there before)
(Sorry, the cursor doesn't show in the screenshot as I expected -- but "Page 1" is clickable and "Page 2" isn't.)

nmadruga:
Hi @CarlosNZ thanks for having another look into this PR : )

> -- Sections appearing out of order?

Yes, that's a problem there. I had already worked on a fix - added to a separated PR (but forgot to create yesterday) here: #134.

>  Also -- isn't there meant to be valid/invalid icons for the sections once we've filled them in?

Yes. It should do this. There is already logic to do that, for some reason not working right now. I'll have a look after morning break : ) I also want to make sure I'm following the rules we set for Progress bar in the description of issue #81.

> Also -- it's odd that Page 1 has an active link when I'm on page 1, and page 2 doesn't (and I've been there before)
(Sorry, the cursor doesn't show in the screenshot as I expected -- but "Page 1" is clickable and "Page 2" isn't.)

Hmmmm. Not sure how to fix that exactly. I made the `User registration` as a `linear` application, which means you can only move forward (to **Page 2**) once all required fields in **Page 1** are filled and valid. That's why the next page in the progress bar is not clickable, the same if there were more sections further they would be all unclickable.

Not sure if the current page should or shouldn't be clickable though. What do you think? Not clickable?

CarlosNZ:
> Not sure if the current page should or shouldn't be clickable though. What do you think? Not clickable?

Yeah, I definitely think the current page should be not clickable. But can make it a new issue if it's getting to complicated in this one.

One other suggestion -- do you think we should add additional pages to the Company Registration template? I feel like I'm not able to fully check the functionality of a non-linear application with more than one page in a section.

nmadruga:
> Yeah, I definitely think the current page should be not clickable. But can make it a new issue if it's getting to complicated in this one.

It's ok. I'll add this as part of this one. Shouldn't take me too long.

> One other suggestion -- do you think we should add additional pages to the Company Registration template? I feel like I'm not able to fully check the functionality of a non-linear application with more than one page in a section.

Ok. I'll add more pages to it on another Back-end PR.

nmadruga:
This PR is ready for another review : )

I have reviewed the progress states expected in the issue, and linked other issues to be dealing with more specific changes - not part of this PR.

src/containers/Application/ApplicationPageWrapper.tsx
```diff
@@ -44,10 +57,42 @@ const ApplicationPageWrapper: React.FC = () => {
...
+    const progressStructure = buildProgressInApplication({
+      elementsState,
+      responses: responsesFullByCode,
+      templateSections,
+      isLinear: application?.isLinear as boolean,
+      currentSection: currentSection?.index as number,
+      currentPage: Number(page),
+    })
+    setProgressInApplication(progressStructure)
+    setProcessingValidation(false)
```
andreievg, line 76: Hmm I am not sure why `processingValidation` is required

src/containers/Application/ApplicationPageWrapper.tsx
```diff
@@ -44,10 +57,42 @@ const ApplicationPageWrapper: React.FC = () => {
...
         replace,
       })
+
+      if (sectionCode && page)
+        setCurrentSection(templateSections.find(({ code }) => code === sectionCode))
+    }
+  }, [isApplicationLoaded, sectionCode, page])
+
+  useEffect(() => {
+    const progressStructure = buildProgressInApplication({
```
andreievg, line 67: `comment`, with multiple effects, it would be good to name them somehow, a quick comment with their intended function would be great.

src/containers/Application/ApplicationPageWrapper.tsx
```diff
@@ -44,10 +57,42 @@ const ApplicationPageWrapper: React.FC = () => {
...
+      return false
     }
-  }, [application])
+    setProcessingValidation(true)
+    const validation = validatePage({
+      elementsState: elementsState as ApplicationElementStates,
+      responses: responsesFullByCode as ResponsesFullByCode,
+      sectionIndex: currentSection?.index as number,
+      page: Number(page),
+      checkEmpty: true,
```
andreievg, line 90: what is the use case of checkEmpty parameter, it's not apparent from reading the code

src/utils/helpers/validatePage.tsx
```diff
@@ -0,0 +1,113 @@
...
+  page,
+  checkEmpty = false,
+}: validatePageProps): ProgressStatus => {
+  let count = 1
+  const elementsInCurrentPage = Object.values(elementsState)
+    .filter(({ section }) => section === sectionIndex)
+    .reduce((pageElements: ElementState[], element) => {
+      if (element.elementTypePluginCode === 'pageBreak') count++
+      if (count !== page) return pageElements
+      if (element.category === TemplateElementCategory.Question) return [...pageElements, element]
```
andreievg, line 58: not a required change, this is stylistic preference, but to me logically it makes more sense to check in this fashion:

if pageBreak increment
if not question type category return pageElement
if current page return pageElements + element
return pageElements

src/utils/helpers/validatePage.tsx
```diff
@@ -0,0 +1,113 @@
...
+  ApplicationElementStates,
+  ElementState,
+  ProgressStatus,
+  ResponsesFullByCode,
+  ResponseFull,
+} from '../../utils/types'
+
+import { TemplateElementCategory } from '../../utils/generated/graphql'
+
+export const PROGRESS_STATUS: { [key: string]: ProgressStatus } = {
```
andreievg, line 11: i think this is use case for enum ?

src/utils/helpers/validatePage.tsx
```diff
@@ -0,0 +1,113 @@
...
+  const elementsInCurrentPage = Object.values(elementsState)
+    .filter(({ section }) => section === sectionIndex)
+    .reduce((pageElements: ElementState[], element) => {
+      if (element.elementTypePluginCode === 'pageBreak') count++
+      if (count !== page) return pageElements
+      if (element.category === TemplateElementCategory.Question) return [...pageElements, element]
+      return pageElements
+    }, [])
+
+  const responsesStatuses = elementsInCurrentPage.reduce(
```
andreievg, line 62: Just a preliminary question before looking at code below in depth, don't we just want to do this:
(btw nested conditional operators are hard to read)

(considering that elementsInCurrentPage have responses already linked)

filter elementsInCurrentPage by isVisible visibleElementsInPage
filter visibleElementsInPage by isRequired let's call it requiredAndVisibleElementsInPage

if(visibleElementsInPage find isValid === false) then NOT_VALID
if(requiredAndVisibleElementsInPage find by value === undefined ) then INCOMPLETE

otherwise VALID

andreievg (changes requested):
@nmadruga Partial Review, still have `useGetResponsesAndElementState` to review and UI

src/containers/Application/ApplicationPageWrapper.tsx
```diff
@@ -111,4 +164,83 @@ function processRedirect(appState: any): void {
...
+  responses,
+  templateSections,
+  isLinear,
+  currentSection,
+  currentPage,
+}: buildProgressInApplicationProps): ProgressInApplication {
+  if (!elementsState || !responses) return []
+  let sectionsStructure: ProgressInApplication = templateSections.map((section) => {
+    // Create an array with all pages in each section
+    const pages = Array.from(Array(section.totalPages).keys(), (n) => n + 1)
```
andreievg, line 187: looks pretty cool.

src/containers/Application/ApplicationPageWrapper.tsx
```diff
@@ -111,4 +164,83 @@ function processRedirect(appState: any): void {
...
+    const checkIsCurrentPage = (page: number) =>
+      section.index === currentSection && page === currentPage
+    const checkIsPreviousPage = (index: number, page: number) =>
```
andreievg, line 190: don't really need the `check` bit in the method naming here, but doesn't hurt to keep

src/containers/Application/ApplicationPageWrapper.tsx
```diff
@@ -111,4 +164,83 @@ function processRedirect(appState: any): void {
...
+      index < currentSection ? true : index === currentSection && page <= currentPage
+
+    // Build object to keep each section progress (and pages progress)
+    const progressInSection: ProgressInSection = {
+      code: section.code,
+      title: section.title,
+      canNavigate: isLinear ? section.index <= currentSection : true,
+      isActive: section.index === currentSection,
+      pages: pages.map((number) => ({
```
andreievg, line 199: pageNumber ?

src/containers/Application/ApplicationPageWrapper.tsx
andreievg, line 190: sectionIindex ?

src/containers/Application/ApplicationPageWrapper.tsx
```diff
@@ -111,4 +164,83 @@ function processRedirect(appState: any): void {
...
+  currentPage,
+}: buildProgressInApplicationProps): ProgressInApplication {
+  if (!elementsState || !responses) return []
+  let sectionsStructure: ProgressInApplication = templateSections.map((section) => {
+    // Create an array with all pages in each section
+    const pages = Array.from(Array(section.totalPages).keys(), (n) => n + 1)
+    const checkIsCurrentPage = (page: number) =>
+      section.index === currentSection && page === currentPage
+    const checkIsPreviousPage = (index: number, page: number) =>
+      index < currentSection ? true : index === currentSection && page <= currentPage
```
andreievg, line 191: i see you quite like conditional operators. typically we would just add brackets around `index === currentSection && page <= currentPage` with an or in between.

I think what we want for canNavigate logic, for linear (does not need to be aware of current or previous page):

the first not valid or incomplete page from the start = set flag for cant navigate, then every page and section after that in linear mode, can't be navigated to

src/containers/Application/ApplicationPageWrapper.tsx
```diff
@@ -111,4 +164,83 @@ function processRedirect(appState: any): void {
...
+
+    // Build object to keep each section progress (and pages progress)
+    const progressInSection: ProgressInSection = {
+      code: section.code,
+      title: section.title,
+      canNavigate: isLinear ? section.index <= currentSection : true,
```
andreievg, line 197: canNavigate to next section if currentSection last page is Valid in linear mode

src/containers/Application/ApplicationPageWrapper.tsx
```diff
@@ -111,4 +164,83 @@ function processRedirect(appState: any): void {
...
+      isActive: section.index === currentSection,
+      pages: pages.map((number) => ({
+        pageName: `Page ${number}`,
+        canNavigate: isLinear ? checkIsPreviousPage(section.index, number) : true,
+        isActive: checkIsCurrentPage(number),
+        status: checkIsCurrentPage(number)
```
andreievg, line 203: can you comment, what this logic is here, shouldn't it just be :`validatePage(...)` ?

I would do pages outside of the progressInSection object like this:

```

const pages = ...

progressInSection: {
..
status: getCombinedStatus(pages),
pages
}

```

src/utils/helpers/validatePage.tsx
```diff
@@ -0,0 +1,113 @@
...
+
+export const PROGRESS_STATUS: { [key: string]: ProgressStatus } = {
+  NOT_VALID: 'NOT_VALID',
+  VALID: 'VALID',
+  INCOMPLETE: 'INCOMPLETE',
+}
+
+export const getCombinedStatus = (
+  array: { status: ProgressStatus }[] | undefined,
+  strictly: boolean = false
```
andreievg, line 19: what is strictly for this method ?

src/utils/helpers/validatePage.tsx
```diff
@@ -0,0 +1,113 @@
...
+import { TemplateElementCategory } from '../../utils/generated/graphql'
+
+export const PROGRESS_STATUS: { [key: string]: ProgressStatus } = {
+  NOT_VALID: 'NOT_VALID',
+  VALID: 'VALID',
+  INCOMPLETE: 'INCOMPLETE',
+}
+
+export const getCombinedStatus = (
+  array: { status: ProgressStatus }[] | undefined,
```
andreievg, line 18: I think it's much simpler to just do `some` here:

pages.some(({status} )=> status === NOT_VALID) return NOT_VALID
pages.some(({status} )=> status === INCOMPLETE) return INCOMPLETE
return VALID

andreievg (changes requested):
I think only UI remains now

src/containers/Application/ApplicationPageWrapper.tsx
andreievg, line 67: I think we should also check for responsesLoading and return if it's true, (would need to add it to watched variables array)

andreievg (changes requested):
Missed a comment for logic, still have to look at ProgressBar UI

andreievg:
@nmadruga @CarlosNZ. 

There was a conversation above about clickable pages, not sure the result of it. 

This is something we can ask Debbie about, but I think we should still have sections and pages highlightable as if you can click them, even if canNavigate is false, and show a message when trying to navigate to them ? Can be a generic message for linear "Can only move to this page if requirements of previous page are fullfilled"

CarlosNZ:
Could I just make one final UI comment if it's not too late? I find it quite disconcerting not being able to tell (visually) what page I'm on within a Section. Can the current page show slightly different in the Sidebar (maybe bold or indented?) so there's some indication of the current page. With sections, you've made the current section show in a bigger size type, and that works great as a visual indicator.

andreievg:
@CarlosNZ 

> Could I just make one final UI comment if it's not too late? I find it quite disconcerting not being able to tell (visually) what page I'm on within a Section. Can the current page show slightly different in the Sidebar (maybe bold or indented?) so there's some indication of the current page. With sections, you've made the current section show in a bigger size type, and that works great as a visual indicator.

https://github.com/openmsupply/application-manager-web-app/pull/82#discussion_r520942397

I think the only remaining thing si to make sure the header is moved to be a pageNumber as text inside list item

src/utils/types.ts
```diff
@@ -94,10 +101,33 @@ interface TemplateSectionPayload {
...
+  canNavigate: boolean
+  isActive: boolean
+  pages?: ProgressInPage[]
+}
+
+type ProgressInApplication = ProgressInSection[]
+
+type ProgressStatus = 'VALID' | 'NOT_VALID' | 'INCOMPLETE'
+
+type ReviewCode = 'PR'
```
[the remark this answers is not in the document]
nmadruga: Hmmm, that is just to create the structure to be sent to the `Progress bar`, so each section has it's code (to be used in the URL when the user clicks the section/page link) and to create the structure with a `Summary And Review` section I had to add some code to it. Can add a more descriptive name to it.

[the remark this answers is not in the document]
nmadruga: I can't think of a better solution for now, so have simply changed to use a constant to be used for as the summary section code. Check [commit](https://github.com/openmsupply/application-manager-web-app/pull/127/commits/85cd6e55e5f398d3b9d3f769ac33e2ef9a7cd0ef)

src/containers/Application/ApplicationPageWrapper.tsx
nmadruga: It was required to call the `ProgressBar` component when the structure have finish the creation. It doesn't seem to be a problem to remove it now, so I've removed [here](https://github.com/openmsupply/application-manager-web-app/pull/127/commits/2336794c474455557585527bfbfde6221859600e)

src/containers/Application/ApplicationPageWrapper.tsx
nmadruga: No problem. I'll do that now

src/containers/Application/ApplicationPageWrapper.tsx
nmadruga: The `checkEmpty` is used for the **strict** validation. The default then is **loose** validation, which means that the validatePage is not checking for empty responses. I'll move this to a helper file and add some comments to this.

nmadruga: Oh, sorry, it's already in a helper file. I'll just add a comment explaining what I just did here. Then we can improve this function as needed.

src/utils/helpers/validatePage.tsx
nmadruga: Thanks. I have changed it slightly to what you have suggested: [here](https://github.com/openmsupply/application-manager-web-app/pull/127/commits/bd0a16c3cc244a73542f8880aa206894619243cc)

src/utils/helpers/validatePage.tsx
nmadruga: > filter elementsInCurrentPage by isVisible visibleElementsInPage

nmadruga: > otherwise VALID

There is just one problem with that: we can have notRequired elements that have been filled and are INVALID

src/utils/types.ts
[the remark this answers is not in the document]
andreievg: I think that works for now, at least no confusion now what it's about

src/containers/Application/ApplicationPageWrapper.tsx
andreievg: thanks

src/utils/helpers/validatePage.tsx
andreievg: I think that's good, if something in the form is entered incorrectly (even though it's not required), I think it should be indicated as incorrect

nmadruga: I added more comments should be easier to read now. Thanks for all the help with this function much appreciated! Commit [here](https://github.com/openmsupply/application-manager-web-app/pull/127/commits/4cf4f1b3a292fc1359dd0a99a817faf237c0babc)

src/containers/Application/ApplicationPageWrapper.tsx
nmadruga: I'll remove. : )

src/containers/Application/ApplicationPageWrapper.tsx
nmadruga: Yes, the conditional helps me to think, but I agree is not the easiest to read. I'll change to what you suggest.
I will also consider what you comment on the status for `canNavigate` flag.

src/containers/Application/ApplicationPageWrapper.tsx
nmadruga: Moved to outside of the scope as suggested: [here](https://github.com/openmsupply/application-manager-web-app/pull/127/commits/b6e437b958e873fd0182e53fa8fa4174cc4a91c9)
 
The reason why the `validatePage` is done 2 ways is differing if it would run in a `strict` or `loose` mode.
1. `linear` and `previous page` it would run a STRICT validation
2. `non-linear` and `previous page` is should run a LOOSE validation

Now in case the user has been to the `Summary` page I think we would need to run a STRICT validation over all the application, which is not implemented yet. I think it would be better if we leave this to be dealt in the issue #48

src/containers/Application/ApplicationPageWrapper.tsx
nmadruga: > the first not valid or incomplete page from the start = set flag for cant navigate

I think this would be more complicated since we can have any page with status INVALID, that's why I'm checking first for `isPreviousPage`. Although if we want to display all validation in the progress bar after the user has been to the `Summary section` I don't think this code would work. But I'm not sure how we can change it to consider this overall validation.

src/utils/helpers/validatePage.tsx
nmadruga: I was reconsidering this one. I think this is not needed anymore. I was thinking we would show the section state (combined of pages statuses) different when we ran the validation. After our meetings and the document that @CarlosNZ created that the combined status would always return:
> Any INVALID page in Section → Section state = ❌
> Any NEUTRAL page in Section → Section state = Neutral
> ELSE:  Section state = ✅

src/utils/helpers/validatePage.tsx
nmadruga: Yes! I agree (only saw this after I commented on the previous one).
Change [here](https://github.com/openmsupply/application-manager-web-app/pull/127/commits/c55bfbf478f1f7b772b0ce4a506fe4cbb246e5b8)

nmadruga:
> There was a conversation above about clickable pages, not sure the result of it.

For now the current page is disabled and in linear application all next pages are disable was well.

> This is something we can ask Debbie about, but I think we should still have sections and pages highlightable as if you can click them, even if canNavigate is false, and show a message when trying to navigate to them ?

I have created an issue to change the pages to be all enabled and show the message to users: #142

nmadruga:
> I think the only remaining thing si to make sure the header is moved to be a pageNumber as text inside list item

Attempt to address that [here](https://github.com/openmsupply/application-manager-web-app/pull/127/commits/6b4c963c10c0de86f5665f6bb220c0759729a260)

src/containers/Application/ApplicationPageWrapper.tsx
andreievg: Doesn't `canNavigate` only apply to linear mode? So in linear mode you can only navigate to the first page that is invalid/incomplete ?

src/containers/Application/ApplicationPageWrapper.tsx
```diff
@@ -111,4 +167,84 @@ function processRedirect(appState: any): void {
...
+    const pageNumbers = Array.from(Array(section.totalPages).keys(), (n) => n + 1)
+    const isCurrentPage = (page: number) => section.index === currentSection && page === currentPage
+    const isPreviousPage = (sectionIndex: number, page: number) =>
+      sectionIndex <= currentSection || page <= currentPage
+
+    const pages = pageNumbers.map((pageNumber) => ({
+      pageName: `Page ${pageNumber}`,
+      canNavigate: isLinear ? isPreviousPage(section.index, pageNumber) : true,
+      isActive: isCurrentPage(pageNumber),
+      status: isCurrentPage(pageNumber)
```
andreievg, line 199: I still don't quite understand the need for condition like this here.

If you forget about `strict` mode setting in validationPage -> it just return valid, invalid or incomplete.

Then the logic seem quite simple to me:
```
getStatue = () => {
draftPageStatus = validatePage
if(isLinear || strict) return draftPageStatus === valid : valid ? not valid

return draftPageStatus
}
```

And for canNavigate
```
canNavigate = true
...
let sectionsStructure: ProgressInApplication = templateSections.map((section) => {
...
const pages = pageNumbers.map((pageNumber) => {
...
status = getPageStatus()
returnObject = {
...
status,
canNavigate
...
}
...
canNavigate = status !== not valid
...
return returnObject
}

```

src/utils/helpers/validatePage.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+  // Generate array with statuses of responses of question to verify
+  const responsesStatuses = verifyQuestions.reduce(
+    (responsesStatuses: { status: ProgressStatus }[], element: ElementState) => {
+      const { isValid } = responses[element.code] as ResponseFull
+
+      return [
+        ...responsesStatuses,
+        {
+          status: isValid ? PROGRESS_STATUS.VALID : PROGRESS_STATUS.NOT_VALID,
```
andreievg, line 80: I don't see any returns for incomplete, I still think the logic here can be further simplified with suggested from here https://github.com/openmsupply/application-manager-web-app/pull/127#discussion_r529871184

andreievg (changes requested):
Further changes please, looking better though

nmadruga:
@andreievg I have added the changes as you suggested after our talk. I have updated the `validatePage` describing what is the validation considering, and I think the statuses are looking better now with more `INCOMPLETE` status on pages. Hopefully this is good to go. I think any further changes might be better done in separated PR, just because this one is becoming too big, which isn't helpful.

src/containers/Application/ApplicationPageWrapper.tsx
```diff
@@ -111,4 +168,79 @@ function processRedirect(appState: any): void {
...
+    const progressInSection: ProgressInSection = {
+      code: section.code,
+      title: section.title,
+      canNavigate: isLinear ? section.index <= currentSection : true,
+      isActive: section.index === currentSection,
+      pages,
+      status:
+        section.index === currentSection
+          ? PROGRESS_STATUS.INCOMPLETE
+          : getCombinedStatus(pages.map(({ status }) => status)),
```
andreievg, line 229: nice

andreievg (approved):
It's looking good, thanks @nmadruga for all the changes, sorry for all the requests.

I still want to go through with testing and make an issue for extra bits.

src/utils/helpers/validatePage.tsx
```diff
@@ -0,0 +1,78 @@
...
+        : pageElements
+    }, [])
+
+  // Generate array with current status of responses for each question to verify
+  const responsesStatuses = visibleQuestions.reduce(
+    (responsesStatuses: ProgressStatus[], { code, isRequired }: ElementState) => {
+      const { text, isValid } = responses[code] as ResponseFull
+      if (isRequired && (!text || isValid === null))
+        return [...responsesStatuses, PROGRESS_STATUS.INCOMPLETE]
+      if (!text || isValid === null) return [...responsesStatuses, PROGRESS_STATUS.VALID]
```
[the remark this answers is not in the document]
nmadruga: Added [here](https://github.com/openmsupply/application-manager-web-app/pull/127/commits/6870ea4e3c11586120e7796752b1748b507c51f7)

src/containers/Application/ApplicationPageWrapper.tsx
```diff
@@ -111,4 +168,79 @@ function processRedirect(appState: any): void {
...
+        status: isCurrentPage(pageNumber) ? PROGRESS_STATUS.INCOMPLETE : status,
+      }
+    })
+
+    // Build object to keep each section progress (and pages progress)
+    // with section status as combined statuses of pages
+    const progressInSection: ProgressInSection = {
+      code: section.code,
+      title: section.title,
+      canNavigate: isLinear ? section.index <= currentSection : true,
```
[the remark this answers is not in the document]
nmadruga: Yes, good point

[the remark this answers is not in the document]
nmadruga: Made a change to try to address this [here](https://github.com/openmsupply/application-manager-web-app/pull/127/commits/7e8a56e9dfa2ed3b1222dcce8c689afec6d28cd7)

Issue #124, opened 2020-11-16
Evaluate template element parameters in UI


CarlosNZ:
We've currently implemented the expression evaluator for visibility condition, required, editable and Validation.

However, we still need to use it to evaluate the actual parameters for each template element. (eg. "label", "placeholder", etc.)

CarlosNZ:
Also -- recommend doing #130 at the same time.

CarlosNZ:
Regarding priority -- we will need this if we want to have dynamically created option lists for drop-downs etc.

CarlosNZ:
Okay, having thought about this some more, I think I agree with you @andreievg that the plugin itself should evaluate any dynamic parameter fields. This is mainly because there are loads of possible parameters for plugins and for the majority it wouldn't make sense to have a dynamic value, and it'd be cumbersome to have to express all fields in that format unnecessarily.

Having said that, it does mean that different plugins can choose to interpret certain fields as dynamic or not independently. So one plugin might evaluate `label` as a dynamic query, whereas another one might not. (Of course, we should aim to be consistent as much as possible).

So, I would suggest that each plugin stores the names of the fields that should be evaluated dynamically in the `pluginConfig.json` file, and we have standardised boilerplate at the top of each plugin that handles the evaluation using a common method. Does that sound okay?

CarlosNZ:
One further question that arises from the previous comment:

Assuming some of the parameter fields can be defined as "dynamic expressions", should it be set for the whole plugin type (in which case, defining it in `pluginConfig.json` seems appropriate), or per template_action (i.e. each Action instance can define arbitrary parameters as dynamic), in which case it'd need to defined with an additional field in `template_action`?

nmadruga:
@CarlosNZ do you consider this is already done? Or anything remaining to be done regarding validation of parameters?

CarlosNZ:
@nmadruga -- No, it's not done. We'll need to do it soon if we want to be able to have dynamic choice lists in Drop-downs (for example)

CarlosNZ:
Okay, here is my detailed suggestion for how this should work:

- For each element plugin, the `pluginConfig.json` file gets an extra field called `dynamicParameters` where the parameters that should be supplied as dynamic expressions are defined.  
  - e.g. for infoText, the field might be just: `"dynamicParameters": [ "label" ]`  
  - for the dropdownChoice plugin, the field would probably be: `"dynamicParameters": [ "label", "options" ]`
[...]

PR #128, opened 2020-11-19
#87 Summary view in plugin

### LINKED ISSUE #87: Element plugin: ShortText - SummaryView
-- Create "SummaryView" component for ShortText plugin
-- Change "Summary" page to use this component for each element

CarlosNZ:
Fixes #87 and #68.

- Creates a new "SummaryViewWrapper" component and "SummaryView" in shortText plugin.
- Summary page only shows `isVisible` items
- Summary page shows error icon next to invalid responses
- Submit button and Edit icon only appear when application is 'Draft' status

Additionally:
- Added `parameters` to elements fragment and `isValid` responses fragment, as these are required in plugin display (also handy for Progress bar logic). (Note: the "Label" field for plugin display should be taken from `parameters.label`, not `title`, as it was previously)


src/components/Application/ApplicationSummary.tsx
```diff
@@ -1,17 +1,20 @@
...
   onSubmitHandler: () => void
+  appStatus: any
 }
 
 const ApplicationSummary: React.FC<ApplicationSummaryProps> = ({
   sectionsAndElements,
   onSubmitHandler,
+  appStatus: { status: status },
 }) => {
+  console.log('appStatus', status)
```
nmadruga, line 17: Please remove the output

src/elementPlugins/SummaryViewWrapper.tsx
```diff
@@ -0,0 +1,33 @@
...
+    parameters,
+    category,
+    code,
+    elementTypePluginCode: pluginCode,
+    isEditable,
+    isRequired,
+    isVisible,
+  } = element
+
+  if (!pluginCode || !isVisible || category === 'INFORMATION') return null
```
nmadruga, line 18: Maybe a good idea to add a comment here about why we are not rendering the INFORMATION type.
Also would be good to use this type as part of a map e.g.:
```
const questionType: { [key: string]: TemplateElementCategory } = {
  INFORMATION: 'INFORMATION',
  QUESTION: 'QUESTION' 
}
```

Or if that's simpler (actually I think this is simpler really) just import the `TemplateElementCategory` and use as `TemplateElementCategory.INFORMATION`. That's better that spreading the string. Harder to do maintanence or check for any type problems in the future.

src/elementPlugins/SummaryViewWrapper.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+  const { SummaryView }: PluginComponents = pluginProvider.getPluginElement(pluginCode)
+
+  // TO-DO: Provide a Default SummaryView (just label/text) if SummaryView not provided in plugin
```
nmadruga, line 22: Maybe a good idea to already create an issue to do that? Also, please when creating new issues use the label to link with the epic - so it's easier to track them to be listed inside the epic, so we know how much we have done in each area.

src/elementPlugins/pluginProvider.tsx
```diff
@@ -57,6 +57,7 @@ class pluginProvider {
 
 function getLocalElementPlugin(folderName: string) {
   const result: PluginComponents = {}
+  // TO-DO: optimize so it only imports the component type (Application, Template, Summary) that is required
```
nmadruga, line 60: Also another issue to be done? I recommend adding these with the `Epic#6 - Question plugins` tag.

src/elementPlugins/shortText/src/SummaryView.tsx
```diff
@@ -0,0 +1,17 @@
+import React from 'react'
+import { Header, Icon } from 'semantic-ui-react'
+import { SummaryViewProps } from '../../types'
+
+const SummaryView: React.FC<SummaryViewProps> = ({ parameters, value }) => {
+  return (
+    <>
```
nmadruga, line 7: better to use a Container here

src/elementPlugins/shortText/src/SummaryView.tsx
```diff
@@ -0,0 +1,17 @@
...
+      <Header as="h3" content={parameters.label} />
+      <p>
+        {!value?.isValid ? <Icon name="exclamation circle" color="red" /> : null}
+        {value?.text}
+      </p>
```
nmadruga, line 12:
```suggestion
<Header as="h3" content={parameters.label} />
      <Input
        disabled
        defaultValue={value?.text}
        icon={!value?.isValid ? <Icon name="exclamation circle" color="red" /> : null}
      />
```

It's better if we always use Semantic-ui element - so it's easier to move on to layout and styling soon.

src/elementPlugins/types.ts
```diff
@@ -22,6 +22,16 @@ interface ApplicationViewProps extends ApplicationViewWrapperProps {
...
 
+interface SummaryViewProps {
+  parameters: any
+  value: ResponseFull | null
+}
+
+interface SummaryViewWrapperProps {
+  element: any
+  value: ResponseFull | null
+}
```
nmadruga, line 33: Those two types are very similar now, would they require different input? If that's the case can it be more finely defined?

src/utils/hooks/useGetResponsesAndElementState.tsx
```diff
@@ -109,19 +109,19 @@ const useGetResponsesAndElementState = (props: useGetResponsesAndElementStatePro
     const isEditable = evaluateExpression(element.isEditable, evaluationParameters)
     const isRequired = evaluateExpression(element.isRequired, evaluationParameters)
     const isVisible = evaluateExpression(element.visibilityCondition, evaluationParameters)
-    // const isValid = evaluateExpression(element.validation, evaluationParameters)
+    // TO-DO: Evaluate element paremeters (in 'parameters' field, but unique to each element type)
```
nmadruga, line 112: I haven't understood this comment

src/utils/hooks/useGetResponsesAndElementState.tsx
```diff
@@ -71,7 +71,7 @@ const useGetResponsesAndElementState = (props: useGetResponsesAndElementStatePro
       const code = response?.templateElement?.code
       if (code) {
         currentResponses[code] = response?.value?.text
-        currentFullResponses[code] = response?.value
+        currentFullResponses[code] = { isValid: response?.isValid, ...response?.value }
```
nmadruga, line 74: Is this to make sure `isValid` isn't undefined? I wonder if there is a way to define that the return is never undefined.

nmadruga (commented):
Hey @CarlosNZ 
Just a couple of comments here. This is tidying the use of the newly added parameter - isValid (I'm not sure if this will be used in the progress structure though - I was getting from the elementState object) and adding the SummaryViewWrapper. Very good and needed.

src/elementPlugins/SummaryViewWrapper.tsx
CarlosNZ: Yup, issue exists -- #129. Have just added Epic tag.

src/elementPlugins/shortText/src/SummaryView.tsx
CarlosNZ: Sure, but this is not an `<Input>` element, and there doesn't seem to be a `<Text>` element like there is in Material-UI. I couldn't really see what the generic "text" element is for Semantic, so assumed there isn't one? Open to suggestions, but don't really think we should be using elements that are semantically incorrect.

src/elementPlugins/shortText/src/SummaryView.tsx
CarlosNZ: Yeah, but what Semantic component would you suggest? There didn't seem to be an appropriate one. Something like `Segment.Group` would seem appropriate around the whole list, but not sure what for individual elements. If there's no distinguishing boundary between them, does it need it?

src/utils/hooks/useGetResponsesAndElementState.tsx
CarlosNZ: What we were discussing today -- #124

src/elementPlugins/pluginProvider.tsx
CarlosNZ: The whole pluginProvider needs to be re-written (it's currently just placeholder as it has a hard-coded manifest), so these kind of improvements should be done with that. Have made an issue for the wider re-write: #133

src/utils/hooks/useGetResponsesAndElementState.tsx
CarlosNZ: No, it's so that the "Full" responses object has the full Response shape (in an object, `text` field, etc.) wheras the "response" just has the text string.

I think we're planning to refactor this at some point to only have 2 objects returned from this method rather than 3 at some point (#130) so this can be tidied up/got rid of in there.

src/elementPlugins/SummaryViewWrapper.tsx
CarlosNZ: Okay, good suggestion, ta. Done.

CarlosNZ:
> (I'm not sure if this will be used in the progress structure though - I was getting from the elementState object

I think that's ultimately the same thing in a more roundabout way. This `isValid` gets saved to "isValid" field of `template_element`, which is read by `useGetResponsesAndElementState` which updates the elementState object. :)

But yeah, it makes sense to get isValid from the elementState object -- it should always be up to date.

src/elementPlugins/types.ts
CarlosNZ: Okay, have made the second one more specific. For the first I've created a new type called "BasicObject", which is essentially an object of any shape, same as in back-end. Do you think this is necessary, or should we leave "parameters" as `any`. (Parameters is an object but properties could be anything)

src/elementPlugins/SummaryViewWrapper.tsx
nmadruga: Thanks!

src/elementPlugins/pluginProvider.tsx
nmadruga: Cool. No problem. Just wanted to make sure we don't leave the TODO unattended.

src/elementPlugins/shortText/src/SummaryView.tsx
nmadruga: I think the `Container` component is good. It doesn't add any spaces in the group it contains, I think that's alright for now.

src/elementPlugins/shortText/src/SummaryView.tsx
nmadruga: The benefits of using a existing UI component is that you already get the features of disabling, the icon inside the box, the border. I think the semantic is correct, showing an `Input` disabled for visibility, although I think you are also correct that this is not how it looks in the UI design. I suggest then creating your own component - and using the `Label` or `Header` to display the value filled by the user, and an optional `invalid` icon that shows next to it. So you make your own component, and pass the props to decide how it is presented. That's how I would do it.

src/elementPlugins/shortText/src/SummaryView.tsx
CarlosNZ: Container is for containing groups of elements, I'm not sure I understand why we'd want to wrap single elements in their own container.

I guess we should probably start making some standards/rules/style guides for this stuff cos I can see it getting very confusing and inconsistent. Things like making sure "Label" always gets a "Header" component (for example, probably not in actuality) and "Description" is always body text, etc. And what kind of Groups and Containers we wrap things in, etc. I guess we'll probably wait till the designs are close to the final versions to do that, right?

src/elementPlugins/types.ts
nmadruga: Yes. That's better, thank you. I think it would be good at some point to track all possible properties (that we allow plugins to have - maybe not the keys but the types of the objects' properties). But not for now.

src/elementPlugins/shortText/src/SummaryView.tsx
nmadruga: For the record, here is how the Summary view is currently UI is designed:
Screen Shot 2020-11-09 at 3 28 57 PM

src/utils/hooks/useGetResponsesAndElementState.tsx
nmadruga: Thank you

src/utils/hooks/useGetResponsesAndElementState.tsx
nmadruga: Ok. No problem. I was wondering what not just `currentFullResponses[code] = response` instead. But don't worry about this for now since it will be refactored soon : )

src/elementPlugins/shortText/src/SummaryView.tsx
CarlosNZ: Right, so from that I would interpret it to be that each Section would have a (collapsible) `Container`, each Page a `Segment.Group`,  then each summary item is just a (sub) `<Header>` and a `body` text. Or maybe just bold and standard body text for each item.

FWIW -- all the code examples on the Semantic-UI site just use `<p>` for body text.

But this seems more like a Style discussion, which is maybe a bit premature?

nmadruga (changes requested):
Thanks for the changes @CarlosNZ. 
I have just added some suggestion on the way to display the component on the Summary view, it should take much effort, just seems worth to get it tidier so it can be the base of next question plugins components to copy from.

src/elementPlugins/shortText/src/SummaryView.tsx
nmadruga: > I'm not sure I understand why we'd want to wrap single elements in their own container.

src/elementPlugins/shortText/src/SummaryView.tsx
nmadruga: No problem to wait for some future only style issue. 

I just thought that making it more similar to existing UI style that we could make (something that doesn't take more than 15 minutes effort) was worth. From my point of view it would also make the code easier to change and read. But don't worry, if you don't feel like doing it for now, it's fine. Was just a suggestion : )

nmadruga (approved):

nmadruga:
I saw one existing problem with the `Summary` button still showing after the application is submitted. But I'll create a separated issue for that. I think we need to show some intermediate page between when the application is submitted and then showing the summary page of the application with status COMPLETED.

src/elementPlugins/shortText/src/SummaryView.tsx
CarlosNZ: I agree in principle, I just think we should probably have a chat and establish some standards before going too far down this road. Thanks

src/elementPlugins/shortText/src/SummaryView.tsx
CarlosNZ: Trying Segment and Segment.Group, but they both style each question in their own "box", which is not what we want.

I think of Container in terms of Flexbox, and I wouldn't put Container around Flexbox Items unless they, in turn, contained things that needed formatting distinct from the parent.

I think I'll just leave as is now -- have made an issue to discuss these issues: https://github.com/orgs/openmsupply/projects/6#card-49947515

Issue #126, opened 2020-11-18
Plugin -- move "general" functionality to ApplicationViewWrapper


CarlosNZ:
Move boilerplate, including update methods, validation methods, etc. to Wrapper component.

Also -- as suggested [here](https://github.com/openmsupply/application-manager-web-app/pull/125#pullrequestreview-533922233)

- Password hashing logic
- Maybe second entry validation on keystroke could be optional via parameter ?
- When field A shows message for validation of other fields, error will persist unless something is entered in filed A (I think this will be dealt with here: #126)

And maybe, if time:
- Multiple validations

andreievg:
Proposal:

Plugin exposes a method `validate`, which is used by the wrapper.

Plugin ApplicationView **parameter change**:

`onUpdate(jsonValue)` <- is called on every keystroke etc..
`save(jsonValue)` <- is called when save needs to happen (i.e. onBlur)
`validationState` <- from validation method evaluation int the wrapper

Plugin **validate method**:

`Params`:
* jsonValue
* parameters <- from templateElememt
* evaluate() <- this method already has all of the object provided to it, i.e. `evaluate = (validationExpression) => { evaluateExpression(validationExpression, { objects: [thisResponse, ...allResponses], APIfetch: fetch }) }`

`Return`
* `{isValid, validationMessage, ..other things that plugin might need to pass from it's own validation method}` <- Passed through to plugin.


`Maybe second entry validation on keystroke could be optional via parameter ?` -> would need to store something in jsonValue to indicate second entry

PR #132, opened 2020-11-23
#126 Plugin generalisation & Dropdown plug

CarlosNZ:
Addresses #126 & #90.

Requires back-end PR/branch: https://github.com/openmsupply/application-manager-server/pull/125 to update the template.

Worked on #126 collaboratively with @andreievg, then just tweaked with a few bug-fixes tidy-ups.

In Summary:
- ApplicationViewWrapper performs validation check, but plugin *can* have additional validation logic if required for specific cases (still to do -- for just uses `defaultValidate`.
- Update and Save logic is also performed in the Wrapper, and those methods `OnUpdate`, `OnSave` are passed down to the plugin and the plugin decides when to call them (onChange, etc.)
- The plugin itself is responsible for:
   - rendering based on passed-in props/parameters
   - event handling (but mainly using events to call the above methods)
   - displaying and updating it's local "value" (user input), which is (probably) also sent to the Wrapper via onUpdate and onSave.

Also, created an initial version of "DropdownChoice" plugin, and added it to the UserRego template -- now all the questions in this template should be visible.

Note, I haven't really made any effort stylistically, it's a bare-bones Semantic selector. Just interested in Functionality at this point.


src/elementPlugins/ApplicationViewWrapper.tsx
```diff
@@ -1,30 +1,80 @@
...
+    setValidationState(validationResult)
+
+    return validationResult
+  }
+
+  const onSave = async (jsonValue: ResponseFull) => {
+    const validationResult: any = await onUpdate(jsonValue.text)
+    console.log('On Save', {
+      variables: {
+        id: currentResponse?.id as number,
```
nmadruga, line 61: Hmmm since we are doing the mutation here using the response `id` I don't think this can be `null`. Can we change the type to always expect the response to be of type `ApplicationResponse` please

nmadruga: After that many places that are using this variable in here (and all plugins `ApplicationView` functions) can stop using the `currentResponse?.`. We can start shaping the variables to be checked before they reach a certain point in the App, and then we don't need to keep checking for `undefined` or nulls

src/elementPlugins/dropdownChoice/src/SummaryView.tsx
```diff
@@ -0,0 +1,17 @@
+import React from 'react'
+import { Header, Icon } from 'semantic-ui-react'
+import { SummaryViewProps } from '../../types'
+
+const SummaryView: React.FC<SummaryViewProps> = ({ parameters, value }) => {
+  return (
+    <>
```
nmadruga, line 7: Better to use `Container` or `Segment.Group` here

src/elementPlugins/dropdownChoice/src/TemplateView.tsx
```diff
@@ -0,0 +1,16 @@
+import React, { useState } from 'react'
+import { Input } from 'semantic-ui-react'
+import { TemplateViewProps } from '../../types'
+
+// TODO type should be global
```
nmadruga, line 5: Is this still todo? What type should be global?

src/elementPlugins/dropdownChoice/src/TemplateView.tsx
```diff
@@ -0,0 +1,16 @@
...
+const TemplateView: React.FC<TemplateViewProps> = ({ parameters, onUpdate }) => {
+  const [placeholder, setPlaceholder] = useState(parameters.placeholder)
```
nmadruga, line 7:
```suggestion
  const [placeholder, setPlaceholder] = useState<string>(parameters.placeholder)
```

src/elementPlugins/dropdownChoice/src/TemplateView.tsx
```diff
@@ -0,0 +1,16 @@
...
+
+  const onChange = (_: any, { value }: any) => {
```
nmadruga, line 9:
```suggestion
  const onChange = (_: any, { value }: InputOnChangeData) => {
```

src/elementPlugins/shortText/src/ApplicationView.tsx
```diff
@@ -7,52 +7,25 @@ const ApplicationView: React.FC<ApplicationViewProps> = ({
...
     setValue(e.target.value)
   }
 
   function handleLoseFocus(e: any) {
     // TO-DO if password (i.e 'maskedInput'), do HASH on password before sending value
-    onUpdate({ value: { text: value }, isValid })
+    onSave({ text: value })
   }
 
   return (
```
nmadruga, line 31: This one is `Form.input`, should we simply change it to `Input` for now, since its form would be in an external component?

src/elementPlugins/shortText/src/ApplicationView.tsx
```diff
@@ -7,52 +7,25 @@ const ApplicationView: React.FC<ApplicationViewProps> = ({
...
-    if (isValid) setValidationMessageDisplay('')
-    else setValidationMessageDisplay(validationMessage)
-  }, [isValid])
+    onUpdate(value)
+  }, [])
 
   function handleChange(e: any) {
+    onUpdate(e.target.value)
     setValue(e.target.value)
   }
```
nmadruga, line 24: This is so different from what is in the `dropdownChoice/TemplateView.tsx`:

```
const onChange = (_: any, { value }: InputOnChangeData) => {
    setPlaceholder(value)
    onUpdate({ placeholder: value })
  }
```

Should we keep one pattern? I would rather the version from `dropdownChoice`

nmadruga: Similar to the one below: `handleLoseFocus`

src/utils/types.ts
```diff
@@ -98,7 +98,7 @@ type ResponseFull = {
   text: string | null | undefined
   optionIndex?: number
   reference?: any // Not yet decided how to represent
-  isValid: boolean | null
+  isValid?: boolean | null
```
nmadruga, line 101: Can this be null?

src/elementPlugins/dropdownChoice/src/ApplicationView.tsx
```diff
@@ -0,0 +1,63 @@
...
+    default: defaultIndex,
+  } = templateElement?.parameters
+
+  useEffect(() => {
+    onUpdate(value)
+  }, [])
+
+  function handleChange(e: any, data: any) {
+    const { value } = data
+    console.log('Event', data)
```
nmadruga, line 29: Missing `log`. Found!

nmadruga: Same comment as in `shortText`, better to keep one pattern for function definition. I most like this one from `dropdownChoice/TemplateView.tsx`:
```
const onChange = (_: any, { value }: InputOnChangeData) => {
    setPlaceholder(value)
    onUpdate({ placeholder: value })
  }
```

src/elementPlugins/dropdownChoice/src/ApplicationView.tsx
```diff
@@ -0,0 +1,63 @@
...
+    return {
+      key: option,
+      text: option,
+      value: option,
+      index,
+    }
+  })
+
+  return (
+    <>
```
nmadruga, line 45: `Container` or `Semantic.Group` here?

src/elementPlugins/dropdownChoice/src/ApplicationView.tsx
```diff
@@ -0,0 +1,63 @@
...
+      <Header content={label} />
+      {description}
+      <Dropdown
+        fluid
+        selection
+        defaultValue={options[defaultIndex]}
```
nmadruga, line 51: Don't know what could be wrong but this is not working as expected.
Before you open the dropdown it doesn't show the default value:
Screen Shot 2020-11-24 at 2 24 27 PM

Then after it opens the dropdown it is showing the default and then all the options after:
Screen Shot 2020-11-24 at 2 24 38 PM

nmadruga (changes requested):
Wow! I'm sorry I have been soo picky on you! 😜 
I totally agree that we can leave more style changes for later, but code-wise that's the chance we get to try to make the code the easier it can be for us during development and to maintain. That's why all these suggestions. I hope it doesn't seem unnecessary for you, but the clear the code is the easier will be for next iterations. Hopefully.

Other than that, it looks good. I'm glad you have worked in all of that. I'll have another go after the types are there to help me understand what are the changes related to issue #126 that was a little hard to see so far.

nmadruga:
Hey! One overall comment. Haven't we agreed on a **new** name for the plugins? I thin here will be a good place to change the folder's name to what was agreed (and I can't remember what it was).

src/elementPlugins/ApplicationViewWrapper.tsx
```diff
@@ -1,30 +1,80 @@
...
 
-  if (!pluginCode || !isVisible) return null
+  const { elementTypePluginCode: pluginCode } = templateElement
+  const { validation: validationExpression, validationMessage } = templateElement?.parameters
 
   const [responseMutation] = useUpdateResponseMutation()
+  const [pluginMethods, setPluginMethods] = useState({
+    validate: (validationExpress: any, validationMessage: any, evaluator: any) =>
+      console.log('notLoaded'),
+  })
```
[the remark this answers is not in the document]
CarlosNZ: I'm not sure exactly what you mean, perhaps you can show me tomorrow. 

Sorry -- some of this code is @andreievg 's since we did it together, so I'm not entirely clear about some of his syntax.

src/elementPlugins/ApplicationViewWrapper.tsx
```diff
@@ -1,30 +1,80 @@
...
+  const [validationState, setValidationState] = useState<any>({})
```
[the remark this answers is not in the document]
CarlosNZ: Blame @andreievg 😜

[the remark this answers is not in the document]
CarlosNZ: Okay, have updated types. There's still one `any` at line 48 that I can't seem to fix. The type should be `ValidationState` as its what the `defaultValidate` method returns, but typescript is complaining that it is type `void`. Don't really understand this, any ideas?

src/elementPlugins/ApplicationViewWrapper.tsx
```diff
@@ -1,30 +1,80 @@
...
+
+  useEffect(() => {
+    if (!pluginCode) return
+    // TODO use generic or plugin specific
+    setPluginMethods({
+      validate: defaultValidate,
+    })
+  }, [])
+
+  const onUpdate = async (value: string | null | undefined) => {
```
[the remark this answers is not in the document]
CarlosNZ: Okay, have defined a `LooseString` type.

src/elementPlugins/shortText/src/ApplicationView.tsx
```diff
@@ -7,52 +7,25 @@ const ApplicationView: React.FC<ApplicationViewProps> = ({
...
-  useEffect(() => {
-    setResponses({ thisResponse: value, ...allResponses })
-  }, [value])
-
-  useEffect(() => {
-    if (isValid) setValidationMessageDisplay('')
-    else setValidationMessageDisplay(validationMessage)
-  }, [isValid])
+    onUpdate(value)
+  }, [])
```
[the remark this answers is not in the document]
CarlosNZ: Yes, that's what is intended -- once on load to populate the validationState of the element.

src/elementPlugins/ApplicationViewWrapper.tsx
```diff
@@ -1,30 +1,80 @@
...
+        value: jsonValue,
+        isValid: validationResult.isValid,
+      },
     })
   }
 
-  const newProps = { evaluator: evaluateExpression, onUpdate, ...props }
+  if (!pluginCode || !isVisible) return null
+
+  const newProps = { onUpdate, onSave, validationState, ...props }
```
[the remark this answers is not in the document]
CarlosNZ: Ah good, I was trying to do something like this but couldn't get the syntax right. 👍

src/elementPlugins/ApplicationViewWrapper.tsx
CarlosNZ: I'm not sure I understand. The `currentResponse` is typed as `ApplicationResponse | null` (in types.ts) and it *can* be `null` as all responses start off as `null`, right? Maybe you can show me what you mean?

src/elementPlugins/dropdownChoice/src/SummaryView.tsx
CarlosNZ: Again, both those components stylise the element in their own box, which is not what we want for the Summary page (according to current designs)

src/elementPlugins/dropdownChoice/src/TemplateView.tsx
CarlosNZ: I don't know  -- this is left over from Andrei's original prototype, I haven't changed this at all, it's just copied from shortText folder. AFAIK, we're not really dealing with the TemplateViews at the moment, as that will be for the template builder.

src/elementPlugins/dropdownChoice/src/TemplateView.tsx
CarlosNZ: Same here. Ignore TemplateView for now.

src/elementPlugins/dropdownChoice/src/TemplateView.tsx
CarlosNZ: Same.

src/elementPlugins/shortText/src/ApplicationView.tsx
CarlosNZ: Form.input and Input have slightly different properties, and this is what Andrei configured originally. I tried Input, and it wasn't as appropriate for some reason (though I can't now remember what the reason was). This will be the kind of thing we establish in our ongoing style/layout discussion.

src/elementPlugins/shortText/src/ApplicationView.tsx
CarlosNZ: We didn't touch TemplateView -- what we've done in ApplicationView is intended to be the standard.

src/utils/types.ts
CarlosNZ: I think so. All Response parameters start off as NULL when the Application is created, right?

src/elementPlugins/shortText/src/ApplicationView.tsx
CarlosNZ: I think 🤷‍♂️

src/elementPlugins/dropdownChoice/src/ApplicationView.tsx
CarlosNZ: See comment above.

src/elementPlugins/dropdownChoice/src/ApplicationView.tsx
CarlosNZ: See above.

src/elementPlugins/dropdownChoice/src/ApplicationView.tsx
CarlosNZ: Yes, good point. I've made an change now to set the `value` to the default on load. I'm not actually sure what the `defaultValue` prop does, it seems like whatever the "value" prop is determines what shows. 

Also, the back-end didn't actually have a default set for this question, and I've made it so there's nothing pre-selected if no default (as opposed to the first item). I think this is the correct behaviour, since we need to be able to have nothing selected sometimes to force the user to make a conscious choice.

CarlosNZ:
> Hey! One overall comment. Haven't we agreed on a **new** name for the plugins? I thin here will be a good place to change the folder's name to what was agreed (and I can't remember what it was).

I think it was "Form Element Plugins". Does that sound right? How about we wait till this is approved and then I'll change the folder name just before merging? Just in case renaming the folder makes all its files show as diff.

src/elementPlugins/shortText/src/ApplicationView.tsx
```diff
@@ -7,52 +7,25 @@ const ApplicationView: React.FC<ApplicationViewProps> = ({
...
-
-  useEffect(() => {
-    setResponses({ thisResponse: value, ...allResponses })
-  }, [value])
-
-  useEffect(() => {
-    if (isValid) setValidationMessageDisplay('')
-    else setValidationMessageDisplay(validationMessage)
-  }, [isValid])
+    onUpdate(value)
```
andreievg, line 18: Can and should we move this to wrapper ?

src/elementPlugins/shortText/src/ApplicationView.tsx
andreievg: @nmadruga i actually used Form.input because I saw some one using it for the original form (I think it was you). it might give us more control over layout

src/elementPlugins/dropdownChoice/src/ApplicationView.tsx
```diff
@@ -0,0 +1,64 @@
...
+    placeholder,
+    options,
+    default: defaultIndex,
+  } = templateElement?.parameters
+
+  const [value, setValue] = useState(initialValue ? initialValue?.text : options[defaultIndex])
+
+  useEffect(() => {
+    onUpdate(value)
+    if (!initialValue) onSave({ text: value, optionIndex: options.indexOf(value) })
```
andreievg, line 26: hmm any reason to save on start up ? Wouldn't it be exact same value as already stored ? Or is it for validation ?

src/elementPlugins/dropdownChoice/src/ApplicationView.tsx
```diff
@@ -0,0 +1,64 @@
...
+}) => {
+  const {
+    label,
+    description,
+    placeholder,
+    options,
+    default: defaultIndex,
+  } = templateElement?.parameters
+
+  const [value, setValue] = useState(initialValue ? initialValue?.text : options[defaultIndex])
```
andreievg, line 22: what would be the difference between condition operator and `initialValue?.text || options[defaultIndex]`

src/elementPlugins/dropdownChoice/src/ApplicationView.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+  useEffect(() => {
+    onUpdate(value)
+    if (!initialValue) onSave({ text: value, optionIndex: options.indexOf(value) })
+  }, [])
+
+  function handleChange(e: any, data: any) {
+    const { value } = data
+    onUpdate(value)
```
andreievg, line 31: do we need on update if onSave is called ?

andreievg (commented):
Just some questions (which may raise change requests), overall looks mergeable

src/elementPlugins/ApplicationViewWrapper.tsx
nmadruga: No problem, doesn't need to be done here. I've been working with to reuse more common types between the components (and ApplicationViewWrapper) and I can consider adding this in there.

src/elementPlugins/shortText/src/ApplicationView.tsx
CarlosNZ: It doesn't seem to work. I think something to do with the fact that `value` here is local to this component, and so we'd have to use `props.initialValue` in the Wrapper, and it's not behaving correctly when I do. Maybe just leave as is for now?

src/elementPlugins/dropdownChoice/src/ApplicationView.tsx
CarlosNZ: It's because the DropDown can have a default selection -- in which case we need to save this value immediately, as it's already got a valid response before the user has even done anything.

The shortText plugin currently doesn't handle a default value, but if it did, we'd need to do the same for that too.

Although, I've now changed it to actually check there is a default value provided before saving:
```
if (!initialValue && defaultIndex) onSave({ text: value, optionIndex: options.indexOf(value) })
```

src/elementPlugins/dropdownChoice/src/ApplicationView.tsx
CarlosNZ: Yeah, I guess that works and is a bit more concise. Thanks.

src/elementPlugins/dropdownChoice/src/ApplicationView.tsx
CarlosNZ: Yes, because `onUpdate` regenerates `allResponses` based on the new value, which we need. :)

src/elementPlugins/shortText/src/ApplicationView.tsx
andreievg: Sure

src/elementPlugins/dropdownChoice/src/ApplicationView.tsx
andreievg: Doesn't onSave call onUpdate equivalent ?

CarlosNZ: Ah yes, oops didn't notice that. Have removed extra onUpdate. :)  ac83836

src/elementPlugins/shortText/src/ApplicationView.tsx
andreievg: Just realised this would show validation error that wont' necessarily be reflected in progress bar, which would require save, which might lead to some side effects. Need to have a deeper think about it.

The scenario I was thinking about, is this: Everything is valid, but user goes back a page, changes something that will affect validation in further page (page 3), loads page 3, invalid status would show on the question but is not saved (so not shown in progress bar etc..)

nmadruga (approved):

nmadruga:
Please remember to renamed the folder when you finish this one, otherwise we might forget about it later.

Issue #95, opened 2020-11-06
Epic #12 - Login/Logout


nmadruga:
After registered with Email confirmed the Applicant can login.
During the login if the user has companies associated there is a second step to select a company.
All pages now should be verified for authentication & user permission.
The JWT token retrieved from Backend is stored as persisted even if page reloads. It should be available for X time.
After logged in user can see the Dashboard which should direct to 2 possible options: Join a company or create a company.
On the top user menu user can Sign out.

PR #144, opened 2020-11-26
#115 Login page/functionality

### LINKED ISSUE #115: Create login view for user
UI - Should be a 2 step login:
* First page where user enters `username` and `password` -> What is aimed for this issue
* Second page where we list user's companies and the user selects a company to login (To be done in another issue!)

CarlosNZ:
Fixes #115.

**Note**: requires branch from [backend PR 129](https://github.com/openmsupply/application-manager-server/pull/129) as it includes a new, simplified user registration template

Functionality as follows:
- Check for JWT (`isLoggedIn`) at top level, and if not present redirect to `/login` (Main site content wrapped in new "AuthenticatedContent" wrapper, which only returns main site components if logged in, else redirect to /login.
- Login page: checks username and password using `login` endpoint, and if correct, saves JWT to localStorage. Then redirects to Home
- Logout button: clears all local storage and forces a reload.
- ~~Placeholder for `/register` page.~~
- UPDATE: Clicking "Create new account" logs user in as the special 'nonRegistered' user (which should have minimal privileges) and re-directs to UserRegistration application. Do we think this is the best way to handle this?

~~Issue to solve: How do we show "UserRego" form on `/register` page without loading other things within the "AuthenticatedContent" wrapper?~~ (See UPDATE above)


src/containers/Main/SiteLayout.tsx
```diff
@@ -26,6 +26,7 @@ import { ApplicationCreate, ApplicationList, ApplicationPageWrapper } from '../A
 import UserRegister from '../User/UserRegister'
 import { ApplicationProvider } from '../../contexts/ApplicationState'
 import ApplicationOverview from '../Application/ApplicationOverview'
+import isLoggedIn from '../../utils/helpers/loginCheck'
```
nmadruga, line 29: Not used here I think?

src/containers/User/UserArea.tsx
```diff
@@ -47,6 +48,9 @@ const UserArea: React.FC = () => {
               {user}
             </Button>
           ))}
+        <Button basic color="blue" onClick={logOut}>
+          Log out
+        </Button>
```
nmadruga, line 53: This made me think we need a issue just to work on the UserArea layout. Not a big priority but needs to be done. Do you mind creating this issue please? I'm planning to go through all issues and epics on the weekend so we can try visualising our milestones and what need to be done in the next future.

src/containers/User/UserRegister.tsx
```diff
@@ -1,166 +1,18 @@
...
-  isErrorMessage: boolean
-}
-
-interface User {
-  username: string
-  password: string
-  email: string
-}
+import React from 'react'
+import { useHistory } from 'react-router'
```
nmadruga, line 2: Could you use the `useRouter` instead? There all these hooks are combined in one place. I think it's easier to have it all contained in one place like that. You can destruct the `push` to be used here.

src/containers/User/UserRegister.tsx
```diff
@@ -1,166 +1,18 @@
...
-          success
-          hidden={!snackbar.showMessage}
-          header={snackbar.messageTitle}
-          content={snackbar.messageText}
-          error={snackbar.isErrorMessage}
-        />
-      </Segment>
-    </Segment.Group>
-  )
+  return <p>Re-directing to user registration application...</p>
```
nmadruga, line 15: I tried going to the login page and stopping the server. It doesn't show any problems on the login page - which I think is fine, but the problem is when I click the `Create new account` icon, it is showing this message.

Maybe we just need to deal with this checking - server is unreachable in any page should display some icon or take the user to unreachable page. I think this can also be another issue. Do you have more thought around this?

src/containers/User/Login.tsx
```diff
@@ -0,0 +1,95 @@
+import React, { useState } from 'react'
+import { useHistory } from 'react-router'
```
nmadruga, line 2: useRouter here too?

src/containers/User/Login.tsx
```diff
@@ -0,0 +1,95 @@
...
+    event.preventDefault()
+    const passwordHash = hashPassword(password)
+    const loginResult = await attemptLogin(username, passwordHash)
+    if (!loginResult.success) setIsError(true)
+    else {
+      setIsError(false)
+      // localStorage.setItem('username', username)
+      // localStorage.setItem('templatePermissions', loginResult.templatePermissions)
+      localStorage.setItem('persistJWT', loginResult.JWT)
+      console.log('Log in successful!')
```
andreievg, line 24: good time to set username and templatePermissions

https://github.com/openmsupply/application-manager-web-app/blob/0da5d6491908ef9676537689e8be4319892d8ad1/src/utils/hooks/useGetUserPermissions.ts#L21-L22

src/containers/User/Login.tsx
```diff
@@ -0,0 +1,95 @@
...
+        </Grid.Column>
+      </Grid>
+    </Container>
+  )
+}
+
+export default Login
+
+function hashPassword(password: string) {
+  // TO-DO Implement password hashing
```
nmadruga, line 76: Maybe create an issue for this already? Link to Epic Login/logout #95. Thanks!

src/utils/helpers/loginCheck.ts
```diff
@@ -0,0 +1,6 @@
+const isLoggedIn = () => {
+  const JWT = localStorage.getItem('persistJWT')
```
andreievg, line 2: i think we need to check for either JWT being present or username in JWT = 'nonRegistered' (since even nonRegistered user will have persisted JWT,  can ask me why that is, I haven't documented that bit yet)

nmadruga (approved):
Looking pretty. And simple. That's great!

I made just few comments, more regarding other issues to create and just some minor changes. 👍

src/containers/Main/SiteLayout.tsx
andreievg, line 29: not used I think

andreievg (changes requested):
Looks good, very precise. Haven't tested yet, a couple of minor requests.

src/containers/User/UserArea.tsx
CarlosNZ: 👍 #148

src/containers/User/UserRegister.tsx
CarlosNZ: Got it. I had trouble using it before, but I think I get how it works now.

src/containers/User/UserRegister.tsx
CarlosNZ: Good call -- I've added some simple error handling and now the page will display the network error message if there is one.

We can add some fancier error pages later on, I guess (along with 404 pages etc.)

src/containers/User/Login.tsx
CarlosNZ: 👍 #149

src/utils/helpers/loginCheck.ts
CarlosNZ: Yeah, good point. And we should find a way to make sure we nuke the localStorage for the "nonRegistered" user after user registration form is submitted.

src/containers/User/UserRegister.tsx
CarlosNZ: Hey @nmadruga , The match, history, location and params properties weren't defined in the `RouterResult` interface in `useRouter`, so using them caused a type error. I've added them in, but can't figure out what their type should be other than `any`. Any ideas?

src/utils/helpers/loginCheck.ts
CarlosNZ: Have added username to localStorage, but as we mentioned in meeting -- we need to allow 'nonRegistered' JWT to be considered "logged in", or it won't even be able to load the UserRego page.

Have made an issue to log out nonRegistered on Application Submit though. #150

andreievg (approved):

Issue #143, opened 2020-11-26
Unify types and reuse of evaluated responses object on SummaryView and ApplicationView 


nmadruga:
Reuse what was done in **hook** `useGetElementsAndResponsesQuery` (to return an object with one `elementState` type per element for the application and the evaluated responses in other objects `ResponseFull` and `ResponsesByCode`types) in a new **hook** `useGetSectionElementAndResponses` to the a very similar thing for a section.
The idea is to merge this two hooks one day if possible, but for now only doing very similar things in both and calling a unique helper to evaluate all elements that were queried by each hook.
This same hook would be used to list all **elements & responses** in a single page and in the summary, which would required changing the types currently received by `ApplicationViewWrapper` and SummaryViewWrapper`.

CarlosNZ:
Yeah, definitely some functionality regression here compared with #127. The progress bar PR seems to be working fine, but in this one the on-page validation is broken. There's no error message popping up on invalid fields, and the "Next" button doesn't seem to move the page forward whether or not the fields are valid.

PR #151, opened 2020-11-30
#143 unify types and use single hook for application page

nmadruga:


src/utils/types.ts
```diff
@@ -50,12 +50,19 @@ interface AppStatus {
   outcome: string
 }
 
+interface ElementAndResponse {
+  element: ElementState,
+  response: ResponseFull
+}
```
CarlosNZ, line 56: This is not being used anymore, is it?

src/utils/helpers/validatePage.tsx
```diff
@@ -41,16 +41,17 @@ interface validatePageProps {
...
 }: validatePageProps): ProgressStatus => {
   let count = 1
 
   // Filter visible elements in the current page
   const visibleQuestions = Object.values(elementsState)
-    .filter(({ section, isVisible }) => isVisible && section === sectionIndex)
+    .filter(({ sectionIndex, isVisible }) => isVisible && currentSectionIndex === sectionIndex)
     .reduce((pageElements: ElementState[], element) => {
-      if (element.elementTypePluginCode === 'pageBreak') count++
+      // TODO: Use the page property from ElementState
```
CarlosNZ, line 53: Did you want to do this for this PR? Probably makes sense if you have time.

CarlosNZ (commented):
Great! (I would say that, I was there 😉 )

Just wondering if you want to quickly do the "TO-DO" you left. I think it makes sense to do it as part of this.

src/utils/types.ts
CarlosNZ: And if it is -- that comma is unnecessary (VSCode should remove it anyway)

CarlosNZ:
Oh, there's one other bug we've inadvertently caused here.

When going back to a page that had previously been filled in, the previous responses are missing. They're being saved to the database okay, but not showing up in the UI.

CarlosNZ:
> Oh, there's one other bug we've inadvertently caused here.
> 
> When going back to a page that had previously been filled in, the previous responses are missing. They're being saved to the database okay, but not showing up in the UI.

Okay, fixed this in d17d9c1.

src/utils/types.ts
CarlosNZ: Have removed, along with SectionPages and SectionElements, which are no longer used.

src/utils/helpers/validatePage.tsx
CarlosNZ: Done this in 994f63f

src/utils/types.ts
nmadruga: Yep. Cool! Agreed

nmadruga:
I'll have another look on this PR and check if can be merged.

nmadruga:
There is a problem with going back to the previous page -> The plugins don't get loaded at first. I need to reload the page to make it load correctly. I've created another issue to solve that: #155.

CarlosNZ:
> There is a problem with going back to the previous page -> The plugins don't get loaded at first. I need to reload the page to make it load correctly. I've created another issue to solve that: #155.

It's already solved! 😜
https://github.com/openmsupply/application-manager-web-app/pull/160

nmadruga:
That's nice : )

Can you approve this PR to merge please?

CarlosNZ (approved):

PR #152, opened 2020-11-30
#130 One ResponsesByCode object instead of two

### LINKED ISSUE #130: useGetResponsesAndElementState hook -- should only return 2 objects, not 3
Currently it returns:
- responsesByCode ([code]:[text])
- responsesFullByCode (same, but full response object rather than text)
- elementsState (evaluated states (isVisible etc) by code)

These last two can be combined together.

Recommend doing this issue at the same time as #126 since they'll be poking around in the same area.

CarlosNZ:
Fixes #130. 

Needs back-end PR/branch: https://github.com/openmsupply/application-manager-server/pull/130

Removed old "responsesByCode" object and type, and renamed the previous "responsesFullByCode" to just "responsesByCode". It now contains the "full" response objects rather than just the `.text` string.

NOTE: This PR has latest master merged into it, so contains changes not currently in #143. Probably best to review and merge that first, or wait till that branch has had master merged into it, otherwise this diff is too big.


nmadruga (approved):
LGTM!

Issue #155, opened 2020-11-30
Problem to load elements when going to previous page (or clicking the edit button)


nmadruga:
It seems like a problem with loading the formElementPlugin. The reason needs to be investigated, it started to be visible (or may be related to changes on [PR#143](#151)

Reproducible:
1. Start a new application (User registration)
2. Fill all fields in first page, click Next
3. After loading the Summary, click to go back (on the browser)
4. See problem:
<img width="1425" alt="Screen Shot 2020-12-01 at 11 34 51 AM" src="https://user-images.githubusercontent.com/16461988/100674842-b51e3c00-33ca-11eb-9e73-87c5979a2fb4.png">

Issue #153, opened 2020-11-30
Bug fix -- Dropdown plugin -- default selection doesn't affect visibility condition of other elements


CarlosNZ:
Currently, the visibility condition of Q7/8/9 of the Test Rego application is dependent on the answer to Q6. The default for Q6 is "Manufacturer", which shows up correctly on page load. However, Q7 does NOT appear, which it should if Q6 has selected "Manufacturer". It only appears if you select something else in Q6, then switch it back.

PR #160, opened 2020-11-30
#153 #155 Bug fixes

CarlosNZ:
2 small fixes -- #153 & #155

**NOTE**: If it's not already merged, this should go with back-end PR: https://github.com/openmsupply/application-manager-server/pull/130

#153 was due to the old "0 === false" trap in javascript. Needed to change to an explicit `defaultIndex !== undefined` check, because "defaultIndex" can often be 0, which would return false. :(

#155 There's probably a more thorough fix involving making sure `validationState` is properly calculated before rendering the child plugin. However, for now, having a fall-back (`isValid: true`) options will do the trick.

### DIFF
--- a/src/formElementPlugins/ApplicationViewWrapper.tsx
+++ b/src/formElementPlugins/ApplicationViewWrapper.tsx
@@ -82,6 +82,7 @@ const ApplicationViewWrapper: React.FC<ApplicationViewWrapperProps> = (props) =>
             onUpdate={onUpdate}
             onSave={onSave}
             validationState={validationState || { isValid: true }}
+            // TO-DO: ensure validationState gets calculated BEFORE rendering this child, so we don't need this fallback.
             {...props}
           />
         }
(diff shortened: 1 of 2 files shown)


src/formElementPlugins/ApplicationViewWrapper.tsx
```diff
@@ -81,7 +81,7 @@ const ApplicationViewWrapper: React.FC<ApplicationViewWrapperProps> = (props) =>
           <ApplicationView
             onUpdate={onUpdate}
             onSave={onSave}
-            validationState={validationState}
+            validationState={validationState || { isValid: true }}
```
nmadruga, line 84: Maybe a good idea to add a comment here to explain. From my understanding there is still a problem happening, we just don't want to get much involved with the solution for now.

nmadruga (commented):
Just a comment, I don't have the details of what is causing the problem, but it might be good to describe here as a comment - so we know what to fix in the near future.

src/formElementPlugins/ApplicationViewWrapper.tsx
CarlosNZ: Yeah okay. Done now in 4e35cdf

```
// TO-DO: ensure validationState gets calculated BEFORE rendering this child, so we don't need this fallback.
```

nmadruga (approved):
Thanks for adding the comment. Approved now

PR #164, opened 2020-12-01
#162 show asterix on required question

### LINKED ISSUE #162: Automatically add asterisk in required elements fields on the UI

nmadruga:
Fixes #162 

Very minor! Just showing one asterisk after the label on required questions.
Also shows a label on the pages (with visible, required questions) explaining what the asterisk means.

[EDITED - 02/12/2020]

- Using `required` property from semantic-ui form elements on `ApplicationViewWrapper` and `SummaryViewWrapper`. 
- Updated `ApplicationView` and `SummaryView` files of **shortText** as example on how the asterisk would show.


CarlosNZ (changes requested):
Hey, I hate to be a pain the butt, but I don't think we should be hard-coding the asterisk into the label field all the way up in the `useGetResponsesAndElementState` hook. 

There is already an HTML and Semantic flag for "required" form fields, so all we need to do is turn this on in each form field and it'll be rendered automatically: https://react.semantic-ui.com/collections/form/#field-variations-required-field 

(The `Form.Field` component will probably be able go in ApplicationViewWrapper and SummaryViewWrapper around the plugin)

Also, when we come to dynamically evaluate some of these parameters (like label) #124 the evaluation won't take place until (probably) the plugin view or plugin wrapper components, so it won't work to hard-code the string further up.

CarlosNZ:
Just FYI -- it looks like the built-in "required" flag needs to be used with the component structure shown in the demo (Form->Form.Field->Label/Input) so we'll need to re-factor the ApplicationView for ShortText to use `<Input>` and `<Label>` rather than `<Form.Input>`. Maybe limit this issue to just the higher-level changes (refactoring, page required check, etc.) and do the Form/Field/Input re-factoring in a different issue?

nmadruga:
> Hey, I hate to be a pain the butt, but I don't think we should be hard-coding the asterisk into the label field all the way up in the useGetResponsesAndElementState hook.

Not at all. I'm glad you gave this suggestion. My thinking was - since we aren't using the Form component settings to do any validation why to use some property built-in to do this. Thanks so much for looking into this. It seems so much better!

> (The Form.Field component will probably be able go in ApplicationViewWrapper and SummaryViewWrapper around the plugin)

I'll try this out now.

> Maybe limit this issue to just the higher-level changes (refactoring, page required check, etc.) and do the Form/Field/Input re-factoring in a different issue?

No problem. I'll do that, perhaps already do this change on the [PR#142](#166) to update the **shortText** to be using the Input on the plugin as you suggest. So this PR only contains the changes to the wrapper.

src/containers/Application/ApplicationPageWrapper.tsx
```diff
@@ -7,9 +7,11 @@ import useGetResponsesAndElementState from '../../utils/hooks/useGetResponsesAnd
 import { ElementsBox, NavigationBox } from './'
 import validatePage, { getCombinedStatus, PROGRESS_STATUS } from '../../utils/helpers/validatePage'
 import { SummarySectionCode } from '../../utils/constants'
+import getPageElements from '../../utils/helpers/getPageElements'
```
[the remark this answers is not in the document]
nmadruga: Not required. I just thought it would be better to be done outside this presentation component. I can remove this change if you think would conflict too much with other work.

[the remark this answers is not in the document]
CarlosNZ: No it's all good. :)

src/formElementPlugins/ApplicationViewWrapper.tsx
```diff
@@ -74,18 +75,24 @@ const ApplicationViewWrapper: React.FC<ApplicationViewWrapperProps> = (props) =>
...
-            validationState={validationState || { isValid: true }}
-            // TO-DO: ensure validationState gets calculated BEFORE rendering this child, so we don't need this fallback.
-            {...props}
-          />
-        }
+        {isRequired ? (
+          <Form.Field required>{PluginComponent}</Form.Field>
+        ) : (
+          <Form.Field>{PluginComponent}</Form.Field>
+        )}
```
CarlosNZ, line 95:
```suggestion
<Form.Field required={isRequired}>{PluginComponent}</Form.Field>
```
Bit simpler, right? ;)

src/formElementPlugins/shortText/src/ApplicationView.tsx
```diff
@@ -30,9 +30,9 @@ const ApplicationView: React.FC<ApplicationViewProps> = ({
 
   return (
     <>
+      <label>{label}</label>
       <Form.Input
```
CarlosNZ, line 34: Oh, okay, so you can leave it as `<Form.Input>` rather than `<Input>`. Do you know what the difference is?

CarlosNZ (changes requested):
Nice, just one suggested change.

src/formElementPlugins/SummaryViewWrapper.tsx
```diff
@@ -15,10 +15,18 @@ const SummaryViewWrapper: React.FC<SummaryViewWrapperProps> = (props) => {
...
     <ErrorBoundary pluginCode={pluginCode}>
       <React.Suspense fallback="Loading Plugin">
-        <SummaryView parameters={parameters} value={value} />
+        {isRequired ? (
+          <Form.Field required inline>
+            {PluginComponent}
+          </Form.Field>
+        ) : (
+          <Form.Field inline>{PluginComponent}</Form.Field>
+        )}
```
CarlosNZ, line 29:
```suggestion
<Form.Field required={isRequired} inline>{PluginComponent}</Form.Field>
```
Same suggestion as other PR.

CarlosNZ (changes requested):

src/formElementPlugins/shortText/src/ApplicationView.tsx
nmadruga: I think the main difference is that with the Input we have to change how we do the validation - it expects only a `boolean` for the error and then another label on the bottom with the `validationMessage`. I have changed to be like this, and then changed back to use Form.Input, so there was something else. I'll check what it was.

src/formElementPlugins/SummaryViewWrapper.tsx
nmadruga: Nice! I wasn't sure if I could do that. Thanks

src/formElementPlugins/shortText/src/ApplicationView.tsx
nmadruga: Here is the difference:

With `<Input>`

Screen Shot 2020-12-02 at 4 48 16 PM

With `<Form.Input>`

Screen Shot 2020-12-02 at 4 43 52 PM

src/formElementPlugins/ApplicationViewWrapper.tsx
```diff
@@ -88,11 +88,7 @@ const ApplicationViewWrapper: React.FC<ApplicationViewWrapperProps> = (props) =>
   return (
     <ErrorBoundary pluginCode={pluginCode}>
       <React.Suspense fallback="Loading Plugin">
-        {isRequired ? (
-          <Form.Field required>{PluginComponent}</Form.Field>
-        ) : (
-          <Form.Field>{PluginComponent}</Form.Field>
-        )}
+<Form.Field required={isRequired}>{PluginComponent}</Form.Field>
```
CarlosNZ, line 91: Indentation problem. Saving should fix the formatting.

src/formElementPlugins/SummaryViewWrapper.tsx
```diff
@@ -20,13 +20,7 @@ const SummaryViewWrapper: React.FC<SummaryViewWrapperProps> = (props) => {
...
     <ErrorBoundary pluginCode={pluginCode}>
       <React.Suspense fallback="Loading Plugin">
-        {isRequired ? (
-          <Form.Field required inline>
-            {PluginComponent}
-          </Form.Field>
-        ) : (
-          <Form.Field inline>{PluginComponent}</Form.Field>
-        )}
+<Form.Field required={isRequired} inline>{PluginComponent}</Form.Field>
```
CarlosNZ, line 23: Indentation again. Is this because you used "Commit suggestion" button?

CarlosNZ (approved):
Indentation issue. Saving it again in VSCode fixes it for me.

Approving cos you can merge it as soon as this is done.

src/formElementPlugins/shortText/src/ApplicationView.tsx
CarlosNZ: OH, okay. That's weird, because we were already using `<Form.Input>` and NOT getting the red borders, right? 🤷‍♂️

But looks good 👍

Issue #158, opened 2020-11-30
Summary Wrapper -- validation


nmadruga:
The validation should run on the plugins SummaryView and display correctly on the summary page - for application (for DRAFT application),

CarlosNZ:
Here is my understanding of how the Summary page should handle validation etc. Please correct me before I get to far into this issue (and #48 which I may do at the same time).

When loading the Summary page, each question has an `isValid` value of either `true`, `false`, or `undefined`, which is loaded using ApolloQuery. We also create a local boolean called `isStrict`, which is initially set to `false`.

The SummaryViewWrapper for each question then **re-runs** validation.
- If result is `true`/`false`, write result to database if changed (shouldn't happen very often).
- If undefined, leave as is.

The SummaryViewWrapper passes this value to each SummaryView plugin, which displays error state for `false` and normal display for `true` or `undefined`. (Error state is an error icon, plus the ValidationMessage)

When the user clicks "Submit", we change the `isStrict` value to `true`. The SummaryView entities then re-render and any that are `undefined` (and required) now show in error state.

If all required questions pass strict validation, the form Submission proceeds, else an error message shows along with the updated error states.

The actual **re-calculation** of validity is only done during page load. The rest of the logic on this page just uses the result of that calculation.

Note that `isStrict` is a local value only, so will be reset if refreshing or navigating away from the page.

andreievg:
Agree with all

> When the user clicks "Submit", we change the isStrict value to true. The SummaryView entities then re-render and any that are undefined (and required) now show in error state.

I think showing required but not filled in, should be responsibility of wrapper ?

nmadruga:
> When the user clicks "Submit", we change the isStrict value to true. The SummaryView entities then re-render and any that are undefined (and required) now show in error state.

I think the idea to use `isStrict` as a local variable is still undecided. We might want to keep showing the fields with problems to the user after they enter the **strict mode**. So the alternative would be to send the validation to the server once the submit button is clicked, and then the other pages would show the invalid icon on `undefined` fields too.

Issue #142, opened 2020-11-26
Linear application: Show message about validation when click 'Next' or other page in the progress bar


nmadruga:
Should also enable pages in the progress bar on linear applications. If the user clicks in a page that is not allowed to move to yet: show the generic message: "Can only move to this page if requirements of previous page are fullfilled".
All errors on invalid fields would be visible at this point (after a STRICT validation ran).

PR #167, opened 2020-12-02
#158 Summary validation

CarlosNZ:
Does #158, but doesn't complete #48. 

~~Note: will resolve conflicts once @nmadruga 's remaining PRs are merged. There's some changes in here that are from another PR, but not entirely sure which, so probably hold off reviewing until the other PRs are merged and I've had a chance to resolve issues.~~ 
**Resolved merge changes, ready to review now**

This branch implements:
- Performs new validation on Summary page load
- Changes to "Strict" mode on clicking Submit and displays all undefined & required fields as errors

Still to do (in new PR)
- Run `validatePage` method on whole application on Submit and only submit if it passes, otherwise show error modal. (Currently it submits regardless, but will utilise similar checking to #142 once that's merged)

Also changed:
- Moved common layout code for Summary page (i.e. Grid/Segment structure) to SummaryViewWrapper rather than duplicating in each plugin.
- tidied up default validation code and extracted `defaultValidate` to separate file so it can be reused in Summary

NOTE: Have made some additional layout tweaks in #168 so maybe save layout-related comments to that PR review?


src/formElementPlugins/ApplicationViewWrapper.tsx
```diff
@@ -26,14 +26,14 @@ const ApplicationViewWrapper: React.FC<ApplicationViewWrapperProps> = (props) =>
...
   })
   const [validationState, setValidationState] = useState<ValidationState>({} as ValidationState)
 
   useEffect(() => {
     if (!pluginCode) return
-    // TODO use generic or plugin specific
+    // TODO use plugin-specific validation method if defined
     setPluginMethods({
       validate: defaultValidate,
     })
```
nmadruga, line 39: I don't understand exactly the idea with using the `useEffect` without any variable being watched. Can you add some comment on top of this one, so it's clear what this block is doing please?

src/formElementPlugins/SummaryViewWrapper.tsx
```diff
@@ -1,12 +1,58 @@
...
+    validate: (
+      validationExpress: IQueryNode,
+      validationMessage: string,
+      evaluatorParameters: EvaluatorParameters
+    ): any =>
+      defaultValidate(validationExpression, validationMessage, {
+        objects: [responses],
+        APIfetch: fetch,
+      }),
+  })
```
nmadruga, line 27: I think this could be separated in `initiateGenericValidation` function and here only calls that. Thinking this would improve readability of this method.

src/formElementPlugins/SummaryViewWrapper.tsx
```diff
@@ -1,12 +1,58 @@
...
+
+  useEffect(() => {
+    if (!pluginCode) return
+    // TODO use plugin-specific validation method if defined
+    setPluginMethods({
+      validate: defaultValidate,
+    })
+  }, [])
```
nmadruga, line 35: Same as `ApplicationViewWrapper` maybe add a comment to explain what this effect is for

src/formElementPlugins/SummaryViewWrapper.tsx
```diff
@@ -1,12 +1,58 @@
...
+      setValidationState({ isValid: false, validationMessage: 'Field cannot be blank' })
+      return
+    }
+
+    pluginMethods
+      .validate(validationExpression, validationMessage, { objects: [responses], APIfetch: fetch })
+      .then((validationResult: ValidationState) => {
+        setValidationState(validationResult)
+      })
+  }, [isStrictValidation])
```
nmadruga, line 55: That's cool. So this is doing a `strict` validation without sending the data to the server. Which I think would prevent the application to show all problems when trying to go back to the application page. Is that correct?

src/utils/types.ts
```diff
@@ -63,6 +64,17 @@ interface ElementState extends ElementBase {
...
+  fetch: Function
+  endpoint: string
+}
+
+interface EvaluatorParameters {
+  objects?: object[]
+  pgConnection?: any // Any, because not likely to be used in front-end
+  graphQLConnection?: IGraphQLConnection
+  APIfetch?: Function
+}
```
nmadruga, line 77: Nice to have that

src/utils/types.ts
```diff
@@ -63,6 +64,17 @@ interface ElementState extends ElementBase {
   isRequired: boolean
   isVisible: boolean
 }
+export interface IGraphQLConnection {
```
nmadruga, line 67: Can you add this `export` to the list on the top please?

nmadruga (commented):
First iteration on this PR. I made a few comments, will be running to see the changes now.

src/formElementPlugins/ApplicationViewWrapper.tsx
CarlosNZ: From [React Docs](https://reactjs.org/docs/hooks-effect.html)

> If you want to run an effect and clean it up only once (on mount and unmount), you can pass an empty array ([]) as a second argument. This tells React that your effect doesn’t depend on any values from props or state, so it never needs to re-run. This isn’t handled as a special case — it follows directly from how the dependencies array always works.
> 
> If you pass an empty array ([]), the props and state inside the effect will always have their initial values. While passing [] as the second argument is closer to the familiar componentDidMount and componentWillUnmount mental model, there are usually better solutions to avoid re-running effects too often. Also, don’t forget that React defers running useEffect until after the browser has painted, so doing extra work is less of a problem.

In other words, it means the Effect only runs once, but after all useState variables have been initialised -- which is important if you're going to `set` them within the useEffect.

src/containers/Application/ApplicationOverview.tsx
```diff
@@ -73,10 +79,13 @@ const ApplicationOverview: React.FC = () => {
     <Container>
       <Header as="h1" content="REVIEW AND SUBMIT" />
       <Form>
-        {sectionsPages.map((sectionPages) => (
+        {sectionsPages.map((sectionPage) => (
```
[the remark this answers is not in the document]
CarlosNZ: Okay, I don't fully get what you're saying, but I've changed it back.

src/formElementPlugins/ApplicationViewWrapper.tsx
CarlosNZ: It's the equivalent of `componentDidMount` in class-based Compoments.

src/formElementPlugins/SummaryViewWrapper.tsx
CarlosNZ: See above.

src/formElementPlugins/SummaryViewWrapper.tsx
CarlosNZ: I agree that this isn't that easy to make sense of (I didn't write it originally ;) ) but we already have `defaultValidate` function, so having a whole other function that just initiates this function doesn't seem to help with confusion for me.

src/containers/Application/ApplicationOverview.tsx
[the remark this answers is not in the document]
nmadruga: Hahaha just the different between `sectionPage` and sectionPageS` the later meaning "pages from a section", which you have renamed to "page of a section". : ) To make this clearer

src/formElementPlugins/ApplicationViewWrapper.tsx
nmadruga: Cool! Can you add a single comment on top stating `// Similar to componentDidMount`, please? I think that helps reading the code, at least to me.

src/formElementPlugins/SummaryViewWrapper.tsx
nmadruga: Alright, but it's still convoluted there. Could you at least separate the types definition from the `validate` function definition on the useState declaration? In other words, create a type for `ValidateProps` and only list the properties received and pass down to the `defaultValidate`?

src/formElementPlugins/SummaryViewWrapper.tsx
nmadruga: Yes. Can the comment be added here as well please?

nmadruga (changes requested):
test

src/formElementPlugins/SummaryViewWrapper.tsx
CarlosNZ: Okay, I've defined a type for ValidateFunction and ValidateObject, which helps a bit. Let me know if that's okay.

nmadruga (approved):
Very nice!

PR #168, opened 2020-12-02
#129 Default SummaryView

### LINKED ISSUE #129: Plugins -- default "SummaryView"
Since many plugins will have exactly the same Summary View (just `label`, `text`), there should be a default Summary view returned by the SummaryViewWrapper component.

The wrapper checks for a SummaryView component in the plugin folder, and if present, it displays it. But if not, it displays the default one.

CarlosNZ:
Fixes #129 

Implements a DefaultSummaryView, which will be adequate for most plugins. Plugins will only need to implement their own SummaryView if they need to display more info than just the "Label" and the "response.text". Now we can make changes to the summary view layout without having to repeat the changes for every plugin.

Added a FallbackComponent prop to the `<ErrorBoundary>` component, so will display this DefaultView if error (i.e. can't import SummaryView from plugin).

Question: The little asterisk icon for "required" fields doesn't seem to be showing up like it does in the Application View. Any idea why, @nmadruga ?


src/components/Application/SectionSummary.tsx
```diff
@@ -43,7 +43,7 @@ const SectionSummary: React.FC<SectionSummaryProps> = ({
                           isStrictValidation={isStrictValidation}
                         />
                       </Grid.Column>
-                      <Grid.Column floated="right" width={5}>
+                      <Grid.Column floated="right" width={3}>
```
nmadruga, line 46: I think it's better to leave as `5` just because it was looking better on mobile view

src/formElementPlugins/SummaryViewWrapper.tsx
```diff
@@ -59,26 +58,33 @@ const SummaryViewWrapper: React.FC<SummaryViewWrapperProps> = (props) => {
...
 
-  // TO-DO: Provide a Default SummaryView (just label/text) if SummaryView not provided in plugin
+  const DefaultSummaryView: React.FC = () => {
+    return (
+      <>
+        <label>{parameters.label}</label>
+        <Input fluid readOnly disabled transparent value={response ? response?.text : ''} />
+      </>
+    )
+  }
```
nmadruga, line 68:
```suggestion
  const DefaultSummaryView: React.FC = () => {
    return (
      <Form.Field required={isRequired}>
        <label>{parameters.label}</label>
        <Input fluid readOnly disabled transparent value={response ? response?.text : ''} />
      </Form.Field>
    )
  }
```

To fix the problem with the required fields

src/components/Application/SectionSummary.tsx
CarlosNZ: Actually, I think it was the `floated` properties that were messing it up. I've removed them now, and it responds properly in mobile view. The float and the Grid were conflicting, I think, and we don't need any float in this situation.

src/formElementPlugins/SummaryViewWrapper.tsx
CarlosNZ: Ah, of course! I forgot the DefaultSummaryView component wasn't with the existing <Form.Field>. Thanks 🙏

src/components/Application/SectionSummary.tsx
CarlosNZ: It's still not perfect, but I think I need to properly understand how <Grid> works responsively and what all its props mean before going to deep down this rabbit hole.

nmadruga: Cool. I had a quick check now and I agree. And anyway we shouldn't make it look great in the mobile view yet. Just the most responsive we can get without much effort. All good!

nmadruga (approved):
Good addition

PR #171, opened 2020-12-06
#124 Dynamic parameters in template elements

CarlosNZ:
Implements #124, specifically [this comment](https://github.com/openmsupply/application-manager-web-app/issues/124#issuecomment-736943099)

Note: please review with back-end [PR #124F](https://github.com/openmsupply/application-manager-server/pull/134) as it provides template changes to demonstrate the features here.

The only difference from the solution described in #124 is that the dynamic values are only re-evaluated on losing focus rather than on every keystroke. This can be changed if it's deemed necessary.

Have set the following fields to be dynamically evaluated:

- **textInfo**: `title, text`
- **shortText**: `label`
- **dropdownChoice**: `label, options`

After re-initialising the database with the changes in 124F, start a new "Test Register" application to see demonstration of dynamic parameters:

**Page 1**
- "Last name" question label shows response to Q1
- Additional Text Box showing concatenation of names responses
- Dynamic choice list, the choices being previous responses on this page

**Page 2**
- Dynamic choice list using an external API lookup (https://jsonplaceholder.typicode.com/users)
- Text box showing selection from previous lookup, including the index number (field made visible with "magicword" response"

### Additional notes:

- I've added an additional field, "config" to the `PluginComponents` object return by the pluginProvider, containing the contents of the pluginConfig.json file. Not sure this the best way to get this info, but we can handle it better when we make [needed improvements to the pluginProvider](https://github.com/orgs/openmsupply/projects/6#card-50691093).
- Small additional bug fix: visibility_condition was not being dynamically updated properly, due to a missing variable (`elementsState`) in the useEffect dependency array in ApplicationViewWrapper
- Tidied up some types to do with `pluginComponents`, `pluginConfig`, `ElementPluginParameters`

### DIFF
--- a/src/formElementPlugins/ApplicationViewWrapper.tsx
+++ b/src/formElementPlugins/ApplicationViewWrapper.tsx
@@ -1,10 +1,11 @@
 import React, { useEffect, useState } from 'react'
 import { ErrorBoundary, pluginProvider } from '.'
 import { ApplicationViewWrapperProps, PluginComponents, ValidationState } from './types'
-import { IQueryNode } from '@openmsupply/expression-evaluator/lib/types'
 import { useUpdateResponseMutation } from '../utils/generated/graphql'
 import { EvaluatorParameters, LooseString, ResponseFull } from '../utils/types'
 import { defaultValidate } from './defaultValidate'
+import evaluateExpression from '@openmsupply/expression-evaluator'
+import { IQueryNode } from '@openmsupply/expression-evaluator/lib/types'
 import { Form } from 'semantic-ui-react'
 
 const ApplicationViewWrapper: React.FC<ApplicationViewWrapperProps> = (props) => {
@@ -30,6 +31,16 @@ const ApplicationViewWrapper: React.FC<ApplicationViewWrapperProps> = (props) =>
     ): any => console.log('notLoaded'),
   })
   const [validationState, setValidationState] = useState<ValidationState>({} as ValidationState)
+  const [evaluatedParameters, setEvaluatedParameters] = useState({})
+  const [parametersLoaded, setParametersLoaded] = useState(false)
+
+  const {
+    ApplicationView,
+    config: { dynamicParameters },
+  }: PluginComponents = pluginProvider.getPluginElement(pluginCode)
+
+  const dynamicExpressions =
+    dynamicParameters && extractDynamicExpressions(dynamicParameters, parameters)
 
   useEffect(() => {
     if (!pluginCode) return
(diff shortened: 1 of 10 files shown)


src/formElementPlugins/pluginProvider.tsx
```diff
@@ -62,7 +67,8 @@ class pluginProvider {
 }
 
 function getLocalElementPlugin(folderName: string) {
-  const result: PluginComponents = {}
+  const result: PluginComponents = {} as PluginComponents
+  result.config = require(`./${folderName}/pluginConfig.json`)
```
andreievg, line 71: Will rethink this when making plugins work on prod

andreievg (approved):
Looks pretty good, something to change when we start making custom validations:

this is a little bit out of scope, but I didn't get a chance to comment earlier sorry:

https://github.com/openmsupply/application-manager-web-app/blob/009f3d0681ba56e8cbb30c46b0d718117528c31d/src/formElementPlugins/ApplicationViewWrapper.tsx#L80-L84

In order for custom validation to happen, validation method should take parameters of form element configurations and the original idea (and code I think), provided an evaluator method with parameters prefilled (just have to pass it an evaluation condition)

CarlosNZ:
> In order for custom validation to happen, validation method should take parameters of form element configurations and the original idea (and code I think), provided an evaluator method with parameters prefilled (just have to pass it an evaluation condition)

Yeah, I had a hard time fighting with this to try and make typescript play nicely without using `any` everywhere. You mind if we address this in another issue? (Or leave it until we actually implement a custom validation method?)

We still have this left unchanged:
```
useEffect(() => {
    // Runs once on component mount
    if (!pluginCode) return
    // TODO use plugin-specific validation method if defined
    setPluginMethods({
      validate: defaultValidate,
    })
  }, [])
```
So there'll still be a way to specify a custom validation method in that "To-DO" won't there?

src/formElementPlugins/ApplicationViewWrapper.tsx
```diff
@@ -27,6 +35,14 @@ const ApplicationViewWrapper: React.FC<ApplicationViewWrapperProps> = (props) =>
       console.log('notLoaded'),
   })
   const [validationState, setValidationState] = useState<ValidationState>({} as ValidationState)
+  const [evaluatedParameters, setEvaluatedParameters] = useState({})
+  const [parametersLoaded, setParametersLoaded] = useState(false)
```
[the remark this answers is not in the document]
CarlosNZ: How about `parametersReady`? Have also added comment. b149511

src/formElementPlugins/pluginProvider.tsx
CarlosNZ: Yes, for sure. I wasn't particularly happy with it, but didn't see a straightforward way to return the config info the the Wrapper without a significant refactoring of the pluginProvider.

andreievg:
>Or leave it until we actually implement a custom validation method?)

sounds good

andreievg (approved):

PR #198, opened 2020-12-07
#196 Fix problem with invalid not-visited page

### LINKED ISSUE #196: Bug -- Validation of Linear forms - problems with multi-pages/multi-sections
These 2 bug only became apparent after creating [the new "ReviewTest" template](https://github.com/openmsupply/application-manager-web-app/pull/195) -- a **linear** template with multiple pages in multiple sections.

In order to reproduce, just start a new "Review Test form" application. You'll immediately see that the validation for Page 2 and Section 2 is showing in "error" even though we haven't advanced to those pages yet:

Screen Shot 2020-12-07 at 5 14 33 PM

After filling first page advance to Page 2. Notice the "Next" button is reading "Summary", even though there is another Section to go. Clicking it takes us to the Summary page, completely skipping Section 2.

Attn: @nmadruga

nmadruga:
Using back-end [PR#135](https://github.com/openmsupply/application-manager-server/pull/142)

Fixes #196 

Please, merge only after [PR#135B](https://github.com/openmsupply/application-manager-web-app/pull/195) is merged.

Simple fix to the problem of always showing one invalid page for the first page on a section - it was considering that the first page should always run a STRICT validation, since it wasn't checking for status of previous section, and always considering the previous section was visited.

PS: The second bug described on the same issue, was a configuration problem. So nothing to fix for that one.

### DIFF
--- a/src/containers/Application/ApplicationPageWrapper.tsx
+++ b/src/containers/Application/ApplicationPageWrapper.tsx
@@ -136,8 +136,6 @@ const ApplicationPageWrapper: React.FC = () => {
     }
 
     const statuses = Object.values(pageElementsStatuses)
-    console.log('Validation', statuses)
-
     const validation = getCombinedStatus(statuses)
 
     // Run STRICT validation for linear and LOOSE for non-linear application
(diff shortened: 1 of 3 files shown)


CarlosNZ:
> PS: The second bug described on the same issue, was a configuration problem. So nothing to fix for that one.

You mean template config? The indexing per section, is that what you're referring to?

Because when I start the indexing again from 0 in Section 2, I all the section 2 questions mixed in with the Section 1 questions, like so:

Screen Shot 2020-12-08 at 1 05 08 PM

CarlosNZ:
Seems almost fixed -- but one small issue: when I've finished filling in the first page, and the last question loses focus, but before I click "Next", the error icon for page 2 appears still:

Screen Shot 2020-12-08 at 1 07 46 PM

nmadruga:
> when I've finished filling in the first page, and the last question loses focus, but before I click "Next", the error icon for page 2 appears still:

You are editing Page 1, not Page 2. I guess you also have something wrong in page 2? The current page, so far is always showing the `neutral` state. I think Debs design have changed to show the error if something is incorrect (on Loose validation) on the page. So I think we still need an issue for that. I would rather create a new one.

src/containers/Application/ApplicationPageWrapper.tsx
```diff
@@ -26,6 +26,7 @@ import {
   ValidationMode,
 } from '../../utils/types'
 import { TemplateElementCategory } from '../../utils/generated/graphql'
+import { ValidationContext } from 'graphql'
```
CarlosNZ, line 29: `ValidationContext` and `SummarySectionCode` aren't used at all.

CarlosNZ: Also  `ProgressInPage` type.

CarlosNZ (approved):
Nice, functionality seems pretty solid now. I really like how the ticks appear as soon as you've finished the last question on a page.

Just a couple of unused imports to remove, plus also wondered:

- Line 123 & 268: Any particular reason to use `let` rather than `const`?

nmadruga:
> Nice, functionality seems pretty solid now. I really like how the ticks appear as soon as you've finished the last question on a page.

Yes. I changed to be concise with the latest UI design, so the validation show for the current page as well, when one validation problem is found (on Loose validation) or when the whole page is valid.

nmadruga:
> Line 123 & 268: Any particular reason to use let rather than const?

No. Line 268 was using something that required but I have removed now, and line 123 I checked and there is no problem on changing to `const`. I'm doing this now.

Issue #194, opened 2020-12-07
Validity check updates (Summary/Submit)


nmadruga:
Should first load all plugin specific validation same place that will be dealing with the visibility criteria.

Carl suggested to move the validation back to a field instead of part of the `parameters` field.

---
Updating with **agreed solution** (simplified, will comment back here if unexpected hurdles arise):

- Move "validation" and "validationMessage" back to their own columns in template_element table (and adjust templates accordingly)
- Write new method called `validateAll()` that re-runs validation on all responses and updates the `isValid` field (in database). Returns `true` (All valid) or `false` (At least one not valid)
   - has parameter "strict", so can be used for loose validation if necessary
   - takes array of responses as parameter, so can be used to validate whole application, or just one page if required
- Run `validateAll()` when trying to go to Summary page, don't allow if `false`
- Run `validateAll()` when trying to Submit -- don't allow if `false`
- DON'T run `validateAll` when individual responses change, this behaviour remains the same.

CarlosNZ:
> Carl suggested to move the validation back to a field instead of part of the `parameters` field.

Yeah, I think validation and validationMessage can go back to being their own fields (instead of part of parameters), as that makes them more in line with "isVisible", "isEditable" and "isRequired" and can be evaluated along with them in `useGetResponsesAndElementStates`.

PR #199, opened 2020-12-08
#192 Remove summary from progress

### LINKED ISSUE #192: Remove "Review and Submit" from progress bar and display as button

nmadruga:
Fixes #192.

### Changes
- Remove option "Review and Summary" from Progress bar and ability to move to summary - which simplified the code.
- In `ProgressBar` component have also moved functions to inside the component, the code looks way clearer.
- Using constant for buttons and titles, to easily be replace later.
- In `NavigationBox` component have also moved functions to inside the component for simplicity.
- `NavigationBox` only show `Previous` or `Next` buttons and Modal for not allowing user to continue.
- Show the **Submit application** button on `ApplicationPageWrapper` and take the user to Summary page.
- A little styling on `ApplicationPageWrapper` to look more like the design.

### TO DO
- **Submit application** button should also run the validation and don't go to the Summary page if any elements is invalid (for linear or non-linear applications), this would be done in another issue: #194.


CarlosNZ:
Cool, can you merge the current `master` into this and then resolve the conflicts with your existing #196 (which is basically approved) and then I'll have a proper look at this?

nmadruga:
Done 👍

andreievg (commented):
I had a quick scan of the code, looks great, I like the precision. 

I'll have to test it and check the code in more detail, Today.

src/containers/Application/NavigationBox.tsx
```diff
@@ -16,71 +19,65 @@ interface ModalProps {
...
-              }}
-            />
-          </Grid.Column>
-        </Grid.Row>
-      </Grid>
+  const nextPageButtonHandler = (_: any): void => {
+    // Run the validation on the current page
+    const status = validateCurrentPage()
+    if (!status) {
+      setShowModal({ open: true, ...VALIDATION_FAIL })
```
andreievg, line 54: should we show the same modal when future page / section is pressed (but it can't be navigated) ? or maybe we can remove 'selectable' look on pages that can't be navigated

andreievg (changes requested):
Just a couple of small questions and probably just one change request

src/utils/constants.ts
```diff
@@ -1 +1,13 @@
...
+const TITLE_COMPANY_PLACEHOLDER = 'ABC Company'
+const TITLE_PROGRESS = 'Introduction'
+
+export {
+    BUTTON_NEXT,
+    BUTTON_PREVIOUS,
+    BUTTON_SUBMIT,
+    TITLE_COMPANY_PLACEHOLDER,
+    TITLE_PROGRESS
+}
```
CarlosNZ, line 13: Do you think we want to have to keep track of a bunch of individual imports/exports for different strings? Wouldn't it be easier to just have one `export/import strings` with all the strings and then just putting them in place as `strings.BUTTON_NEXT`, etc.?

src/components/Application/ProgressBar.tsx
```diff
@@ -44,16 +44,10 @@ const ProgressBar: React.FC<ProgressBarProps> = ({
...
-                  serialNumber,
-                  sectionCode,
-                  pageCode,
-                  canNavigate,
-                  push,
-                  validateCurrentPage,
-                })
-              }
+              onClick={() => {
+                if (canNavigate || validateCurrentPage())
```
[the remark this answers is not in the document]
nmadruga: Yes. That's the case, for example, when in linear mode after completing all sections, jump to section 1 to change something and then change to section 4 to change another thing. In this case we should allow the user to go to another section, even for a linear application.

src/containers/Application/NavigationBox.tsx
```diff
@@ -1,12 +1,15 @@
 import React, { useState } from 'react'
-import { Button, Grid, Header, Icon, Message, Modal } from 'semantic-ui-react'
-import PageButton from '../../components/Application/PageButton'
+import { Button, Container, Header, Icon, Label, Modal } from 'semantic-ui-react'
 import { useRouter } from '../../utils/hooks/useRouter'
 import { TemplateSectionPayload } from '../../utils/types'
 import { VALIDATION_FAIL } from '../../utils/messages'
+import { BUTTON_PREVIOUS, BUTTON_NEXT } from '../../utils/constants'
```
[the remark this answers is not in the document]
nmadruga: Hmmm I can't see the advantage of doing that. On the code I think it would look more convoluted.

src/containers/Application/NavigationBox.tsx
nmadruga: I think we can ask what Debs think and create a separated issue to do that. But that's a good point! I'll create a note in the board as placeholder for an issue.

nmadruga: [Note](https://github.com/orgs/openmsupply/projects/6#card-50955479)

src/utils/constants.ts
nmadruga: Ok. I'm convinced now hahah 2 x 1. I'll change to be like that : )

src/containers/Application/NavigationBox.tsx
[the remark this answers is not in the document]
nmadruga: Made a commit for that, Carl also suggested the same : )

src/utils/constants.ts
CarlosNZ: If you think it makes it harder to read in code just rename it to a single character on import.
`import { strings as 's' } from 'contants'`

Then can use `s.BUTTON_NEXT` ? (Or `str`, `c`, `_` -- whatever you fancy)

src/containers/Application/NavigationBox.tsx
andreievg: Sure, it's just seemed weird to me that pages look like they are selectable, but nothing happens when they are pressed.

src/components/Application/ProgressBar.tsx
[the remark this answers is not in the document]
andreievg: But I can go to anywhere in the application mode if current page is valid (with this implementation).

I think, you may want to replace or with and ? (not sure if that work though)

CarlosNZ (approved):

Issue #213, opened 2020-12-11
Don't create responses for non-QUESTION element types.


CarlosNZ:
It's currently creating them for other types, including Page Break and TextInfo

Issue #193, opened 2020-12-07
Redirect to last page with invalid elements when user loads Summary of application invalid


nmadruga:

CarlosNZ:
Does this mean when trying to go to the Summary page directly via URL?

Issue #135, opened 2020-11-24
Redirect Draft application to page in progress


nmadruga:
As described in the #81, when entering a `Linear` application in DRAFT: 

"Redirects to last page that has invalid or unanswered elements - if there is one"

nmadruga:
Think about #186 while doing this one

CarlosNZ:
Just checked and this is _almost_ implemented by #215. That will re-direct back to the last page with an **invalid** response, but not the first incomplete. But it should be a fairly small adjustment to make this work.

PR #215, opened 2020-12-13
#194 #193 #48 Validity Updates

CarlosNZ:
This should address #194, #193 and #48 (and probably #213). Please review with accompanying [back-end PR 150](https://github.com/openmsupply/application-manager-server/pull/150).

Quite a few changes implemented here:

- Moved `validation` and `validationMessage` back to their own fields in `templateElement` instead of being parameters
- Created stand-alone `revalidateAll` function, which re-runs validation on all elements passed into it (would expect all in application), and returns a boolean stating if all responses passed validation, and an array listing the failed ones. (Has parameters for "strict" or not (default strict) and whether database should be updated (default true -- although update not actually done within this function, it just returns the responses needing updating)
- Removed previous Summary-loading validation, as these changes supercede that (basically, undoing #194 😔)
- Don't show any error icons on Summary page (since it shouldn't be possible for it to load if there are errors)
- Moved modal notification from NavigationBox into ApplicationPageWrapper, so it can be used more generally (it now passes `showModal` and `setShowModal` to NavigationBox as props. _TO-DO: we should create a more general modal alert system that can be used all throughout the application._
- Small tweak to `updateResponse` mutation to allow updating without changing response value
- Small change to `useLoadTemplate` so that only "QUESTION" type `id`s are returned (Issue #213). This was to prevent empty responses being created for non-QUESTION elements. I'm not sure this is the _best_ way to handle this, as if the `useLoadTemplate` hook is ever used to provide info for more than just creating the responses (I don't think it currently is) then we might need to return INFORMATION type `id`s too. @nmadruga you might have a suggestion for a better way to fix this?
- Run `revalidateAll` when clicking button to advance to Review & Summary, and only allow if passing. Failure shows modal notification
- Run `revalidateAll` when attempting to load Summary page. If fails, it re-directs to the page with the first invalid response. (I realise this means than when clicking "Review & Summary" from the previous page, it will run two re-validations in sequence, but that seems safer than not doing it in one or other of those scenarios)
- Run `revalidateAll` when clicking "Submit" and only submit form if it passes. If not, it re-directs to the first invalid response as above. _TO-DO: we need some better notification here so the user knows why they're suddenly whisked away when they thought they were submitting, but I thought it best to leave until we've implemented the more general modal alert system (see To-Do above)_
- Slightly out of scope, but small tweak: combine "messages" strings into single object so can be imported with single import (same as "constants")


nmadruga:
Sorry I ran out of time today. Will review this first thing tomorrow.

nmadruga:
>  TO-DO: we should create a more general modal alert system that can be used all throughout the application.

Created an issue for that: #220

nmadruga:
> Small change to useLoadTemplate so that only "QUESTION" type ids are returned (Issue #213). This was to prevent empty responses being created for non-QUESTION elements. I'm not sure this is the best way to handle this, as if the useLoadTemplate hook is ever used to provide info for more than just creating the responses (I don't think it currently is) then we might need to return INFORMATION type ids too. @nmadruga you might have a suggestion for a better way to fix this?

Using this hook to retrieve only Questions is can actually a good thing in my opinion, could also use this for example calculating the total number of questions in the application before the user starts filling fields, I think that's ok.

src/containers/Application/ApplicationOverview.tsx
```diff
@@ -118,3 +160,23 @@ const showProcessingModal = (processing: boolean, submitted: boolean) => {
...
+    const page = elementsState[code].page
+    if (sectionIndex < firstErrorSectionIndex) {
+      firstErrorSectionIndex = sectionIndex
+      firstErrorPage = page
+    } else
+      firstErrorPage =
+        sectionIndex === firstErrorSectionIndex && page < firstErrorPage ? page : firstErrorPage
+  })
+  return { firstErrorSectionIndex, firstErrorPage }
+}
```
nmadruga, line 182: Very cool! I think this may be better included in the same file as `validateAll` so it's accessible for other client classes to use.

src/containers/Application/ApplicationOverview.tsx
```diff
@@ -71,13 +83,44 @@ const ApplicationOverview: React.FC = () => {
...
+
+    // If invalid responses, re-direct to first invalid page
+    if (!revalidate.allValid) {
+      console.log('Some responses invalid')
+      const { firstErrorSectionIndex, firstErrorPage } = getFirstErrorLocation(
+        revalidate.validityFailures,
+        elementsState as ApplicationElementStates
+      )
+      // TO-DO: Alert user of Submit failure
+      push(`/application/${serialNumber}/S${firstErrorSectionIndex + 1}/Page${firstErrorPage}`)
```
nmadruga, line 107: Ahh, I know this might require quite a big change, but we can't rely on the code being the same as the `index + 1`. This should be the section code instead. If `elementState` doesn't have the `sectionCode` perhaps is better to change it to have it, or we can change the URL to always use the sectionIndex, that could also be done.

nmadruga: Although since the index start from 0 it might be weird on the URL

src/formElementPlugins/SummaryViewWrapper.tsx
```diff
@@ -82,18 +48,14 @@ const SummaryViewWrapper: React.FC<SummaryViewWrapperProps> = (props) => {
   )
 
   return (
-    <Grid columns={2}>
+    <Grid columns={1}>
```
nmadruga, line 51: We don't need to use the Grid here anymore, the ErrorBoundary can be straight at the root level of this component

src/utils/helpers/revalidateAll.tsx
```diff
@@ -0,0 +1,73 @@
...
+export const revalidateAll = async (
+  elementsState: ApplicationElementStates | undefined,
+  responsesByCode: ResponsesByCode | undefined,
+  strict = true,
+  shouldUpdateDatabase = true
+): Promise<RevalidateResult> => {
+  const validate = defaultValidate // To-Do: import custom validation methods
+
+  // console.log('Elements', elementsState)
+  // console.log('responses', responsesByCode)
```
nmadruga, line 18: Remove comments please

src/utils/helpers/revalidateAll.tsx
```diff
@@ -0,0 +1,73 @@
...
+  if (elementsState) {
+    const elementCodes = Object.keys(elementsState).filter(
+      (key) =>
+        responsesByCode && // Typescript requires this
+        elementsState[key].category === 'QUESTION' &&
+        elementsState[key].isVisible === true &&
+        // Strict/Loose validation logic:
+        ((strict && elementsState[key].isRequired) ||
+          (strict && responsesByCode[key]?.isValid !== null) ||
+          (!strict && responsesByCode[key]?.isValid !== null))
```
nmadruga, line 28:
```suggestion
          (responsesByCode[key]?.isValid !== null))
```

src/utils/helpers/revalidateAll.tsx
```diff
@@ -0,0 +1,73 @@
...
+    )
+
+    const validationExpressions = elementCodes.map((code) => elementsState[code].validation)
+
+    const evaluatedValidations = []
+
+    for (let i = 0; i < elementCodes.length; i++) {
+      const code = elementCodes[i]
```
nmadruga, line 36: I think this would have been more elegant using instead an array that keeps both: elementCode and validation expression, so the first `. map` can be changes to: `const validationExpressions = elementCodes.map((code) => ({ code, expression: elementsState[code].validation }))` then you iterate over this array using both properties: `validationExpressions.map(({code, expression}) => ... )`

src/utils/helpers/revalidateAll.tsx
```diff
@@ -0,0 +1,73 @@
...
+      const expression = validationExpressions[i]
+      evaluatedValidations.push(
+        validate(expression, elementsState[code]?.validationMessage as string, {
+          objects: [responses],
+          APIfetch: fetch,
+        })
+      )
+    }
+    const resultArray = await Promise.all(evaluatedValidations)
+    // console.log('Result', resultArray)
```
nmadruga, line 52: So cool the use of Promises like this : )

Also another comment to be removed

src/utils/helpers/revalidateAll.tsx
```diff
@@ -0,0 +1,73 @@
...
+      elementCodes.forEach((code, index) => {
+        if (!resultArray[index].isValid) {
+          validityFailures.push({
+            id: (responsesByCode && responsesByCode[code].id) || 0,
+            isValid: false,
+            code,
+          })
+        }
+      })
+    }
```
nmadruga, line 64:
```suggestion
  const validityFailures: ValidityFailure[] = shouldUpdateDatabase
      ? elementCodes.reduce((validityFailures: ValidityFailure[], code, index) => {
          if (!resultArray[index].isValid) {
            return [
              ...validityFailures,
              {
                id: (responsesByCode && responsesByCode[code].id) || 0,
                isValid: false,
                code,
              },
            ]
          } else return validityFailures
        }, [])
      : []
```
This can be a nice use for the `reduce` method, which will only return the elected elementCodes in the expected format of validityFailures.

src/utils/helpers/revalidateAll.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+    return {
+      allValid: resultArray.every((element) => element.isValid),
+      validityFailures,
+    }
+  }
+  // Typescript requires final return -- it should never be reached
+  return { allValid: false, validityFailures: [] }
```
nmadruga, line 72: This is just because `elementState` can be undefined... it might be the case it reaches this point. Should change to not allow receiving elementState as undefined?

nmadruga (changes requested):
Nice work Carl, this is looking great. I think it makes the code easier to follow to. I've made just a fews suggestions, some style code changes and the biggest one for me is: trying to unify the `SummaryViewWrapper` with the Application and Review would be better to receive the validation to be run inside each plugin to be received as a single function.

src/containers/Application/ApplicationOverview.tsx
```diff
@@ -29,16 +31,26 @@ const ApplicationOverview: React.FC = () => {
...
+  const [responseMutation] = useUpdateResponseMutation()
+
+  useEffect(() => {
+    // Fully re-validate on page load
+    if (!appStatus) return
+    if (appStatus?.status !== 'DRAFT' && appStatus?.status !== 'CHANGES_REQUIRED') {
+      // Show summary, even if it no longer validates, as it would
+      // have been valid when submitted.
+      setIsRevalidated(true)
+      return
```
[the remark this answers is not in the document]
CarlosNZ: Interesting. I see your point about going back to "DRAFT" once the user starts making changes. Although I thought that "CHANGES REQUIRED" would be kind of the same thing -- it's effectively a "Draft" status as far as functionality goes, it's just a distinction for things like Actions, etc. Maybe bring it up at next meeting?

src/containers/Application/ApplicationOverview.tsx
```diff
@@ -71,13 +83,44 @@ const ApplicationOverview: React.FC = () => {
...
+        variables: {
+          id: changedElement.id,
+          isValid: changedElement.isValid,
+        },
+      })
+    })
+
+    // If invalid responses, re-direct to first invalid page
+    if (!revalidate.allValid) {
+      console.log('Some responses invalid')
```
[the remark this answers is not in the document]
CarlosNZ: Oh, I wasn't thinking we would remove this. I think legit messages that are useful bits of information are okay to log to console -- there's a few floating round in there currently. 

The relevant TO-DO is there though, a couple of lines further down:
`// TO-DO: Alert user of Submit failure`

src/containers/Application/ApplicationOverview.tsx
CarlosNZ: Sure thing. Done 5d3ebaa

CarlosNZ: I think I will probably use it again for #135 👍

src/containers/Application/ApplicationOverview.tsx
CarlosNZ: Oh, my bad, I thought the URL was just using "S" + sectionIndex. Yes, I'll fix this for sure.

CarlosNZ: Done 682b09d :)

src/utils/helpers/revalidateAll.tsx
CarlosNZ: This logic was doing my head in. Just to be clear, it's supposed to be KEEPING for validation:
- if strict: all required and all not null
- if not strict: all not null.

That's right, isn't it?

src/utils/helpers/revalidateAll.tsx
CarlosNZ: Yeah, I agree that this would be nicer. The reason I didn't is that the `Promise.all` requires an array of promises, and it won't work if the promise is only a field within an object in the array.

Does that make sense, or is the a simple way around this that you can think of?

src/utils/helpers/revalidateAll.tsx
CarlosNZ: Very nice, thank you 😎

src/utils/helpers/revalidateAll.tsx
CarlosNZ: Yeah, I couldn't get it to work. The `elementState` object should never be undefined by the time it gets to this function, but it _can_ be undefined elsewhere (like in `useGetResponsesAndElementState` for example), so I wasn't able to insist that this function take `elementState` that is never undefined. Any ideas?

src/components/Application/SectionSummary.tsx
```diff
@@ -40,7 +38,6 @@ const SectionSummary: React.FC<SectionSummaryProps> = ({
                           element={element}
                           response={response}
                           allResponses={allResponses}
```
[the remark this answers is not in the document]
CarlosNZ: I'm not entirely sure what you mean here. There's no validation being done in Summary elements at all now. 

But we still need the responses for dynamic parameters (e.g. labels), which will be evaluated for both Summary and Review display. (Unless we update them to have "fixed" values after submission, but that would be quite a big job which would require some significant planning)

src/utils/helpers/revalidateAll.tsx
nmadruga: As far as I'm understanding the code you get as array of promises with `evaluatedValidations` which is created inside this loop. What I'm saying is that this loop can be run using the .map function instead of the traditional for loop

src/utils/helpers/revalidateAll.tsx
nmadruga: You can type to the `ApplicationElementState` if you are sure it would only reach the call in case it was defined.
Example:
```
    const revalidate = await revalidateAll(
      elementsState as ApplicationElementStates,
      responsesByCode
    )
```
And change the `revalidateAll` to expect only the defined object

src/components/Application/SectionSummary.tsx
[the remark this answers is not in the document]
nmadruga: Ohh. Ok! Good it, I'll see how to pass those to the `SummaryViewWrapper` on the Review page then. Thanks for the explanation, it wasn't clear to me why it was still expecting these parameters.

src/utils/helpers/revalidateAll.tsx
CarlosNZ: Ah -- I think I get you now. Good suggestion, have updated it, seems cleaner. Thanks :) fabc757

src/utils/helpers/revalidateAll.tsx
CarlosNZ: Got it! Thanks for clarifying :) 179cbdc

CarlosNZ:
Thanks for the great suggestions (and patient explanations), @nmadruga, I really appreciate it. I think all your comments have been addressed now. 👍

src/utils/helpers/revalidateAll.tsx
```diff
@@ -0,0 +1,86 @@
...
+      elementsState[code]?.validationMessage as string,
+      {
+        objects: [responses],
+        APIfetch: fetch,
+      }
+    )
+  })
+
+  const resultArray = await Promise.all(validationExpressions)
+  console.log('resultArray', resultArray)
```
nmadruga, line 44: Please remove this last console.log

CarlosNZ: Done a5f34c1

nmadruga (approved):
Great!

Issue #202, opened 2020-12-09
Trigger updates


CarlosNZ:
I think these things need to be done

1. Add trigger field to review_assignment table (with same trigger logic as other triggers) [back-end]
2. Update application_trigger_states View to properly handle multiple reviews per application (and include review_assignment triggers) [back-end]
3. Update `useTriggerProcessing` hook (and query) to accommodate above changes and work properly for application AND review requests [front-end]
   - Maybe add timeout (#185) too, if not too much work
4. Split "trigger" enums into per-table enum types: [back-end]
   - i.e. an "ApplicationTrigger" type which only has "onApplicationCreate", "onApplicationSubmit" etc. This will ensure that you can't put a "onReviewSubmit" on an application table trigger. 

Can split into multiple issues if you like, but only the `useTriggerProcessing` one will take any significant time, the others are fairly trivial (yet important).

PR #218, opened 2020-12-14
#156 review application page

nmadruga:
Fixes #156 

~~Based on PR #205, please merge that one before this.~~ **DONE**
Based on PR #219, please merge that one before this.

Requires [back-end PR152](https://github.com/openmsupply/application-manager-server/pull/152) which updates the application_trigger_states view.

## Changes
- Some layout changes to make the Review look as UI designs. Not very precise, just more similar.
- create new hook `useLoadReview` that create a new structure (similar to what is done in `ApplicationOverview` with sections pages and its elements, response and review to be displayed, it uses other hooks:
  - `useLoadApplication` to load the template sections - not all will be part of the review, so we need to first get all sections of the application.
  - `useGetElementsAndResponsesState` - will get all elements in application and its responses, validate the state of each element and return 2 objects one with all elements and another with all sections
  - `useGetReview` - the Graphql hook to retrieve all review_responses (and later more review information) for this review
- Created the component `ReviewSection` to display each section of the application with review buttons. 
- Created a few more constants to be replacing loose strings inside the containers/components.

PR #219, opened 2020-12-14
#202 Trigger processing updates

CarlosNZ:
Addresses #202. Requires [back-end PR152](https://github.com/openmsupply/application-manager-server/pull/152) which updates the application_trigger_states view.

I tried to do away with the need for the application_trigger_states view and make `useTriggerProcessing` query the respective tables directly, using @directives to choose which to call, but for some reason it didn't work with re-polling query. No idea why, but probably worth looking into in more detail at some point.

Anyway, have changed `useTriggerProcessing` so it can retrieve the trigger status for either an application, a review, or a review_assignment. It's working as it used to with the `useLoadApplication` hook. I've added similar functionality to the `useLoadApplication` hook, but this hasn't been properly tested yet as we don't have any reviews yet. This should be checked and updated accordingly with ongoing Review work.

Haven't added any trigger checking for Review assignments yet.

Note: have made the assumption that we'd only be wanting to check the trigger status of one table at a time. If loading an application, we'd check the application table entry; if loading a review, we'd check the review table entry. Does this make sense? Is there any situation where we'd need to check triggers on multiple entities at once?


src/utils/hooks/useTriggerProcessing.tsx
```diff
@@ -1,16 +1,41 @@
 import { useState, useEffect } from 'react'
 import { useGetTriggersQuery } from '../../utils/generated/graphql'
 
-type triggerTypes = 'applicationTrigger' | 'reviewTrigger'
+type TriggerType = 'applicationTrigger' | 'reviewAssignmentTrigger' | 'reviewTrigger'
```
nmadruga, line 4: Any reason not to use the Graphql type here?

src/utils/hooks/useTriggerProcessing.tsx
```diff
@@ -1,16 +1,41 @@
...
-  const { serialNumber, trigger } = props
+interface TrigerQueryProps {
+  serialNumber?: string
+  reviewAssignmentId?: number
+  reviewId?: number
+  triggerType?: TriggerType
+}
+
+const useTriggerProcessing = (props: TrigerQueryProps) => {
+  const { serialNumber, reviewAssignmentId, reviewId } = props
```
nmadruga, line 14:
```suggestion
const useTriggerProcessing = ({ serialNumber, reviewAssignmentId, reviewId, triggerType }: TrigerQueryProps) => {
```
Than also change elsewhere using `props.triggerType`

src/utils/hooks/useTriggerProcessing.tsx
```diff
@@ -1,16 +1,41 @@
...
+  }
+
+  // If triggerType not provided, infer it from the supplied ID
+  const triggerType = props.triggerType
+    ? props.triggerType
+    : serialNumber
+    ? 'applicationTrigger'
+    : reviewAssignmentId
+    ? 'reviewAssignmentTrigger'
+    : 'reviewTrigger'
```
nmadruga, line 32: I think this and the `type` defined up here would be better separated in its own helper method, just to be clear what's going on and helping to read this code. The method could be in the same file, if that is simpler.

src/utils/hooks/useTriggerProcessing.tsx
```diff
@@ -19,7 +44,21 @@ const useTriggerProcessing = (props: { serialNumber: string; trigger: triggerTyp
 
   useEffect((): any => {
     if (data?.applicationTriggerStates?.nodes[0]) {
-      const triggerRequested = data?.applicationTriggerStates?.nodes[0][trigger]
+      let triggerRequested
+      switch (triggerType) {
+        case 'applicationTrigger':
+          triggerRequested = data?.applicationTriggerStates?.nodes[0]?.applicationTrigger
```
nmadruga, line 50: Probably need to check first if `data?.applicationTriggerStates?.nodes` is not empty

nmadruga (commented):
I don't really understand the need for `hooks/temp.ts`, and made a few comments. I don't know how to test this too, but have merge with PR #218 so you can lead on the tests or just let me know how I could test this.

CarlosNZ:
> I don't really understand the need for `hooks/temp.ts`

This is a mistake, it shouldn't be in here. I've removed it now. 😊

CarlosNZ:
> I don't know how to test this too, but have merge with PR #218 so you can lead on the tests or just let me know how I could test this.

How I've been testing it is:
- first check that Applications and Reviews load as expected.
- manually edit the database to set triggers to either "Processing" or "Error" (they're the only two that won't immediately revert due to trigger processing)
- reload the related Applications or Reviews -- you should see the perpetual loading icon.
- while it's still trying to load, set the related trigger back to `null` -- the application or review should immediately finish loading.

src/utils/hooks/useTriggerProcessing.tsx
CarlosNZ: We don't have a GraphQL type for this. We have a type for the individual triggers (`onApplicationCreate`, etc.), but not on a per-table basis.

src/utils/hooks/useTriggerProcessing.tsx
CarlosNZ: Isn't that what this is doing at the top of the function?
```
if (data?.applicationTriggerStates?.nodes[0])
```

nmadruga: No, I think it would crash there if there was no applicationTriggerStates

CarlosNZ: Ah, okay looks like this will work:
`if (data?.applicationTriggerStates?.nodes?.[0])` 😎

Done a5cf6b6a

nmadruga: Ah, I'm not sure this will solve this. Sorry! This is again checking for null, but you still try to access an index of the nodes array without checking if this array is empty! I've seen this happening before in many places with GraphQL queries returning empty arrays, is better to have a safe guard for this here. We will improve the error handling at some point, but we still want to make this easier for development, if something unexpected happens it would be easier to see the error instead of a hook that crashes which would happen in this case.

src/utils/hooks/useTriggerProcessing.tsx
```diff
@@ -1,25 +1,62 @@
...
+      let triggerRequested
+      switch (inferredTriggerType) {
+        case 'applicationTrigger':
+          triggerRequested = data?.applicationTriggerStates?.nodes[0]?.applicationTrigger
+          break
+        case 'reviewAssignmentTrigger':
+          triggerRequested = data?.applicationTriggerStates?.nodes[0]?.reviewAssignmentTrigger
+          break
+        case 'reviewTrigger':
+          triggerRequested = data?.applicationTriggerStates?.nodes[0]?.reviewTrigger
```
nmadruga, line 54: Also better to set this `data?.applicationTriggerStates?.nodes[0]?` to a variable to make the code cleaner

src/utils/hooks/useTriggerProcessing.tsx
```diff
@@ -31,3 +68,17 @@ const useTriggerProcessing = (props: { serialNumber: string; trigger: triggerTyp
...
+  serial: string | undefined,
+  reviewAssignmentId: number | undefined
+) {
+  return triggerType
+    ? triggerType
+    : serial
+    ? 'applicationTrigger'
+    : reviewAssignmentId
+    ? 'reviewAssignmentTrigger'
+    : 'reviewTrigger'
```
nmadruga, line 83: I think you could be using the type here to return instead of strings

src/utils/hooks/useTriggerProcessing.tsx
```diff
@@ -1,25 +1,62 @@
...
     fetchPolicy: 'no-cache',
   })
 
   useEffect((): any => {
-    if (data?.applicationTriggerStates?.nodes[0]) {
-      const triggerRequested = data?.applicationTriggerStates?.nodes[0][trigger]
+    if (data?.applicationTriggerStates?.nodes?.[0]) {
+      let triggerRequested
+      switch (inferredTriggerType) {
+        case 'applicationTrigger':
```
nmadruga, line 47: Same thing here, if you have a `type` why not use it? It is good for any changes that you might have it would point out for using the wrong string right away. : )

nmadruga: And I'm pretty sure there is a more elegant solution for switch in the ES6, have a look on that. And you can probably merge the code for setting the error and processing to `false` in case of bad parameters.

nmadruga (commented):
Few more comments to be addresses please

src/utils/hooks/useTriggerProcessing.tsx
CarlosNZ: > but you still try to access an index of the nodes array without checking if this array is empty!

That's what the `?.` before the index is for. It won't try and access the index if it doesn't exist!

Okay, I double-checked this and it seems fine: https://jsfiddle.net/CarlosNZ/z0e6jwg5/16/

src/utils/hooks/useTriggerProcessing.tsx
CarlosNZ: Okay, assigned the whole long thing to `triggers` a63c0e5

src/utils/hooks/useTriggerProcessing.tsx
CarlosNZ: Okay, I've added the `TriggerType` as the output of the function, so the returned values have to comply with that.

Is that what you mean, or were you meaning something else?

src/utils/hooks/useTriggerProcessing.tsx
CarlosNZ: > Same thing here, if you have a `type` why not use it? It is good for any changes that you might have it would point out for using the wrong string right away. : )

This is being typed implicitly now, based on last change. `inferredTriggerType` has to be `TriggerType` as that's what its defining function returns. Can confirm that if you try and change one of these "case" strings to something non-conforming, typescript rejects it.

CarlosNZ: > And I'm pretty sure there is a more elegant solution for switch in the ES6, have a look on that.

The trigger field names match the "TriggerType" options, so can just refer to the trigger directly in one line!
```
if (triggers[inferredTriggerType] === null) setIsProcessing(false)
```

Good suggestion, thanks 👍

src/utils/hooks/useTriggerProcessing.tsx
nmadruga: Yes, that's fine. I was suggestion to also use the format  `TriggerType. applicationTrigger` for types, but not problem. Just using as the return will also flag for possible problem with types.

nmadruga (approved):
Thanks for all the changes.

Approving now!

PR #222, opened 2020-12-15
#135 Redirect application to first incomplete page

CarlosNZ:
Implements #135 -- just a fairly small extension to #194 (so you'll need [that PR's back-end branch](https://github.com/openmsupply/application-manager-server/pull/150) to test against, if it's not already merged)

This PR does two additional things:
1. Ensures that any empty response (null, undefined or empty string) all get set to "invalid" (for required questions) when re-validating all (change to `revalidateAll.tsx`). This ensures that when attempting to load Summary, you'll be taken back to first incomplete page (strict check), not just page with an invalid element (loose check).
2. Loading "Draft" application re-directs user to first invalid/incomplete page (instead of section 1, page 1) -- uses the same `revalidateAll` and `firstErrorLocation` functions as #194. (This does NOT write validity back to database, so `null` validations will persist in each element.

### DIFF
--- a/src/containers/Application/ApplicationPageWrapper.tsx
+++ b/src/containers/Application/ApplicationPageWrapper.tsx
@@ -23,7 +23,7 @@ import validatePage, {
   PROGRESS_STATUS,
 } from '../../utils/helpers/validatePage'
 import getPageElements from '../../utils/helpers/getPageElements'
-import { revalidateAll } from '../../utils/helpers/revalidateAll'
+import { revalidateAll, getFirstErrorLocation } from '../../utils/helpers/revalidateAll'
 import strings from '../../utils/constants'
 import messages from '../../utils/messages'
 
@@ -79,7 +79,7 @@ const ApplicationPageWrapper: React.FC = () => {
   // 1 - ProcessRedirect: Will redirect to summary in case application is SUBMITTED
   // 2 - Set the current section state of the application
   useEffect(() => {
-    if (isApplicationLoaded) {
+    if (elementsState && responsesByCode) {
       processRedirect({
         ...appStatus,
         serialNumber,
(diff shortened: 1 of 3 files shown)


src/containers/Application/ApplicationPageWrapper.tsx
```diff
@@ -79,7 +79,7 @@ const ApplicationPageWrapper: React.FC = () => {
   // 1 - ProcessRedirect: Will redirect to summary in case application is SUBMITTED
   // 2 - Set the current section state of the application
   useEffect(() => {
-    if (isApplicationLoaded) {
+    if (elementsState && responsesByCode) {
       processRedirect({
```
nmadruga, line 83: Don't need to use await here?

src/containers/Application/ApplicationPageWrapper.tsx
```diff
@@ -290,14 +292,26 @@ function processRedirect(appState: any): void {
...
+
+    if (revalidate.validityFailures.length > 0) {
+      const { firstErrorSectionCode, firstErrorPage } = getFirstErrorLocation(
+        revalidate.validityFailures,
+        elementsState as ApplicationElementStates
+      )
+      push(`/application/${serialNumber}/${firstErrorSectionCode}/Page${firstErrorPage}`)
+    } else {
+      const firstSection = templateSections[0].code
+      replace(`/application/${serialNumber}/${firstSection}/page1`)
```
nmadruga, line 313:
```suggestion
      replace(`/application/${serialNumber}/${firstSection}/Page1`)
```

nmadruga (commented):
I'll be testing this one now. Getting all pieces together looks good.

Make just 2 suggestions on this one for you to have a look please

src/containers/Application/ApplicationPageWrapper.tsx
CarlosNZ: I don't think so, cos if there is a redirect required, then the called method does the `push`, so it'll over-rule anything else in progress. And if there's no redirect, then we don't need to wait for anything.

nmadruga: Ah! Got it. That's good.

nmadruga (approved):
LGTM

Issue #145, opened 2020-11-26
Make user info available in local state


CarlosNZ:
So it can be passed to the evaluator, or other parts of the app.

User info that should be available:
- First & Last name
- username
- email
- date of birth?
- permissions?
- Org name

Also, info about current application:
- serial no.

andreievg:
Can these be deduced via DB (if user id is available) ? 

Also can extend useGetUserPermissions (and login endpoint) to set 'userInfo' which can be available through response from login and userPermission end points:
https://github.com/openmsupply/application-manager-web-app/blob/0da5d6491908ef9676537689e8be4319892d8ad1/src/utils/hooks/useGetUserPermissions.ts#L21-L22

Can add extra bits to userInfo when deemed necessary. It can be used similar to how userPermissions context key is used:

https://github.com/openmsupply/application-manager-web-app/blob/0da5d6491908ef9676537689e8be4319892d8ad1/src/components/Main/AppMenu.tsx#L13-L15

https://github.com/openmsupply/application-manager-web-app/blob/0da5d6491908ef9676537689e8be4319892d8ad1/src/components/Main/AppMenu.tsx#L34-L37

PR #223, opened 2020-12-16
#145 User state info

CarlosNZ:
Implements #145. Requires [back-end PR 155](https://github.com/openmsupply/application-manager-server/pull/155)

### Changes:

- Change UserState context area to treat `currentUser` as a User object rather than just username string
- Update login function to acquire and set "User" info to local storage rather than "username"
- Update loginCheck function to check for JWT & "user" (rather than "username")
- Update `useGetUserPermissions` method (which runs on loading the User Area) to fetch User info (from localStorage) and set `currentUser` in Context Provider.
- Update all evaluator calls to pass `currentUser` as 2nd object in object array (back-end template updated to display "Current User" in the Feature Showcase template):  
Screen Shot 2020-12-16 at 4 04 42 PM


src/utils/hooks/useGetUserPermissions.ts
```diff
@@ -10,15 +10,16 @@ const createAuthorisationHeader = (JWT: string) => ({
...
-  const [username, setUsername] = useState<string>('')
+  const [user, setUser] = useState<User>()
 
   useEffect(() => {
     const JWT: string = localStorage.getItem(LOCAL_STORAGE_JWT_KEY) || ''
+    const user: User = JSON.parse(localStorage.getItem('user') || '')
 
     fetch(userPermissionsUrl, { headers: createAuthorisationHeader(JWT) })
       .then((res: any) => res.json())
       .then(({ username, templatePermissions, JWT }: FullUserPermissions) => {
```
nmadruga, line 21:
```suggestion
      .then(({ templatePermissions, JWT }: FullUserPermissions) => {
```

src/containers/User/UserRegister.tsx
```diff
@@ -12,7 +12,7 @@ const UserRegister: React.FC = () => {
     .then((loginResult) => {
       console.log('Result', loginResult)
       localStorage.setItem('persistJWT', loginResult.JWT)
-      localStorage.setItem('username', loginResult.username)
+      localStorage.setItem('user', JSON.stringify(loginResult.user))
```
nmadruga, line 15: Do we need to keep the current user in place, in case we reload the page? But in this case - after reloading I can't see where it is updating the context to use the current user in the evaluation

src/utils/helpers/loginCheck.ts
```diff
@@ -1,7 +1,7 @@
 const isLoggedIn = () => {
   const JWT = localStorage.getItem('persistJWT')
-  const username = localStorage.getItem('username')
-  return JWT && username
+  const user = localStorage.getItem('user')
+  return JWT && user
 }
```
nmadruga, line 5: Maybe here should also update the user in the context? I haven't tested yet, but I'm assuming reloading would fail?

src/utils/hooks/useGetResponsesAndElementState.tsx
```diff
@@ -7,13 +7,17 @@ import {
   TemplateElement,
   Application,
 } from '../generated/graphql'
+import { useUserState } from '../../contexts/UserState'
```
nmadruga, line 10: Interesting! I didn't know you would have access to the context in here. That's cool!

nmadruga (commented):
I tested and reloading works - I'm not sure how this is doing. Do we need to store the `user` in the localStorage? I'm not sure if we need. But perhaps it's not clear to me how the context is being updated.

src/containers/User/UserRegister.tsx
CarlosNZ: > But in this case - after reloading I can't see where it is updating the context to use the current user in the evaluation

The UserArea get "user" and "templatePermissions" from `useGetUserPermissions`, which it then sets to context (near the top of UserArea component)

> Do we need to keep the current user in place, in case we reload the page?

I was doing it so the "user" data could persist between login and loading `useGetUserPermissions`, but now that I think about it, we could save the userdata in Context in the login component and then it'd be already available everywhere else.

Although we still need to persist in in localStorage so that the user data is available next visit/reload.

Do you think it's a problem?

CarlosNZ: Okay, I've made the changes I suggested above. I think it's clearer now and keeps the user info/context separate from the Permissions fetching.

So, in summary:
- After successful login, the Login component writes the user info to Context and localStorage
- When reloading, and "loginCheck" confirms that user is already logged in (i.e. has valid JWT and user data), the UserArea fetches the user data from localStorage and stores it in context (it only does this if it can't find a currentUser already in context).

src/utils/helpers/loginCheck.ts
CarlosNZ: Yes, I was going to make the context update go here, but it's not a React component, so can't use Hooks, etc.

See my above comment for how it's being handled now.

src/utils/hooks/useGetResponsesAndElementState.tsx
CarlosNZ: Yes, we can have Application state here too if we want it, so that can also be used in evaluation expressions. :)

CarlosNZ:
> I tested and reloading works - I'm not sure how this is doing.

I've changed it a bit now, so see [this comment](https://github.com/openmsupply/application-manager-web-app/pull/223#discussion_r544796616) for a summary of what's happening.

> Do we need to store the `user` in the localStorage? I'm not sure if we need. 

The stored JWT only has username info in it, so we'd need to make a query to fetch all the user data from the database on load. Keeping it locally seemed preferable and simpler, since it's not dynamic data so once we've got it once (i.e. at login) it's never going to change. It gets flushed on log out along with the JWT.  I don't see an immediate problem with storing the user data (there's no sensitive stuff like passwords or anything), but I'm happy to be convinced why it's a bad idea.

src/containers/Main/SiteLayout.tsx
```diff
@@ -51,7 +51,9 @@ const SiteLayout: React.FC = () => {
                 <Home />
               </Route>
               <Route exact path="/login">
-                <Login />
+                <ApplicationProvider>
+                  <Login />
+                </ApplicationProvider>
```
nmadruga, line 56: Why the need for adding the ApplicationProvider here?

src/containers/User/UserRegister.tsx
nmadruga: I understood now. That's what I was thinking, if the user already comes from Graphql query and that runs on page reaload why the need to store it locally, but I see that the information could delay to display. I think this is good.

nmadruga:
> [NM] Do we need to store the user in the localStorage? I'm not sure if we need.

> [CS] The stored JWT only has username info in it, so we'd need to make a query to fetch all the user data from the database on load. Keeping it locally seemed preferable and simpler, since it's not dynamic data so once we've got it once (i.e. at login) it's never going to change. It gets flushed on log out along with the JWT. I don't see an immediate problem with storing the user data (there's no sensitive stuff like passwords or anything), but I'm happy to be convinced why it's a bad idea.

I think the only down side is the complexity, because we are keeping here and there we need to check (and update) both places whenever we need to update something in the user settings. I guess that when we add permissions this is going to be more dynamic so we might find a problem with that. My thinking was to make it simpler the cache is persisted in local storage, so if the user data is on cache the query would get from there (instead of from the server). No it is just a matter of waiting the super fast query to finish to display the user settings in place.

andreievg:
@CarlosNZ I just saw this. 

> The stored JWT only has username info in it, so we'd need to make a query to fetch all the user data from the database on load. Keeping it locally seemed preferable and simpler, since it's not dynamic data so once we've got it once (i.e. at login) it's never going to change. It gets flushed on log out along with the JWT. I don't see an immediate problem with storing the user data (there's no sensitive stuff like passwords or anything), but I'm happy to be convinced why it's a bad idea.

I think we still need to do a query for permissions on every load of page. And it's possible permissions would change, and it seem easer way to just query on reload to account for all use cases. I vote for just storing JWT

CarlosNZ:
> I think we still need to do a query for permissions on every load of page. And it's possible permissions would change, and it seem easer way to just query on reload to account for all use cases. I vote for just storing JWT

So we currently fetch permission info from database on load. You think we should fetch user info at the same time (so put it in `useGetUserPermissions`) ?

src/containers/Main/SiteLayout.tsx
CarlosNZ: So it can access the User context to write the user info to context. But if we're not going to do that here, then we can take this provider away.

andreievg:
Was hoping that login and userPermission endpoint return the same things (although it would happen only on success for login endpoint). Maybe `getUserPermission` method in back end is misleading, and should be getUserInfo (which also includes permissions)

And when either end point is queried (either on login or on reload), userInfo is returned and set in context.

CarlosNZ:
> Was hoping that login and userPermission endpoint return the same things (although it would happen only on success for login endpoint). Maybe `getUserPermission` method in back end is misleading, and should be getUserInfo (which also includes permissions)
> 
> And when either end point is queried (either on login or on reload), userInfo is returned and set in context.

Got it. So -- to summarise:

- Rename `getUserPermission` endpoint to `getUserInfo` and have it return user info with permissions
- Login component gets JWT and username from /login endpoint
   - sets JWT to localStorage
- When logged in and app loads (after successful login or reload when JWT already stored)
  - UserArea makes call to `/getUserInfo` which returns Permissions **and** Userinfo based on username
  - UserArea writes user info to User context

Sound right?

andreievg:
Just a quick clarification:

> When logged in and app loads (after successful login or reload when JWT already stored)
> UserArea makes call to /getUserInfo which returns Permissions and Userinfo based on username

I think either way on reload (even if JWT is missing) we do `getUserInfo`, so that 'nonRegistered' user is `loaded` JWT etc.. for GraphQL permissions

And since `/getUserInfo` return the same result as `/login` endpoint, we store JWT token returned by `/getUserInfo` to local storage. I think most of this is functioning, just the extra bits with user info and renaming endpoint. Does that make sense ?

CarlosNZ:
Okay, have re-done according to summary in [above comment](url).

Note: changes to back-end too, will need to re-Pull.

Ready for re-review.

src/utils/types.ts
```diff
@@ -107,9 +108,10 @@ interface EvaluatorParameters {
   APIfetch?: Function
 }
 
-interface FullUserPermissions {
+interface FullUserInfo {
   username: string
```
andreievg, line 112: technically speaking this wont' be needed anymore

src/utils/hooks/useGetUserInfo.ts
```diff
@@ -1,26 +1,24 @@
 import { useEffect, useState } from 'react'
```
andreievg, line 1: would this method get re-fired after login ?

andreievg: I think the reason I first thought that login url and login logic should return and set the same stuff as this hook, was because this hook happen above everything else (the first hook to fire up). And wasn't sure how to re-fire it on login

src/utils/types.ts
CarlosNZ: 💨 Gone

src/utils/hooks/useGetUserInfo.ts
CarlosNZ: Yes, it gets called by UserArea, which can't load until log in is successful.  Is that what you meant?

(It's showing what it's supposed to in the UI and all the user data is getting loaded after login)

andreievg: hmm, I see why you wanted to store username in local storage now. I originally thought of slightly different workflow. With this way we do need to store something in local storage apart from JWT (or access JWT to figure out username is nonRegistered)

andreievg: As per discussion look good to go

andreievg (approved):

PR #229, opened 2020-12-17
#214 Fix button click validation not waiting for in-focus element to update

### LINKED ISSUE #214: Bug: Clicking Next while still focused in Text input field doesn't validate
Same for clicking "Review & Summary". If you click the buttons while still in an input box, you'll get the Invalid error.

This must be because the result of the input is not written to database until onLoseFocus. 

--- 

### Proposed solution (UPDATED from below comments and discussion):

- Create a new state value in ApplicationState Context Area called `inputElementsActivity` with the following shape, and a "set" action for each timestamp field:
```
{
  elementEnteredTimestamp: <Date> 
  elementLostFocusTimestamp: <Date> 
  elementsStateUpdatedTimestamp: <Date>
  areTimestampsInSequence: <boolean>
}
```
- ApplicationViewWrapper imports `setElementEnteredTimestamp` and `setElementLostFocusTimestamp` methods, and creates one new method `setIsActive()`, which simply runs `setElementEnteredTimestamp` if true, and `setElementLostFocusTimestamp` if false. The `setIsActive` method is passed to the plugin as a prop.
- Plugins have an `onFocus` event which calls `setIsActive()`
- The `setActive(false)` call can be made at the end of `onSave` (need to make sure async calls within onSave (such as `responseMutation`) are `await`ed).
- Add `setElementsStateUpdatedTimestamp` to `useGetResponsesAndElementState` and call it last in the `useEffect` responding to changes in `responsesByCode`
- Whenever one of these "set" actions runs in ApplicationState, the `areTimestampsInSequence` boolean is recalculated.
- On the Navigation (Prev/Next) buttons, Progress bar links, and the Review & Sumary button handlers, add a check to `areTimestampsInSequence` value -- if `false` don't proceed. Maybe add a `while (areTimestampsInSequence === false)` loop (with appropriate timeout) so operation can proceed once update is finished.

CarlosNZ:
~~**UPDATE**: Just a heads-up that I've realised that I didn't add this check for the Progress bar links. Will update this PR with that once #206/#230 is merged.~~  **DONE now** -- _CJS_.

---

Fixes #214.

Differences from the [proposed solution](https://github.com/openmsupply/application-manager-web-app/issues/214#issue-762257633):

- Due to the issue raised in [this comment](https://github.com/openmsupply/application-manager-web-app/issues/214#issuecomment-747044357), we have to update `elementsStateUpdatedTimestamp` in two different places:
   - In "ApplicationPageWrapper" -- useEffect in response to changes in `elementsState`
   - In "ApplicationViewWrapper" -- at the end of `onSave`, but only if there's no change in the response (which is why it won't fire in the above situation)
 - Getting the button click handlers to wait until `areTimestampsInSequence` was a bit tricky. Open to better suggestions, but my solution was:
    - new `useState` hook for `[buttonClicked, setButtonClicked]`, which is what the click event calls.
    - `useEffect` hook that watches `areTimestampsInSequence` AND `buttonClicked`, which calls the previous click handler function ONLY IF both are true.
    - `setButtonClicked(false)` on page validation error or page change.

Note that the `dropDown` plugin doesn't require any changes, as any selection causes immediate lose focus event.


src/containers/Application/NavigationBox.tsx
```diff
@@ -54,6 +64,13 @@ const NavigationBox: React.FC<NavigationBoxProps> = (props) => {
     } else sendToPage(currentSection.code, nextPage)
   }
 
+  //
```
nmadruga, line 67: Missing a comment here? I think that's a good idea to describe the use here for this `useEffect`

src/containers/Application/ApplicationPageWrapper.tsx
```diff
@@ -166,6 +183,12 @@ const ApplicationPageWrapper: React.FC = () => {
     return application?.isLinear ? validation === (PROGRESS_STATUS.VALID as ProgressStatus) : true
   }
 
+  useEffect(() => {
+    if (areTimestampsInSequence && summaryButtonClicked) {
+      handleSummaryClick()
+    }
+  }, [areTimestampsInSequence, summaryButtonClicked])
```
nmadruga, line 190: Same thing here, deserves a comment

src/containers/Application/ApplicationPageWrapper.tsx
```diff
@@ -183,8 +206,10 @@ const ApplicationPageWrapper: React.FC = () => {
       })
     })
 
-    if (!revalidate.allValid) setShowModal({ open: true, ...messages.VALIDATION_FAIL })
-    else push(`/application/${serialNumber}/summary`)
+    if (!revalidate.allValid) {
+      setShowModal({ open: true, ...messages.VALIDATION_FAIL })
+      setSummaryButtonClicked(false)
+    } else push(`/application/${serialNumber}/summary`)
```
nmadruga, line 212: Should do the same in the `else` statement?

nmadruga (changes requested):
All seems good to me.

There is a tiny requested change, please. And just 2 places to add comments, it's not clear to me how this is working but I guess I'll see now while testing.

src/containers/Application/ApplicationPageWrapper.tsx
CarlosNZ: No need I think, since the Summary page (where it gets sent) has no interest in the state of Summary button. (And if you tried to back to this page, the state is reset.)

nmadruga (approved):

CarlosNZ:
@andreievg you might want to have a look over this to check that it's doing what we envisioned (and note my caveats in the post up top).

CarlosNZ:
Okay, I'll merge this now since @andreievg isn't back for a few more days. Any additional changes required can be added as bug fix issues next week.

PR #230, opened 2020-12-17
#206 Fix problem with validation check on progress bar

### LINKED ISSUE #206: Bug: Fix problem with allowing user to go to another page (after the first one is valid) on linear Application

nmadruga:
Fix #206

### Changes
- Simply check for the previous page of the **aimed** page the user wants to go. Using the same `validatePreviousPage` in both `NavigationBox` and `ProgressBar` for reusability, although it made the process a bit less clear when click **Next**. 

### Steps to reproduce bug (on `master`):
1. Create a new "Review Test form" application
2. Fill all elements in page 1, loose focus on last on so it will validate
3. Now click on Section 2 **Product Info**, and it will wrongly take you there!

### DIFF
--- a/src/components/Application/ProgressBar.tsx
+++ b/src/components/Application/ProgressBar.tsx
@@ -1,7 +1,13 @@
 import React from 'react'
 import { Accordion, Container, Grid, Header, Icon, Label, List, Sticky } from 'semantic-ui-react'
-import { ProgressInApplication, ProgressInPage, ProgressStatus } from '../../utils/types'
+import {
+  CurrentPage,
+  ProgressInApplication,
+  ProgressInPage,
+  ProgressStatus,
+} from '../../utils/types'
 import strings from '../../utils/constants'
+import { useRouter } from '../../utils/hooks/useRouter'
 
 interface SectionPage {
   sectionIndex: number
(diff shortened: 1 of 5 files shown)


CarlosNZ (changes requested):
Looks good, but just to clarify what we just discussed IRL:

- The name `validatePreviousPage` is confusing as it's not clear if you're validating the page previous to the one you're on, or the one your trying to go to.
- Suggest breaking into two functions:
   - `validatePage` which validates any page you pass to it (with a default to the current, perhaps?)
   - a function to get the Section and Page of the one you wish to validate (previous to the one you're trying to link to, isn't it?) which you then pass to `validatePage`.

Make sense?

src/containers/Application/ApplicationPageWrapper.tsx
```diff
@@ -121,17 +120,33 @@ const ApplicationPageWrapper: React.FC = () => {
     setPageElements(elements)
   }, [responsesLoading, currentSection, page, elementsState])
 
-  const validateCurrentPage = (): boolean => {
-    if (!application || !currentSection || !page) {
-      console.log('Problem to validate - Undefined parameters')
+  const validatePreviousPage = (sectionCode: string, pageNumber: number): boolean => {
```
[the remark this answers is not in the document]
nmadruga: I'll try to work on this change now

src/components/Application/ProgressBar.tsx
```diff
@@ -32,6 +39,11 @@ const ProgressBar: React.FC<ProgressBarProps> = ({
     return status ? indicator[status] : null
   }
 
+  const getPreviousPageIsValid = (sectionCode: string, pageNumber: number) => {
```
CarlosNZ, line 42: I kind of like functions that return booleans to start with `is` -- so `isPreviousPageValid`. 

No biggie though, I just think it reads clearer (as opposed to `get` which suggests I'm expecting a spefiic value back).

nmadruga:
Forgot to say here - I have done the changes as suggested.

CarlosNZ (approved):
LGTM! 
Just the one tiny comment -- not essential.

src/components/Application/ProgressBar.tsx
nmadruga: Oh, I agree. That's a good idea.
Will change this now and start merging PRs!

PR #233, opened 2020-12-29
#182 Show the start page with title and message defined for template

nmadruga:
Fix #182 

Using [Backend-#182F](https://github.com/openmsupply/application-manager-server/pull/162).

## Changes
- Layout on start application page
- Fetches `startTitle` and `startMessage` for template to be displayed on start application page.
- Add every string as constant to the constants file.
- Show application page if start message is defined.
- [Existing behaviour] Create the application when 'Get started' button is pressed.

## To do next:
- [#232] If `startTitle` or `startMessage` not defined - go straight to start page OR last invalid (or incomplete) page.
- [#236] Show progress on start page of Draft application:
  - Show progress on each section and validation icons.
  - Hide 'Get started' button when application already exists.
  - Show 'Resume' button on the last section with invalid (or incomplete) questions.

PR #234, opened 2021-01-01
#138 submission page

### LINKED ISSUE #138: Display submission page after submit an application
**Updated description on 29/12 after creation of Epic#24**

## Scope
- Show submission message on Summary page upon successful submission
- After the application is submitted (and the trigger finished to run) go to the submission page after submission
- Show alert (modal?) if submission unsuccessful before re-directing to first invalid page (this part already done)
- Link to summary page and back to template list (e.g. Drug registrations) when the trigger has finished processing and the submission page is shown as it would for a completed application - the submission page should continue to be accessible.
- When submission page is accessed before the application is submitted, redirects to first invalid page. 

Overview-of-Navigation- -URL-structure-Submission-page

nmadruga:
Fixes #138 

~~Branched off PR #233, please merge that one before this one!~~ DONE

To be used with back-end running PR [138F](https://github.com/openmsupply/application-manager-server/pull/163)

## Changes
- Add documentation about new **Submission page**.
- Break down hook to get appStatus from `useLoadApplication` to `useGetApplicationStatus`, it also returns application current stage and each stage (with description) of the application template.
  - New query: `getApplicationStatus` and new hook: `useGetApplicationStatus` which will wait for triggers to be completed before showing the current status of the application.
- Create new container `ApplicationSubmission` and connect to route from /application/{serial_number}/submission URL
- Make layout of **Submission page** similar to UI design. With links back to **Summary page** and **Dashboard**.
- Add fragment templateStage.

## Other unrelated changes
- Change hooks `useLoadApplication`, `useUpdateApplication` and `useGetResponsesAndElementState` to use same received type: **UseGetApplicationProps** since they were the same.
- Create link to go back to list of application from the application **Start page**.

## Problems
- There is an error "Uncaught Invariant Violation: Missing field 'serial' while computing key fields" happening when starting a new application that I tried to find the source but couldn't. Think it's related to an Apollo warning that shows sometimes about a different in references. But this still need some investigation - the app continues to work, but the error is unclear.


src/components/Application/ApplicationStart.tsx
```diff
@@ -18,7 +19,8 @@ const ApplicationStart: React.FC<ApplicationStartProps> = (props) => {
   return template ? (
     <Segment.Group style={{ backgroundColor: 'Gainsboro', display: 'flex' }}>
       <Button
-        as="href"
+        as={Link}
+        to={`/applications`}
```
CarlosNZ, line 23: Shouldn't this link to the specific application type?
`/applications?type={applicationType}` Something like that?

I know we haven't implemented list query routing yet, but it can't hurt to send it to what will be the right place.

CarlosNZ (changes requested):
I've only done a partial review, but there is a bug that needs fixing:
When I click "Submit" on the UserRegistration form, I get an error that the username is not unique. And when I check the database, the user has been created already.

So something wrong with the order in which these events are happening.

It should be (I think):
OnClick (submit button)
- revalidate all (existing behaviour)
- If okay, set application Trigger => OnApplicationSubmit (which triggers Action to create user), then go to /submission page
- If not okay, show error.

It must be doing the revalidate after setting the Trigger, which is why the user already exists when it does the username unique check.

src/containers/Application/ApplicationSubmission.tsx
```diff
@@ -0,0 +1,102 @@
...
+    if (statusLoading) return
+    // Check if application is in Draft or Changes required status and redirect to the summary page
+    // Note: The summary page has its own redirection logic to ay specific page (with invalid items).
+    if (
+      appStatus?.status === ApplicationStatus.Draft ||
+      appStatus?.status === ApplicationStatus.ChangesRequired
+    ) {
+      push(`/application/${serialNumber}/summary`)
+    }
+  }, [appStatus])
```
CarlosNZ, line 44: The problem is somewhere in here. The "appStatus" is still "Draft", so it's re-directing back to Summary.

It shouldn't be allowed to acquire the status until all Triggers are reset, so it's jumping the gun here somehow. Will investigate further...

CarlosNZ: Okay -- it's a cache issue. Even when the Status has been set to "COMPLETED" in database, this hook is still returning the cached value of "DRAFT". Need to force a re-fetch from server

src/components/Application/ApplicationStart.tsx
nmadruga: Yes. I guess that's better, I will change to this now

src/utils/hooks/useLoadTemplate.tsx
```diff
@@ -46,14 +46,13 @@ const useLoadTemplate = (props: useLoadTemplateProps) => {
       return
     }
 
-    const { id, code, name, startMessage, submissionMessage, templateSections } = template
+    const { id, code, name, startMessage, templateSections } = template
```
CarlosNZ, line 49: Just wondering what the reason for returning "submissionMessage" from AppStages is, rather than here?

nmadruga: I just thought that it would be better as a single-hook call in the Submission page - that gets all that is related to it. And calling this hook would be only needed when you actually want to see all the elements listed for a template.

src/containers/Application/ApplicationSubmission.tsx
nmadruga: I guess this is fixed with by bug fix done in [5fad684](https://github.com/openmsupply/application-manager-web-app/pull/234/commits/5fad684b765ea8317e31410ebaa1a5a78b84b78c)?

src/utils/hooks/useLoadTemplate.tsx
CarlosNZ: Right, and the submission message isn't used anywhere else, is it. Cool.

src/components/Application/ApplicationStart.tsx
nmadruga: Done: [ebc4a9f](https://github.com/openmsupply/application-manager-web-app/pull/234/commits/ebc4a9f06ba22c76c3f3706f83266848219ae7a6)

src/containers/Application/ApplicationSubmission.tsx
CarlosNZ: Yup! 👍🏼

CarlosNZ (approved):
Cool -- let's get this merged! :)

nmadruga:
🎉

Issue #237, opened 2021-01-05
Add Markdown interpretation using third party library and use for start, submission pages


nmadruga:
TODO

Add template with markdown for start and submission messages.

Issue #182, opened 2020-12-07
Add start message and submission message for Templates


nmadruga:
**Updated description on 29/12 after creation of Epic#24**

## Scope
- Add two new TEXT fields to `template` table: `start_message`, `submission_message`
   - Submission message should have a default: "Thank you... etc.", but NOT the start message
- Write start and submission messages for existing templates (put in `insertData.js`)

CarlosNZ:
Suggestion: If "start message" is null in database, the application goes straight to first page.
We can use this for "user registration", so new users don't have to go through the "Start application" page.

CarlosNZ:
Here's my suggestion for breaking this into separate issues:

## Issue 1 (Effort 1?)
- Add two new TEXT fields to `template` table: `start_message`, `submission_message`
   - Submission message should have a default: "Thank you... etc.", but NOT the start message
- Write start and submission messages for existing templates (put in `insertData.js`)

## Issue 2 (Effort 2?)
- Front-end update to query these fields in the getTemplate query
- Display start message in ApplicationStart (ideally using evaluator, though that could wait until we create [stringSubstitution operator](https://github.com/openmsupply/application-manager-server/issues/158))
- If start_message is NULL, automatically create the application and go to the first page (we'd use this for userRegistration)

## Issue 3 (Effort 2-3?)
- Show submission message on Summary page upon successful submission (this is what we agreed we'd do, right? Not a new page?)
- Hide Submit button and Edit buttons after successful submission
- Show alert (modal?) if submission unsuccessful before re-directing to first invalid page (this part already done)

nmadruga:
@CarlosNZ Thank you, I'm adding this new Epic now. I'll keep this one as an issue (issue 3 I think) and create a new one where I'll update with the list of issues as suggested in here. Will update this one shortly to point to it as well.

PR #240, opened 2021-01-05
#237 Markdown module

CarlosNZ:
Implements #237.

This is branched off #182, so I haven't used the module in a Submission page yet, but this should be super-easy to do in the Submission page PR using the new Markdown component.

Essentially, I've created new component called "MarkdownBlock" in `semanticReactMarkdown.tsx` which uses the [ReactMarkdown library](https://www.npmjs.com/package/react-markdown) to convert the input Markdown text to HTML, but with additional rendering over-rides which allow us to specify what kind of SemanticUI component should be returned instead of straight HTML.

The only component type I've defined so far is "Message", which is implemented on the Start page. Additional renderers can be added to semanticReactMarkdown.tsx as required. If no semantic component is specified, the module will return straight HTML.

I've also added the module to the TextInfo plugin, which required minimal changes. You can see it in action on the "UserRegistration" form -- the markdown in the info box shows the **bold** text correctly now.

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,10 +23,10 @@
     "@openmsupply/expression-evaluator": "^1.2.1",
     "apollo3-cache-persist": "^0.8.0",
     "graphql": "^15.3.0",
-    "marked": "^1.2.7",
     "query-string": "^6.13.5",
     "react": "^16.13.1",
     "react-dom": "^16.13.1",
+    "react-markdown": "^5.0.3",
     "react-router-dom": "^5.2.0",
     "serve": "^11.3.2"
   },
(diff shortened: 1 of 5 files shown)


nmadruga (approved):
Very good! It's looking so nice.
Interesting to use this to set how `Messages` should look like. Perhaps we should have different categories of messages here and show it differently for error, information or warning. All with the existing option available in Semantic.

CarlosNZ:
>  Perhaps we should have different categories of messages here and show it differently for error, information or warning.

Yeah you can do that -- I've allowed for the info or error prop to be passed along if desired, so can specify which type of Message box to display.

Issue #149, opened 2020-11-27
Password hashing (and salting)


CarlosNZ:
Part of [Epic #12](https://github.com/openmsupply/application-manager-web-app/issues/95)

### Back-end:

1.  ~~Add new column to `user` table -- `password_salt`~~ (Don't need -- bcrypt incorporates salt into hash string)
2. ~~Amend "createUser" Action to take `passwordSalt` parameter and write this value to above field~~ (Don't need)
3. ~~Update template with expression to extract `passwordSalt` field from response object for Action~~ (Don't need)
4. Create new endpoint `/createHash`, which takes in password, generates random salt, hashes password+salt (bcrypt), then returns `hash` (Might not need this -- initial hashing can probably be done in front-end password plugin, since salt is generated simultaneously using `bcrypt`)
5. Amend `/login` endpoint to take username & password, and perform hashing using stored salt. (as opposed to calculating hash in browser)

### Front-end:

1. Amend `login` method to not do hashing (send password to server)

### Form Element plugin: (To do in separate issue #253)

Create a new formElement plugin called "passwordCreate", which works as follows:
1. Displays similar to `shortText`, with same parameters, except user input is masked "•••••••"
2. Checkbox next to input field to "Show password" (Changes masked chars to text)
4. Show a second input field for password confirmation.
3. On Input Lose Focus (Application View),
 - Check both password inputs match, set error state if not.
 - If they do: send password to above `/createHash` endpoint, then put returned `hash` into response.text. Ensure actual password is NEVER stored in response or it could be saved to DB.
4. Summary view: Display password response as "•••••••" (it'll have to be arbitrary length, since original password won't be accessible from here)
5. Note that password cannot be reloaded if the form is refreshed, or paged back from Summary, as it's not stored remotely. The user will have to re-enter it, so we need to set the `initialValue` back to "" for password types on page load. (This should fine, as user registration should be filled in and submitted in one go, as they won't have any way to re-access their application form yet -- otherwise we'll have to do a lot more work to keep the password available locally without sending it to the server)

**Note**: seems like varying opinions on whether to do server-side or client-side hashing. However I found this to be the most convincing argument for server-side: https://myers.io/2016/09/08/why-you-shouldnt-hash-on-the-client/

CarlosNZ:
And are there any preferred libraries for hashing, random-string generating, etc.? [bcrypt](https://www.npmjs.com/package/bcrypt)?

andreievg:
[...]

PR #241, opened 2021-01-05
#216 update user area with current user

### LINKED ISSUE #216: Update the userArea to login with another user
Update the UserArea to use the information from user updated on the User context. And update the code to allow login with a different user (after logout the current one!).

Ideally make the user area more similar to what is in the UI design. Don't spend too long on this layout change! This should be just a tiny layout refactor to look a little bit better. 
Actually making the User Area look exactly like designed in **Epic#10 - Layout and menus** #148.

Design:
Screen Shot 2020-12-14 at 10 51 00 AM

nmadruga:
Fixes #216 

### Changes
- Layout on UserArea to look more similar to UI Designs
- Add selection to change current user on the go
- Move hook `useGetUserInfo` to a utility function, it receives the dispatch method to set a new user after a successful fetch
- [EDITED] Move `login` and `logout` function to be exported from the **UserState** context. It also checks on load if there is a persisted user (in local storage) to login the user.

### Changes unrelated
- Update query to get Current user information to also store the `id` and uses the id on the existing Reviewer query.
- Update a few hardcoded strings to be using a constant or message define in the specific files


src/containers/User/Login.tsx
```diff
@@ -73,7 +73,7 @@ const Login: React.FC = () => {
 
 export default Login
 
-function hashPassword(password: string) {
+export function hashPassword(password: string) {
```
CarlosNZ, line 76: I wouldn't bother with this, since password hashing will be done on the Server, as per #149.

CarlosNZ (changes requested):
Just before I look at this any deeper:
- There's a problem been introduced with the order things are loading, I think. When I log in, the Home page (Home.tsx) doesn't show the first name in "Hello XXX, welcome to the Dashboard"., and the name doesn't show in the Users dropdown.  If I switch users, it shows the new user correctly, but when I reload the page, the usernames aren't showing again.

src/containers/User/Login.tsx
nmadruga: Ok, I'll remove that

nmadruga:
Cool. It took me sometime on this one, but I think it's good to be re-reviewing now.

Thanks for spotting this problem, Carl.

src/containers/Main/AppWrapper.tsx
```diff
@@ -30,15 +30,21 @@ const AppWrapper: React.FC = () => {
...
     return (
       <>
         <UserArea />
         <SiteLayout />
         <Footer />
       </>
     )
-  else return <Redirect to={{ pathname: '/login', state: { from: location.pathname } }} />
+  } else {
+    console.log('NOT LOGGED')
```
CarlosNZ, line 44: Delete log.

CarlosNZ (approved):
Nice, working really well now and all seems very logical and clean. Thx.

2 console.logs to remove.

src/containers/Main/AppWrapper.tsx
```diff
@@ -30,15 +30,21 @@ const AppWrapper: React.FC = () => {
 }
 
 const AuthenticatedContent: React.FC = () => {
-  if (isLoggedIn())
+  if (isLoggedIn()) {
+    console.log('LOGGED')
+
```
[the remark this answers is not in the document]
nmadruga: Thanks! Removing it now

PR #238, opened 2021-01-05
#159B Evaluator refactor

CarlosNZ:
Fixes [back-end issue #159](https://github.com/openmsupply/application-manager-server/issues/159). Refactors evaluator queries to accommodate changes made in [back-end PR #161](https://github.com/openmsupply/application-manager-server/pull/161)

NOTE: Must run `yarn install` to upgrade evaluator package version. (If you go back to another branch, you'll need to run it again to revert to previous version -- these changes are breaking changes and incompatible between evaluator version 1.2.1 and 1.3.0 (the latest))

Needs to use [back-end PR branch #166](https://github.com/openmsupply/application-manager-server/pull/166) in conjunction to test.

Issue #232, opened 2020-12-29
Display start page for draft application with existing start message


nmadruga:
## Scope:
- ~~Front-end update to query these fields in the getTemplate query~~
- ~~Display start message in ApplicationStart (ideally using evaluator, though that could wait until we create [stringSubstitution operator](https://github.com/openmsupply/application-manager-server/issues/158))~~ (Done basic display, just need to add evaluation)
- If start_message is NULL, automatically create the application and go to the first page (we'd use this for userRegistration)

CarlosNZ:
Note: `stringSubstitution` is now in PR: https://github.com/openmsupply/application-manager-server/pull/164

PR #242, opened 2021-01-06
#232 Start & Submission page improvements

CarlosNZ:
Implements remaining scope of #232, namely:
- Start (and Submission) message is evaluated with the expression evaluator before being displayed (as Markdown)
- If no start message is defined, application loads with no start page

Requires [back end branch #232F](https://github.com/openmsupply/application-manager-server/pull/168) to see its functionality:
- launch the **FeatureShowcase** application to see the message using expression evaluation.
- launch **UserRegistration** application to see start page skip.

~~**NOTE**: This still uses the old array format for the evaluator objects. It will need to be updated after #238 is merged, so hopefully that will be approved before this one and I'll merge it back into here and do those changes before this is merged.~~ (Done)


CarlosNZ:
Will also update Submission page to use the evaluator once #234 is merged.

nmadruga (changes requested):
I'm finding problem when trying to load the applications after the Start page has been displayed. The error showing on the console is this:
Screen Shot 2021-01-08 at 12 54 10 PM

CarlosNZ:
> I'm finding problem when trying to load the applications after the Start page has been displayed.
> It can create the application but is failing to show the initial page. The error showing on the console is this:
> Screen Shot 2021-01-08 at 12 54 10 PM

Is there a particular template this is happening with, or all of them?

nmadruga:
Yey! Finally working. The browser was confusing the evaluatorExpression package versions, it would work fine on another private instance that I was using to debug, but not for the regular Google Chrome instances. Solved now after clear my browser cached data.

nmadruga (approved):
Thanks for putting all of this together. Looks great!

I'm seeing some errors in the console which are related to 2 bugs in development introduced before this PR, so not problem with that. I think this one is good.

Issue #86, opened 2020-11-04
Add Markdown interpret for Elements plugin


nmadruga:
Maybe in the ApplicationWrapper we add this interpreter? Since it's going to be required for different fields in different elements. Some parameters that could be using the markdown interpret:

### Short Text Input
- **label**: `string` -- Text that shows in the HTML "label" attribute of the form element

### Text Information
- **text**: `string` -- body text to display

CarlosNZ:
Yes, great idea, and should be easy to do. If we import it in the ApplicationViewWrapper, then we can pass it down to every plugin.

We'd probably also want it in Summary (and maybe Template) Views too, right?

CarlosNZ:
To be done after #237

PR #252, opened 2021-01-09
#86 Markdown for plugins

CarlosNZ:
Implements #86.

Already had Markdown display for TextInfo plugin, but I've moved the import to the Wrapper(s) and the component is passed to plugin components as a prop (as discussed).
- applied Markdown interpreter to `label` elements in Application and Summary views.
- added new 'noParagraph' renderer to Markdown module -- this prevents labels being returned wrapped in `<p></p>` tags, which messes up the label display.

### DIFF
--- a/src/utils/helpers/semanticReactMarkdown.tsx
+++ b/src/utils/helpers/semanticReactMarkdown.tsx
@@ -31,7 +31,7 @@ const MarkdownBlock: React.FC<MarkdownBlockProps> = (props) => {
         return <Message children={input.children} info={info} error={error} />
       },
       heading: ({ node: { children }, level }: any) => {
-        const headingText = children[0].value
+        const headingText = children?.[0]?.value
         return (
           <Message.Header>
             {info && <Icon name="info circle" />}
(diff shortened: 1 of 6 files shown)


src/utils/helpers/semanticReactMarkdown.tsx
```diff
@@ -41,6 +41,12 @@ const MarkdownBlock: React.FC<MarkdownBlockProps> = (props) => {
         )
       },
     },
+    // Prevents returned HTML from being wrapped in <p></p> tags
+    noParagraph: {
+      paragraph: ({ node: { children } }: any) => {
+        return <span>{children[0].value}</span>
```
nmadruga, line 47: I'm not sure about this one. It's guarantee this will always be receiving an array that has a value? It might be saffer to check that before and return null otherwise (with a log message for us).

nmadruga (changes requested):
Just one point of concern, otherwise looks pretty simple. I'm not sure how to test this though

src/utils/helpers/semanticReactMarkdown.tsx
CarlosNZ: Okay, I've changed it to:
```
return <span>{children?.[0]?.value}</span>
```
So if any of the array or its element properties are missing, it'll just return an empty `<span></span>`

Same on the "heading" one above. 

c7a883a That okay?

nmadruga (approved):

Issue #254, opened 2021-01-10
Sticky headers on Summary and Review sections


CarlosNZ:
As discussed in [this comment](https://github.com/openmsupply/application-manager-web-app/issues/217#issuecomment-755819029), adding the Sticky headers is more complicated, so moving this to separate issue to be done later.

Issue #217, opened 2020-12-14
Collapsibility for application/review sections


nmadruga:
Should be used in Application Summary and Review pages. 
The current section should sticky to the top, collapsed sections don't stick.

Current design layout:
![Screen Shot 2021-01-06 at 3 17 28 PM](https://user-images.githubusercontent.com/5456533/103721363-56378c80-5032-11eb-8325-06e824f90da6.png)

CarlosNZ:
It's fairly simple to implement the Drop-down accordion, but I'm having trouble with the Sticky functionality.

The problem is, the Sticky requires a "context" attribute, and the [examples in Semantic](https://react.semantic-ui.com/modules/sticky/#usage-above-content) use `createRef` to define refs. But the "ref" attribute doesn't work in functional components (see [here](https://reactjs.org/docs/refs-and-the-dom.html#refs-and-function-components), and I don't really understand how createRef and useRef work enough to figure it out.

Will see what you guys think at next meeting.

CarlosNZ:
As discussed, made new issue for Sticky functionality: #254

PR #255, opened 2021-01-10
#217 Collapsible sections - Summary and Review pages

CarlosNZ:
Implements #217, but only includes the collapsible sections. The Sticky functionality has been moved to a separate issue (#254) as discussed.

Collapsible section headers visible on Application Summary and Review pages.


nmadruga (approved):
Perfect.

It's very simple implementation which I think is how we want to go. For the future I think it would be preferable to persist the opened sections and show all sections expanded for Applicant which I have added to the Epic#10 for tyding up the UI later.

CarlosNZ:
>  For the future I think it would be preferable to persist the opened sections and show all sections expanded for Applicant which I have added to the Epic#10 for tyding up the UI later.

Great, I've done the first part -- sections default to "open" on application Summary page.

Issue #253, opened 2021-01-10
Element plugin: Password entry


CarlosNZ:
To be done after #149.

## Agreed solution:

Create a new formElement plugin called "password", which works as follows:
1. Displays similar to `shortText`, with same parameters, except user input is masked "•••••••"
2. Checkbox next to input field to "Show password" (Changes masked chars to text)
4. Show a second input field for password confirmation.
3. On Input Lose Focus (Application View),
 - Check both password inputs match, set error state if not.
 - If they do: send password to above `/createHash` endpoint, then put returned `hash` into response.text. Ensure actual password is NEVER stored in response or it could be saved to DB.
4. Summary view: Display password response as "•••••••" (it'll have to be arbitrary length, since original password won't be accessible from here)
5. Note that password cannot be reloaded if the form is refreshed, or paged back from Summary, as it's not stored remotely. The user will have to re-enter it, so we need to set the `initialValue` back to "" for password types on page load. (This should fine, as user registration should be filled in and submitted in one go, as they won't have any way to re-access their application form yet -- otherwise we'll have to do a lot more work to keep the password available locally without sending it to the server)

Issue #173, opened 2020-12-06
Epic #14 - Application Submit


CarlosNZ:
On clicking "**Submit**" button:

- Re-validation (new calculation), so we can be sure everything is valid NOW (there may be some significant delay between filling in form and submitting it) (**make new issue**)
   - If Validation fails, show modal, then re-direct user to first invalid question.

If Validation is okay:

- Set ApplicationTrigger to `onApplicationSubmit` (**Done**)
- Wait until Application Trigger resets to `NULL` (with Timeout) ([Issue card](https://github.com/orgs/openmsupply/projects/6#card-49487563))
   - If timeout reached, inform user via modal of a problem, and inform system admin (via back-end Action?) **TO-DO**

If Submission Okay so far:
- Inform user of successful submission via modal (or new page), including template "Success" message (check email, whatever) **TO-DO**

After user clicks "OK" on modal:
- If user is `nonRegistered`, log out. #150 
- Else, redirect to Dashboard **TO-DO**

---

When loading Summary page:
- Currently, it performs re-validation and displays result in Summary. This is **under discussion**.

### Sprint 8
- [x]  #103 Redirection to the summary in case the application is submitted - Effort 2
- [x] #85 Needs to add newly application to cache in order to show on the list of applications - Effort 1
- [x] #45 Mutate application from cache - Effort 4
- [x] #114 Bug fixes for application loading/submitting - Effort 1
- [x] #113 Create application mutation in single-block- Effort 1

### Sprint 9
- [x] #68 Display application View page when click an application with Submitted status - Effort 1

### Sprint 11
- [x] #143 Unify types and reuse of evaluated responses object on SummaryView and ApplicationView - Effort 3
- [x] #122 Create link to sections from the Summary page - Effort 1
- [x] #158 Summary Wrapper -- validation - Effort 2

### Sprint 12:
- [x] #48 check validation of progress structure before submission - Effort 1
- [x] #193 Redirect to last page with invalid elements when user loads Summary of application invalid - Effort 1

PR #256, opened 2021-01-10
#149 Use password hash for login

CarlosNZ:
Implements #149. As discussed in that issue, I've left the "password plugin" to a [separate issue](#253) as there are additional complexities involved with that.

Requires [back-end PR #173](https://github.com/openmsupply/application-manager-server/pull/173) -- most of the logic changes are in there. This front-end is basically just naming/tidying of existing code.

Logging in with one of the existing users (password: "123456") should work with hash comparison on back-end.

### DIFF
--- a/src/containers/User/Login.tsx
+++ b/src/containers/User/Login.tsx
@@ -17,8 +17,8 @@ const Login: React.FC = () => {
 
   const handleSubmit = async (event: any) => {
     event.preventDefault()
-    const passwordHash = hashPassword(password)
-    const loginResult = await attemptLogin(username, passwordHash)
+    const loginResult = await attemptLogin(username, password)
+    console.log('Login Result', loginResult)
     if (!loginResult.success) setIsError(true)
     else {
       setIsError(false)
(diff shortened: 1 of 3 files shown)


src/containers/User/Login.tsx
```diff
@@ -17,8 +17,8 @@ const Login: React.FC = () => {
 
   const handleSubmit = async (event: any) => {
     event.preventDefault()
-    const passwordHash = hashPassword(password)
-    const loginResult = await attemptLogin(username, passwordHash)
+    const loginResult = await attemptLogin(username, password)
+    console.log('Login Result', loginResult)
```
nmadruga, line 21: I think this can be removed now

nmadruga (changes requested):
Nice and tidy, I'm having a look on the back-end now.

For now can I just ask something? Can you please also add the encrypted password replace `const hardcodedPassword = '123456'` on the `UserSelection` container so we can continue to login with different test users.

CarlosNZ:
> For now can I just ask something? Can you please also add the encrypted password replace `const hardcodedPassword = '123456'` on the `UserSelection` container so we can continue to login with different test users.

This shouldn't need changing -- the password is sent to the server, which compares its hash against the stored one. I think you'll find it's still working okay, no?

Or did you mean something else?

nmadruga:
> This shouldn't need changing -- the password is sent to the server, which compares its hash against the stored one. I think you'll find it's still working okay, no?

Yes! Sorry, what a confusion/ That's all good! I'll test again and see if this can be already approved

nmadruga (approved):

PR #258, opened 2021-01-11
#247 Bug fix -- appStatus getting wrong status

### LINKED ISSUE #247: Bug: Validation running on Submitted application (Summary view)
After submitting an application and clicking on the button to see the Summary page:
Screen Shot 2021-01-07 at 5 12 02 PM

When it loads the summary page is running a validation and showing a "not unique username" problem:
Screen Shot 2021-01-07 at 5 11 31 PM

CarlosNZ:
Fixes #247 

Okay, I _think_ this is fixed now.  🤞  The appStatus query should always get the most recent status from the network and re-direct appropriately, in both the "Summary" page, the application "Home" page, and the "Submission" page.

Please give it a good test and see if you can find any situations where it still fails or re-directs incorrectly.


src/containers/Application/ApplicationOverview.tsx
```diff
@@ -74,7 +76,7 @@ const ApplicationOverview: React.FC = () => {
     if (isApplicationLoaded && elementsState && responsesByCode) {
       revalidateAndUpdate().then(() => setIsRevalidated(true))
     }
-  }, [responsesByCode, elementsState])
+  }, [responsesByCode, elementsState, appStatus])
```
nmadruga, line 79: i think this `appStatus` can be removed now along with the `console.log` commented on the top.

src/containers/Application/ApplicationSubmission.tsx
```diff
@@ -48,7 +48,7 @@ const ApplicationSubmission: React.FC = () => {
 
   useEffect(() => {
     if (!isApplicationLoaded || statusLoading) return
-    console.log('appStatus', appStatus)
+    // console.log('appStatus', appStatus)
```
nmadruga, line 51: I think it would be better to remove this `console.log` too, so we don't end-up having many commented logs.

nmadruga (approved):
Looks good! I tested a little bit and all seemed fine. I think we will continue to test this as we work with it everyday. I'm glad that this has solved the issue. Finally!

Just 2 comments to address please

src/containers/Application/ApplicationOverview.tsx
CarlosNZ: No, the lack of this `appStatus` is the main thing that was causing the problem! 😄

Because the first time `appStatus` comes back `undefined`, in which case `isRevalidated` never gets set to `true`, and the application doesn't load in the UI. And so we need to make sure this `useEffect` re-runs when `appStatus` is eventually updated with it's proper value (in which case it skips revalidation and sets `isRevalidated` to `true`).

Does that make sense? It's been doing my head in for a while.

Issue #161, opened 2020-11-30
Dynamically display template options based on user permissions


nmadruga:
https://github.com/openmsupply/application-manager-web-app/issues/64

Show in the dashboard and headers?

### Proposed solution

1. Create new graphQL query hook to fetch all templates details from Back-end (that have status == AVAILABLE)
  * Alternative would be add condition to fetch using filtered by code with current user permissions
2. Create new custom hook `useListTemplates` to call the graphQL hook and use the current user `templatePermissions` to filter and return a list of template names.
3. Update the `UserState` context to set a `isLoading` flag, which is going to be used to prevent the custom hook to be called before all `templatePermissions` are received for the current user.
4. Update `AppMenu` component to receive the list of templates (codes and userRole - e.g **Apply**,**Review**) - that was fetched by the custom hook and display on the User Area on a horizontal menu (without sub-menus for now).
![Screen Shot 2021-01-11 at 10 43 57 AM](https://user-images.githubusercontent.com/16461988/104136304-1db0fd80-53fa-11eb-923d-ee999aff2c03.png)
5. The same custom hook is called on the `Dashboard` container to display each template name with a link to the list of applications with the filter by template type (code). The dashboard functionality will be more broadly worked as part of the [Epic#23 - User's Dashboard](#249).

nmadruga:
@CarlosNZ  and @andreievg please have a loo on the **Proposed solution** on this issue description to see if you have any other suggestions or I can continue with this approach.

One question: When we show the `User registration` page, for a non-registered user we don't want to show any other area of the web app. So I'm thinking if we should wrap the current `SiteLayout` container with an `Authorised` container to first wait for the logged user to finish loading (and fetch the templatePermissions). And have another `Unauthorised` container that doesn't display other parts of the app layout (user area, or list applications) that would only allow displaying the `User Registration` for a user logged as **non-authorised** user.

CarlosNZ:
So for now the menu will just show ALL available template types? We're not breaking down into "Categories" just yet? 
But yeah, seems good. I think the menu bar will get very cluttered once we have more than 3-4 template types, so we'll need to show them by category at some point.

[...]

Issue #150, opened 2020-11-27
Log out nonRegistered user after any application form submit


CarlosNZ:
Assuming there is a modal appearing after Submit, this logout should occur on modal close.

CarlosNZ:
Needs #145 done first so we know how the username is exposed.

PR #259, opened 2021-01-11
#150 Log out nonRegistered user

CarlosNZ:
Implements #150.

After submitting any application form, if currentUser.username === 'nonRegistered', run logout() method. This still allows the "/submission" page to show, but can't load anything else.

The buttons which link to other parts of the application on the /submission page are hidden if logged out.

Currently, the Sidebar still shows up, but this will be taken care of once we do the suggestion from [here](https://github.com/openmsupply/application-manager-web-app/issues/161#issuecomment-757552343)


nmadruga (approved):
That's great!

I would suggest using a constant to store the `nonRegistered` username in a unique place. That's all.

CarlosNZ:
> I would suggest using a constant to store the `nonRegistered` username in a unique place. That's all.

Great. Done! cf9299b

PR #261, opened 2021-01-11
#253 Password entry plugin

CarlosNZ:
Implements #253 - password entry plugin. Requires back-end branch from [PR#175](https://github.com/openmsupply/application-manager-server/pull/175)

_Please only review once #256 is approved._

There were some challenges for this one, in that a standard validation expression wouldn't work, as the actual password can never be exposed to the Wrapper for saving, only the hash. Solution was to do "custom validation" inside the plugin and return it as a "customValidation" field in the object returned to the wrapper. This will be the way we do custom validations for other plugins that require it.

### Changes:

- Updated Element-Type-Specs documentation
- Simplified "defaultValidate" method in ApplicationViewWrapper. There is now just the one "generic" validation method (which uses `validation` expressions), and any custom validations that are needed for particular plugins should be done inside the plugin.
- `onSave` method in ApplicationViewWrapper now behaves differently for "generic" validation plugins and plugins that return a "customValidation" field.
- code for plugin's ApplicationView is based on 'shortText', but significantly modified to handle the unique requirements of this plugin

Unrelated change: added `&& appStatus` to check for processRedirect in ApplicationPageWrapper.


documentation/internal/Element-Type-Specs.md
```diff
@@ -33,20 +33,53 @@ _Free-form, single-line text input element_
...
+
+Note that, as the password itself is never stored (only the hash), this fields input is reset if the page is reloaded.
+
+#### Response type
+
+_This describes the expected object that will be stored in the `application_response` table `value` field from the user's response_
+
+```
+{
+  text: <string> (Only used to display '•••••••' on the summary page)
```
nmadruga, line 75: What is this text storing? Just the number of characters in \* ?

src/formElementPlugins/ApplicationViewWrapper.tsx
```diff
@@ -10,10 +10,9 @@ import {
   ElementPluginParameters,
   ElementPluginParameterValue,
   ValidateObject,
```
nmadruga, line 12: Not used

src/formElementPlugins/ApplicationViewWrapper.tsx
```diff
@@ -103,20 +92,36 @@ const ApplicationViewWrapper: React.FC<ApplicationViewWrapperProps> = (props) =>
...
           value: jsonValue,
-          isValid: validationResult.isValid,
+          isValid,
         },
       })
-    if (jsonValue.text == allResponses[code]?.text) {
-      setApplicationState({
-        type: 'setElementTimestamp',
-        timestampType: 'elementsStateUpdatedTimestamp',
-      })
```
nmadruga, line 119: I'm trying to understand why this code is on the `IF` statement (when it's a generic validation) but not in here?

nmadruga (commented):
haven't finished this Review yet. I still need read the new `ApplicationView` for password. But so far a few comments and here, some weird results I got while trying out. Would like to see if this can be improved in this issue or we create a separated one:

1 - When remove the text from first input it's displaying both errors. Perhaps should show only one - on the first? What do you think?
Screen Shot 2021-01-15 at 10 32 05 AM

2 - After entering the correct password on the first input it doesn't update the error until the input loose focus. It should remove the error before you need to loose focus.
Screen Shot 2021-01-15 at 10 32 19 AM

documentation/internal/Element-Type-Specs.md
CarlosNZ: Yes, it's just '•••••••' regardless, not even dependent on the number of characters (which is what it should be, since not even the number of chars should be stored outside the plugin).

It serves two purposes:
1. Something to display on the Summary page
2. The revalidateAll method checks the `text` field of the Response and if blank considers that a fail, so this ensures it will pass that check.

src/formElementPlugins/ApplicationViewWrapper.tsx
CarlosNZ: This was from an older issue PR #229. Normally the `elementsStateUpdatedTimestamp` is set when `elementsState` is updated, but if the value hasn't changed it needs to be updated here.

And since the `elementsState` isn't updated (since it's external "value" doesn't change really) then I guess it would make sense to include this in the "internal validation" block as well. Thanks for bringing it to my attention.

Done in 6734ce6

CarlosNZ:
> 1 - When remove the text from first input it's displaying both errors. Perhaps should show only one - on the first? What do you think?
> Screen Shot 2021-01-15 at 10 32 05 AM

I think it's okay to show both messages in this case. However, I've just pushed a version where it only shows the bottom error if the first input is valid. Not sure which I prefer to be honest, so let me know what you think.

> 2 - After entering the correct password on the first input it doesn't update the error until the input loose focus. It should remove the error before you need to loose focus.

Yes, I thought this was going to be too tricky to fix for now, but I've managed to come up with a solution that works, I think. The logic in this plugin is pretty convoluted due to the limitation of the Wrapper not having access to the password value but still needing to get correct validation status from it, so let me know if you need some further explanation, or have some suggestions for improvement.

These changes in fcaba0c.

documentation/internal/Element-Type-Specs.md
nmadruga: Alright! Thanks, I was just curious

nmadruga:
> I think it's okay to show both messages in this case.

No problem. I agree, is just that the current UI seems a bit strongly prompting. But I think that both fields have a problem, so that's all right to show both errors.
 
> v

nmadruga (approved):
I'm still seeing the error on the first input after matching with the `confirm password` (and before removing focus of input), but I don't think that's a bit deal. So approving!

Thanks for working with it, looks great.

Issue #249, opened 2021-01-07
Epic #23 - Dashboard


nmadruga:
### Description
When the user logs in that's the first page they see and where they can get to each Template they have access.
Now this Dashboard isn't configurable, it would always show the user most recent activity and notifications.
There are links to open specific applications of User's interest from the Dashboard. And links to the list of applications using certain filters.

### Documentation
- [ ] Documentation on how to configure dashboards #725
- [ ] TODO: Add as part of [Epic #38](338) Public docs & tutorials

### Sprint 31:
- [x] Add dashboard layout (from demo) using categories - not filter filters #639
- [x] Options to filter list by number using the fields from Application view list for reviewer and assigner #640
- [x] Display on the dashboard the number of application based on the filters #641

### Sprint 32:
- [x] Add categories and filters into template schema inserts [Back-end#322](https://github.com/openmsupply/application-manager-server/issues/322)
- [x] #698 PR dashboard functionality from demo
- [x] Ask Debbie to refine dashboard look

### Sprint 40:
- [x] #824 
- [x] #830

Issue #78, opened 2020-10-29
Epic #10 - App Layout and Styles


nmadruga:
## Description
Top bar: App logo and name; 
User menu: Name, company, notifications, Log-out; 
Left nav bar: available options/categories and each options that the Applicant can view apply. All options will be linked to what permissions the Applicant has on the JWT Token - the minimal set is: Create a company and Join a Company.

### Documentation:
- ~~#184 Display application types in navigation based on current user permissions~~ Duplicated
- [x] #139 Component consistency/Style discussion

### Sprint 8:
- ~~#64 Front-end query server to get template types and actionable (based on template type) that the current user is allowed and display on App navigation~~ (Relocated to epic 11)

### Sprint 16:
- [x] #161 Dynamically display template options based on user permissions - Effort 3

### Sprint 28:
- [x] Remove many Semantic themes (that aren't in use) to speed up the first compilation for the front-end bundle #191 **PR #612**
- [x] Global font and colours #586 **PR #612**
- [x] App container & User Area -- layout & styles #148
#### Break-down to replace element tags styles for each existing page in the app
- [x] Update layout and use global styles: Application create page #593 PR #619
- [x] Update layout and use global styles: Application home page #597 PR #622
- [x] Add specific title to progress bars on SectionProgress line on ApplicationHome #621
- [x] Update layout and use global styles: List page #592

### Sprint 30:
- ~~Improve front-end hot reloading to only transpile CSS files with new changes #613~~ INVALID
- [x] Update layout and use global styles: Login page #591
- [x] Update layout and use global styles: Application page + Progress Bar #594 Part 1 - PR #647
- [x] Update layout and use global styles: Review home page #599 PR #650

### Sprint 31:
- [x] Update layout and use global styles: Application Summary page #595 PR #671
- [x] Update layout and use global styles: Review page + Submission modal #598 PR #673
- [x] Update layout and use global styles: Application Submission page #596 PR #679
- [x] Update layout and use global styles: Application + Summary + Template question plugins #600
- [x] Element plugins -- add "help" field #645 ~**In PR**

### Sprint 32:
- ~~Progress bar: UI when page has no required elements -- to consider #298~~ WONTFIX
- ~~Improvements to sticky section titles on ReviewPage and ApplicationSummary #588~~ INVALID
- [x] Quick fix: List layout issue (Edit draft) #661

### Sprint 33:
- [x] Change links in the list to go to the Review home page for reviewers and hide the expansion icon #703
[...]

PR #262, opened 2021-01-12
#161 dynamically display accessible templates

### LINKED ISSUE #161: Dynamically display template options based on user permissions

nmadruga:
Fixes #161 

### Changes
- Move Authenticated area to separated component
- Create `type` and update the `interface` to store each template permission on one object
- Removed the AppMenu from the `SiteLayout`, since it's now build based on the current user permissions.
- Dynamically build the `AppMenu` on the `UserArea` with properties to link to each template list (using filter `type` and `user-role` with the first in the list of permissions). Add one initial `home` link to display the user Dashboard.

## TODO
Still need to list the Template names on the dashboard. Should be done later as part of the [Epic#23 - Users Dashboard](#249).


CarlosNZ:
Quick initial comment: I think we need some way to access our templates and other links from the main page. We don't want to have to type URLs in.

Perhaps for now we can use the Footer area to add quick links to:
- application list
- all the template types that we had before (Feature showcase, etc.) (If we don't have permission to see them, at least we can still use the link to test the 404 page).

src/containers/User/UserArea.tsx
```diff
@@ -1,47 +1,73 @@
...
-          <Grid.Column style={{ width: '30%' }}>
+          <Grid.Column style={{ width: '80%' }}>
+            {error}
+            {menuItems && <Menu fluid inverted pointing secondary {...menuItems} />}
             <Segment inverted>{strings.TITLE_COMPANY_PLACEHOLDER}</Segment>
           </Grid.Column>
-          <Grid.Column style={{ width: '50%' }}>
-            <Segment inverted />
-          </Grid.Column>
           <Grid.Column style={{ width: '20%' }}>
```
CarlosNZ, line 54: Same here. Isn't the total "width" in semantic units 16? So the first column should be `width={12}` and this one `width={4}`?

src/contexts/UserState.tsx
```diff
@@ -1,10 +1,12 @@
+import Log from 'apollo3-cache-persist/lib/Log'
```
CarlosNZ, line 1: Remove this? (What is it for?)

src/components/Main/AppMenu.tsx
```diff
@@ -1,41 +1,40 @@
...
-  items: Array<Array<String>>
+  templatePermissions: TemplatesDetails
 }
 
-const AppMenu: React.FC<AppMenuProps> = (props: AppMenuProps) => {
-  const { pathname } = useRouter()
-  const {
-    userState: { templatePermissions },
-  } = useUserState()
+// THIS IS NOT WORKING YET - Code is in UserArea. To be moved here.
```
CarlosNZ, line 11: Okay, I'm ignoring this file for now. Any comments are in UserArea.tsx

CarlosNZ (changes requested):
Nice, thanks. Would really like to get those links in the menu bar working properly. Perhaps my suggestion might help? Apologies if not ;)

src/contexts/UserState.tsx
nmadruga: Oops. Mistake.

src/containers/User/UserArea.tsx
```diff
@@ -1,47 +1,73 @@
...
+        active: type === code,
+        onClick: () => push(`/applications?type=${code}&user-role=${permissions[0]}`),
+      }))
+      const homeItem = {
+        key: `app_menu_home`,
+        content: <Header as="h5" inverted icon="home" />,
+        active: location.pathname === '/',
+        onClick: () => push('/'),
+      }
+      setMenuItems({ items: [homeItem, ...items] })
```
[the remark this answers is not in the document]
nmadruga: >  Is is using the "shorthand" style.

Yep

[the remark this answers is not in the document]
nmadruga: > And I think it would be better to use the link attribute where possible, so they show up link conventional links -- maybe this will work properly too with the above suggestion?

I'm doing this because I'm setting this Menu items in the `useEffect`, that's why the shorthand option seemed alright, but if it's not so clear... Ok, I had an idea. I will try it.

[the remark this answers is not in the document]
nmadruga: Ah, I think it's not going to be very helpful... Maybe let's continue with using the `shorthand` I think once we get used is a good practice. Helps to preset the items on a component. Also this code should be moved to the AppMenu component at some point so it would be better allocated. 

I'll change to use the **Link** now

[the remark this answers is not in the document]
nmadruga: ok ok, I'll change to use <Menu.Item>, the **Link** is not working with the `shorthand` solution... I guess I can check for {filteredTemplates} on the render.

src/containers/User/UserArea.tsx
```diff
@@ -1,47 +1,73 @@
...
     replace('/login')
   }
 
-  return currentUser ? (
+  return (
     <Sticky>
       <Segment inverted vertical>
         <Grid inverted>
-          <Grid.Column style={{ width: '30%' }}>
+          <Grid.Column style={{ width: '80%' }}>
```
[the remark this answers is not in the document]
nmadruga: Yea, your are right. I'll change to that.

src/components/Main/AppMenu.tsx
nmadruga: Comment removed. Now it's using the AppMenu and Links got fixed, the UserArea needed to be inside the `Router` so moved to inside `SiteLayout` container.

src/containers/User/UserArea.tsx
nmadruga: Done: [441d675](https://github.com/openmsupply/application-manager-web-app/pull/262/commits/441d67506d3f338022fe9fe4914467d424cc610b)

src/containers/User/UserArea.tsx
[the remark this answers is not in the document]
nmadruga: Done: [a1adc13](https://github.com/openmsupply/application-manager-web-app/pull/262/commits/a1adc13548833f0d0b1cdc334884d4d5a3676fcd)

src/components/Main/AppMenu.tsx
```diff
@@ -1,40 +1,41 @@
...
+    templatePermissions && (
+      <Menu fluid inverted pointing secondary>
+        <Menu.Item
+          key={`app_menu_home`}
+          content={<Header as="h5" inverted icon="home" />}
+          active={pathname === '/'}
+          as={Link}
+          to={'/'}
+        />
+        {templatePermissions.map(({ name, code, permissions }) => (
```
CarlosNZ, line 27: Nice, I find this much easier to read and understand than when using the Shorthand syntax within the .map. Looks great.

CarlosNZ (dismissed):
Great, thanks for the changes. LGTM!

src/containers/User/UserArea.tsx
```diff
@@ -1,47 +1,55 @@
...
+    userState: { currentUser, isLoading, templatePermissions },
     logout,
   } = useUserState()
   const { replace } = useRouter()
+  const { error, loading, filteredTemplates } = useListTemplates(templatePermissions, isLoading)
 
   const handleLogOut = async () => {
     await logout()
     replace('/login')
   }
```
CarlosNZ, line 21: Problem with the way you've re-structured the routing: When you click "Logout", the Login component is rendered _inside_ the other stuff, so the head and footer are still visible on the page. After refreshing the './login' page it appears correctly (because it hits the /login route in AppWrapper first, presumably)

CarlosNZ (changes requested):

src/containers/User/UserArea.tsx
CarlosNZ: In this case (logging out), it's probably acceptable to use a javascript over-ride to force a full refresh:
`window.location.href = "/login"` (Tested it, it works)

nmadruga: That's for this suggestion. I have added it now, it still has a problem, but I have added a comment so we can go back to improve this when we are working on the tyding up of User Are in Epic#10 #78

CarlosNZ: Great. I don't even think it needs improving, I think when logging out its acceptable, even desirable, to force a refresh like this as it resets any React state that might be hanging around and does a proper "reset".

Maybe make the comment say `// Force full application reload on log out` or something like that? 

Anyway all good, approved. :)

CarlosNZ (approved):

Issue #179, opened 2020-12-06
Decision on response in the suggestion panel after clicking the review button


nmadruga:
In the suggestion area the reviewer select an option and add optional comment. When closing the suggestion area with some selection a `review_response` should be created (or updated if there was already one).
This issue includes creating a hook to send comments after entered by reviewer.

Scope:
a.Show the modal
b. Make mutation
c. Show the latest comment below the response - Already started by #156.

![Screen Shot 2021-01-05 at 2 15 40 PM](https://user-images.githubusercontent.com/16461988/103595532-8b73aa00-4f60-11eb-8f12-c4618f1a5f0d.png)
![Screen Shot 2021-01-05 at 2 15 28 PM](https://user-images.githubusercontent.com/16461988/103595536-8dd60400-4f60-11eb-80d9-50e643e21c3e.png)

PR #265, opened 2021-01-13
#225 Documentation for filters on list of applications

### LINKED ISSUE #225: Documenting List applications headers and functionality
Some columns should be base and would show for all template - independent on the user role.
Should we have columns specific for template. e.g Drug registration: `drug name` and `company`, Company registration would have `company`, `user`.  OR we would have the same columns for all templates, and use the `Application name` to set specific information on each template. - For iteration 1 we should keep it simple and use the specific `application name` 

There will be differences on what are the headers and what is showing in each cell based on the user role. 
Discussed that url should continue to be general one that has different filters to display based on a application template or application category. Also we will be using the user role as one of the filters to display applications to make simpler on which columns and actions will be displayed based on the user role. 

[EDITED 11/01/2020]

## Filters:
- type (default: first from list of template-type in `templatePermissions` list)
- user-role (default: first role of template-type in `templatePermissions` list)
- category
- stage
- sort-by
- page
- per-page
- search
- date (default:"Today")
- outcome (default: "pending")

### User-role related filters
- last-active-date [Applicant, Reviewer, Consolidator] - dates per user last action
- deadline-date [Applicant]
- action* [Applicant, Reviewer, Consolidator] - different values
- consolidator [Reviewer]
- assigned [Consolidator]

\* The **action** filter should list applications in the same `status`, by showing the main action that the user can make in order to continue this application workflow. And the possible values are defined per `user-role`.
E.g: **Action** `view` to list all application which the user doesn't have any available action (already submitted).

### Combined filters
- For different filters are use in the same URL the query will be using **AND** logic between different filters.
E.g: `applications/type=CompRego&date=today`
- For the same filter with more than one value the query will be using **OR** logic between values in same filter.
E.g: `applications/type=CompRego&date=this-week,last-week`

## Workflow:
1. Parse the url query
2. Each query should correspond to a filter on the GraphQL query
   - The query might be different depending on the user role
3. Get back all the applications to be displayed
4. Update URL with any changes needed
5. Check user role to display columns and cells with all application retrieved.

## Up-to-date document
More details about each filter and URL structure for listing application in this [document](https://docs.google.com/document/d/1zjEcxgw48MIii2KDytf5g7qHJi3VPqq4OV9Dmxz5tNw/edit)

nmadruga:
Fixes #225 

Documentation about filters to be done in [Epic#16 - List and Filter applications](#98)

Based on document: https://docs.google.com/document/d/1zjEcxgw48MIii2KDytf5g7qHJi3VPqq4OV9Dmxz5tNw


documentation/internal/Filters-for-Applications.md
```diff
@@ -0,0 +1,292 @@
...
+- Reviewer
+  `start-review`
+  `continue-review`
+  `review-updates`
+  `view` (Submitted)
+- Consolidator
+  `assign`
+  `consolidate`
+  `review-updates`
+  `view` (Submitted)
```
CarlosNZ, line 122: I mentioned this in the doc too, but I don't really understand how "actions" can be filters. Surely actions are links of things you can do, and are determined by your permissions and role?

documentation/internal/Filters-for-Applications.md
```diff
@@ -0,0 +1,292 @@
...
+
+Example: `last-active-date=2021-01-01`
+Considered in implementation: **Yes** [#228 - Dates in columns]
+Can have combined values: **Yes** e.g: `last-active-date=today,last-week`
+
+Options:
+
+- Pre-defined string: `today` (See more formats on Formats & Rules)
+- Single date: `YYYY-MM-DD`
+- Period (start:end): `YYYY-MM-DD:YYYY-MM-DD`
```
CarlosNZ, line 195: We should also be able to specify no start or end date by omitting one of the range boundaries
For example, `:2021-01-14` would mean all dates before today, and `2021-01-01:` would mean all dates after 1st Jan 2021.

documentation/internal/Filters-for-Applications.md
```diff
@@ -0,0 +1,292 @@
...
+
+##### last-active-date
+
+Example: `last-active-date=2021-01-01`
+Considered in implementation: **Yes** [#228 - Dates in columns]
+Can have combined values: **Yes** e.g: `last-active-date=today,last-week`
+
+Options:
+
+- Pre-defined string: `today` (See more formats on Formats & Rules)
```
CarlosNZ, line 193: I assume (for implementation) we'd just have a look-up table with a set of functions to convert pre-defined strings into the standard `YYYY-MM-DD:YYYY-MM-DD` format for further processing?

documentation/internal/Filters-for-Applications.md
```diff
@@ -0,0 +1,292 @@
...
+
+##### per-page
+
+Example: `per-page=20`
+Considered in implementation: **Yes** [#210 - Pagination]
+Can have combined values: **No**
+
+Options:
+
+- Number between 10 - 100
```
CarlosNZ, line 231: Why enforce a limit at all? I mean, we'd provide some "pre-set" options in the pagination controller, but there's no reason why the query interpretor has to have limits.

documentation/internal/Filters-for-Applications.md
```diff
@@ -0,0 +1,292 @@
...
+Can't have special caracters.
+Are case insensitive. So `thisTHAT` = `THISthat` = `thisthat`.
+
+#### Labels
+
+When combined labels are used (with separator `,`) the query will be using the OR connector.
+
+#### General
+
+When combined filters are used (with separator `&`) the query will be using the AND connector.
```
CarlosNZ, line 254: Probably should specify (somewhere) that if a filter is not provided, it will fetch ALL (i.e there's no such restriction on the query).

documentation/internal/Filters-for-Applications.md
```diff
@@ -0,0 +1,292 @@
...
+#### Applicant
+
+- Deadline date
+- Progress
+
+#### Reviewer
+
+- Company
+- Applicant
+- Consolidator
```
CarlosNZ, line 278: Also "Review stage"?

documentation/internal/Filters-for-Applications.md
```diff
@@ -0,0 +1,292 @@
...
+- Sections columns:
+  - Section title
+  - _Assigned to you_ / _Assigned to A Nother_
```
CarlosNZ, line 281:
```suggestion
  - _Assigned to you_ / _Assigned to another_
 ```
And why _italics_?

File: documentation/internal/Filters-for-Applications.md [hunk 8]
```diff
@@ -0,0 +1,292 @@
...
+- Status / Actionable
+
+#### Applicant
+
+- Deadline date
+- Progress
+
+#### Reviewer
+
+- Company
```
CarlosNZ, line 276: We actually have no way to tie an application to an organisation currently (applications only have user_id), so I've made an issue for it: [Back-end #179](https://github.com/openmsupply/application-manager-server/issues/179)

CarlosNZ: Also, I think we need to keep getting in the habit of using "Organisation" rather than "Company".

documentation/internal/Filters-for-Applications.md
```diff
@@ -0,0 +1,292 @@
...
+### Columns per user-role
+
+---
+
+### Common columns (For all user-roles)
+
+- Last active date
+- Application name
+- Stage
+- Status / Actionable
```
CarlosNZ, line 267: This is what I mean about the actions (above). It makes sense to have a list of "actionables" for each row like this, but not for filtering by it. 😕

CarlosNZ (changes requested):
Sweet, thanks for doing all this. Good to have a comprehensive reference all in one place.

One other thing if you can be bothered: since this article is so long, it'd be handy to have a few in-page links at the top to the main sections. See the [Query Syntax](https://github.com/openmsupply/application-manager-server/wiki/Query-Syntax#development) page for how to add them (there's an Installation and Development link in the top paragraph).

nmadruga:
New things to include to the document: 
**Sort-by**: Can sort more than one column at the same time. Would have a default sorted column? e.g. last-active-date? 
**Serial** column/filter to be included

nmadruga:
> since this article is so long, it'd be handy to have a few in-page links at the top to the main sections.

Sure thing. I'll do this now! Have checked on Query-syntax page, good to learn.

documentation/internal/Filters-for-Applications.md
```diff
@@ -0,0 +1,292 @@
...
+#### Label filters:
+
+##### type
+
+**This is a compulsory filter**
+In case none is included in URL will be deduced from 1st user's persmissions
+
+Example: `type=user-registration`
+Considered in implementation: **Yes** [#131 - Headers]
+Can have combined values: **No**
```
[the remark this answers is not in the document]
nmadruga: Done! Well spotted.

documentation/internal/Filters-for-Applications.md
```diff
@@ -0,0 +1,292 @@
...
+
+---
+
+### Filters
+
+---
+
+#### Label filters:
+
+##### type
```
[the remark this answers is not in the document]
nmadruga: Ok. Will do that now. Thanks for the suggestion

documentation/internal/Filters-for-Applications.md
```diff
@@ -0,0 +1,292 @@
...
+
+![Applicant-List](images/Filters-for-Applications-Applicant-List.png)
+
+---
+
+### Filters
+
+---
+
+#### Label filters:
```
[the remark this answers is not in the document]
nmadruga: In the document?

That was just a way which I think was helpful to help determining invalid values and acceptable values (even though no results) based on the `type of filter`.

- Labels = pre-defined strings (no spaces so not using quotes. I think that when we have a predefined 2+ words labels we can use the snake case to display on the URL for those)
- Strings = letters, numbers, special characters, spaces, etc

documentation/internal/Filters-for-Applications.md
```diff
@@ -0,0 +1,292 @@
...
+
+Example: `sort-by=stage:asc`
+Considered in implementation: **Yes** [#131 - Headers]
+Can have combined values: **Yes** `sort-by=stage:asc,application-name:desc`
+
+Options:
+
+- `column-name:direction`
+- direction [Optional] `asc`/`desc`
+- if no direction is included use default: `asc`
```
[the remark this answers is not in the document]
nmadruga: Yes, that's a good point. I think we can give a default order using a column that is not mirrored in the URL, for example sorting by `desc` date on the initial list that is displayed without any filters (just not to be a random order). 
This default was only for when the user clicks on a column for the first time, I haven't thought to much about it, now that you commented I do agree it should be `desc`. I'll add to the docs.

documentation/internal/Filters-for-Applications.md
nmadruga: Yes, actions are a set of possible things that each user-role can do, but I was thinking it would be useful to search and see only applications that you can `make-updates` for example. I know this is the same as a filter that would search for application with status == `Changes required` but since this is not visible in the UI I thought using actions would be a better filter for users. Note that there isn't one status filter for the same reason.

documentation/internal/Filters-for-Applications.md
nmadruga: I'll add that to the docs now

documentation/internal/Filters-for-Applications.md
nmadruga: I was thinking about just storing that on the Front-end. And converting on the front-end before calling the query.

documentation/internal/Filters-for-Applications.md
nmadruga: Ok. removing the limits

documentation/internal/Filters-for-Applications.md
nmadruga: Cool. I'll add this info to the docs

documentation/internal/Filters-for-Applications.md
nmadruga: Added. Well spotted

documentation/internal/Filters-for-Applications.md
nmadruga: Agreed. The filter name is already `org` so all good.

documentation/internal/Filters-for-Applications.md
nmadruga: Oh, sorry. Just saw after, there is one section **Common columns (For all user-roles)** which already list columns that are common for all user-roles.

documentation/internal/Filters-for-Applications.md
nmadruga: See my comment previously about that, I'm open to other ideas. Just thought using `actions` would reflect more what the user sees in the UI.

documentation/internal/Filters-for-Applications.md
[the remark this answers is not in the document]
CarlosNZ: Cool. I think it looks a little neater if these blocks just have line breaks (rather than paragraph breaks), but not a big deal either way. Up to you.
```
Example: `type=user-registration`  (2 spaces at line end forces a line break)  
Considered in implementation: **Yes** [#131 - Headers]  
Can have combined values: **No**  
Options:
- `template-code` - Of application type in list (and header)
```

documentation/internal/Filters-for-Applications.md
CarlosNZ: But wouldn't "review stage" only be visible to reviewers and above? The applicant shouldn't be able to see the review stage, right?

documentation/internal/Filters-for-Applications.md
CarlosNZ: Still wondering about the italics here. 😕

CarlosNZ (approved):
Great, very comprehensive.

I made a couple of minor comments, but only change them if you want to. Feel free to merge whenever you're ready.

documentation/internal/Filters-for-Applications.md
nmadruga: The `Stage` is regarding the Application stage, not sure what do you mean by "Review stage", are you saying "Review status"? In that case it would display in the `Status` column based on the user-role. So for applicant it is the `application.status`, for reviewer it should be `review.status` and for consolidator `consolidation.status`.

documentation/internal/Filters-for-Applications.md
nmadruga: Just to emphasises it is a string a dynamic label (as others).

Issue #116, opened 2020-11-13
Create login view for company


nmadruga:
To be done after #115 and using the extended end-point of login (listing the companies associated with the user).

UI - Should be a 2 step login:
* First page where user enters `username` and `password` (Done in a previous issue)
* Second page where we list user's companies and the user selects a company to login -> **What is aimed for this issue**

CarlosNZ:
Suggestion, though we should probably get Deb to design this page/process:

First login screen like this:
![Screen Shot 2020-11-26 at 12 06 47 PM](https://user-images.githubusercontent.com/5456533/100290304-e5876400-2fdf-11eb-8ccd-f6c59962e37d.png)
changes into this (without loading a new page) after username/password is authenticated:
![Screen Shot 2020-11-26 at 12 07 28 PM](https://user-images.githubusercontent.com/5456533/100290355-00f26f00-2fe0-11eb-9e29-7f34339cb59f.png)

CarlosNZ:
Just putting this here for discussion.

I'll implement this using the new permissions PR and my password hashing branch as a basis.

From my understanding, here is a rough outline of the **current** login/re-load process (sorry about the shoddy formatting, I find draw.io kinda awkward to use):
![Login procedure-Current Process](https://user-images.githubusercontent.com/5456533/104393092-7d4d0b80-55a8-11eb-89a1-b0860ff654f9.png)

And this is what I **suggest**:
![Login procedure-Suggested Process](https://user-images.githubusercontent.com/5456533/104393661-de291380-55a9-11eb-8d92-899555554c89.png)

Original draw.io diagram available [here](https://drive.google.com/open?id=10x5MythfzfNGUGUeCOrnx9Cf9JiDlOtl&authuser=carl%40msupply.foundation&usp=drive_fs)

### Comments

- I'd like to minimize the current double-up between the `/login` and /userInfo` endpoints, they're currently almost doing exactly the same thing. I think the login endpoint should be limited to:
   - authenticating login
   - providing a list of available organisations
   - JWT with userId and orgId (if selected/available)
   - minimal user info (only needed for displaying name on "Select Organisation" form)
- the `userInfo` endpoint shouldn't return JWT -- we already have it at this point
- the `currentUser` will also have `orgId` and `orgName` fields
- would like to rename the `login` method in UserState to `onLogin` to avoid confusion. It doesn't actually log in, it's the procedure that occurs after successful log in, so `onLogin` seems more appropriate.

[...]

PR #266, opened 2021-01-13
#116 Login for organisation

CarlosNZ:
Implements #116. Requires back-end from [PR #180](https://github.com/openmsupply/application-manager-server/pull/180)

Review after #256.

### Summary of changes

- UI and functionality for handle multi-step login process:
   - After username/password login, user is presented with list of available orgs
   - Full login process only completes after org is submitted (and verified) (i.e. write JWT to local storage, setCurrentUser in context)
   - Skip org login if no orgs available, and auto-login if only one org
- Changed `attemptLogin` function to a generic "POST request" function (and moved to new `fetchMethods.ts` file)
- UserState context doesn't re-fetch from `user-info` on login, since the information has already been returned. (It re-fetches when reloading based on stored JWT)
- Renamed UserState `login` method to `onLogin` so it's less confusing (it runs AFTER successful login, it doesn't actually log in itself)
- Organisation name appears in UI (in UserArea, etc.) (replaces placeholder "ABC Company")
- The UserSelection switcher automatically selects the user's first org in list to log in with

Please test to ensure no regressions in existing functionality. Users `nmadruga` & `carl` have multiple orgs, `andrei` has one, and `valerio` has none, so try logging in with each.


src/containers/Application/ApplicationSubmission.tsx
```diff
@@ -65,7 +65,9 @@ const ApplicationSubmission: React.FC = () => {
     <Loading />
   ) : serialNumber && appStatus ? (
     <Segment.Group style={{ backgroundColor: 'Gainsboro', display: 'flex' }}>
-      <Header textAlign="center">{strings.TITLE_COMPANY_PLACEHOLDER}</Header>
+      <Header textAlign="center">
+        {currentUser?.organisation?.orgName || strings.TITLE_NO_ORGANISATION}
+      </Header>
```
nmadruga, line 70: Hmmm perhaps we should move this surrounding to a common component?

src/containers/Review/ReviewPageWrapper.tsx
```diff
@@ -97,7 +97,7 @@ const ReviewPageWrapper: React.FC = () => {
 
       <Form>
         {reviewSections.map((reviewSection) => {
-          const { id, firstName, lastName } = currentUser as User
+          const { userId: id, firstName, lastName } = currentUser as User
           const assignedToYou = reviewSection.assigned?.id === id
```
nmadruga, line 101:
```suggestion
          const { userId, firstName, lastName } = currentUser as User
          const assignedToYou = reviewSection.assigned?.id === userId
```

src/utils/helpers/fetchMethods.ts
```diff
@@ -0,0 +1,22 @@
+// Generic GET/POST methods for re-use throughout app
```
nmadruga, line 1: Very nice to have! I wonder if it makes sense to create a folder just to store these fetch-related functions... I see that the `fetchUserInfo` could be using this helper at some point, no problem with it for now. We can change it to use when we see a commonality. I'm just thinking it would be good to also have the other fetching (with the constant to define the route) in separated files. Which is now inside the component `login`.

src/containers/User/Login.tsx
```diff
@@ -1,66 +1,135 @@
...
+                  onChange={handleSelection}
                 />
-              </Form.Field>
+              )}
               <Container>
-                <Button type="submit" onClick={handleSubmit}>
-                  {strings.LABEL_LOG_IN}
+                {!user && <Link to="/register">{strings.LINK_LOGIN_USER}</Link>}
+                <Button floated="right" type="submit" onClick={handleSubmit}>
+                  {!user ? strings.LABEL_LOG_IN : 'Proceed'}
```
nmadruga, line 131: Please add `Proceed` string to constants file

src/containers/User/Login.tsx
```diff
@@ -1,66 +1,135 @@
...
+                      type="password"
+                      value={password}
+                      onChange={(event) => setPassword(event.target.value)}
+                    />
+                  </Form.Field>
+                </>
+              )}
+              {user && (
+                // To-Do: Add to Messages
+                <p>{`Welcome back, ${user.firstName}. Please select your organisation.`}</p>
```
nmadruga, line 114: Also can this one be done already please? I think it is best to keep all these strings in a file already so we know all that we need to localise when the time comes

src/containers/User/Login.tsx
```diff
@@ -1,66 +1,135 @@
...
-                  name="password"
-                  type="password"
-                  value={password}
-                  onChange={(event) => setPassword(event.target.value)}
+              {!user && (
+                <>
+                  <Form.Field>
+                    <label>{strings.LABEL_LOGIN_USERNAME}</label>
+                    <input
+                      placeholder="Username"
```
nmadruga, line 93: Also good to have in the constants file

src/containers/User/Login.tsx
```diff
@@ -1,66 +1,135 @@
...
+                      name="username"
+                      type="text"
+                      value={username}
+                      onChange={(event) => setUsername(event.target.value)}
+                    />
+                  </Form.Field>
+                  <Form.Field>
+                    <label>{strings.LABEL_LOGIN_PASSWORD}</label>
+                    <input
+                      placeholder="Password"
```
nmadruga, line 103: And this one also please

src/containers/User/Login.tsx
nmadruga: Oh, actually can already use `{strings.LABEL_LOGIN_USERNAME}`

src/containers/User/Login.tsx
nmadruga: And `{strings.LABEL_LOGIN_PASSWORD}` here

src/containers/User/Login.tsx
```diff
@@ -1,66 +1,135 @@
...
+      const loginResult: LoginPayload = await attemptLogin({ username, password }, loginURL)
+
+      if (!loginResult.success) setIsError(true)
+      else {
+        setIsError(false)
+        setUser(loginResult.user)
+        setJWT(loginResult.JWT)
+        setTemplatePermissions(loginResult.templatePermissions)
+        setOrgList(loginResult?.orgList as OrganisationSimple[])
+      }
```
nmadruga, line 44: Consider moving this to a `attemptLogin` that has the `const loginURL = config.serverREST + '/login'` also. And uses the `fetchMethod` utility plus receive the `onSuccess` and `onFailure` methods to set up the same as what is done here in the if...else statement. Also, consider using a single state for setting (user, JWT, templatePermissions and orgList).

src/containers/User/Login.tsx
```diff
@@ -1,66 +1,135 @@
...
-    const loginResult = await attemptLogin(username, password)
-    if (!loginResult.success) setIsError(true)
-    else {
-      setIsError(false)
-      login(loginResult.JWT)
-      if (history.location?.state?.from) push(history.location.state.from)
-      else push('/')
+      if (verifyOrgResult.success) {
+        finishLogin(verifyOrgResult)
+      }
```
nmadruga, line 53: Similar thing as suggested before to create utility `attemptOrgLogin` that receives `onSuccess` and `onFailure` and also defines `const loginOrgURL = config.serverREST + '/login-org'` internally.

src/containers/User/UserRegister.tsx
```diff
@@ -1,26 +1,41 @@
...
+        const { JWT, user, templatePermissions } = loginResult
+        onLogin(JWT, user, templatePermissions)
+      })
+      .then(() => {
+        console.log('Trying to load new URL')
+        push('/application/new?type=UserRegistration')
+      })
+      .catch((err) => {
+        setNetworkError(err.message)
+      })
```
nmadruga, line 37: Suggested `attemptLogin` can be reused here.

nmadruga (changes requested):
Very good! I think the code looks great and is a very cool feature to have.
Tested and seems all good to me, just made some suggestions to try to move out some of the fetching methods to utility functions (that use the `fetchMethods` generic one - really like that) and a few strings missing from the `constants` file.

src/containers/Application/ApplicationSubmission.tsx
CarlosNZ: That'd be a pretty small component, wouldn't it? Shall we save it till we've got a few more things like this and do it more generally? 
Personally, I find this expression not too unwieldy -- it's pretty clear what it is IMHO.

src/containers/Review/ReviewPageWrapper.tsx
CarlosNZ: Yes, good catch -- this is what it does sometimes when you use the "Rename Symbol" feature in VSCode. Must have missed this one.

src/utils/helpers/fetchMethods.ts
CarlosNZ: > I'm just thinking it would be good to also have the other fetching (with the constant to define the route) in separated files. Which is now inside the component `login`.

Which component in `login` are you referring to?

CarlosNZ:
Thanks for the suggestions, @nmadruga . The attemptLogin method is working nicely as you suggested, and I was able to combine the four state variables into one `loginPayload` object.

See what you think ;)

src/containers/Application/ApplicationSubmission.tsx
nmadruga: No problem. We can leave this for [Epic#10](#78). I was thinking more like the whole page, with common title.

src/containers/User/Login.tsx
```diff
@@ -1,68 +1,142 @@
...
+        password,
+        onLoginSuccess,
+        onLoginFailure: () => setIsError(true),
+      }).catch((error) => {
+        setNetworkError(error.message)
+      })
+    } else {
+      // Organisation login
+      const { orgId } = loginPayload?.orgList?.[selectedOrgIndex] as OrganisationSimple
+      attemptLoginOrg({
```
nmadruga, line 50:
```suggestion
      await attemptLoginOrg({
```

I guess this one should also await?

nmadruga (approved):
Very nice! I'm happy with the result.

src/containers/User/Login.tsx
CarlosNZ: > I guess this one should also await?

Actually, neither need `await`, since there's nothing afterwards which relies on waiting sequentially for them. Have removed from both (and added `.catch` for earlier one).

Issue #207, opened 2020-12-11
List applications with specific cells on each column


nmadruga:
[Updated on 12/01/2021]
## Scope
- List applications based on the `type` and `user-role` (filters)
- Don't do any other filtering based on the URL filters yet (to be done in #251)
- Don't do any ordering of columns yet (to be done in #209) or pagination of list (#210)
- The main focus of this issue is to display a different list of applications based on the current user (user-role) with different cells per columns (as shown in the Designs).
- Some columns that display for all user-roles:
  - Instead of `Drug name` column will be replaced with `Application name` regardless of template type
  - On the `actionable` column an icon would be displayed for applications that required attention (after a review/consolidation with "Changes requested" was submitted) 

#### Applicant
- Replace `date` with `Last active date` to display date when application was last edited
- Combine the application stage and status on the `stage` column. This change should be described on the `stage` filter values #225.
- Show a link to open the application labeled with one of the Applications possible actions* on the column `actionable` for applications with statuses.
- Should show a bin icon on last column (no title) for applications with status as **Draft**
- Show current progress on column `status`. Maybe should be implemented as part of #236.

\* Application actions:
`edit-draft`: **Draft** application status
`make-updates`: **Changes-required** application status
`renew`: **Expired** or **Withdrawn** application statuses
 _empty_ **Submitted** or **Re-submitted** application statuses

![Screen Shot 2020-12-11 at 3 12 34 PM](https://user-images.githubusercontent.com/16461988/101852809-55dadb80-3bc3-11eb-9dee-f58bfa864f8a.png)

#### Reviewer
- Similar to Applicant column `date`, but display date when application were **Assigned** or **Edited** by Reviewer.
![Screen Shot 2020-12-11 at 3 17 45 PM](https://user-images.githubusercontent.com/16461988/101853229-211b5400-3bc4-11eb-940b-909286305a78.png)
- Display columns `Company` and `Username` of Applicant.
- Display username of who have assigned reviewers in the `Consolidator` column. 
- Display the applicant stage on the `stage` column. This may be changed to also add the `status` as for Applicant?
- Similar to Applicant, should show a link to open the review labeled with one of the Reviewer possible actions*.

\* Reviewer actions: 
`start-review`: One review assigned to you without a review associated with application + stage
`continue-review`: One review assigned to you and **In progress** review status
`review-updates`: **Changes Required** after consolidator have asked for changes OR Applicant has Re-submitted. 
_empty_ (Review submitted)

[...]

Issue #251, opened 2021-01-08
Filter application using URL parsed filters


CarlosNZ:
We need some specific options to use more than one filter for the same type. 
Base64 regex or specific our parsing for types in the URL (an earlier document with the schema should be ready before this issue can be started.

CarlosNZ:
So the idea I had earlier about using `null` values for the "ALL" variables is not going to work well enough -- it only seems to work with certain types of fields.

However, I've discovered that the `filter` object can just be a single variable, so we can dynamically build the filter object and pass that as a single variable, like so:
<img width="1423" alt="GraphQL filtering" src="https://user-images.githubusercontent.com/5456533/105015314-4e520080-5aa6-11eb-801f-d94652a71331.png">

🥳

CarlosNZ:
We didn't really nail down the "Search" query, as in: what fields does it search, and what kind of match?

How about, initially search matches:
- application name: search string anywhere in name
- stage, status: search string matches start
- applicant, reviewer, consolidator: search string matches start of First name, Last name, username

CarlosNZ:
Some other considerations I've come across when doing this. Some will probably require discussion, some will be separate issues:

- Should "applicant", "reviewer" and "consolidator" filters match username, firstName AND lastName fields? (I think Yes)
- Search filter (see above comment)
- Different roles will fetch different fields -- we should use @directives in getApplications query for this (new issue)
- We should "pre-filter" the queries so any unallowed ones (i.e. applicant trying to filter by reviewer) get ignored (new issue)

Issue #226, opened 2020-12-16
Update URL for lists to handle redirects and documentation


nmadruga:
Should update the filters handling on the list page and add more rules to the documentation:
Example: If user role is null - should update the URL

Use the user context to store the each application template type and your user role per template type. If you go to a specific application type url without the user role, we should change the url to use your first user role for that.

nmadruga:
Redirect to first type in the list of current user when they go to the `/applications` URL

PR #267, opened 2021-01-13
#131 Headers for application list

### LINKED ISSUE #131: Header columns for Application list
Currently the application list is just displaying all fields of the `application` table. This isn't great as things like `id` and `isLinear` are not things we'd want to show to user.

Suggestion: Define an object which maps field names to "display names" (e.g. "status" -> "Status"). If fields are omitted from this object (like "isLinear"), they don't get displayed.

Then we'd just need an additional `.filter` before the .`map` at line 30 and 38 of ApplicationList which only passes through fields which exist in the reference object, and amend the existing `.map` to pass the mapped name to the Header.

nmadruga:
Fixes #131 

~~Branch off #262. Please merge that one first before merging this one.~~ DONE

### Changes
- Add conversion for snake-case keys in the URL to camelCase (to be used as variables) in `useRouter`
- Update `ApplicationList` to first check for filter keys `type` and `user-role` in URL. 
  - If `type` not defined is not making the query for application (To be improved)
  - If `type` is defined but no `user-role` will get the permissions from `UserState` related to this `type`.
- New utility method `findUserRole` to map user-roles ('applicant', 'reviewer', 'consolidator') based on group of permissions
- New utility method to list all column headers to be displayed by `user-role`
- Update `ApplicationList` to display headers based on user-type. Using now a `switch...case` to find application properties to display in each cell. (To be properly implemented in #207). Where each cell should have a proper `render` function defined in the `ColumnDetails` object which is also based on the User-role and selected properties (from GraphQL query)

### Unrelated changes:
- Create folder `application` inside helpers to keep helpers related to an application
- Change helpers extension to `.ts` instead of `.tsx` when not used for Rendering

### Testing
For now most users have only user-role `applicant` available for the "Company Registration". So log in with some applicant user and click the header for "Company Registration" to see the list.


src/utils/helpers/translations/mapColumnsByRole.ts
```diff
@@ -0,0 +1,106 @@
...
+    APPLICATION_COLUMN.LAST_ACTIVE_DATE,
+    APPLICATION_COLUMN.APPLICATION_NAME,
+    APPLICATION_COLUMN.APPLICANT,
+    APPLICATION_COLUMN.STATUS,
+  ],
+}
+
+export default (userRole: string): Array<ColumnDetails> => {
+  const columns: Array<ColumnDetails> = allColumns.filter(({ headerName }) =>
+    columnsPerRole[userRole] ? columnsPerRole[userRole].includes(headerName) : false
```
CarlosNZ, line 103: I think a much more efficient mapping would be:
```
const columns: Array<ColumnDetails> = columnsPerRole[userRole].map((key) => allColumns[key] )
```

And then instead of `allColumns` being an array, it should be an object with the unique name as the key, i.e.:
```
const allColumns = {
  LAST_ACTIVE_DATE: {
    headerName: APPLICATION_COLUMN.LAST_ACTIVE_DATE,
    filters: 'last-active-date',
  },
  DEADLINE_DATE: {
    headerName: APPLICATION_COLUMN.DEADLINE_DATE,
    filters: 'deadline-date',
  },
  APPLICATION_NAME: {
    headerName: APPLICATION_COLUMN.APPLICATION_NAME,
    filters: 'search',
  },
  APPLICANT: {
    headerName: APPLICATION_COLUMN.APPLICANT,
    filters: 'applicant',
  },
  COMPANY: {
    headerName: APPLICATION_COLUMN.COMPANY,
    filters: 'org',
  },
  CONSOLIDATOR: {
    headerName: APPLICATION_COLUMN.CONSOLIDATOR,
    filters: 'consolidator',
  },
  STAGE: {
    headerName: APPLICATION_COLUMN.STAGE,
    filters: ['search', 'stage'],
  },
  STATUS: {
    headerName: APPLICATION_COLUMN.STATUS,
    filters: 'search',
  },
  ACTIONABLE: {
    headerName: '',
    filters: ['search', 'action'],
  },
}
```

And then the columnsPerRow would just list the keys for each role type, e.g.:
```
applicant: [
    LAST_ACTIVE_DATE,
    DEADLINE_DATE,
    APPLICATION_NAME,
    STAGE,
    STATUS,
    ACTIONABLE,
  ]
```

It'd still return an array of column objects, but it's a linear operation to build it.

src/utils/hooks/useListApplications.tsx
```diff
@@ -6,13 +6,19 @@ import {
...
+
+const useListApplication = ({ type }: UseListApplicationsProps) => {
   const [applications, setApplications] = useState<ApplicationDetails[] | undefined>()
   const [loading, setLoading] = useState(true)
   const [error, setError] = useState('')
 
   const { data, loading: apolloLoading, error: apolloError } = useGetApplicationsQuery({
+    variables: { code: type as string },
     fetchPolicy: 'network-only',
+    skip: !type,
```
CarlosNZ, line 21: I assume we have an issue to set a default type based on user permissions, right? (So we won't need to skip)

src/utils/hooks/useRouter.tsx
```diff
@@ -34,6 +35,17 @@ interface RouterResult {
   location: any
 }
 
+const replaceHifenKeys = (parsedQuery: { [key: string]: any }) => {
```
CarlosNZ, line 38:
```suggestion
const replaceSnakeCaseKeys = (parsedQuery: { [key: string]: any }) => {
```
I assume you meant "hyphen", right? I'd prefer `replaceSnakeCaseKeys`, but if not it'd be `replaceHyphenKeys` ;)

src/utils/hooks/useRouter.tsx
```diff
@@ -34,6 +35,17 @@ interface RouterResult {
   location: any
 }
```
CarlosNZ, line 37: Can you add a comment explaining what it does, its parameters and its return shape?

src/utils/hooks/useRouter.tsx
CarlosNZ: Also, this is the kind of function that would really suit some unit tests. I don't think we have Jest installed on front-end yet, but maybe make an issue to configure Jest and add this as one of the functions we should write tests for.

src/utils/hooks/useRouter.tsx
```diff
@@ -34,6 +35,17 @@ interface RouterResult {
...
 
+const replaceHifenKeys = (parsedQuery: { [key: string]: any }) => {
+  if (Object.keys(parsedQuery).length === 0) return parsedQuery
+  const replacedHifenKeys = Object.keys(parsedQuery).map((key) => {
+    const convertedKey = key.replace(/-([a-z])/g, function (m, w) {
```
CarlosNZ, line 41: Could you give `m` and `w` more descriptive names please?

src/utils/hooks/useRouter.tsx
```diff
@@ -34,6 +35,17 @@ interface RouterResult {
...
+      return w.toUpperCase()
+    })
+    return { [convertedKey]: parsedQuery[key] }
+  })
+  return replacedHifenKeys.reduce((a, b) => Object.assign({}, a, b))
```
CarlosNZ, line 46: Might be nicer to use `Object.fromEntries()` instead of `reduce`?
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/fromEntries

src/utils/hooks/useRouter.tsx
CarlosNZ, line 38: Also -- we're leaving mapping the values to another PR? (Presumably because the delimiters and date parsing will be more complex?)

src/utils/hooks/useRouter.tsx
```diff
@@ -44,6 +56,9 @@ export function useRouter(): RouterResult {
   // Memoize so that a new object is only returned if something changes
 
   return useMemo(() => {
```
CarlosNZ, line 58: Nice!

CarlosNZ (changes requested):
Good start for this Epic! Thanks :)

src/utils/helpers/translations/mapColumnsByRole.ts
```diff
@@ -0,0 +1,106 @@
...
+    headerName: APPLICATION_COLUMN.COMPANY,
+    filters: 'org',
+  },
+  {
+    headerName: APPLICATION_COLUMN.CONSOLIDATOR,
+    filters: 'consolidator',
+  },
+  {
+    headerName: APPLICATION_COLUMN.STAGE,
+    filters: ['search', 'stage'],
```
CarlosNZ, line 46: Since this is an array of filters, wouldn't it be simpler to make them all arrays, even if only one item? Saves having to do an additional type check when interpreting them later on, no?

src/containers/Application/ApplicationList.tsx
```diff
@@ -1,14 +1,41 @@
...
+        const newRole = findUserRole(permissions)
+        // TODO: Call helper to build similar URL query with the new userRole
+        if (newRole) push(`/applications?type=${type}&user-role=${newRole}`)
+      }
+    }
+    if (type && userRole) {
+      const columns = mapColumnsByRole(userRole)
+      setHeaders(columns.map(({ headerName }) => headerName))
+    }
+  }, [type, userRole])
```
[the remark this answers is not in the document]
nmadruga: Hmmm I'm a big confused with `....&type="applicant"` did you mean `....&user-role="applicant"`?

Thanks for spotting this, I'll have a look now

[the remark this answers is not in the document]
nmadruga: Done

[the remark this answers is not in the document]
CarlosNZ: > Hmmm I'm a big confused with `....&type="applicant"` did you mean `....&user-role="applicant"`?

Ah yes, I did. 🤦‍♂️

src/containers/Application/ApplicationList.tsx
```diff
@@ -23,29 +50,49 @@ const ApplicationList: React.FC = () => {
           <List.Item key={`app_selected_parameter_${value}`} content={key + ' : ' + value} />
         ))}
       </List>
-      {applications && applications.length > 0 && (
+      {userRole && headers && applications && applications.length > 0 && (
         // TODO: Implement sortable columns -> Requires reducer
```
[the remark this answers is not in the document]
nmadruga: > Sorting should be done in the GraphQL right? 

Yes. But the sorted columns would be defined be what headers were clicked in the UI - which should be store in states (not reducers sorry). I'll update this comment.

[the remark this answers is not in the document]
nmadruga: > Clicking a "sort" column should just update the query, just like changing a filter, or pagination settings. (so nothing fancy needs to be done to sort in here I would think)

Yes. That's right. We would update the URL and that would be used in the query. So, not even a state (and my previous comment).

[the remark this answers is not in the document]
nmadruga: > Because the sorting should take into account the entire result set, not just the ones currently returned for the current page, right?

Yes, the comment was old. I'll update now trying to summary was was commented here.

src/containers/Application/ApplicationList.tsx
```diff
@@ -23,29 +50,49 @@ const ApplicationList: React.FC = () => {
...
-                {Object.entries(application).map(([key, value]) => {
-                  switch (key) {
-                    case 'name':
+                {headers.map((headerName) => {
+                  // TODO: Move this to render property to be defined for each ColumnDetails in mapColumnsByRole
+                  switch (headerName) {
+                    case APPLICATION_COLUMN.LAST_ACTIVE_DATE:
+                    case APPLICATION_COLUMN.DEADLINE_DATE:
+                      return (
+                        <Table.Cell>
```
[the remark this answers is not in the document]
nmadruga: Thanks. Fixed now: [1c96990](https://github.com/openmsupply/application-manager-web-app/pull/267/commits/1c96990927f90ca5bbefc9060d3a70976f0766c7)

src/containers/Application/ApplicationList.tsx
```diff
@@ -23,29 +50,49 @@ const ApplicationList: React.FC = () => {
...
           </Table.Header>
           <Table.Body>
             {applications.map((application, index) => (
               <Table.Row key={`${application.serial}`}>
-                {Object.entries(application).map(([key, value]) => {
-                  switch (key) {
-                    case 'name':
+                {headers.map((headerName) => {
+                  // TODO: Move this to render property to be defined for each ColumnDetails in mapColumnsByRole
+                  switch (headerName) {
```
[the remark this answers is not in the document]
nmadruga: yes, that's right. I haven't thought yet how exactly this would be done yet - I'm planning to update #207 with the approach before I start the implementation. But briefly this would be replaced to just display the component returned by the current `header` which would expect the current application to be passed and return how this cell would be displayed.

src/utils/graphql/queries/getApplications.query.ts
```diff
@@ -1,8 +1,10 @@
 import { gql } from '@apollo/client'
 
 export default gql`
-  query getApplications {
-    applications {
+  query getApplications($code: String!) {
+    applications(
+      filter: { template: { code: { equalTo: $code }, and: { status: { equalTo: AVAILABLE } } } }
```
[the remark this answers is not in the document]
nmadruga: I don't think that this is dependent on if the filter is passed or not. We would need to query GraphQL To get all possible `labels` for certain columns. Example: Action, Stage, Status. This would be available for filter selection at some point, and also for checking for valid filter values (I guess)

src/utils/helpers/translations/findUserRole.ts
```diff
@@ -0,0 +1,26 @@
+import { PermissionPolicyType } from '../../generated/graphql'
```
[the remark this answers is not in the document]
nmadruga: No problem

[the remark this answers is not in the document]
nmadruga: Done: [41027ec](https://github.com/openmsupply/application-manager-web-app/pull/267/commits/41027ec85bc7074525af0458dc790516f7a2ebe0)

src/utils/helpers/translations/mapColumnsByRole.ts
```diff
@@ -0,0 +1,106 @@
+export enum APPLICATION_COLUMN {
```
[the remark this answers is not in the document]
nmadruga: Done: [881b31d](https://github.com/openmsupply/application-manager-web-app/pull/267/commits/881b31dd29d3f55558a4fb36e0310149e64f2780)

src/utils/graphql/queries/getApplications.query.ts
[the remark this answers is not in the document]
CarlosNZ: What I mean is, say for a query like:
```
query MyQuery($status: ApplicationStatus) {
  applications(filter: { status: { equalTo: $status } }) {
    nodes {
      name
      isActive
      status
      stage
    }
  }
}

```

If we don't supply a value for `$status`, what happens? What we _want_ to happen is that it would ignore that filter and return ALL applications, but it seems like GraphQL throws an error. So we'll need to figure out how to solve this problem. I guess the whole "filter" string would need be dynamically created and inserted into the GraphQL string, right?

src/utils/helpers/translations/mapColumnsByRole.ts
nmadruga: Yea, that doesn't work exactly like that. I'm still trying to figure out how to create a type to list the possible columns, which is how I think this could work, just saying each constant of a user role is an array of that type. But I still need a little more research to make this one work.

src/utils/hooks/useListApplications.tsx
nmadruga: We don't have a specific one for that. We just have the #251 that would deal with that, that's what I was thinking. If you feel we should have a issue just to deal with this problem, let me know.

src/utils/hooks/useRouter.tsx
nmadruga: This is just a wrapper for all the default available properties when using these:
```
  const params = useParams()
  const location = useLocation()
  const history = useHistory()
  const match = useRouteMatch()
```

I don't see why we need to document this one, which can be quite extensive. It is simply wrapping around used the Router individual hooks.

src/utils/hooks/useRouter.tsx
nmadruga: Carl, I don't think this needs to be specified since it's just a utility minor function. I'm not entirely sure on what are the 2 properties there, I think it's ok to leave it like that.

src/utils/hooks/useRouter.tsx
nmadruga: Not sure what was the idea, can you try out and show how do you think this can be replace please?

src/utils/hooks/useRouter.tsx
nmadruga: Yes

src/utils/helpers/translations/mapColumnsByRole.ts
nmadruga: Sure. I'll review this while I'm thinking about the next issue #207

src/utils/hooks/useRouter.tsx
CarlosNZ: I meant just for the hyphenKeys function, not the whole hook.

src/utils/graphql/queries/getApplications.query.ts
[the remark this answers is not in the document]
nmadruga: Oh, I see. I wasn't getting this idea. You can pass optional parameters to the GraphQL query, so in case the parameter isn't defined you won't add to the query. I'm not sure how we would do that with different filters. I'll see if I can come up with a solution for that - anyway it's not included in this issue.

src/utils/helpers/translations/mapColumnsByRole.ts
nmadruga: Ok. I made quite a big change now. I'm not so happy with the way we can have types using enums, but I guess that's how things are. Please have another look and see what do you think. Mind that there is a new folder called `model` where I tried to keep this constants that are being used as types, but aren't types. So I guess models?

src/utils/hooks/useRouter.tsx
nmadruga: Oh! Ok. Is just that you commented is flagging just a empty line in this file

src/utils/graphql/queries/getApplications.query.ts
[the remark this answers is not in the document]
CarlosNZ: I think I've figured out how this can work.

We need to enable `connectionFilterAllowNullInput` in the advanced filtering plugin (see [here](https://github.com/graphile-contrib/postgraphile-plugin-connection-filter#connectionfilterallownullinput))

Then we can just set the value of the variable to `null` and it will ignore that filter. Nice ;)

src/utils/hooks/useRouter.tsx
CarlosNZ: `Object.fromEntries()` turns an array of pairs of values (in sub-array) into and key-value pairs in a new object.
The 4 lines above would be:
```
    })
    return [ [convertedKey], parsedQuery[key] ] // Put in array instead of object here
  })
  return Object.fromEntries(replacedKeys)
```

Personally I find it a bit nicer to read, but not a big deal. Up to you.

src/utils/hooks/useRouter.tsx
CarlosNZ: Actually, I put you wrong: apparently hyphens is `kebab-case`. `snake_case` is underscores. 🤦‍♂️ 😆

nmadruga:
I'll continue this review first thing tomorrow

CarlosNZ:
Seems to be a bug: when I click the "Company Registrations" link at the top, the URL it tries to load is:
```http://localhost:3000/applications?type=CompRego1&user-role=0```

I can manually type `http://localhost:3000/applications?type=CompRego1&user-role=applicant` and it loads okay.

**Comment by nmadruga** [comment]
> Seems to be a bug: when I click the "Company Registrations" link at the top, the URL it tries to load is:
http://localhost:3000/applications?type=CompRego1&user-role=0

> I can manually type http://localhost:3000/applications?type=CompRego1&user-role=applicant and it loads okay.

I saw this bug happening yesterday. I have fixed on #207 but not here. Sorry, I'll add the same thing here to fix that.

File: src/utils/helpers/translations/findUserRole.ts [hunk 16]
```diff
@@ -0,0 +1,32 @@
...
+  // Compare array of permission checking if are the same
+  const matching = Object.entries(userRoles).filter(([role, permissionList]) => {
+    console.log('role', role)
+
+    const difference = permissionList.filter((x) => !comparePermissions.includes(x))
+    return difference.length === 0
+  })
+  console.log('Possible user-roles:', Object.keys(matching))
+  return matching.length > 0 ? Object.keys(matching)[0] : undefined
+}
```
CarlosNZ: This is the method that is returning `0` instead of `applicant`.

It seems overly confusing to me, I think it could be simpler. This seems to work (although probably needs some additional checks:
```
export default (permissions: Array<PermissionPolicyType>): string | undefined => {
  const comparePermissions = permissions.map((permissionType) => permissionType.toUpperCase())

  // Compare array of permission checking if are the same
  const matching = Object.entries(userRoles).filter(([role, permissionList]) => {
    const common = permissionList.filter((permission) => comparePermissions.includes(permission))
    return common.length > 0
  })
  return matching?.[0]?.[0] || undefined
```

src/utils/graphql/queries/getApplications.query.ts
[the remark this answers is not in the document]
nmadruga: Cool! As agreed this would be included as part of issue #226.

CarlosNZ (commented):
Cool, looks good (apart one from that one bug)

src/utils/helpers/translations/findUserRole.ts
CarlosNZ: Ah, I see your other comment about fixing this elsewhere. Cool. I do think the method could be simpler, though, but I guess I'll take another look at it over on that PR.

src/utils/hooks/useRouter.tsx
nmadruga: I can change to that, no problem

src/containers/Application/ApplicationList.tsx
```diff
@@ -1,14 +1,42 @@
...
+
+  useEffect(() => {
+    if (type && templatePermissions) {
+      if (!userRole) {
+        const found = Object.entries(templatePermissions).find(([template]) => template === type)
+        if (found) {
+          const [template, permissions] = found
+          const newRole = findUserRole(permissions)
+          // TODO: Call helper to build similar URL query with the new userRole
+          if (newRole) push(`/applications?type=${type}&user-role=${newRole}`)
```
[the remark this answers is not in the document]
nmadruga: Yep, that was the problem. Already fixed now.

src/utils/hooks/useRouter.tsx
nmadruga: Done (8da61d1)[https://github.com/openmsupply/application-manager-web-app/pull/267/commits/8da61d14f8b8801d1fa511df2f5ae31237db0d0d]

src/containers/Application/ApplicationList.tsx
```diff
@@ -20,32 +48,59 @@ const ApplicationList: React.FC = () => {
       {Object.keys(query).length > 0 && <h3>Query parameters:</h3>}
       <List>
         {Object.entries(query).map(([key, value]) => (
-          <List.Item key={`app_selected_parameter_${value}`} content={key + ' : ' + value} />
+          <List.Item key={`ApplicationList-parameter-${value}`} content={key + ' : ' + value} />
         ))}
       </List>
-      {applications && applications.length > 0 && (
-        // TODO: Implement sortable columns -> Requires reducer
+      {userRole && headers && applications && applications.length > 0 && (
```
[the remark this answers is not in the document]
nmadruga: Cool suggestion. I'll add this now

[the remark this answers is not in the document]
nmadruga: Done d20d9ba

src/utils/helpers/translations/findUserRole.ts
nmadruga: Done b5dc196

src/utils/hooks/useRouter.tsx
```diff
@@ -34,6 +28,22 @@ interface RouterResult {
...
 }
 
+/**
+ * @function: replaceSnakeCaseKeys
+ * For each filter key in the query (which will use snake-case: e.g. user-role)
+ * will convert to camelCase (e.g userRole) before returning query to components.
+ * - @param parsedQuery - URL query after parsed to object
+ * - @returns Object with all query {key: value} with keys in camelCase format.
+ */
+const replaceSnakeCaseKeys = (parsedQuery: { [key: string]: any }) => {
```
CarlosNZ, line 38: Change to `replaceKebabCaseKeys` remember? ;)

CarlosNZ (approved):
Just need to rename "snakeCase" to "kebabCase" function, otherwise good to go :) 

Also, need to fix conflict in types file (I used "Keep both" when merging master locally).

PR #268, opened 2021-01-18
#207 Applications list with specific cells

### LINKED ISSUE #207: List applications with specific cells on each column

nmadruga:
Fixes #207 

Branch of #267, please merge that one first.

### Changes
- Add new Context `ListState` to keep state related to the current list - for now just `reset` and `setApplications` list
- Separate query for `ApplicationStageStatusAll` from `getApplications`, to be called in sequence. The new query received a list of applications `serialNumber` to be filtered, so it won't retrieve all applications stages and status as before.
- Create many components in `components/List` used in each application cell of the list... The behaviour of each cell is dependent on the user-role viewing the application. E.g: Applicant sees the status related to the application and Consolidator/Reviewer will see the status related to Reviews (**Not implemented yet**).
- New folder `utils/data` to store some static enum to be used while constructing the list: USER_ROLES,  APPLICATION_COLUMNS, COLUMNS_PER_ROLE.
- Update `ApplicationList` to use the `ListState`, show the columns and cells dynamically (as user-role and type defined). And add new **Button** to link to the `ApplicationStart` page.

### Unrelated changes:
- Update `useLoadApplication` to use a similar approach on loading the stage and status of the current application.

### Existing problems:
- When changing the list type, it re-renders a couple of times. So the page is flickering. Need some help with that.


src/components/List/ApplicantCell.tsx
```diff
@@ -0,0 +1,8 @@
+import React from 'react'
```
CarlosNZ, line 1: Just a personal preference, but since these individual cell components are so small I feel it would be easier to have them in one file. Like a library of components in list like in `mapColumnsByRole.`
If there's any complex logic required for any of them, that could be extracted to another file (or at the bottom of this one). What do you think?

src/components/List/StatusCell.tsx
```diff
@@ -0,0 +1,49 @@
...
+enum ACTIONS {
+  EDIT_DRAFT = 'Edit draft',
+  MAKE_CHANGES = 'Make changes',
+  RENEW = 'Renew',
+}
+
+const StatusCell: React.FC<CellProps> = ({ application }) => {
+  const { serial, stage } = application
+  const status = stage?.status
+  switch (status) {
```
CarlosNZ, line 16: I don't seem to be getting the correct `status` here. Made a new CompRego application, status is showing as draft in database. The application shows in the list but the Status field is blank.
Console output for serial, stage, status:
```
Status 8259 
Object { id: 2, name: "Screening", status: undefined, date: "2021-01-19" } undefined
```

src/components/List/StatusCell.tsx
```diff
@@ -0,0 +1,49 @@
+import React from 'react'
+import { Link } from 'react-router-dom'
+import { Grid, Icon, Label, Progress } from 'semantic-ui-react'
+import { ApplicationStatus } from '../../utils/generated/graphql'
+import { ApplicationStage, CellProps } from '../../utils/types'
```
CarlosNZ, line 5: ApplicationStage unused

src/containers/Application/ApplicationList.tsx
```diff
@@ -1,15 +1,26 @@
 import React, { useEffect, useState } from 'react'
-import { Container, Table, List, Label, Header, Progress, Message } from 'semantic-ui-react'
+import {
+  Container,
+  Table,
+  List,
+  Label,
+  Header,
+  Progress,
```
CarlosNZ, line 8: Header and Progress currently unused

src/contexts/ListState.tsx
```diff
@@ -0,0 +1,55 @@
+import React, { createContext, useContext, useReducer } from 'react'
```
CarlosNZ, line 1: Just a question -- what is the reason for having a List Context state? I was under the impression that the "List" data would only ever be needed on the List page/component, but Context makes it available elsewhere. What is the advantage of having it as opposed to  just having a local state object in "ApplicationList"?

src/utils/hooks/useListApplications.tsx
```diff
@@ -1,61 +1,108 @@
...
+  setListState: Dispatch<ListActions>
 }
 
-const useListApplication = ({ type }: UseListApplicationsProps) => {
-  const [applications, setApplications] = useState<ApplicationDetails[] | undefined>()
+interface ApplicationDetailsMap {
+  [serial: string]: ApplicationDetails
+}
+
+const useListApplication = ({ type, setListState }: UseListApplicationsProps) => {
```
CarlosNZ, line 20: Probably want to name it `useListApplications` -- with an `s`, like the filename.

src/utils/hooks/useLoadApplication.tsx
```diff
@@ -24,20 +28,31 @@ const useLoadApplication = (props: UseGetApplicationProps) => {
   })
 
   useEffect(() => {
-    if (data && data.applicationBySerial && data.applicationStageStatusAlls) {
+    if (data?.applicationBySerial) {
       const application = data.applicationBySerial as Application
 
-      const applicationDetails = {
+      let applicationDetails: ApplicationDetails = {
```
CarlosNZ, line 34: Why change to `let` here?

src/utils/helpers/translations/mapColumnsByRole.ts
```diff
@@ -11,52 +25,60 @@ import COLUMNS_PER_ROLE from '../../model/columnsPerUserRole'
...
-  headerName: string
-  filters: Array<string>
-  render?: Function
-}
-
-const allColumns = {
+const allColumns: { [key in APPLICATION_COLUMNS]: ColumnDetails } = {
+  SERIAL_NUMBER: {
+    headerName: 'Serial number',
+    filters: ['serial'],
```
CarlosNZ, line 31: I'm wondering if these filters need to be defined here. We're going to have to have a mapping method to convert URL query filters to actual GraphQL filters, so they'll all need to be defined elsewhere anyway, and I don't think it matter which columns they go with. What was your thinking for having this `filters` field?

CarlosNZ (changes requested):
Cool -- I like the general way this is going. Main problem seems to be that the `status` is not being fetched, and looking through the code I can't see where it would be.

Also, in `getApplications.query`, shouldn't we get rid of the `applicationStageStatusAlls` node as we discussed yesterday, otherwise its going to fetch the entire list of stage/statuses regardless of the applications filters.

src/components/List/ApplicantCell.tsx
nmadruga: I disagree. Sorry, I really think it's easier to see things when they are separated in its own file, no matter how big they are. It's not like the bundle size increases, and if we keep thing inside folder to me makes it compact when you are not working with that area of components.

src/utils/helpers/translations/mapColumnsByRole.ts
nmadruga: I thought that returning which ones are the filters (for each column) in the array of columns
https://github.com/openmsupply/application-manager-web-app/blob/48c8899fe539e62ad8392c552f06784b0aa9a16c/src/utils/helpers/translations/mapColumnsByRole.ts#L81-L83
would be useful to map something that should be displayed in the Cell. For example, using the `search` filter would query for a new list of applications and highlight the **start of words** considered in the query, for specific columns.
I'm happy to remove from there if you think it's not helping if the filtering.

src/utils/hooks/useLoadApplication.tsx
nmadruga: To update the stage (if found) after:
https://github.com/openmsupply/application-manager-web-app/blob/48c8899fe539e62ad8392c552f06784b0aa9a16c/src/utils/hooks/useLoadApplication.tsx#L48-L53

src/utils/hooks/useListApplications.tsx
nmadruga: Yes. Definitely. I'm changing it now.

nmadruga: Done 273e77d

src/contexts/ListState.tsx
nmadruga: As explained in person: When changing from one type to another the useListApplications hook was being called many times and the query for the previous type was re-running. So I decided to `reset` the list before querying for another one, and used the Context. I can investigate further why this was happening and try not to use the Context, I thought it would be useful to store other filters, but if you think we can simply use the local state and URL we can try that. By the way the Context was only available for the ApplicationsList, since that was the only Container with the Context provider wrapping it.

src/containers/Application/ApplicationList.tsx
nmadruga: Thanks! Changing that now

src/components/List/StatusCell.tsx
nmadruga: And this one too

src/components/List/StatusCell.tsx
nmadruga: I will investigate this (and try to remove the Context) will get back to you when finished.

src/utils/helpers/translations/mapColumnsByRole.ts
CarlosNZ: I think this will be taken care of with what I'm doing with the filtering handling. But maybe just leave it for now and I'll remove them in my PR once I'm sure they're not needed.

src/components/List/StatusCell.tsx
CarlosNZ: Any joy with this? (I've noticed a few things when doing the filtering, if you wanted to discuss it first?)

src/contexts/ListState.tsx
CarlosNZ: Got it. I think it might be worth re-visiting this once we have all the filtering and everything else working, but all good for now.

src/utils/hooks/useLoadApplication.tsx
CarlosNZ: But it's fine to mutate a property of a `const`, it only needs `let` if you're reassigning the reference to the object itself.

src/components/List/StatusCell.tsx
nmadruga: Have solved this problem now. And removed the use of `ListContext`

nmadruga:
@CarlosNZ Ready for re-review. Unfortunately that problem that I showed you with the useRouter being called many times and making the list flicker is still happening. But I'm a bit clueless on how to solve that one for now. I'll create a separated issue.

CarlosNZ:
> @CarlosNZ Ready for re-review. Unfortunately that problem that I showed you with the useRouter being called many times and making the list flicker is still happening. But I'm a bit clueless on how to solve that one for now. I'll create a separated issue.

Need to merge latest master in here. Currently incompatible with the current back-end master (login stuff)

CarlosNZ (approved):
Cool -- cleaner without ListContext I think.

Any additional changes can either be dealt with in what I'm working on now, or in #283.

Cheers :)

src/utils/hooks/useLoadApplication.tsx
nmadruga: Oops! That's a good point 😆 I'll change to const : )

Issue #89, opened 2020-11-04
Element plugin: Radio Buttons


nmadruga:

CarlosNZ:
Have added Checkbox to this issue too, as "Radio" is basically a subset of Checkboxes. They'll be two different plugins, but with basically the same configuration/functionality.

PR #271, opened 2021-01-18
#89 Radio button plugin

CarlosNZ:
Implements #89 -- radio button plugin. Requires branch from back-end [PR#182](https://github.com/openmsupply/application-manager-server/pull/182)

Presents a choice list to user in the form of a radio button group. Only one can be selected. Default selection can be defined in template.

Also updated docs for this plugin and other plugins

Demonstrated on Page 2 & 3 of **Feature Showcase** template -- one has a default, one doesn't.

### Additional changes (not directly connected to this issue):
- Added a few "quick links" to "Home" page, for getting to the list and specific template start pages quickly. These are just for development ease.
- A new method `getDefaultIndex` in ApplicationViewWrapper which is passed into all plugins. Allows "default" to be specified as either an index or a value -- this method will return the index in either case. (We might want to have a default "value" if querying an API, say, and know some of the items in the list, but not what position to expect them)
- Added the above method to "dropdownChoice" plugin
- Markdown interpretation for labels and descriptions on all plugins (as per the docs). Can see an example of Markdown in description in Q6 (radio button selector on Page 2)


documentation/internal/Element-Type-Specs.md
```diff
@@ -24,7 +24,9 @@ _Ongoing authoritative reference of Template Question/Element types, including i
 
 **Note**: all parameter fields can also have a dynamic query object instead of a primitive. The [`evaluateExpression`](https://github.com/openmsupply/application-manager-server/wiki/Query-Syntax) function will return literal strings (or numbers, booleans) as is. The types described for the parameters below are the type that is expected to be _returned_ from a query expression.
 
-### Short Text Input
+\* parameters marked with \* can be defined as dynamic expressions -- these are specified in the `pluginConfig.json` file of each plugin.
+
+### Short Text Input _(not yet implemented -- highest priority)_
```
nmadruga, line 29: Maybe this was a mistake on the conflict resolution - I think the `Short Text Input` is already done. Isn't it?

src/components/Home.tsx
```diff
@@ -9,6 +10,18 @@ const Home: React.FC = () => {
...
+      <Header as="h2">Quick Links (for Dev)</Header>
+      <List>
+        <List.Item>
+          <Link to="/applications">Applications list</Link>
+        </List.Item>
+        <List.Item>
+          <Link to="/application/new?type=TestRego">Feature showcase application</Link>
+        </List.Item>
+        <List.Item>
+          <Link to="/application/new?type=TestRego">User registration application</Link>
```
nmadruga, line 22:
```suggestion
          <Link to="/application/new?type=UserRegistration">User registration application</Link>
```

src/formElementPlugins/ApplicationViewWrapper.tsx
```diff
@@ -165,6 +166,18 @@ const ApplicationViewWrapper: React.FC<ApplicationViewWrapperProps> = (props) =>
...
+Allows the default value in template to be either an index or string
+value. Number is assumed to be index, else it returns the index of the 
+specified value in the options array. Functions is passed as prop to
+element plug-ins so can be used by any plugin.
+*/
+const getDefaultIndex = (defaultOption: string | number, options: string[]) => {
+  if (typeof defaultOption === 'number') {
+    return defaultOption
+  } else return options.indexOf(defaultOption)
+}
```
nmadruga, line 179: I think this function would throw an error if you (by mistake on the template data) send a index out-of-range in the `options` array. There should be a safe guard for that.

nmadruga (commented):
Very good. It's good to have the docs updated at the same time. Thanks!

Just a few comments to have a look please. I'll be testing this now.

src/formElementPlugins/ApplicationViewWrapper.tsx
CarlosNZ: I just checked what happens, and if the returned index number is outside the range, it just has nothing selected, and no run-time error. (The optional chaining operator in `setValue(options?.[getDefaultIndex(defaultOption, options)])` ensures it doesn't try to set a non-existent value.

So having an illegal default value (index or name) behaves the same as having no default value. This seems like the appropriate behaviour don't you think?

nmadruga: Yes. That's good

nmadruga (approved):
Good to go (just one point in the Back-end)

Issue #176, opened 2020-12-06
Show error 404 page on application and review pages


nmadruga:
When making the call for `useLoadApplications` should check if user has access to a certain application.
The load trigger shouldn't run if the application is not found. For now it's just running until it finds the trigger.

Test with non-existing application or user without permission to see applications of type

nmadruga:
Also do for Consolidator

nmadruga:
Create a fallback component - like the one in the Router - and display as 404 page

PR #286, opened 2021-01-20
Remove need for extra hook getApplicationStatus

nmadruga:
When trying to load the Review page on `master` it isn't working after a few changes on `status`:
- Fixed that and removed the whole `getApplicationStatus` hook which can be done easily in the same query in `getApplication`, and just be adding a new state `appStages` to `useLoadApplication` and returning it as part of the results.
- Changed all containers that were using this hook to simply use `useLoadApplication` to get:
  - `status` from `application.stage.status`
  - `stages` (to display all stages on the template) from `appStages`


src/utils/hooks/useGetApplicationStatus.tsx
```diff
@@ -1,65 +0,0 @@
...
-    serialNumber,
-    // triggerType: 'applicationTrigger',
-  })
-
-  const { data, loading, error } = useGetApplicationStatusQuery({
-    variables: {
-      serial: serialNumber,
-    },
-    skip: triggerProcessing || !isApplicationLoaded,
-    fetchPolicy: networkFetch ? 'no-cache' : 'cache-first',
```
CarlosNZ, line 29: I think this is causing the problem I mentioned above. The status needs to ensure it gets the fresh value from the network immediately after submission, hence the dynamic fetchPolicy here. You'll need to do something similar again to the `useGetApplicationQuery` in `useLoadApplication` now that you've consolidated them into one Hook.

CarlosNZ (changes requested):
As per my comments -- one regression due to Submission not getting fresh Status info from Server.

Thanks for this PR though, nice to simplify things.

src/containers/Application/ApplicationOverview.tsx
```diff
@@ -137,13 +133,13 @@ const ApplicationOverview: React.FC = () => {
     }
   }
 
-  return error || statusError ? (
-    <Message error header={strings.ERROR_APPLICATION_OVERVIEW} list={[error, statusError]} />
-  ) : loading || statusLoading || responsesLoading ? (
+  return error ? (
```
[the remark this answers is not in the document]
nmadruga: Sure thing

src/utils/hooks/useGetApplicationStatus.tsx
nmadruga: Very nice! I wasn't sure what was causing this either. Will fix now 👍

src/containers/Application/ApplicationPageWrapper.tsx
```diff
@@ -61,16 +61,12 @@ const ApplicationPageWrapper: React.FC = () => {
...
-    appStages,
-    templateSections,
-    isApplicationLoaded,
-  } = useLoadApplication({
-    serialNumber: serialNumber as string,
-  })
+  const { error, loading, application, templateSections, isApplicationLoaded } = useLoadApplication(
+    {
+      serialNumber: serialNumber as string,
+      networkFetch: true,
```
CarlosNZ, line 67: I don't think you need `networkFetch: true` for this one. The idea was that we only need to force a network refresh when the application is likely to change in the back-end between fetches. Only when submitting the application is really an issue, the rest of the time the cached version should be fine. Maybe just comment out the line here (and in Overview?) so we can quickly switch it back on if/when we need to.

nmadruga: Yes, good point. I was considering that we could have an status update here, but there shouldn't be one while the Applicant is working with the application.

nmadruga:
I also needed to do something extra to fix a problem: After submitting the application the status was not getting updated int he cached Application. Needed to apply this change in `useLoadApplication` in to correctly redirect the application to the `summary` page after it has been submitted:
https://github.com/openmsupply/application-manager-web-app/blob/b95e56a643979d49a1ff2b6a85e6a4cdd86a30e5/src/utils/hooks/useLoadApplication.tsx#L33

src/utils/hooks/useLoadApplication.tsx
```diff
@@ -30,7 +30,7 @@ const useLoadApplication = ({ serialNumber, networkFetch }: UseGetApplicationPro
       serial: serialNumber,
     },
     skip: triggerProcessing || isApplicationLoaded,
-    fetchPolicy: networkFetch ? 'no-cache' : 'cache-first',
+    fetchPolicy: networkFetch ? 'network-only' : 'cache-first',
```
CarlosNZ, line 33: Cool, I'm glad this works. I remember trying to use `network-only` before and it just not working, so I left it as `no-cache`. No idea why, but glad it seems to be working now.

CarlosNZ (approved):
Just update from master and good to go!

nmadruga:
Thanks for this Review Carl! I'll end up closing this issue and opening another PR now with changes on both (this) and issue #176 to redirect pages to <NoMatch> component. I ended up working on top of this one and accidentally committed my changes in here. Reverting this will be trouble so best to keep with all changes in another one.

PR #287, opened 2021-01-20
#251 URL queries to filters, sorting and pagination

CarlosNZ:
**NOTE**: Please review #289 instead as that supersedes much of this. Although please use this page as a guide for what's changed.

Initial implementation of Filter/sorting handling #251.

~~This will need to be re-visited once we do [#185](https://github.com/openmsupply/application-manager-server/issues/185) and have direct access to more columns for sorting.~~ Done #289 

## How it works:
- There are four variables generated from the URL queries that get sent to the `getApplications.query`:
   - `filters` -- converted from url query key-values into a single "filter" object to query
   - `sortFields` -- Array of fields to sort by (in priority order)
   - `paginationOffset` -- the record to start from, computed from the `page` and `per-page` queries
   - `numberToFetch` -- from `per-page`

Had to re-factor a chunk of `useListApplications` so that the applications collection would be kept in an Array rather than an object, otherwise sorting wouldn't work. Although this two-step process (to get applications then match them to their stage/status) will be unnecessary once we do [#185](https://github.com/openmsupply/application-manager-server/issues/185).

As per the [spec](https://github.com/openmsupply/application-manager-web-app/wiki/Filters-for-Applications), the following filters are implemented so far:

### Filters

(generated in `buildQueryFilters.ts`)

- **type** (template code)
- **stage**
- **status**
- **outcome**
- **applicant** (partial matches to username, firstName or lastName)
- **last-active-date** (as per spec, with "today", "yesterday" and "this-week" labels handled)
- **search** (matches any part of application name, username, firstName, lastName; matches start of stage)  
  (Note: it's not possible to do partial match searches on Status, as it's an `enum` and so only does exact matches)

### Sort fields

(in `buildQueryVariables.ts`)

Note: only native (i.e. non-computed) columns on the application table can be sorted by. This will improve when we do  [#185](https://github.com/openmsupply/application-manager-server/issues/185).

Any of the following fields can be sorted by (using the `field:asc/desc` format), which as you can see, most of which are not very useful:
Screen Shot 2021-01-21 at 11 35 45 AM

### Pagination

- page
- per-page

### Examples

Some URLs to try (not very exciting with only a small number of applications in DB, but you'll get the idea):

- http://localhost:3000/applications?type=ReviewTest&user-role=applicant&stage=screening
- http://localhost:3000/applications?type=TestRego&user-role=applicant&status=draft
- http://localhost:3000/applications?type=TestRego&user-role=applicant&outcome=approved&stage=automatic 
- http://localhost:3000/applications?type=TestRego&user-role=applicant&outcome=approved,pending
- http://localhost:3000/applications?type=TestRego&user-role=applicant&search=craig
- http://localhost:3000/applications?type=TestRego&user-role=applicant&last-active-date=today
- http://localhost:3000/applications?type=TestRego&user-role=applicant&last-active-date=2021-01-20:
- http://localhost:3000/applications?type=TestRego&user-role=applicant&last-active-date=2021-01-20:2021-01-24
- http://localhost:3000/applications?type=TestRego&user-role=applicant&last-active-date=:2021-01-20 (Should show no results)
   (I tested when the database_init was on 2021-01-21, so adjust the above dates accordingly.)
- http://localhost:3000/applications?type=TestRego&user-role=applicant&last-active-date=2021-01-01:&page=1
- http://localhost:3000/applications?type=TestRego&user-role=applicant&last-active-date=2021-01-01:&page=2 (No results)
- http://localhost:3000/applications?type=TestRego&user-role=applicant&page=2&per-page=1 (1 result)
- http://localhost:3000/applications?type=TestRego&user-role=applicant&sort-by=serial
- http://localhost:3000/applications?type=TestRego&user-role=applicant&sort-by=serial:desc


nmadruga:
Wow! Ok, I lot was done here. I'll start to have a look now. I think you have covered part of future issues here as well. So might be best to consider these for `this sprint` or `next sprint` and one of us can work quickly to finish what might be missing for example: sort and pagination : )

nmadruga (commented):
Going to review #289 which has changes from both now

PR #289, opened 2021-01-21
#185B Application list view

### LINKED ISSUE #284: Bug: not displaying current user menu items after login - only after refreshing the page

CarlosNZ:
**Note**: supersedes #287 -- so review this one instead, although please use the description on there as a guide for what has been done.

Front-end complement to [Back-end PR#187](https://github.com/openmsupply/application-manager-server/pull/187) -- refactor front-end to query Application from view table.

- Sorting by all relevant columns now possible (and works)
- New query `getApplicationsList.query` -- old getApplications still there, but we may get rid of it at some point
- refactor TableCell components to parse new format
- remove need for separate queries for applications and their stage/status data, so a lot of the code to process that is removed
- filters on enums (i.e. status, outcome) now don't pass illegal values to DB query (which causes runtime error)

### Off-issue changes:

- handle redirect when no `type` provided in url (as well as userRole). So `/applications` will re-direct and add a default type and user role
- fix #284 -- small change to set `isLoading` back to false after user update


nmadruga:
> handle redirect when no type provided in url (as well as userRole). So /applications will re-direct and add a default type and user role

Hmm so this one also covers what was meant to be done in #226

CarlosNZ:
> Hmm so this one also covers what was meant to be done in #226

Yeah I think so, I wasn't entirely clear of the scope of that one. I just extended your default UserRole check to first get a default Template type if not specified.

nmadruga:
> fix #284 -- small change to set isLoading back to false after user update

❤️

src/utils/helpers/application/buildQueryFilters.ts
```diff
@@ -0,0 +1,119 @@
...
+    in: splitCommaList(values)
+      .map((value) => value.toUpperCase())
+      .filter((value) => {
+        return Object.values(enumList).includes(value)
+      }),
+  }
+}
+
+const parseDateString = (dateString: string) => {
+  if (mapNamedDates?.[dateString]) return mapNamedDates[dateString]
```
nmadruga, line 87: Why not using `mapNamedDates[dateString]`? I'm confused what was the problem.

src/utils/helpers/application/buildQueryFilters.ts
```diff
@@ -0,0 +1,119 @@
...
+}
+
+const datePlusDays = (offset = 0, dateString: string | undefined = undefined) => {
+  if (dateString) return DateTime.fromISO(dateString).plus({ days: offset }).toISODate()
+  return DateTime.local().plus({ days: offset }).toISODate()
+}
+
+interface NamedDateMap {
+  [key: string]: string[]
+}
```
nmadruga, line 104: To the top of the file, please? I think it looks tidier with all types/interface on the top.

src/utils/helpers/application/buildQueryFilters.ts
```diff
@@ -0,0 +1,119 @@
...
+    return [startDate, datePlusDays(1, startDate)]
+  if (endDate === '') return [startDate, undefined] // No end date boundary
+  if (startDate === '') return [undefined, endDate] // No start date boundary
+  return [startDate, endDate]
+}
+
+const datePlusDays = (offset = 0, dateString: string | undefined = undefined) => {
+  if (dateString) return DateTime.fromISO(dateString).plus({ days: offset }).toISODate()
+  return DateTime.local().plus({ days: offset }).toISODate()
+}
```
nmadruga, line 100: I would move these utility methods to the top of the file. I think it's more clear.
```suggestion
const datePlusDays = (offset = 0, dateString: string | undefined = undefined) => {
  return dateString ? DateTime.fromISO(dateString).plus({ days: offset }).toISODate()
  : DateTime.local().plus({ days: offset }).toISODate()
}
```

src/utils/helpers/application/buildQueryFilters.ts
```diff
@@ -0,0 +1,119 @@
...
+// Use this if the values must conform to an Enum type (e.g. status, outcome)
+const inEnumList = (values: string, enumList: any) => {
+  return {
+    in: splitCommaList(values)
+      .map((value) => value.toUpperCase())
+      .filter((value) => {
+        return Object.values(enumList).includes(value)
+      }),
+  }
+}
```
nmadruga, line 84:
```suggestion
// Use this if the values must conform to an Enum type (e.g. status, outcome)
const inEnumList = (values: string, enumList: any) => ({
  in: splitCommaList(values)
    .map((value) => value.toUpperCase())
    .filter((value) => {
      return Object.values(enumList).includes(value)
    }),
})
```

src/utils/hooks/useListApplications.tsx
```diff
@@ -1,98 +1,43 @@
...
-  [serial: string]: ApplicationDetails
-}
-
-const useListApplications = ({ type }: UseListApplicationsProps) => {
-  const [applications, setApplications] = useState<ApplicationDetailsMap>()
+const useListApplications = (urlFilters: QueryObject) => {
+  const [applications, setApplications] = useState<ApplicationList[]>([])
   const [loading, setLoading] = useState(true)
   const [error, setError] = useState('')
+  const { type, sortBy, page, perPage } = urlFilters
```
nmadruga, line 15:
```suggestion
const useListApplications = ({ sortBy, page, perPage, ...queryFilters }: QueryObject) => {
  const [applications, setApplications] = useState<ApplicationList[]>([])
  const [loading, setLoading] = useState(true)
  const [error, setError] = useState('')
```

src/utils/hooks/useListApplications.tsx
```diff
@@ -1,98 +1,43 @@
...
 
-  const { data, error: applicationsError } = useGetApplicationsQuery({
-    variables: { code: type as string },
+  const filters = buildFilter(urlFilters)
```
nmadruga, line 17:
```suggestion
  const filters = buildFilter(queryFilters)
```

src/utils/hooks/useListApplications.tsx
```diff
@@ -1,98 +1,43 @@
...
+  const sortFields = sortBy ? buildSortFields(sortBy) : []
+  const { paginationOffset, numberToFetch } = getPaginationVariables(
+    page ? Number(page) : 1,
+    perPage ? Number(perPage) : undefined
+  )
+
+  const { data, error: applicationsError } = useGetApplicationsListQuery({
+    variables: { filters, sortFields, paginationOffset, numberToFetch },
     fetchPolicy: 'network-only',
     skip: !type,
```
nmadruga, line 27: Can this be removed?

src/utils/helpers/application/buildQueryVariables.ts
```diff
@@ -0,0 +1,35 @@
...
+  stage: 'STAGE_',
+  status: 'STATUS_',
+  outcome: 'OUTCOME_',
+  lastActiveDate: 'LAST_ACTIVE_DATE_',
+}
+
+type PaginationValues = {
+  numberToFetch: number
+  paginationOffset: number
+}
```
nmadruga, line 31: Why not use `interface` and better to add to the top?

nmadruga (commented):
Awesome. A lot was done in here! I tested all URL and tried to add one unexpected key=value. All seems to be working great. I can see that from this base we can build up the remaining list issues so quickly! Thanks for doing this. I made a lot of comments - most are superficial and should help with the format of files. I like to keep reusable function in a separated place so they could be found and used elsewhere. But I can see that in the `buildQueryFilters` a lot is just to be used in there. So maybe the tide up could be the ones used in more than one place could be at the top of the files. That's it - a summary of my review : D

documentation/internal/Filters-for-Applications.md
```diff
@@ -88,6 +89,20 @@ Options:
...
+Example: `status=draft`
+
+Considered in implementation: **Yes** [#251 - Filters]
+
+Can have combined values: **Yes** e.g: `status=draft,submitted`
+
+Options:
+
+- Set of `status` options
+
```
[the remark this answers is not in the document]
CarlosNZ: Yeah sure, okay. I think it can't hurt to have a lot of possible filters, even if some of them don't really end up getting used in our "presets". And I think we'll be adding and tweaking them for the next wee while. :)

src/containers/List/ListWrapper.tsx
```diff
@@ -19,33 +20,46 @@ const ListWrapper: React.FC = () => {
...
+
+    function getDefaultUserRole(templatePermissions: TemplatePermissions, redirectType: string) {
+      const found = Object.entries(templatePermissions).find(
+        ([template]) => template === redirectType
+      )
+      if (found) {
+        const [_, permissions] = found
+        return findUserRole(permissions)
+      }
+    }
```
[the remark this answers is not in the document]
CarlosNZ: Sure, good idea.

src/utils/graphql/queries/getApplications.query.ts
```diff
@@ -1,16 +1,30 @@
 import { gql } from '@apollo/client'
```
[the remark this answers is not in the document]
CarlosNZ: Yeah sure, I was just being cautious in case it was needed elsewhere. It's gone now. 💨

src/utils/hooks/useListApplications.tsx
```diff
@@ -1,98 +1,43 @@
...
   })
 
-  const { data: stagesData, error: stagesError } = useGetApplicationsStagesQuery({
```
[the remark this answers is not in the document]
CarlosNZ: 💨

src/utils/hooks/useListApplications.tsx
```diff
@@ -1,98 +1,43 @@
 import { useEffect, useState } from 'react'
-import {
-  Application,
-  ApplicationStageStatusAll,
-  useGetApplicationsQuery,
-  useGetApplicationsStagesQuery,
-} from '../../utils/generated/graphql'
-import { ApplicationDetails } from '../types'
+import buildFilter from '../helpers/application/buildQueryFilters'
+import buildSortFields, { getPaginationVariables } from '../helpers/application/buildQueryVariables'
```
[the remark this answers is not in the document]
CarlosNZ: Okay, sure, moved them and renamed to `list`. I guess I didn't really understand what "translations" meant, so didn't notice it. "List" is clearer I think.

src/utils/hooks/useListApplications.tsx
```diff
@@ -1,98 +1,43 @@
...
+import { useGetApplicationsListQuery, ApplicationList } from '../../utils/generated/graphql'
+import { QueryObject } from '../hooks/useRouter'
 
-interface UseListApplicationsProps {
-  type?: string
+export type URLQueryFilter = {
```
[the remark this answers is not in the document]
CarlosNZ: Actually don't need this one at all now -- it's the same the `query` object in useRouter. Gone.

src/utils/helpers/application/buildQueryFilters.ts
```diff
@@ -0,0 +1,119 @@
+import { DateTime } from 'luxon'
+import { QueryObject } from '../../hooks/useRouter'
```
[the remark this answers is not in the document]
CarlosNZ: Okay, I've renamed it `BasicStringObject` and put it in 'types', since it highly generic and we seem to use it all over the place.

src/utils/helpers/application/buildQueryFilters.ts
```diff
@@ -0,0 +1,119 @@
...
+import { ApplicationStatus, ApplicationOutcome } from '../../../utils/generated/graphql'
+
+export default function buildQueryFilters(filters: QueryObject) {
+  const graphQLfilter = Object.entries(filters).reduce((filterObj, [key, value]) => {
+    if (!mapQueryToFilterField?.[key]) return filterObj
+    return { ...filterObj, ...mapQueryToFilterField[key](value) }
+  }, {})
+  return graphQLfilter
```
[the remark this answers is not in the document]
CarlosNZ: Yeah, I kind of thought that since the expression is so long and hard to understand that having a variable name describing exactly what it results in might be helpful. Not really fussed either way.

src/utils/helpers/application/buildQueryFilters.ts
```diff
@@ -0,0 +1,119 @@
...
+      or: [
+        { name: { includesInsensitive: value } },
+        { applicantUsername: { includesInsensitive: value } },
+        { applicant: { includesInsensitive: value } },
+        { orgName: { includesInsensitive: value } },
+        { templateName: { includesInsensitive: value } },
+        { stage: { startsWithInsensitive: value } },
+      ],
+    }
+  },
```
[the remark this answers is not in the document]
CarlosNZ: Ah yes, much nicer. I never knew how to use the shorthand return form when it was just returning an object, because it always interpreted the outer {} as the function wrapper. Now I see you can just put () around it! Very nice, thanks.

src/utils/helpers/application/buildQueryFilters.ts
CarlosNZ: Yeah I can see how that is confusing to read. What it's doing is, if the `dateString` is one of the "named" dates (e.g. "today"), just return the converted date value from the mapped object. If not, do some further processing and checks.

I've changed it to this, which is much clearer, I think:
```if (dateString in mapNamedDates) return mapNamedDates[dateString]```

src/utils/helpers/application/buildQueryFilters.ts
CarlosNZ: > I would move these utility methods to the top of the file. I think it's more clear.

Really? I thought we've generally been keeping these "helper" methods down the bottom so they don't get in the way of the main functionality. That's how we've been doing it in React files at least.

src/utils/hooks/useListApplications.tsx
CarlosNZ: But since we need `type` in both this main function and passed into `buildFilter` we still need to do `const { type } = queryFilters` (or refer to `queryFilters.type` later), so it doesn't end up looking cleaner or clearer I think.

src/utils/hooks/useListApplications.tsx
CarlosNZ: Remove what exactly? 🤷‍♂️

src/utils/helpers/application/buildQueryVariables.ts
CarlosNZ: Hmm, I feel like this file contains two different "sections" -- the Sorting processing, and the Pagination process, so I felt it was better to keep the interface in its correct section. (You'd probably put Pagination in its own file, but I couldn't bring myself to make an 8-line file. 😜 )

src/utils/hooks/useListApplications.tsx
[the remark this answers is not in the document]
nmadruga: Thank you

src/utils/hooks/useListApplications.tsx
nmadruga: I was talking about removing the `skip: !type`, but I'm not sure if it can be done

nmadruga (approved):
LGTM

src/utils/hooks/useListApplications.tsx
CarlosNZ: Yeah I think we can remove this now. And that means I can implement your previous suggestion about destructuring in hook parameters, since this function no longer needs type specifically.

Good suggestion.

(Will quickly test before merging)

Issue #291, opened 2021-01-22
Schema changes (implementation)


nmadruga:
Changes to consider Consolidator as another review as per #281.

[UPDATED on 03/02/2021]
The fields on `review` and `review_assignment` previously named `can_make_overall_decision` to be replaced with `is_last_level` as BOOLEAN
On another meeting about `template_permission` the team agreed that the table shouldn't have the field `template_section_id` and we would like to add the `stage` (which is the stage number) in the table.

CarlosNZ:
I have some thoughts about the URL structure in light of recent discussions.

Basically, there's three considerations that have come up that suggest some re-thinking of the URLs for reviews:

1. It’s not a great idea to have the database ID in the URL (the more I read about it, it seems not cool)
2. Reviews and Consolidations are now the same kind of entity in the system
3. Review levels (and their names) should be defined by a template (so can’t assume it’s “Consolidate”) (see [this comment](https://github.com/openmsupply/application-manager-server/issues/188#issuecomment-767868613))

## Proposed solution.

Say we have a template with the following review structure:

**Stage 1**: Screening
- Review Level 1: “Initial approval”

**Stage 2**: Assessment
- Review Level 1: “Review” (split into two sections)
- Review Level 2: “Consolidation”

**Stage 3**: Final Decision
- Review Level 1: "Approval"

The URL structure would be as follows:

`/application/[serial]/review` - review home page, same as current

`/application/[serial]/review/[stage]/[review-name (or level?)]/[number]`  - where `number` is an incrementing value for reviews of this stage/level, starting at 1

So the all the reviews for an application of this template (serial: 12345) would be:

- `/application/12345/review` - home page
- `/application/12345/review/screening/initial-approval/1`
- `/application/12345/review/assessment/review/1`
 - `/application/12345/review/assessment/review/2`
- `/application/12345/review/assessment/consolidation/1`
- `/application/12345/review/final-decision/approval/1`

Regarding “number”, we’d probably just generate it on review creation and store it in an additional field on the review table.

Regarding having `review` in the URL twice if the review level is just called “review”, we could always just drop the first “review” if it’s still clear enough, so:

- `/application/12345/screening/initial-approval/1`
- `/application/12345/assessment/review/1`
 - `/application/12345/assessment/review/2`
- `/application/12345/assessment/consolidation/1`
- `/application/12345/final-decision/approval/1`

(I think I prefer this one)

Thoughts?

nmadruga:
[...]

Issue #204, opened 2020-12-10
Add action to set the first review_status_history to DRAFT after the review is created


nmadruga:
It should mirror the same process as in the application.

The action should be already in there, we need to check. Add to every template if not there (besides User Registration);

also create a documentation list of all DEFAULT actions that should be included in every template and think about how we set one as default (that cannot be deleted. And propose a solution.

CarlosNZ:
Needs to happen after (or simultaneously with) #202

nmadruga:
This issue also needs to add the action to set the first `review_status_history` as DRAFT after the review is created

CarlosNZ:
The action exists, just need to add it to the templates -- will do now. Have added it to the [default list](https://github.com/openmsupply/application-manager-server/issues/184).

PR #292, opened 2021-01-25
#176 Show 404 unreachable page as rollback component

### LINKED ISSUE #176: Show error 404 page on application and review pages

nmadruga:
Fix #176 

~~[UPDATED 02/02] Merged with branch [#291 Updates for schema changes](302), please merge it before this one.~~ DONE

### Changes:
- Show the `NoMatch` component on page when the Application or Review isn't found (returned by: `useLoadApplication` or `useLoadReview` hooks).
  - ApplicationOverview
  - ApplicationPageWrapper
  - ApplicationSubmission
  - ReviewOverview
  - ReviewPageWrapper
- Updated most hooks to return error as `string` or `undefined`.  And a lot of rearranging to use less code to do the same: returning the error and loading status based on all hooks that should retrieve data in sequence back to the container.
-  `useTriggerProcessing` hook now expects a flag `checkTriggers` that will set when it would start running. 
In Applications pages wait the application to load (without status & stage), if no application found the trigger will not run.
In Review pages wait for the review to load (without status), if no review found the trigger will not run.
Then after waiting the trigger result (check if any is running for the current Application/Review) will get the stage/status and return to the container to run. In application that is when the `isApplicationLoaded` is set to `true`. The `isReviewLoaded` isn't yet in use, but should be used in the same way, when we start displaying different status of reviews.
- The `useApplicationStatus` and `useReviewStatus` queries both use **fetchPolicy** network only to get the latest status.

### Unrelated changes
- While working on top of a different PR to remove the need of an specific hook my changes ended up being commit to the base one. So the changes done on PR #286 are showing in here:
  - Removing `useGetApplicationStatus` to be called separately in containers - now is called as part of `useLoadApplication`. 

### Problem
- On application it is still trying to **validate** responses after the application is submitted. Not sure why this is happening, but debugging show that the `ApplicationOverview` call is setting `revalidateAll` flag to `true`.


src/utils/hooks/useLoadApplication.tsx
```diff
@@ -2,32 +2,58 @@ import { useEffect, useState } from 'react'
...
+} from '../types'
 
-const useLoadApplication = (props: UseGetApplicationProps) => {
-  const { serialNumber } = props
+const useLoadApplication = ({ serialNumber, networkFetch }: UseGetApplicationProps) => {
   const [application, setApplication] = useState<ApplicationDetails>()
   const [templateSections, setSections] = useState<TemplateSectionPayload[]>([])
+  const [appStages, setAppStages] = useState<ApplicationStages>()
   const [isApplicationLoaded, setIsApplicationLoaded] = useState(false)
+  const [checkTrigger, setCheckTrigger] = useState(false)
```
CarlosNZ, line 23: So, just to check I'm understanding this correctly:
- we have 3 flags: `isApplicationLoaded`, `checkTrigger` and `triggerProcessing`, which are initially set to `false` (or `true` for `triggerProcessing`)
- after initial application is loaded, `checkTrigger` gets set to `true`
- this allows `useTriggerProcessing` to check the trigger status
- if trigger is ok (not "Processing), then `triggerProcessing` is set to `false`
- this allows `useGetApplicationStatusQuery` to load the status info (fresh from the network) and sets `isApplicationLoaded` to `true`

I think the name `checkTrigger` is a bit confusing and not consistent with the others. I'm not sure what the best solution is here, because `isApplicationLoaded` is kind of ambiguous, since there's two separate stages for loading the application. How about:
- `checkTrigger` becomes `isApplicationLoaded` (or `isApplicationLoadedInitial` -- though probably a bit long)
- `triggerProcessing` stays the same (or `isTriggerProcessing` for full consistency)
- `isApplicationLoaded` becomes `isApplicationReady`

Have a think and see what you reckon. Can have a brief discussion if you want.

src/utils/hooks/useLoadReview.tsx
```diff
@@ -15,8 +20,10 @@ interface UseLoadReviewProps {
 const useLoadReview = ({ reviewId, serialNumber }: UseLoadReviewProps) => {
   const [applicationName, setApplicationName] = useState<string>('')
   const [reviewSections, setReviewSections] = useState<SectionStructure>()
-  const [loading, setLoading] = useState(true)
-  const [error, setError] = useState('')
+  const [reviewStatus, setReviewStatus] = useState<ReviewStatus>()
+  const [isReviewLoaded, setIsReviewLoaded] = useState(false)
+  const [reviewError, setReviewError] = useState<string>()
+  const [checkTrigger, setCheckTrigger] = useState(false)
```
CarlosNZ, line 26: Same as the comment I made in `useLoadApplication`, depending on what you end up doing.

src/containers/Application/ApplicationSubmission.tsx
```diff
@@ -47,23 +36,20 @@ const ApplicationSubmission: React.FC = () => {
   }, [appStages, currentUser])
 
   useEffect(() => {
-    if (!isApplicationLoaded || statusLoading) return
-
+    if (!isApplicationLoaded) return
+    const status = application?.stage?.status
     // Check if application is in Draft or Changes required status and redirect to the summary page
     // Note: The summary page has its own redirection logic to ay specific page (with invalid items).
```
CarlosNZ, line 42:
```suggestion
    // Note: The summary page has its own redirection logic to a specific page (with invalid items).
```

src/utils/hooks/useLoadReview.tsx
```diff
@@ -27,41 +34,39 @@ const useLoadReview = ({ reviewId, serialNumber }: UseLoadReviewProps) => {
...
-    }
-    if (application) setApplicationName(application.name)
-  }, [applicationError, application])
+  const { triggerProcessing, error: triggerError } = useTriggerProcessing({
+    checkTrigger,
+    reviewId,
+    // triggerType: 'reviewTrigger',
+  })
+
+  const { data: statusData, error: statusError, loading: statusLoading } = useGetReviewStatusQuery({
```
CarlosNZ, line 56: Shouldn't this one also pass in `triggerProcessing` like the Application one does? (triggerProcessing is currently not being used for anything)

src/utils/hooks/useLoadReview.tsx
```diff
@@ -74,16 +79,30 @@ const useLoadReview = ({ reviewId, serialNumber }: UseLoadReviewProps) => {
...
-      setLoading(false)
+      setCheckTrigger(true)
     }
-  }, [responsesError, apolloError, data])
+  }, [data])
+
+  useEffect(() => {
+    if (!statusData) return
+    const statuses = statusData?.reviewStatusHistories?.nodes as ReviewStatus[]
+    if (statuses.length > 1) console.log('More than one status resulted for 1 review!')
```
CarlosNZ, line 89: Good to check -- are we still getting any situations when this is happening?

CarlosNZ (changes requested):
Great, good to simplify these processes. Just a couple of things to think about.

I know changing the flag variable names would be a hassle, so feel free to leave it till another time, or we can have a quick chat about it to lock it down if you want.

src/utils/hooks/useLoadReview.tsx
nmadruga: Yes, that was just an attempt to use the trigger for review, but hasn't finished - because there are still #204 issue to add the action to create the first `review_status_history` when the review is created. I though it would be good to do that when using this. But good point, I'll add the `triggerProcessing` check here.

src/utils/hooks/useLoadReview.tsx
nmadruga: Yes, that was mainly for dev purpose. I saw that happening when I didn't use the `is_current_status` on the query

src/utils/hooks/useLoadApplication.tsx
nmadruga: Finally found time to do that! I used exactly your suggestions, so now there are 3 flags:
* isApplicationLoaded
* isTriggerProcessing
* isApplicationReady

And the flags go in the same order as listed here, each next hook will wait until the previous flag is finished. Then the UI only render on `isApplicationReady`. Also did a similar approach for the review loading hooks

src/utils/hooks/useLoadReview.tsx
nmadruga: Done

nmadruga:
I have also merged branch [#291 Updates for schema changes](#302) here, so it would have the latest changes on the schema.

nmadruga:
@andreievg @CarlosNZ 

I have a questions about this situation:
- A Reviewer that can review application number 12345 can load page `application/12345/review`, but what should happen when they try to load page `application/12345` for now what is happening is that the Application summary (as Applicant sees) is loaded. Should we also 404 this page for a Reviewer? And how can we check that, since they have permission to see the application?

CarlosNZ:
>     * A Reviewer that can review application number 12345 can load page `application/12345/review`, but what should happen when they try to load page `application/12345` for now what is happening is that the Application summary (as Applicant sees) is loaded. Should we also 404 this page for a Reviewer? And how can we check that, since they have permission to see the application

I think seeing the Summary view as the Applicant sees it makes sense. If they have permission to review it, then they can obviously see it, so just showing the application seems like the right thing to do.

Maybe, at some point down the line, we could add some additional checks on this page and show a "Go to Review" button on the page if appropriate, but I think how it is now is fine for now.

nmadruga:
Coo. Thanks! I think this one is ready for re-review.

src/utils/graphql/queries/getReview.query.ts
```diff
@@ -5,7 +5,8 @@ export default gql`
     review(id: $reviewId) {
       reviewer {
         id
-        username
```
CarlosNZ: Why not leave `username` in? There might not be an immediate use, but can't hurt to have it available, right?

src/utils/hooks/useTriggerProcessing.tsx
```diff
@@ -3,32 +3,46 @@ import { useGetTriggersQuery } from '../../utils/generated/graphql'
...
+  reviewId: number
+}
+interface TriggerAssignment {
+  reviewAssignmentId: number
+}
+type TriggerQueryProps = { triggerType?: TriggerType } & (
+  | TriggerApplication
+  | TriggerReview
+  | TriggerAssignment
+)
```
CarlosNZ, line 21: Nice :) 
So this combines TriggerType with one of the specific types (app, review, etc). I guess it's not really feasible to ensure that the OR types match the TriggerType?

src/utils/hooks/useTriggerProcessing.tsx
```diff
@@ -3,32 +3,46 @@ import { useGetTriggersQuery } from '../../utils/generated/graphql'
...
 
-  // Ensure at least one the primary identifiers is provided
-  if (!serialNumber && !reviewAssignmentId && !reviewId) {
+  // If triggerType not provided, infer it from the supplied ID
+  const inferredTriggerType = inferTriggerType(triggerType, props)
+
+  if (!inferredTriggerType) {
     setIsProcessing(false)
-    setTriggerError('INVALID QUERY')
+    setTriggerError('Type not defined')
```
CarlosNZ, line 32: Maybe "Invalid trigger type" might be better, since it could be defined, but not one of the proper ones.

CarlosNZ (approved):
Great, it's all working very nicely.

Just one other little additional thing, although it's not in the change list here:
- In ReviewSubmission.tsx, there is the "REVIEW SUBMITTED" string which should probably go to constants. And the spelling on SUBMITTED is not quite correct. (And also, can you save the string in normal case and just add css "text-transform" property if you want it upper-case)

👍

src/utils/hooks/useTriggerProcessing.tsx
nmadruga: I think we can use it at some point, I just wasn't sure how to do, so left it not using it to deduce. Let's improve in a next PR

src/utils/hooks/useTriggerProcessing.tsx
nmadruga: Cool. I'll change to that

nmadruga:
> In ReviewSubmission.tsx, there is the "REVIEW SUBMITTED"

Yes, I was thinking about that incompleted issues there, I will add an issue to work with that one, since it should do more stuff there no only display this label and also when reloading the Review page it shouldn't allow a second review to be done. So all that can be done as part of this new issue.

Issue #270, opened 2021-01-18
Element plugin: Checkboxes


CarlosNZ:
## Requirements:

- Can specify one or many checkboxes in one element. User can select as many as needed.
- Each checkbox can have a default selection state (checked/unchecked)
- A label for each checkbox
- Checkbox can also be a toggle switch or a slider switch (very simple variants with SemanticUI)
- ~~Needs a SummaryView (so it can show the multiple values -- default SummaryView won't do)~~

#### Input parameters:

- **label**: `string`
- **description**: `string`
- **checkboxes**: `array[string | object]` -- array of labels for checkboxes
   - if a string array, it'll just be a group of unchecked checkboxes. However, more specific details can be provided by providing objects instead, with the following parameters:  
   ```
   {
    label: <string> - Text to display next to checkbox (Can be empty string but not omitted)
    text: <string> - value to store in Resonse "text" field and shown in Summary View. Will be same as label if omitted.
    key: <string> - unique code used as index for Response object (for searching using subsequent evaluator expressions)
    selected: true/false (initial state)
   }
  ``` 
- **type**: `checkbox | toggle | slider` -- (default: `checkbox`). Changes the UI display to a toggle switch or a slider switch.
- **layout**: `inline | block` -- (default: `block`) If "inline", multiple checkboxes are laid out horizontally (space depending) [NOTE: Only bother with this if its a simple Semantic option]

#### Output (Response) object (JSONValue):

```
{
  text: <string> -- comma separated list of all selected checkbox "text" values -- for display in Summary view
  values: {
    'key1' : { text: <text value>, isSelected: true/false }
    'key2' : { text: <text value>, isSelected: true/false }
     ...
    }
}
```

#### Limitations of this suggested solution:

- It's straightforward to make a list of string array options from a dynamic evaluator query, but it would be tricky to build an object array. Which means dynamic queries can really only be used for the "simple" version (and can't have default selections, etc.)
- the _order_ of the original options list wouldn't be preserved in the response object. If this was important, you'd have to make sure you provided alphabetically ordered keys.

--- 

### Considerations (updated above based on comment discussion below)

This is a bit trickier than the previous plugins as the main reponse value will be more than just a `text` field. So I think it's worth doing soon as a basis for other "multi-value" plugins.

I suggest the main response object (`jsonValue` in wrapper should (in addition to the existing `isValid` field) should be a `value` field with the following shape:
Object with each property being the index of the checkbox item. Then each index field contains:
 ```
[...]

Issue #191, opened 2020-12-07
Remove many Semantic themes (that aren't in use) to speed up the first compilation for the front-end bundle.


nmadruga:
Also remove all components not in use from file `sematic.less`

Keep the `default` theme

PR #296, opened 2021-01-26
#270 Checkboxes plugin

CarlosNZ:
Implements #270 as per "Agreed solution". Requires [back-end PR #191](https://github.com/openmsupply/application-manager-server/pull/191) to demonstrate the plugin.

See Page 4 of FeatureShowcase template for a "Checkbox showcase"

Something to consider: if nothing is selected for a checkbox or checkbox group, the "Text" value that is displayed on the Summary page is blank. Is this what we want, or should it show something like "\<Nothing selected\>"?

### Additional changes:

- Small bug fix to `radioChoice` plugin -- it wasn't displaying saved values on reload
- Simplified `noParagraph` renderer in `semanticReactMarkdown` (will need to make adjustments to "Message" renderer at some point too)


src/utils/helpers/semanticReactMarkdown.tsx
```diff
@@ -42,11 +42,7 @@ const MarkdownBlock: React.FC<MarkdownBlockProps> = (props) => {
       },
     },
     // Prevents returned HTML from being wrapped in <p></p> tags
-    noParagraph: {
-      paragraph: ({ node: { children } }: any) => {
-        return <span>{children?.[0]?.value}</span>
-      },
-    },
+    noParagraph: { paragraph: 'span' },
```
nmadruga, line 45: That looks simpler : D

documentation/internal/Element-Type-Specs.md
```diff
@@ -152,6 +152,47 @@ _Multi-choice question, with one allowed option, displayed as Drop-down list (Co
...
+- **layout**: `string` -- if "inline", displays checkboxes horizontally in rows. Useful if there are a lot of checkboxes.
+
+#### Response type
+
+```
+{
+  text: <string> -- comma separated list of all selected checkbox "text" values, shown in Summary view (or Review)
+  values: {
+            <key-name-1> : { text: <text value>, isSelected: <boolean>}
+            <key-name-2> : { text: <text value>, isSelected: <boolean>}
```
nmadruga, line 187: Is the `key-name` the key defined in the checkbox Object as the `key`? And why this doesn't result also the label? You don't think it would be used?

documentation/internal/Element-Type-Specs.md
```diff
@@ -152,6 +152,47 @@ _Multi-choice question, with one allowed option, displayed as Drop-down list (Co
...
+- **checkboxes\***: `array[string | checkbox]` -- an array of labels, one per checkbox. For more complexity, an array of Checkbox objects can be provided with the following properties:
+
+  ```
+   {
+     label: <string> - text to display next to checkbox (Can be empty string but not omitted)
+     text: <string> - value to store in Response "text" field and shown in Summary View. Will be same as label if omitted.
+     key: <string | number> - unique code used as key/property name for Response object. Defaults to numerical index of array if omitted
+     selected: <boolean> - initial state of checkbox
+   }
+  ```
```
nmadruga, line 175: All good here. I just wonder if for an extension of this plugin we would have here another option to enter something to link this option to e.g. a `permission_name` or an action.

nmadruga (changes requested):
Again very good. 

I like the possibility of using a slider here, bug with the example given is actually hard to see a real use, I think it would be better to add more options to the slider idk?

And I think I found one issue with display of this plugin in the `SummaryView`, when it's a toggle it doesn't show anything. Can we display something as the result of the user toggle/untoggle? I mean, there are 2 values associated with it, just not sure why it isn't displaying in the UI:

Screen Shot 2021-01-29 at 2 55 33 PM

CarlosNZ:
> I think it would be better to add more options to the slider idk

Sure, I did suggest having a different display value when NOT selected, but Andrei didn't think it was necessary. Happy to take suggestion though, discuss it

> And I think I found one issue with display of this plugin in the SummaryView, when it's a toggle it doesn't show anything. Can we display something as the result of the user toggle/untoggle? 

Yes, I realise this is slightly weird, hence my comment in the original post:

>> Something to consider: if nothing is selected for a checkbox or checkbox group, the "Text" value that is displayed on the Summary page is blank. Is this what we want, or should it show something like "\<Nothing selected\>"?

The intention was that the "Text" field would only show the text from _selected_ items, otherwise how would it look any different to un-selected items? But yeah, it should probably show _something_ in Summary view if nothing is selected. Let me know your thoughts.

> I mean, there are 2 values associated with it, just not sure why it isn't displaying in the UI:

That's the problem, there aren't two values, only a single value for when selected.

documentation/internal/Element-Type-Specs.md
CarlosNZ: > Is the key-name the key defined in the checkbox Object as the key? 

Yes, but the _name_ of the key here is the _string_ that was given to the `key` property in the input array. See the [original post](#270) for the reasoning as to why it's like this rather than an array, along with Andrei's suggestion and discussion.

> And why this doesn't result also the label? You don't think it would be used?

If the `text` property is omitted, the label _becomes_ the text value. You only specify a text property if you want the response to store a *different* text string to what is displayed to the user in the UI.

documentation/internal/Element-Type-Specs.md
CarlosNZ: Yes, we still have to figure out exactly how we're going to link these things (using the "reference" field we've alluded to in the docs but never really fleshed out). But the response object for any plugin can easily be extended at any time to include additional fields. 

I've been bearing it in mind, but I think we won't know *exactly* how it's going to connect back to other entities until we go to implement one -- probably the "select a company to join" option will be the first.

documentation/internal/Element-Type-Specs.md
nmadruga: Ok. Thanks for the explanation. I had a quick look on issue #270 to briefly see the discussion too

nmadruga:
About the toggle: It seems weird to only have one toggle per component and then have these displayed in the summary as `empty` or `value`. Thinking about a real-life component I guess it would be more like a group of toggles. Then the summary show either the list of `values` of the checked ones or it would show the toggles ON and OFF the same way - but disabled. What happens when you add a list of values and select some and not others?

CarlosNZ:
@nmadruga Okay, I've updated it so the Summay View shows a generic message when the "text" value is empty string:
Screen Shot 2021-02-02 at 12 11 02 PM

nmadruga:
Great! Thank you

nmadruga (approved):

Issue #157, opened 2020-11-30
Submit application review


nmadruga:
When all questions have ✅ approval suggestion, or at least one have ❗ rejected suggestion the submission can take place.
Then after the submission can be done the reviewer can submit the review, which should also send a trigger of `ON_REVIEW_SUBMITTED` to the review table (on single hook - resend all `review_response` and `review` with trigger.

Should also deal with showing the validation of the review when the user clicks the submit review button as designed:
<img width="519" alt="Screen Shot 2020-12-07 at 12 32 24 PM" src="https://user-images.githubusercontent.com/16461988/101296417-49374a00-3888-11eb-9ffe-591e420c56dd.png">

PR #299, opened 2021-01-27
#157 submit review

nmadruga:
To replace previous PR - which I thought was reverting changes on master.


CarlosNZ (approved):
Thanks, Nicole. Can I assume that this is just an aggregation of existing changes with conflicts resolved and other fixes? But it will essentially bring master back to what it should have been if your other PR had behaved correctly? If so -- Approved! (I don't really want to re-read thru 25 files worth of changes lol -- I trust you know what you're doing )

Issue #297, opened 2021-01-26
createApplication mutation should include user and org id


CarlosNZ:
Just noticed when submitting a new application that the user_id and org_id fields on the application table are NULL. They should be filled with the appropriate values as part of the `updateApplication` mutation.

PR #300, opened 2021-01-27
#297 Fix CreateApplication mutation

CarlosNZ:
Very small update --  adds optional variables `userId` and `orgId` to createApplication mutation, and ensures that those values are passed through from ApplicationCreate.tsx

### DIFF
--- a/src/containers/Application/ApplicationCreate.tsx
+++ b/src/containers/Application/ApplicationCreate.tsx
@@ -5,6 +5,7 @@ import { useRouter } from '../../utils/hooks/useRouter'
 import { Header, Message } from 'semantic-ui-react'
 import useLoadTemplate from '../../utils/hooks/useLoadTemplate'
 import useCreateApplication from '../../utils/hooks/useCreateApplication'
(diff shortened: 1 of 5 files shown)


nmadruga (approved):
Seems good!

I couldn't see the `organisation` on the result of the actions triggered in the back-end, but I guess that would be useful for later on - when we allow a user to see applications from its company right? So now we are storing the company that the user was logged in when created a new application? I tried with `UserRego` and there was a problem to load the application. I don't think it's related to this?? But after I have log out the application loaded fine.

CarlosNZ:
> I couldn't see the `organisation` on the result of the actions triggered in the back-end, but I guess that would be useful for later on - when we allow a user to see applications from its company right?

I've added it to the main "ApplicationData" object (that is generated at the start of an action sequence) in the "Create Org" action PR. 

> So now we are storing the company that the user was logged in when created a new application?

Yes, and if they're logged in without a company, or are `nonRegistered`, the org_id will be `null`.

PR #302, opened 2021-01-28
#291 Updates for schema changes

### LINKED ISSUE #291: Schema changes (implementation)

nmadruga:
Fixes #291 

To be ran with Back-end [PR 291F](https://github.com/openmsupply/application-manager-server/pull/197) 

- Updates graphql query/mutation and enum using new fields in schema.
- Fix problem to load review page.

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
   },
   "dependencies": {
     "@apollo/client": "^3.1.4",
-    "@openmsupply/expression-evaluator": "^1.4.0",
+    "@openmsupply/expression-evaluator": "^1.5.0",
     "apollo3-cache-persist": "^0.8.0",
     "graphql": "^15.3.0",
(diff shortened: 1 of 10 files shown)


CarlosNZ (approved):
Just double-check those name changes to make sure they match the schema. Otherwise, all good 👍

nmadruga:
@CarlosNZ sorry! I had already done the changes but forgot to push to the remote 😝

PR #306, opened 2021-02-02
#236 Start page with progress and button to resume

### LINKED ISSUE #236: Show progress on start page of existing application with status Draft
Show progress on each section and validation icons.
Hide 'Get started' button when application already exists.
Show 'Resume' button on the last section with invalid (or incomplete) questions.

Screen Shot 2021-02-02 at 11 08 45 AM

nmadruga:
Fix #236 

~~Branched off [PR of issue #176 Show 404 unreachable page as rollback component](292) please merge it before this one.~~ DONE
~~Using Backend with latest schema changes and some more inserted data: [Back-end#291F](https://github.com/openmsupply/application-manager-server/pull/197)~~ DONE

### Description
- Display progress on application (for the Applicant) when loading the page `application/<serial>` is loaded and the application is still in DRAFT status.
- If application is SUBMITTED will load the `application/<serial>/summary` page
- Otherwise loads the application start page with progress as:
  - 100% progress in a section means - all `required` have been filled out.
  - When all sections have 100% completed fields (and valid) show the **Review & Summary** button
  - **No progress bar** displayed when no fields are filled in section
  - **Green** progress bar if all filled field have are valid
  - **Red** if at least one field is invalid
  - First section with incomplete/invalid fields show the **Resume** button -> that links a specific page

[UPDATED on 05/02/2021]
### Unrelated changes
- Add helper `getAssignedElements` to simplify `useGetReviewAssignment` hook, it retrieves sections and questions assigned to the reviewer
- Add helper `updateSectionsProgress` to update received by props array of **SectionDetails** with the progress in each section after receiving a promise back from the `revalidateAll` method.

### Existing todos:
- ~~Not sure if should display a Start page when there is no `startMessage` set for the application template. For now is going straight to section 1 page 1. But can change this to go to first incomplete/invalid section. I wasn't sure if we would display a start page to the Applicant anyway (not on start, but when continuing the application).~~ Edge case: Goes straight to 1st page in 1st section
- The link on the **Resume** button isn't working. It show the URL to the correct page but doesn't redirect. I think it's a problem with the `ApplicationStart` component not being part in the Router...


CarlosNZ:
>     * Not sure if should display a Start page when there is no `startMessage` set for the application template. For now is going straight to section 1 page 1.

This is by design. This way things like User Registration can be launched in to immediately without unnecessary start pages. (I think we discussed this when we discussed the Start message)

src/utils/helpers/application/revalidateAll.ts
```diff
@@ -7,23 +7,36 @@ import {
...
-  elementsState: ApplicationElementStates,
-  responsesByCode: ResponsesByCode,
-  currentUser: User,
+interface RevalidateAllProps {
+  elementsState: ApplicationElementStates
+  responsesByCode: ResponsesByCode
+  currentUser: User
+  sectionCode?: string
+  strict?: boolean
+}
```
CarlosNZ, line 16: Changes here look good and useful. Just checking if you've thoroughly regression tested this method in its other places (Summary page access and Submit), so we don't get any unintended regressions slipping through. (Esp related to the removal of the `strict` flag)

src/utils/hooks/useGetSectionProgress.tsx
```diff
@@ -0,0 +1,86 @@
...
+  useEffect(() => {
+    if (loadStart && currentUser && templateSections && elementsState && responsesByCode) {
+      setIsLoadingProgress(true)
+      runScanSectionProgress()
+    }
+  }, [loadStart])
+
+  async function runScanSectionProgress() {
+    let sectionsProgress: SectionsProgress = {}
+    templateSections.forEach(async ({ code, title, index }) => {
```
CarlosNZ, line 40: As discussed, you probably want a `for...in` loop here as it's in an `async` function

src/utils/hooks/useGetSectionProgress.tsx
```diff
@@ -0,0 +1,86 @@
...
+      const validate = await revalidateAll({
+        elementsState,
+        responsesByCode,
+        currentUser,
+        sectionCode: code,
+        strict: false, // Do we still need this flag?
```
CarlosNZ, line 46: Yes?

src/utils/hooks/useGetSectionProgress.tsx
```diff
@@ -0,0 +1,86 @@
...
+        },
+      }
+
+      setSections(sectionsProgress)
+      if (templateSections.length === Object.keys(sectionsProgress).length)
+        setIsLoadingProgress(false)
+    })
+  }
+
+  const getLinkToSection = (validityFailures: any): string => {
```
CarlosNZ, line 70: Am I correct in my understanding that this will get a link to the Section AND Page? If so, maybe call it `getLinkToLocation` or `getLinkToInProgressLocation` ?

src/utils/hooks/useLoadApplication.tsx
```diff
@@ -12,11 +13,13 @@ import {
   ApplicationDetails,
   ApplicationStages,
   TemplateSectionPayload,
+  TemplateType,
   UseGetApplicationProps,
 } from '../types'
 
 const useLoadApplication = ({ serialNumber, networkFetch }: UseGetApplicationProps) => {
   const [application, setApplication] = useState<ApplicationDetails>()
+  const [templateType, setTemplateType] = useState<TemplateType>()
```
CarlosNZ, line 22: Maybe call this `templateInfo`? I find `templateType` to imply that it's a Type (as in Typescript), but it isn't.

src/utils/hooks/useLoadApplication.tsx
```diff
@@ -68,6 +71,15 @@ const useLoadApplication = ({ serialNumber, networkFetch }: UseGetApplicationPro
...
 
+      const { id, code, name, startMessage } = application.template as Template
+
+      setTemplateType({
+        id,
+        code,
+        name: name as string,
+        startMessage: startMessage ? startMessage : undefined,
+      })
+
```
CarlosNZ, line 82: Just wondering why you only want these four fields? Why not just return the whole `application.template` object, even if it's not all used currently?

Also, just a question: where was the PageWrapper getting this info from previously?

src/utils/hooks/useLoadApplication.tsx
```diff
@@ -68,6 +71,15 @@ const useLoadApplication = ({ serialNumber, networkFetch }: UseGetApplicationPro
 
       setApplication(applicationDetails)
 
+      const { id, code, name, startMessage } = application.template as Template
+
+      setTemplateType({
+        id,
+        code,
+        name: name as string,
+        startMessage: startMessage ? startMessage : undefined,
```
CarlosNZ, line 80:
```suggestion
        startMessage,
```
Pretty sure this does the same thing -- if it's undefined the property will carry through the "undefined" value, won't it?

src/utils/hooks/useGetSectionProgress.tsx
```diff
@@ -0,0 +1,86 @@
...
+          info: {
+            title,
+            code,
+          },
+          progress: { total, done, valid, completed: valid && total === done },
+          link: !(total === done && valid) ? getLinkToSection(validityFailures) : undefined,
+        },
+      }
+
+      setSections(sectionsProgress)
```
CarlosNZ, line 64: This is within the `for..each` loop. As well as the async problem mentioned above, wouldn't it be better to collect all the section loop results together and then only do one `setSections` mutation with the whole thing after the loop?

src/containers/Application/ApplicationOverview.tsx
```diff
@@ -86,11 +86,11 @@ const ApplicationOverview: React.FC = () => {
...
   const revalidateAndUpdate = async () => {
-    const revalidate = await revalidateAll(
-      elementsState as ApplicationElementStates,
-      responsesByCode as ResponsesByCode,
-      currentUser as User
-    )
+    const revalidate = await revalidateAll({
+      elementsState: elementsState as ApplicationElementStates,
+      responsesByCode: responsesByCode as ResponsesByCode,
+      currentUser: currentUser as User,
```
CarlosNZ, line 92: I see -- it's because Typescript won't allow `elementsStateaas ApplicationElementStates`. Ugh, it's so ugly 😆

src/components/Application/ApplicationStart.tsx
```diff
@@ -30,8 +37,27 @@ const ApplicationStart: React.FC<ApplicationStartProps> = (props) => {
...
+    const isApplicationCompleted = Object.values(sectionsProgress).every(
+      ({ progress }) => progress?.completed
+    )
+    if (!isApplicationCompleted) findFirstImcompleteSection()
+
+    setIsApplicationCompleted(isApplicationCompleted)
   }, [startMessage, currentUser])
 
+  const findFirstImcompleteSection = () => {
+    const firstIncompleteSection = Object.entries(sectionsProgress)
```
CarlosNZ, line 50: Also -- same comment as in the other file: if this is actually linking to a section AND page, perhaps call it `firstIncompleteLocation`?

src/components/Application/ApplicationStart.tsx
```diff
@@ -60,23 +86,72 @@ const ApplicationStart: React.FC<ApplicationStartProps> = (props) => {
...
+                                error={!progress.valid}
+                              />
+                            )}
+                          </Grid.Column>
+                          <Grid.Column width={2}>
+                            {progress && info.code === firstIncomplete && (
+                              <Button
+                                color="blue"
+                                as={Link}
+                                to={link} // TODO: Link not working - page doesn't refresh!
```
CarlosNZ, line 122: Hmmm, very strange. Would be good to get to the bottom of this in this PR if possible.

CarlosNZ (changes requested):
Cool, pretty big issue! I think there's probably some additional streamlining we can do around validation (progress bar, etc.) but we'll need to talk and plan that for another time.

nmadruga:
> This is by design. This way things like User Registration can be launched in to immediately without unnecessary start pages. (I think we discussed this when we discussed the Start message) I don't think it matters about going to the "first invalid page", since this functionality was intended for short applications that should be done in one go (like User Registration).

Cool. I'll leave it as it is.

src/containers/Application/ApplicationPageWrapper.tsx
```diff
@@ -214,8 +230,14 @@ const ApplicationPageWrapper: React.FC = () => {
 
   return error || responsesError ? (
     <NoMatch />
-  ) : loading || responsesLoading ? (
+  ) : loading || responsesLoading || (loadStart && isLoadingProgress) ? (
     <Loading />
+  ) : loadStart && sections ? (
+    <ApplicationStart
+      template={templateType as TemplateType}
+      sectionsProgress={sections}
```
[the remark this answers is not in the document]
nmadruga: As discussed in person - I'll be looking to change that in a moment - will just finish replying to your other review suggestions first

src/containers/Application/ApplicationPageWrapper.tsx
```diff
@@ -192,11 +208,11 @@ const ApplicationPageWrapper: React.FC = () => {
...
     setSummaryButtonClicked(false)
-    const revalidate = await revalidateAll(
-      elementsState as ApplicationElementStates,
-      responsesByCode as ResponsesByCode,
-      currentUser as User
-    )
+    const revalidate = await revalidateAll({
+      elementsState: elementsState as ApplicationElementStates,
+      responsesByCode: responsesByCode as ResponsesByCode,
+      currentUser: currentUser as User,
```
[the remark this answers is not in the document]
nmadruga: It doesn't let you do that : (

src/utils/helpers/application/revalidateAll.ts
```diff
@@ -7,23 +7,36 @@ import {
...
+    : Object.keys(elementsState)
+
+  // Now filter only questions, visible and which ones should be valid
+  const filteredQuestions = elementCodes.filter(
     (key) =>
       responsesByCode && // Typescript requires this
       elementsState[key].category === 'QUESTION' &&
       elementsState[key].isVisible === true &&
       // Strict/Loose validation logic:
-      ((strict && elementsState[key].isRequired) || responsesByCode[key]?.isValid !== null)
```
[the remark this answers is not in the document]
nmadruga: As discussed earlier - I'll put it back and use with `strict` more for this use

src/utils/helpers/application/revalidateAll.ts
nmadruga: I have tested. It seemed ok, but I think it would be nice if you could check too.

I have reverted the change, so the `strict` check is back

src/utils/hooks/useLoadApplication.tsx
nmadruga: This info was not required for PageWrapper, and it's only now to load the `ApplicationStart` component. Before the component was called by `ApplicationCreate` and this information was getting retrieved by `useLoadTemplate`

nmadruga: I don't like so much the idea of returning the Template object because it has so many unusable fields in it, that's why I'm always converting into a local type. I think this can stay the same way

src/utils/hooks/useLoadApplication.tsx
nmadruga: It can be `string` | `undefined` | `Null` => I'm just updating to be either `string` or `undefined`

src/containers/Application/ApplicationPageWrapper.tsx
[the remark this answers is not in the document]
nmadruga: Done: 53c0803

nmadruga:
@CarlosNZ I have done quite a big refactor here, basically:
- Removed use of `SectionProgress` from components
- Renamed the type `TemplateSectionPayload` to `SectionDetails`
- Updated everywhere that was constructing or using the sections to use the new type (and renamed the variables)
- Renamed type `TemplateType` to `TemplateDetails` and everywhere using it (and renamed the variables)
- Create a new **helper** `updateSectionsProgress` to build an `SectionDetails` array from the received one adding the progress fields to it.
- Improved (on this new helper) the loop to retrieve promises and return the result (using a dispatch state hook) when finished.
- Now the `applicationPageWrapper` always set the progress for all sections and only use it (to display in the `ApplicationStart` component if the page is loaded with `application/<serial>`.

nmadruga:
I still haven't figured out the link not working though. I will have a look after on this too.

src/utils/hooks/useGetSectionProgress.tsx
nmadruga: Done: df2a4cf

src/components/Application/ApplicationStart.tsx
nmadruga: Finally!!! Done: c7bd60a

nmadruga:
Ok. Now I think it's all ready for a re-review. Sorry it has sooo many changes. I had to change a lot of files to change the types from TemplateSectionPayload to SectionDetails and since I was doing that I tried to make it tidier.

But you might need to have another look in many files, so no rush.

CarlosNZ (changes requested):
Heya, code looks good -- tidier and easier to follow, so thanks for doing that.

However, I'm getting a crash when I try to load the start page of an in-progress application.

To reproduce:
- Start an application
- Go back to the homepage
- Go to the application start page either via URL or List.

I get a white screen, with lots of console errors. :(

CarlosNZ:
> However, I'm getting a crash when I try to load the start page of an in-progress application.

Possibly to do with back-end schema changes in latest `master`?

nmadruga:
Thanks @CarlosNZ I'll have a look now

nmadruga:
I have ran the generated file and all seems good to me. Please try again after logout/login to refresh the cached queries/data

CarlosNZ:
> I have ran the generated file and all seems good to me. Please try again after logout/login to refresh the cached queries/data

Great, thanks. Working nicely now. Must have been and out-dated graphql.tsx at my end.

CarlosNZ (approved):
Great, working nicely now and all looks good.

Now that this is more well-established, it'd be good to get some internal docs regarding the application start/load process etc.and what all the different variables mean (and what they're used for).

PR #317, opened 2021-02-09
#203 Review start page improvements

### LINKED ISSUE #203: Update Review start page based on status
If review exists and status = 'Changes Require' or 'Review Pending' then show continue review. Before moving to the `ReviewPageWrapper` we can do the duplication of responses and change of status of review to draft for this one

And if review is submitted show `View`.

nmadruga:
Fix #203

~~Branched off PR #306, please merge it before this one.~~ DONE

### Changes
- Check review status is DRAFT in `ReviewPageWrapper` and `ReviewSection` to display options to edit/submit
- Update `ReviewOverview` (Start page) to display:
  - The review status (during development - we can remove for production)
  - Button with label related to possible actions that reviewer can do relate to review (described in [comment](https://github.com/openmsupply/application-manager-web-app/issues/203#issuecomment-773755093))

### Unrelated changes
- Fix problem in `getReviewAssignment.query` to only get `reviewResponses` associated with queried `applicationId`
- Fix and return the `reviewStatus` from `useLoadReview` which was returning undefined before
- Make a refetchQuery for the current `reviewAssignment` to be updated after a new review is created
- Not sure why neeed to create a new enum REVIEW_STATUS - not always needed. Please help : )


src/utils/data/reviewStatus.ts
```diff
@@ -0,0 +1,7 @@
+export enum REVIEW_STATUS {
+  DRAFT = 'Draft',
+  PENDING = 'Pending',
+  CHANGES_REQUESTED = 'ChangesRequested',
```
CarlosNZ, line 4:
```suggestion
  CHANGES_REQUESTED = 'Changes Requested',
```

src/utils/hooks/useGetReviewAssignment.tsx
```diff
@@ -15,8 +16,8 @@ interface UseGetReviewAssignmentProps {
 }
 
 const useGetReviewAssignment = ({ reviewerId, serialNumber }: UseGetReviewAssignmentProps) => {
-  const [assignment, setAssignment] = useState<AssignmentDetails | undefined>()
-  const [assignedSections, setAssignedSections] = useState<string[] | undefined>()
+  const [assignment, setAssignment] = useState<AssignmentDetails>()
+  const [sectionsAssigned, setSectionsAssigned] = useState<SectionDetails[]>()
```
CarlosNZ, line 20: Would you mind adding a short "TO-DO" down around line 50 about handling the case where might have multiple assignments (for multiple levels) as we discussed?

CarlosNZ (approved):
Great, working really well now. 

Just one comment about what you posted:

>     * Not sure why neeed to create a new enum REVIEW_STATUS - not always needed. Please help : )

Not sure what you mean by "not always needed". Can you clarify the problem here?

nmadruga:
> Not sure what you mean by "not always needed". Can you clarify the problem here?

Some places it's ok to compare with the `ReviewStatus` enum from Graphql that has each status in **lower case**, and in another place (in this PR) it needed to be compared with **upper case** and snake case, so there are 2 ways of comparing review status currently, which is not good. But I couldn't figure out why is that so.

Issue #208, opened 2020-12-11
Search applications


nmadruga:
Implement search box for applications

Some restrictions on the search - initially as you type...
Later - debounce: should start the query after a while between the time you type letters

PR #319, opened 2021-02-10
#208 Search application list

CarlosNZ:
Implement #208.

Querying the back-end via search query in URL was already implemented, this just adds a UI search box which updates the URL as you type.

~~Had to expose a little more `useRouter` functionality to the calling component, as we need to handle conversion of the queries back into a kebab-case URL.~~
**UPDATE**: I have simplified this quite a bit in my next PR (Sorting columns #322), so this useRouter messiness will be gone. You may wish to set the base to master on #322 and just review that, since Search and Sort end up sharing quite a bit.

### DIFF
--- a/src/containers/List/ListWrapper.tsx
+++ b/src/containers/List/ListWrapper.tsx
@@ -81,14 +81,14 @@ const ListWrapper: React.FC = () => {
             <Grid.Column width={3}>
               <Search
                 // size="large"
-                placeholder="Search Applications"
+                placeholder={strings.PLACEHOLDER_SEARCH}
                 onSearchChange={handleSearchChange}
                 open={false}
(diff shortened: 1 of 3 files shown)


nmadruga (approved):
Looking good!

I see that the **Clear search** is getting on top of the search box on responsive, but that can be sorted on the App layout epic.

CarlosNZ:
> I see that the **Clear search** is getting on top of the search box on responsive, but that can be sorted on the App layout epic.

Ah yeah that's annoying. I think that's the problem with using `<Grid>` rather than flex-box, cos grid doesn't wrap. Hmmm...

CarlosNZ:
Merged #320 straight to master, so don't need to merge this specifically.

Issue #88, opened 2020-11-04
Element plugin: Text Information - missing SummaryView


nmadruga:

nmadruga:
@andreievg @CarlosNZ 

Should we add the option of using dynamic labels (using the Query syntax on this property) as part of this issue?

CarlosNZ:
> @andreievg @CarlosNZ
> 
> Should we add the option of using dynamic labels (using the Query syntax on this property) as part of this issue?

If we did it the way I was thinking it would work, then no. In my view, the "label" field (or any field) of the plugin should receive all these values pre-evaluated. But that's still up for discussion.

andreievg:
Should also show in summary/review pages

CarlosNZ:
I have a more general question relating to this that it'd be good to get some agreement on before addressing this issue:

It seems to me we’re not being entirely consistent with how we handle the non-question "INFORMATION" elements. I think there’s two basic approaches, and we should probably pick one and stick to it:

1. Always keep them in: responses get made for them, copies are made during review, reviewers are assigned to them, etc., so they exist in the same way as genuine “question” responses. And then we just filter them out in the UI as appropriate.
2. Get rid of them right at the start (i.e. don’t make application_responses for them at all), then we never have to deal with them thereafter.

I think I’m leaning towards (1) as I think there are situations where we need them (i.e. Summary View), and it makes keeping tracking of whether applications are "fully assigned" or not simpler (doesn't it?) but would be good to get some clarity around it since I think we’re starting to get a little slippery with how we handle them.

andreievg:
As per our telegram discussion:

Andrei
```
Just want to point out a difference:

assignments -> get all template elements 

review -> only review application responses that are assigned with assignments (through template elements)

consolidation -> mainly cares about review responses, but when figuring out if can fully approve takes into account application responses

thus, since information template element won't have application response. it should be oki
```

Carl
```
Right, so if there’s no application_responses created for Information elements, then we’re doing Option 2 -- get rid of them early. It looks like that’s what’s happening currently. So we’re all good?
```
So far, option 2 makes the most sense, but want to make sure we dont' miss anything, below:

> I think there are situations where we need them (i.e. Summary View) 

Can we just have the same presentation for summary view as application view (it doesn't need application response)

> And it makes keeping tracking of whether applications are "fully assigned" or not simpler (doesn't it?)

Yep, I think that works with both options (as it relies on template_elements).

[...]

PR #321, opened 2021-02-11
#88 Text info summary (and Default Summary View improvements)

CarlosNZ:
Text info now has a Summary View and it appears on the Summary and Review pages.

Had a problem getting custom Summary Views to work, and after discussing and resolving with @andreievg the best solution was to ensure each plugin has its own SummaryView.tsx file. However we can still utilise the "DefaultSummaryView" component by passing it down to each plugin and it just returns the DefaultSummary component.

Something to think about: Now info-type elements always appear in Summary and Review pages. Would we *always* want this? Do you think we need another boolean field in `template_element` called `show_in_summary`, which would default to `true` but can be turned off sometimes.


src/components/Review/ReviewSection.tsx
```diff
@@ -88,6 +88,7 @@ const ReviewSection: React.FC<ReviewSectionProps> = ({
         </Accordion.Title>
         <Accordion.Content active={isOpen}>
           {Object.entries(pages).map(([pageName, elements]) => {
+            console.log('elements', elements)
```
nmadruga, line 91: Please remove console

src/components/Review/ReviewSection.tsx
```diff
@@ -104,27 +105,53 @@ const ReviewSection: React.FC<ReviewSectionProps> = ({
...
+                                        color="blue"
+                                        style={{ minWidth: 100 }}
+                                        onClick={() => setDecisionArea(review, summaryViewProps)}
+                                      />
+                                    </Grid.Column>
+                                  </Grid.Row>
+                                </Grid>
+                              </Card.Content>
+                            </Card>
                           </Grid.Row>
```
nmadruga, line 155: Not sure why is showing all this diff here. Perhaps there is another PR included in this one too?

nmadruga (approved):
LGTM

src/components/Review/ReviewSection.tsx
nmadruga: There is actually a compilation error happening due to this change :(

nmadruga: I think this is a conflict resolution mistake. It needs to be removed before merging!

CarlosNZ: Weird. The diff I can explain, not so much the compilation error. The reason for the diff is that this whole block used to be within `if (category === TemplateElementCategory.Question) {`, but I removed that as we now show all, not just question types, hence the indentation for the whole block changed.

Have fixed the error now. If you look at the changes with "Hide whitespace changes" selected, you'll see the diff is just the couple of lines at each of the block.

nmadruga (approved):
Great. Thanks!

Issue #209, opened 2020-12-11
Sort applications


nmadruga:

nmadruga:
See related comment: https://github.com/openmsupply/application-manager-web-app/pull/265#discussion_r559304061

CarlosNZ:
Can this be closed now?

nmadruga:
No. Even though you already added the logic for columns to be sorted and UR to return the list sorted we still need to implement it on the UI

CarlosNZ:
Duh, of course. Thx

CarlosNZ:
Just want to clarify a couple of things before I get into this:

- Even though the query can handle sorting by multiple columns, the UI will only allow single-column sort. You click on one, it sorts by that, you click on another, it sorts by that instead. Is that correct?
- The actual sorting is done by the database and returned sorted, so the UI elements here are just for indication and only update the URL, right?

PR #322, opened 2021-02-11
#209 Sort applications in list

CarlosNZ:
Implements #209 -- basic sorting functionality for Application List.

Branched from #319 -- please review that first.

## Functionality

Click a column header to sort by that column (descending). Click again to sort ascending. Click once more to remove sort restrictions.

Logic for managing the sort query state values and syncing with URL is done in "ListWrapper" component. Relevant props are passed down to "Application List" for display.

## Related changes

- Added a generic `updateQuery` method in useRouter to push a new URL based on additional query parameters (passed in as key:value object). This tidies up a lot of the Router messiness introduced in #319.
- new generic method `replaceObjectKeys`, which combines the common functionality of `replaceKebabCaseKeys` and `restoreKebabCaseKeys`, with replacement function passed in as parameter

## Unrelated changes

- Moved all List Cell components into "Cells" subfolder within "List" folder

## Additional notes

- Some columns are not yet mapped to a "sort" field in the GraphQL query. ~~If you click them, the app will crash. Needs to be addressed at some point.~~ (Fixed f4fcd6e5)
- There is still an occasional problem when the list doesn't load and gives the "Problem loading" message, often requiring a re-load to get it working again. I don't *think* it's due to anything I've done here, but we probably need to investigate at some point. (This should be fixed in my next PR for pagination)


nmadruga (commented):
The code is great, thanks for doing that. It's neat!

I think I found one problem with sorting by date. Just clicking on the "Last active date" just makes the list disappear. Something wrong with the generate URL: http://localhost:3000/applications?type=ReviewTest&user-role=applicant&sort-by=last-active-date

CarlosNZ:
> I think I found one problem with sorting by date. Just clicking on the "Last active date" just makes the list disappear.

Yup, this one I just got the mapping string wrong. However, there's also an issue with the "Deadline date" in that it doesn't actually correspond to a field in the GraphQL query, so we don't have a mapping for that one yet, so it was crashing too. I've just added a couple of safeguards for this:
 - if the 'sortName' doesn't match on of the ones in `mapSortFields` (buildQueryVariables) then it will sort by "NATURAL" (this handles the case when a URL is manually entered with an unknown sort field in the query)
 - if the 'sortName' is empty string (which I've made it for "Deadline Date"), the column is unsortable (i.e. won't respond to a click on the header)

nmadruga (approved):
Yep, it was written in the description, sorry. But that's better with the safeguard now. Thanks!

Issue #210, opened 2020-12-11
Use pagination to list applications


nmadruga:
Before starting please describe the approach.

For now just create ` Previous`  and `Next` buttons that would show only when there is a previous or next page.

Pagination size - as one of the filters (URL) with a default value.

PR #324, opened 2021-02-14
#210 Pagination in list (plus misc fixes)

CarlosNZ:
Branched from #322 -- please review that one first.

Implements basic pagination for list using Semantic's built-in [Pagination](https://react.semantic-ui.com/addons/pagination/) component. Suggest quickly making at least 6-7 new draft applications so you can see it in action.

The pagination component is basically using default options, so we can get some input from Deb as to what looks best (it'll be super-easy to change the parameters) at some point.

Whilst doing this I made a few additional  related fixes/improvements to the List functionality:

- Changed other places that update the URL to use the new `updateQuery` method of useRouter
- Bug fix for the list sometimes not loading properly (useEffects running on incorrect changes and trying to load things that weren't available yet) Also, the "Loading" value wasn't being set back to `true` when re-loading the application data due to sorting/pagination, etc, so have changed `useListApplications` to just return Apollo's "loading" property rather than a separate useState value. I assume this won't cause any other problems, @nmadruga ? (Maybe we need seperate "loading" (for initial page load) and "tableLoading" values (for sort/search/pagination reloads) ? )
- moved the "Loading" component to only replace the Table component -- so the whole page doesn't show "Loading" when changing sort/pagination settings. I'd like to get it to only replace the table rows and have the headers remain, but figuring out how to replace part of the table was getting a bit to complicated for this issue (make a new one?)
- fix for query error when "filter" object was empty (had to send a "dummy" filter -- GraphQL is super-annoying like this 🙄)


nmadruga:
Hello starting on this one now : )

> just return Apollo's "loading" property rather than a separate useState value. I assume this won't cause any other problems, @nmadruga ?

Cool, that makes sense

> Maybe we need seperate "loading" (for initial page load) and "tableLoading" values (for sort/search/pagination reloads) ?

Not sure, the first one would be to display the main component and the other one just to show the loading inside the table?
I'm thinking it's not very nice that the table doesn't stay on the page when changing pages, maybe that's what you meant by this: 

> I'd like to get it [Table component] to only replace the table rows and have the headers remain, but figuring out how to replace part of the table was getting a bit to complicated for this issue (make a new one?)

I think that would be good. And probably something to add to the next sprint so we have the list all sorted by then.

nmadruga (commented):
Very good. The Pagination component looks great. I have had a play with it and the only outstanding this to me was: when searching for names it only counts the one in the current page. Which I would think it would be searching in all the list.

nmadruga:
Sorry, had to open a new PR for this one after pushing wrong merge to it. @CarlosNZ please have a look on the search + pagination in the other one, I also past my comment done in here to be reflected in the new PR.

Closing this one...

Issue #325, opened 2021-02-14
Update Dropdown plugin to allow searching in options list


CarlosNZ:
Should be a simple tweak to existing Drop-down plugin.

Add an additional parameter "searchable", which sets whether this option is shown or not.

PR #326, opened 2021-02-15
#325 Add search capability to Dropdown plugin

CarlosNZ:
Very small tweak to make this work. Have added documentation for the new parameter `search`.

To see it working, use back-end branch from [PR #217](https://github.com/openmsupply/application-manager-server/pull/217) -- on page 2 of the Feature Showcase application, the "API lookup" element has the searchable parameter enabled.

Thinking that this is probable enough for demonstration purposes. Rather than have a strict "lookup" element, we can just add (for example) ATC codes and their long-form string together in a list of options, and the user can easily search by either ATC code or the full name. (To be clear, I mean each option is a string containing both the ATC code and the full name, i.e. "N02AJ06 | codeine and paracetamol" )


nmadruga (approved):
Very nice to have this one added. 👍

PR #328, opened 2021-02-17
#210 pagination in list

nmadruga:
~~Branched from #322 -- please review that one first.~~ Merged

Implements basic pagination for list using Semantic's built-in [Pagination](https://react.semantic-ui.com/addons/pagination/) component. Suggest quickly making at least 6-7 new draft applications so you can see it in action.

The pagination component is basically using default options, so we can get some input from Deb as to what looks best (it'll be super-easy to change the parameters) at some point.

Whilst doing this I made a few additional  related fixes/improvements to the List functionality:

- Changed other places that update the URL to use the new `updateQuery` method of useRouter
- Bug fix for the list sometimes not loading properly (useEffects running on incorrect changes and trying to load things that weren't available yet) Also, the "Loading" value wasn't being set back to `true` when re-loading the application data due to sorting/pagination, etc, so have changed `useListApplications` to just return Apollo's "loading" property rather than a separate useState value. I assume this won't cause any other problems, @nmadruga ? (Maybe we need seperate "loading" (for initial page load) and "tableLoading" values (for sort/search/pagination reloads) ? )
- moved the "Loading" component to only replace the Table component -- so the whole page doesn't show "Loading" when changing sort/pagination settings. ~~I'd like to get it to only replace the table rows and have the headers remain, but figuring out how to replace part of the table was getting a bit to complicated for this issue (make a new one?)~~ DONE!
- fix for query error when "filter" object was empty (had to send a "dummy" filter -- GraphQL is super-annoying like this 🙄)


nmadruga:
Hello starting on this one now : )

> just return Apollo's "loading" property rather than a separate useState value. I assume this won't cause any other problems, @nmadruga ?

Cool, that makes sense

> Maybe we need seperate "loading" (for initial page load) and "tableLoading" values (for sort/search/pagination reloads) ?

Not sure, the first one would be to display the main component and the other one just to show the loading inside the table?
I'm thinking it's not very nice that the table doesn't stay on the page when changing pages, maybe that's what you meant by this: 

> I'd like to get it [Table component] to only replace the table rows and have the headers remain, but figuring out how to replace part of the table was getting a bit to complicated for this issue (make a new one?)

I think that would be good. And probably something to add to the next sprint so we have the list all sorted by then.

CarlosNZ:
@nmadruga's comment from [previous PR version](#324):

> Very good. The Pagination component looks great. I have had a play with it and the only outstanding this to me was: when searching for names it only counts the one in the current page. Which I would think it would be searching in all the list.

CarlosNZ:
> Very good. The Pagination component looks great. I have had a play with it and the only outstanding this to me was: when searching for names it only counts the one in the current page. Which I would think it would be searching in all the list.

I see what's happening here. It is searching the whole list, but if you're on a later page, when you type something in search the page selection in the URL doesn't update, so you might be requesting a page higher than there are available. i.e. if you had 5 pages and your on page 4 or 5, if your search now only has 3 pages of results, it'll show "no results" cos it's still requesting page 4.

I'll add an additional check to insist that the requested page can't be higher than the total pages, and if so, set it to the last.

CarlosNZ:
> I'll add an additional check to insist that the requested page can't be higher than the total pages, and if so, set it to the last.

Okay, that's done in dc962ac2

nmadruga (commented):
All good. I had another run yesterday and today.

CarlosNZ:
Hey @nmadruga -- I managed to get the table-only re-loading working nicely now, so won't need to do another issue for it. I reckon this  and general list stability is pretty solid in this PR now.

nmadruga (commented):
All good! It looks very nice now that the list doesn't get recreated.

Approved (although I can't actually approve it after having re-created this PR)!

CarlosNZ:
> All good! It looks very nice now that the list doesn't get recreated.
> 
> Approved (although I can't actually approve it after having re-created this PR)!

Cool -- I'll merge it now :)

CarlosNZ (approved):

Issue #314, opened 2021-02-05
Duplicate application_responses when user clicks Make changes on Application Start page


nmadruga:

andreievg:
Logic should be pretty much the same as creating a new application (create responses for all reviewable questions, but also duplicate existing ones)

Issue #101, opened 2020-11-06
Epic #21 - Notifications


nmadruga:
### Add new issues:
- [ ] Send notification to user for Application submitted, withdrawn, approved, rejected, expired or required changes.

PR #352, opened 2021-02-19
#314 Create application wrapper router

### LINKED ISSUE #341: Create ApplicationWrapper (container) with middleware router for Applications

nmadruga:
Fix #341

New route `application-new` to point to `ApplicationWrapper` nested route and link to new pages re-structured in [Epic#21](https://github.com/openmsupply/application-manager-web-app/issues/101). 
Added some local components to each page - to be replaced by proper ones (in new files) on next PRs.

### DIFF
--- a/src/containers/Application/ApplicationWrapper.tsx
+++ b/src/containers/Application/ApplicationWrapper.tsx
@@ -7,19 +7,19 @@ import { NoMatch } from '../../components'
 const ApplicationWrapper: React.FC = () => {
   return (
     <Switch>
-      <Route exact path="/application_new/new">
+      <Route exact path="/applicationNEW/new">
         <ApplicationCreateNew />
       </Route>
-      <Route exact path="/application_new/:serialNumber">
+      <Route exact path="/applicationNEW/:serialNumber">
         <ApplicationStartNew />
       </Route>
-      <Route exact path="/application_new/:serialNumber/:sectionCode/Page:page">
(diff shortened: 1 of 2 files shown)


CarlosNZ (changes requested):
Doesn't seem to be working -- I keep getting the "Page not found" message.

I think it's because in the SiteLayout the route it `/application-new` but in the Wrapper the routes start with `/application_new` (with an underscore), so none of them will ever resolve.

Also -- just a suggestion, but I wonder if we should put the NEW in All Caps so it stands out and doesn't get mixed up with (say) `/new`. It makes it clear that it's a temp name that needs to be replaced. So the route could be `/applicationNEW` and the components could be like "ApplicationPageNEW", something like that. So the name is whatever it will be plus a "NEW" butted on the end. What do you think?

nmadruga:
> I think it's because in the SiteLayout the route it /application-new but in the Wrapper the routes start with /application_new (with an underscore), so none of them will ever resolve.

Oh, no. Last minute change that ended half-way.

> Also -- just a suggestion, but I wonder if we should put the NEW in All Caps so it stands out

Indeed. Nice suggestion, I'll change to that.

CarlosNZ (approved):
Sweet -- LGTM now :)

andreievg:
Looks good, would prefer use of `path` from `useRouteMatch`, as per
https://reactrouter.com/web/example/nesting

PR #358, opened 2021-02-22
#343 current status to FullStructure

### LINKED ISSUE #343: Add status (application info) in useLoadApplicationNew to FullStructure after trigger
Status needs to be refreshed when the status changes (Ex: after submission)

nmadruga:
Fixes #343 

Branched of #353 please merge it first.

### Changes
- Use `refetch` on `useGetApplication` query to rerun the query when trigger is still processing.
- Change `getApplicationNEW.query` to also return **trigger** and **status** results for the application. The status and stage (with other important fields of stage) are now added to the returned FullStructure.
- Flag `notifyOnNetworkStatusChange` to correctly update the `loading` returned on refetch.
- Set MAX_REFETCH to 10 and `refetchAttempts` state to keep how many refetches have run.
- On useEffect will first check that application is found and then check trigger, if applicationTrigger returns `null` can load the status for the application. Otherwise will set a timeout of 500 milliseconds and do a refetch.

[EDITED 23/02]
- Change the `ApplicationDetails` type to have the field `current` of type **StageAndStatus** that holds only the information required to display Application/Review - and on the List of Applications.
- Had to update a few of the old components/containers to also refer to `status` through `current` field
- Update **fragment** Stage to return only required fields for type **StageAndStatus** on the query of `getApplicationNEW.query`


src/utils/graphql/queries/getApplicationNEW.query.ts
```diff
@@ -26,5 +26,17 @@ export default gql`
         }
       }
     }
+    applicationTriggerStates(condition: { serial: $serial }, first: 1) {
```
andreievg, line 29: would this not be just the 'trigger' field on application ?

src/utils/hooks/useLoadApplicationNEW.ts
```diff
@@ -13,34 +13,56 @@ import { buildSectionsStructure } from '../helpers/structure/buildSectionsStruct
...
+    if (!application || !application.template) setStructureError('No application/template found')
+    if (!data.applicationStageStatusAlls?.nodes) setStructureError('No status found')
+
+    if (structureError) {
       setIsLoading(false)
       return
     }
 
+    // Checking if trigger is running before loading current status
+    if (data.applicationTriggerStates?.nodes) {
```
andreievg, line 54: As per earlier comment, i think just fetching the trigger from application and checking that it is null would be simpler way to handle this

src/utils/hooks/useLoadApplicationNEW.ts
```diff
@@ -13,34 +13,56 @@ import { buildSectionsStructure } from '../helpers/structure/buildSectionsStruct
...
+      const triggerResult = data.applicationTriggerStates?.nodes
+      if (!triggerResult.every((result) => result === null) && refetchAttempts < MAX_REFETCH) {
```
andreievg, line 56: since MAX_REFETCH is being implemented.

* if max refetch is reached then should check it's been reached and set error with return 
*  after successfully getting application with trigger == NULL set refetch attempts to 0

src/utils/hooks/useLoadApplicationNEW.ts
```diff
@@ -64,13 +86,24 @@ const useLoadApplication = ({ serialNumber, networkFetch }: UseGetApplicationPro
       submissionMessage: application.template?.submissionMessage as string,
     }
 
+    const stages = data.applicationStageStatusAlls?.nodes as ApplicationStageStatusAll[]
```
andreievg, line 89: Are we using this anywhere in UI ? I was thinking maybe we would do a seperate query for this when accessing history ?

andreievg (changes requested):
A few minor changes

src/utils/graphql/queries/getApplicationNEW.query.ts
nmadruga: Yes, that's right, now that we aren't using a single hook to check that we can simply use the field from the entity. Good point. When this is all finished we should remove the `applicationTriggerStates` view: New Issue [Back-end#225](https://github.com/openmsupply/application-manager-server/issues/225)

src/utils/hooks/useLoadApplicationNEW.ts
nmadruga: Done: 655c08a & fc37e47

src/utils/hooks/useLoadApplicationNEW.ts
nmadruga: Done: 13b7b1c

src/utils/hooks/useLoadApplicationNEW.ts
nmadruga: We do need to get `stageID` and `stageNumber` to be used on the filter for applications list (that uses the same type `ApplicationDetails`) and when we display the Review page we need to know the stage title:
Screen Shot 2021-02-23 at 10 49 30 AM

 or the application Submission page the list of stages - which could be a second hook indeed:
Screen Shot 2021-02-23 at 10 51 16 AM

andreievg: Sure, looks like it's needed in summary, great.

Although current stage (for ApplicationPage and list), is available through application 

Screen Shot 2021-02-23 at 11 04 09 AM

src/utils/hooks/useLoadApplicationNEW.ts
```diff
@@ -13,34 +13,60 @@ import { buildSectionsStructure } from '../helpers/structure/buildSectionsStruct
...
+    // Checking if trigger is running before loading current status
+    if (application.trigger === null) {
+      setRefetchAttempts(0)
+    } else {
+      if (refetchAttempts < MAX_REFETCH) {
+        setTimeout(() => {
+          console.log('Will refetch loadApplication', refetchAttempts) // TODO: Remove log
+          setRefetchAttempts(refetchAttempts + 1)
+          refetch()
+        }, 500)
```
andreievg, line 65: one more change please, setIsLoading = true somewhere here (to indicate that we are awaiting trigger even though loading = false (i.e. when submission button was pressed)

andreievg (changes requested):
One more little change please

src/utils/hooks/useLoadApplicationNEW.ts
andreievg: or maybe this should be just before  ` if (!application.template) `{ ?

And I'v just noticed, 

```
f (!application) {
      setIsLoading(false)
      return
    }
```

Should have a setStructureError ?

nmadruga: It is already set to true when the hook starts (isLoading)

nmadruga:
Have done another change - which touched a few files on existing ApplicationPage...

src/utils/hooks/useLoadApplication.tsx
```diff
@@ -112,11 +113,13 @@ const useLoadApplication = ({ serialNumber, networkFetch }: UseGetApplicationPro
...
-          name: stage as string,
-          status: status as string,
-          date: statusHistoryTimeCreated.split('T')[0],
+        current: {
+          stage: {
+            id: stageId as number,
+            name: stage as string,
+          },
+          status: status as ApplicationStatus,
+          date: statusHistoryTimeCreated.split('T')[0] as Date,
```
CarlosNZ, line 122: Can we use the DateTime library (Luxon) to handle this rather than string manipulation? The expression will just be something like:
```date: DateTime.fromISO(statusHistoryTimeCreated)```
(I've used it in `buildQueryFilters` if you want to check the syntax)

File: src/utils/hooks/useLoadApplicationNEW.ts [hunk 7]
```diff
@@ -64,13 +90,25 @@ const useLoadApplication = ({ serialNumber, networkFetch }: UseGetApplicationPro
...
       serial: application.serial as string,
       name: application.name as string,
       outcome: application.outcome as string,
+      current: {
+        stage: {
+          id: stageId as number,
+          name: stage as string,
+        },
+        status: status as ApplicationStatus,
+        date: statusHistoryTimeCreated.split('T')[0] as Date,
```
CarlosNZ, line 110: Same here date handling.

src/utils/hooks/useLoadApplicationNEW.ts
```diff
@@ -98,11 +136,11 @@ const useLoadApplication = ({ serialNumber, networkFetch }: UseGetApplicationPro
       sections: buildSectionsStructure({ sections, baseElements }),
     })
     setIsLoading(false)
-  }, [data])
+  }, [data, loading])
```
CarlosNZ, line 139: Is it totally necessary to monitor `loading` here? Isn't a change in `data` enough?  (If so, I'll need to change it on the other hook too)

CarlosNZ (commented):
Cool, just a couple of comments.

src/utils/hooks/useLoadApplication.tsx
CarlosNZ: Actually, why are we restricting it to just a date here, as opposed to the full timestamp? If we make it a DateTime object (`DateTime.fromISO(statusHistoryTimeCreated`), then we can just use the DateTime library to display it in whatever format we want when we need it.

src/utils/hooks/useLoadApplicationNEW.ts
andreievg: I am think of a use case when hook is re-run due to changes to application (when submitting)

src/utils/hooks/useLoadApplicationNEW.ts
andreievg: As per comment in telegram:

should monitor loading, error and data.

in use effect first check loading
then error
then !data (just in case) <- technically shouldn't need to

src/utils/hooks/useLoadApplication.tsx
nmadruga: Sure. I don't think was a good thing to be removing the timestamp from it as well. I'll change to what you suggested now

src/utils/hooks/useLoadApplicationNEW.ts
nmadruga: Done: caa6b13

src/utils/hooks/useLoadApplicationNEW.ts
nmadruga: That is required to run the useEffect again ofter the second try of refetch - the `data` doesn't change because the result was the same, but we still need to set a second refetch to try and get the correct status of the application. So the way to force the useEffect on running again was to watch the loading property - and only continue if `loading` === `false`

src/utils/hooks/useLoadApplicationNEW.ts
nmadruga: Oh, we need extra stuff. For application list we will also need the `statusHistoryTimeCreated` field which isn't an generated column on the application entity - that's when we decided to use the VIEW `application_stage_status_latest` so we didn't have to continue to map other stage/status attributes to it.

src/utils/hooks/useLoadApplicationNEW.ts
nmadruga: I see! I was trying to understand the use case. I'll add as part of the statement that checks trigger - if it needs to re-run will be set to true. Does that sound good?

nmadruga: Done: 2c4c3fe

CarlosNZ (approved):
LGTM :)

andreievg (approved):
Thanks for changes! Ready to merge : )

PR #359, opened 2021-02-22
#344 useGetFullApplication -- hook basics

CarlosNZ:
Implements #344  -- builds full structure and evaluates visibility/required/editable expressions. Haven't added any validation at this point -- check the ApplicationFlow diagram to see what I've implemented (shaded in green)

## Notes:

- Skeleton "ApplicationHome" component (the "start" page) which just calls the hook and prints the structure to the console
- The evaluation updates the structure elements In-place so we don't have to re-build the structure from flattened elements. Hopefully this doesn't cause problems
- Have updated types and structure layout for better consistency:
   - All expression fields have `..Expression` in the name, and when evaluated get written to the same name without expression (i.e. the field `isVisibleExpression` gets saved to `isVisible` after evaluation. This naming and functionality was inconsistent before, so this makes it easier to understand. Also, we shouldn't use the same field for both the unevaluated expression and the evaluated primitive.
   - there is now ElementBaseNEW,  ElementStateNEW, TemplateElementStateNEW and PageElement types to replace/update the existing ones based on changes
- can't currently test that the hook re-evalutes correctly when responses change, so please check that the logic is correct. The `useEffect` hook is currently only watching "data" -- is this correct? 

### Additional
- If anyone can tell me why "network-only" fetchPolicy is throwing an error (but "no-cache" is fine), and it works fine in `useLoadApplicationNEW` hook, I would be very pleased


src/utils/hooks/useGetFullApplicationStructure.tsx
```diff
@@ -0,0 +1,117 @@
...
+    const promiseArray: Promise<ElementStateNEW>[] = []
+    elements.forEach((element) => {
+      promiseArray.push(evaluateSingleElement(element))
+    })
+    return await Promise.all(promiseArray)
+  }
+
+  async function evaluateSingleElement(element: TemplateElementStateNEW): Promise<ElementStateNEW> {
+    const evaluationParameters = {
+      objects: { responses: responsesByCode, currentUser },
```
andreievg, line 87: hmm responseByCode will only be set after this will run.

Would advise extracting this async functions outside of the hook, and pass through `{ responses: responsesByCode, currentUser }` to them

src/utils/hooks/useGetFullApplicationStructure.tsx
```diff
@@ -0,0 +1,117 @@
...
+      // TO-DO: Also send org objects etc.
+      // graphQLConnection: TO-DO
+    }
+    const isEditable = evaluateExpression(element.isEditableExpression, evaluationParameters)
```
andreievg, line 91: I've had some errors (for say application 101 and 102), when they are thrown the whole thing seems to fail, i recommend adding wrapper around evaluation expression with default values

andreievg (changes requested):
Mostly looks good. I had some errors while testing, please see comments below.

Would advise doing something similar to this: https://github.com/openmsupply/application-manager-web-app/compare/%23344-useGetFullApplication-hook-etc...344-suggested-fixes-1

(it's just an example, but works regardless)

src/utils/hooks/useGetFullApplicationStructure.tsx
```diff
@@ -0,0 +1,162 @@
...
+  async function evaluateElements(elements: TemplateElementStateNEW[], evaluationParameters: any) {
+    const promiseArray: Promise<ElementStateNEW>[] = []
+    elements.forEach((element) => {
+      promiseArray.push(evaluateSingleElement(element, evaluationParameters))
+    })
+    return await Promise.all(promiseArray)
+  }
+
+  const evaluateExpressionWithFallBack = (
+    expression: any,
```
andreievg, line 102: any - > IQueryNode

andreievg (changes requested):
Looks good, just gotta do one more check

src/utils/hooks/useGetFullApplicationStructure.tsx
CarlosNZ: Okay, have passed the `responsesObject` into evaluateElements now, so will be getting the correct, up-to-date object. 1f024c5

src/utils/hooks/useGetFullApplicationStructure.tsx
CarlosNZ: 1f024c5

src/utils/hooks/useGetFullApplicationStructure.tsx
CarlosNZ: 👍 0b2bd9c

src/utils/hooks/useLoadApplicationNEW.ts
```diff
@@ -82,13 +82,19 @@ const useLoadApplication = ({ serialNumber, networkFetch }: UseGetApplicationPro
...
           baseElements.push({
-            ...element,
+            // ...element,
             pluginCode: element.elementTypePluginCode,
             sectionIndex: sectionNode?.templateSection?.index,
             sectionCode: sectionNode?.templateSection?.code,
             elementIndex: element.index,
             page: pageCount,
-          } as TemplateElementState)
+            isEditableExpression: element.isEditable,
```
andreievg, line 91: looks like we need `code: element.code` (otherwise responses are not linked)

andreievg (changes requested):
One more please

src/utils/hooks/useLoadApplicationNEW.ts
CarlosNZ: Ah yeah, good catch. I was just eyeballing the previous structure output to check all the fields -- must have missed that one!

d682ddf

src/utils/graphql/queries/getAllResponses.query.ts
```diff
@@ -0,0 +1,13 @@
+import { gql } from '@apollo/client'
+
+export default gql`
+  query getAllResponses($serial: String!) {
+    applicationBySerial(serial: $serial) {
+      applicationResponses {
```
andreievg, line 6: fingers crossed, last thing.

Was testing with a couple of buttons that mutate response (that changes other response to visible/invisible). Was failing to reload + the network thing was failing as per:

> can't currently test that the hook re-evalutes correctly when responses change, so please check that the logic is correct. The useEffect hook is currently only watching "data" -- is this correct?

After adding id and serial (for application) here, the network-only policy worked and query hook was re-running and visibility validation was work : )

andreievg (changes requested):

src/utils/graphql/queries/getAllResponses.query.ts
CarlosNZ: Awesome, thanks 343ac7e

andreievg (approved):
Nice, thanks !

Issue #347, opened 2021-02-19
Revalidation Logic


nmadruga:
Implement **useEffectForValidation** in the `ApplicationPage` and the method **revalidateWithLoadingModal** (to be passed to Progress and Nextbutton)

PR #360, opened 2021-02-23
#346 Create ApplicationPage container

CarlosNZ:
This is a pretty bare-bones skeleton for the Page container. Will need to be updated as ProgressBar, Validation, Navigation components are implemented.

**NOTE**: I have branched from #359, but I've also merged in #358 (as I needed the Status), so probably need that one to be merged to master first, otherwise the diff here will include #358's changes. (Changes specific to this issue are mainly in "ApplicationWrapper" and "ApplicationPage")

Re-direction based on Status is working, and the FullStructure is loaded.

See ApplicationFlow diagram for what is implemented (coloured green). Let me know if there's any additional functionality that should go into this PR.

### DIFF
--- a/src/containers/Application/ApplicationPage.tsx
+++ b/src/containers/Application/ApplicationPage.tsx
@@ -59,11 +59,11 @@ const ApplicationPage: React.FC<ApplicationProps> = ({ structure }) => {
         }}
       >
         <Grid.Column width={4}>
-          <ProgressBar structure={structure} />
+          <ProgressBar structure={fullStructure as FullStructure} />
         </Grid.Column>
         <Grid.Column width={10} stretched>
           <Segment basic>
-            <PageElements structure={structure} responses={responsesByCode} />
+            <PageElements structure={fullStructure as FullStructure} responses={responsesByCode} />
             <NavigationBox />
           </Segment>
         </Grid.Column>
--- a/src/containers/Application/ApplicationPage.tsx
+++ b/src/containers/Application/ApplicationPage.tsx
@@ -1,40 +1,112 @@
-import React, { useEffect } from 'react'
-import { FullStructure } from '../../utils/types'
+import React, { useEffect, useState } from 'react'
+import { FullStructure, ResponsesByCode } from '../../utils/types'
 import useGetFullApplicationStructure from '../../utils/hooks/useGetFullApplicationStructure'
 import { ApplicationStatus } from '../../utils/generated/graphql'
+import { useApplicationState } from '../../contexts/ApplicationState'
+import { useUserState } from '../../contexts/UserState'
 import { useRouter } from '../../utils/hooks/useRouter'
+import { Loading, NoMatch } from '../../components'
+import strings from '../../utils/constants'
+import messages from '../../utils/messages'
+import {
+  Button,
+  ButtonProps,
+  Grid,
+  Header,
+  Message,
+  MessageProps,
+  ModalProps,
+  Segment,
+  Sticky,
+} from 'semantic-ui-react'
 
 interface ApplicationProps {
   structure: FullStructure
+  responses?: ResponsesByCode
 }
 
 const ApplicationPage: React.FC<ApplicationProps> = ({ structure }) => {
+  const [isStrictPage, setIsStrictPage] = useState(null)
   const { error, isLoading, fullStructure, responsesByCode } = useGetFullApplicationStructure(
     structure
   )
-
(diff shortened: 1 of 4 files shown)


src/containers/Application/ApplicationPage.tsx
```diff
@@ -0,0 +1,102 @@
...
+  const { push } = useRouter()
+
+  console.log('Structure', fullStructure)
+
+  useEffect(() => {
+    if (!structure) return
+
+    // Re-direct based on application status and progress
+    if (structure.info.current?.status === ApplicationStatus.ChangesRequired)
+      push(`/application/${structure.info.serial}`)
```
andreievg, line 35: should this be applicationNew for now

src/containers/Application/ApplicationPage.tsx
```diff
@@ -0,0 +1,102 @@
...
+    if (structure.info.current?.status !== ApplicationStatus.Draft)
+      push(`/application/${structure.info.serial}/summary`)
+
+    // TO-DO: Redirect based on Progress (wait till Progress calculation is done)
+  }, [structure])
+
+  if (error) return <NoMatch />
```
andreievg, line 42: I think Nicole had a different output if structure is undefined -> (404) and a way to display errors set in the hook. @nmadruga  ?

andreievg (commented):
Just a couple of comments, looks pretty good, going to branch from this to do revalidation method

src/containers/Application/ApplicationPage.tsx
CarlosNZ: Yes of course :) 

97405ba

src/containers/Application/ApplicationPage.tsx
CarlosNZ: Yup, got it.
```
if (error) return <Message error header={strings.ERROR_APPLICATION_PAGE} list={[error]} />
```

2bebe02

src/containers/Application/ApplicationPage.tsx
```diff
@@ -0,0 +1,102 @@
...
+      push(`/applicationNEW/${structure.info.serial}`)
+    if (structure.info.current?.status !== ApplicationStatus.Draft)
+      push(`/applicationNEW/${structure.info.serial}/summary`)
+
+    // TO-DO: Redirect based on Progress (wait till Progress calculation is done)
+  }, [structure])
+
+  if (error) return <Message error header={strings.ERROR_APPLICATION_PAGE} list={[error]} />
+  if (isLoading) return <Loading />
+
```
nmadruga, line 44: Shouldn't have a check for `fullStructure` not `undefined` here?

nmadruga: With that you wouldn't need to cast to FullStructure type on the render block

src/containers/Application/ApplicationPage.tsx
```diff
@@ -0,0 +1,102 @@
...
+        </Grid.Column>
+        <Grid.Column width={2} />
+      </Grid>
+      <Sticky
+        pushing
+        style={{ backgroundColor: 'white', boxShadow: ' 0px -5px 8px 0px rgba(0,0,0,0.1)' }}
+      >
+        <Segment basic textAlign="right">
+          <Button color="blue" onClick={() => {}}>
+            {/* TO-DO */}
```
nmadruga, line 78: I was thinking it would be included in this issue to add the implementation from `ApplicationPageWrapper` to also be able to submit the application... Should we create another issue to deal with this one?

nmadruga (commented):
Thanks. This looks good, I'll be checking the `useGetFullApplication` in more detail now.

I made a suggestion to create an issue for Submit button, which could be added to this issue or be on a new one.

src/containers/Application/ApplicationPage.tsx
CarlosNZ: Yes, have done that in the next PR #347

src/containers/Application/ApplicationPage.tsx
CarlosNZ: Yeah I think we need to do a new issue for that. It might be quite simple, but there's probably a few things that we need to be careful with, so would rather do that separately I think.

nmadruga: Added #365

nmadruga (approved):

andreievg (approved):

PR #361, opened 2021-02-23
#347 Validation in useGetFullStructure hook

CarlosNZ:
Implements #347 -- validation happens along with evaluation of visibility, etc. Haven't implemented any logic outside this hook -- @andreievg is working on that.

I'm not actually 100% clear on how it's handling undefined/null responses -- it should really also consider whether question is required and visible, will need to clarify that.

### DIFF
--- a/src/utils/hooks/useGetFullApplicationStructure.tsx
+++ b/src/utils/hooks/useGetFullApplicationStructure.tsx
@@ -100,6 +100,7 @@ const useGetFullApplicationStructure = ({
       })
       if (shouldProcessValidation || firstRunProcessValidation)
         setLastValidationTimestamp(Date.now())
+      generateProgressStructure(newStructure) // To-Do
       setFirstRunProcessValidation(false)
       setFullStructure(newStructure)
       setResponsesByCode(responseObject)
@@ -189,4 +190,10 @@ const flattenStructureElements = (structure: FullStructure) => {
     })
   })
   return flattened
+}
+
+const generateProgressStructure = (structure: FullStructure) => {
+  // TO-DO:
+  // Calculate all Progress objects and update structure in-place
+  // Shouldn't return anything
 }
--- a/src/utils/hooks/useGetFullApplicationStructure.tsx
+++ b/src/utils/hooks/useGetFullApplicationStructure.tsx
@@ -174,7 +174,7 @@ const useGetFullApplicationStructure = ({
             currentEvaluationParameters,
             false
           )
-        : new Promise(() => responseObject[element.code]?.isValid)
+        : async () => responseObject[element.code]?.isValid
     const results = await Promise.all([isEditable, isRequired, isVisible, isValid])
 
     const evaluatedElement = {
--- a/src/utils/hooks/useGetFullApplicationStructure.tsx
+++ b/src/utils/hooks/useGetFullApplicationStructure.tsx
@@ -98,6 +98,9 @@ const useGetFullApplicationStructure = ({
         flattenedElements[index].element = evaluatedElement
         flattenedElements[index].response = responseObject[evaluatedElement.code]
       })
+      if (shouldProcessValidation || firstRunProcessValidation)
+        setLastValidationTimestamp(Date.now())
+      setFirstRunProcessValidation(false)
       setFullStructure(newStructure)
       setResponsesByCode(responseObject)
       setIsLoading(false)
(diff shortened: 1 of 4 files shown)


src/utils/hooks/useGetFullApplicationStructure.tsx
```diff
@@ -134,13 +150,21 @@ const useGetFullApplicationStructure = (structure: FullStructure, firstRunValida
       evaluationParameters,
       true
     )
-    const results = await Promise.all([isEditable, isRequired, isVisible])
+    const isValid =
+      shouldProcessValidation || firstRunProcessValidation
+        ? evaluateExpressionWithFallBack(element.validationExpression, evaluationParameters, false)
+        : new Promise(() => responseObject[element.code]?.isValid)
```
andreievg, line 156: should this resolve ? i.e. `new Promise(resolve => resolve(responseObject[element.code]?.isValid))`

src/utils/hooks/useGetFullApplicationStructure.tsx
```diff
@@ -134,13 +150,21 @@ const useGetFullApplicationStructure = (structure: FullStructure, firstRunValida
...
+    setFirstRunProcessValidation(false)
+    const results = await Promise.all([isEditable, isRequired, isVisible, isValid])
+    if (shouldProcessValidation || firstRunProcessValidation) setLastValidationTimestamp(Date.now())
```
andreievg, line 159: just to be safe, maybe this should be down the bottom ?

src/utils/hooks/useGetFullApplicationStructure.tsx
andreievg, line 156: Btw pretty neat with resolving promise after `:`

andreievg (changes requested):
Looks good, just a couple of small requests

andreievg (changes requested):
return lastValidationTimestamp pls

andreievg:
Not part of this PR but ApplicationPage should probably check FullStructure exists instead of loading (since using is loading will cause the whole dom tree to change every time responses are refetched), or just use query hook loading

src/utils/hooks/useGetFullApplicationStructure.tsx
andreievg: actually this is updating on eery question evaluated

src/utils/hooks/useGetFullApplicationStructure.tsx
CarlosNZ: I was trying to make it so that it's the type/state as what `isValid` is in the first case -- i.e. unresolved Promise. All promises for these four variables are resolved in the subsequent `const results = await Promise.all([isEditable, isRequired, isVisible, isValid])` line. Basically, this just setting the value of `isValid` to the same as it previously was, but wrapped in a Promise so it's consistent with the value if its getting evaluated. Does that make sense?

src/utils/hooks/useGetFullApplicationStructure.tsx
CarlosNZ: Yeah true, have moved it (and `setFirstRunProcessValidation`) back up to the same place as the other `set...` calls.

0436e17

src/utils/hooks/useGetFullApplicationStructure.tsx
CarlosNZ: Yeah, as I mentioned above -- it's just a straight value, but wrapped in a Promise just so `isValid` gets handled the same whether or not it gets evaluated/

CarlosNZ:
> return lastValidationTimestamp pls

👍

> Not part of this PR but ApplicationPage should probably check FullStructure exists instead of isLoading (since using is loading will cause the whole dom tree to change every time responses are refetched), or just use query hook loading

Okay, so you mean: `if (!fullStructure) return <Loading />` ? 

Done fd3da68

src/utils/hooks/useGetFullApplicationStructure.tsx
andreievg: As our chat

andreievg (approved):

src/utils/hooks/useGetFullApplicationStructure.tsx
CarlosNZ: Cool, I updated to :
`async () => responseObject[element.code]?.isValid` as per suggestion. That's essentially what I was meaning in the first place. Cheers.

Issue #363, opened 2021-02-23
Show sections of Application in expanded area on the list


nmadruga:
When clicked should show sections with details as in design. The same (or very similar) to what is displayed in ApplicationHomePage and ReviewHomePage. Should do more layout changes so far - and use hook created in #385.

More refined reviews in sections on issues: #391, #392, #393, #394 & #395.

PR #364, opened 2021-02-23
#211 expand rows on applications list

### LINKED ISSUE #211: Expand application sections in list
### Applicant perspective
Screen Shot 2020-12-11 at 3 28 20 PM

### Reviewer perspective
Screen Shot 2020-12-11 at 3 29 17 PM

### Consolidator perspective
Screen Shot 2020-12-11 at 3 28 36 PM
Screen Shot 2020-12-11 at 3 30 21 PM

### Scope
- Show collapsable and expanded icons
- When clicked should show sections with details as in design
- Actions/labels on last column are based on user permission
- NOT to implement the assignment yet - to be done in another issue
- NOT to implement the actions when user clicks options in the list - to be done in another issue

nmadruga:
Fixes #211 

Create expanded row on the list of Applications to be showing sections of application with Progress (To be done in #363).

### Changes
- Each `ApplicationList` that represents one application on the list has also a flag `expanded` to determine if it is expanded on the list
- On the `ApplicationsList` component there is a icon on each application row to allow expansion of display the sections area.
- When expanding/collapsing the sections the table needs to get reloaded - needed to pass a new props to `ApplicationsList` to reload the table, otherwise it was not re-rendering after clicking on the expansion icon.

### DIFF
--- a/src/components/List/ApplicationsList.tsx
+++ b/src/components/List/ApplicationsList.tsx
@@ -1,5 +1,5 @@
-import React from 'react'
-import { Table, Message } from 'semantic-ui-react'
+import React, { useEffect, useState } from 'react'
+import { Table, Message, Segment } from 'semantic-ui-react'
 import { ApplicationList } from '../../utils/generated/graphql'
 import messages from '../../utils/messages'
 import { ColumnDetails, SortQuery } from '../../utils/types'
@@ -20,6 +20,16 @@ const ApplicationsList: React.FC<ApplicationsListProps> = ({
   handleSort,
   loading,
 }) => {
+  const [expandedRows, setExpandedRows] = useState<Set<number>>(new Set())
+  const handleClick = (index: number) => {
+    if (expandedRows.has(index)) expandedRows.delete(index)
+    else expandedRows.add(index)
+    setExpandedRows(expandedRows)
+  }
+
+  useEffect(() => {
+    console.log(expandedRows)
+  }, [expandedRows])
   return (
     <>
       <Table sortable stackable selectable>
@@ -45,13 +55,27 @@ const ApplicationsList: React.FC<ApplicationsListProps> = ({
             </Table.Row>
           ) : (
             applications.map((application, index) => (
-              <Table.Row key={`ApplicationList-application-${application.serial}`}>
-                {columns.map(({ headerName, ColumnComponent }) => (
-                  <Table.Cell key={`ApplicationList-row${index}-${headerName}`}>
-                    <ColumnComponent application={application} />
+              <>
+                <Table.Row
+                  key={`ApplicationList-application-${application.serial}`}
+                  onClick={() => handleClick(index)}
+                >
+                  {columns.map(({ headerName, ColumnComponent }) => (
+                    <Table.Cell key={`ApplicationList-row-${index}-${headerName}`}>
+                      <ColumnComponent application={application} />
+                    </Table.Cell>
+                  ))}
+                  <Table.Cell icon="angle down" />
+                </Table.Row>
+                <Table.Row
+                  key={`ApplicationList-application-${application.serial}-sections`}
+                  colSpan={columns.length}
+                >
+                  <Table.Cell colSpan={columns.length}>
+                    <Segment color="grey" />
                   </Table.Cell>
-                ))}
-              </Table.Row>
+                </Table.Row>
+              </>
             ))
           )}
         </Table.Body>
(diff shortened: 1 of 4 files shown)


src/containers/List/ListWrapper.tsx
```diff
@@ -84,6 +85,24 @@ const ListWrapper: React.FC = () => {
...
+    if (!applicationsRows) return
+    const findApplication = applicationsRows.find(
+      ({ application }) => application.serial === serialNumber
+    )
+    if (findApplication) {
+      const { expanded } = findApplication
+      findApplication.expanded = !expanded
+    }
+    setApplicationsRows(applicationsRows)
+    setReloadTable(true)
```
CarlosNZ, line 98: Ah, I see -- you've used the "ReloadTable" value to handle the re-rendering problem you were showing me earlier.

However, I think the source of the problem is that `setApplicationsRows(applicationsRows)` is just setting `applicationRows` with itself. And it hasn't changed -- the values inside have but it's just pointing to the same object reference.

Anyway, I just tested it and it will work if you just spread the array to force a new copy:
`setApplicationsRows([...applicationsRows])`

Then you can do away with the whole reloadTable value completely.

Also -- why not pass the whole Application (rather than just the serial) to the `handleExpansion` method, then you won't have to do the `.find`.

Then the whole method could be reduced to (this also assumes you have the "isExpanded" field on the main application object, as suggested above):
```
  const handleExpansion = (application: ApplicationListRow) => {
    if (!applicationsRows) return
    application.isExpanded = !(application.isExpanded) // Updates in place inside applicationRows
    setApplicationsRows([...applicationsRows])
  }
```

src/components/List/ApplicationsList.tsx
```diff
@@ -64,4 +78,49 @@ const ApplicationsList: React.FC<ApplicationsListProps> = ({
...
+    <Table.Row
+      key={`ApplicationList-application-${serial}`}
+      onClick={() => handleExpansion(serial as string)}
+    >
+      {columns.map(({ headerName, ColumnComponent }) => (
+        <Table.Cell key={`ApplicationList-row-${index}-${headerName}`}>
+          <ColumnComponent application={application} />
+        </Table.Cell>
+      ))}
+      <Table.Cell icon="angle down" collapsing />
```
CarlosNZ, line 105: Any chance you can add an extra column (or maybe just a `colSpan` in the last cell) to the header row so the grey header background and frame extends right to edge to allow for the extra "drop-down" icon?
Screen Shot 2021-02-25 at 12 32 07 AM

CarlosNZ (changes requested):
Cool, just a couple of suggestions for how to not need the whole "reloadTable" variable/useEffect, plus some simplifications.

Summary:
- spread operator allows you to update `setApplicationRows`
- `isExpanded` field on ApplicationRow would probably be tidier than a nested object
- pass the whole applicationRow to `handleExpansion` (not just serial)

You might even be able to go back to your original Set idea, I'm not sure, but you can definitely do away with "reloadTable".

Cheers :)

Cheers.

src/containers/List/ListWrapper.tsx
nmadruga: I get a bit confused with this difference on updating the object properties which doesn't change the object itself. And didn't know about this option to do the spread to force the  a new copy. Thank you so much for this suggestion. I'll implement it now

src/containers/List/ListWrapper.tsx
```diff
@@ -37,7 +38,7 @@ const ListWrapper: React.FC = () => {
 
   useEffect(() => {
     if (!loading && applications) {
-      setApplicationsRows(applications)
+      setApplicationsRows(applications.map((application) => ({ application, expanded: false })))
```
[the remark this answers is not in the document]
nmadruga: Good idea : ) 👍 👍

nmadruga:
Fixes done. Thanks for suggestions! It looks better.

src/containers/List/ListWrapper.tsx
```diff
@@ -85,24 +86,12 @@ const ListWrapper: React.FC = () => {
     }
   }
 
-  const handleExpansion = (serialNumber: string) => {
+  const handleExpansion = (application: ApplicationListRow) => {
     if (!applicationsRows) return
```
CarlosNZ, line 90: You could *probably* lose this line too, since there's no way `applicationRows` couldn't exist if you've just passed an `application` (which is inside `applicationRows`) to this function.

CarlosNZ (approved):
Great, thanks. The simplified logic is easier to follow. 

Cheers :)

src/containers/List/ListWrapper.tsx
nmadruga: Yep! I see. Thanks

nmadruga: Ah, nah. Then it would required setting to ApplicationListRow[] on the spread which would be less visible. So keeping as it is

PR #366, opened 2021-02-24
#362 Add progress to structure

### LINKED ISSUE #362: Add logic to calculate progress fields for pages and sections in getFullApplicationStructure

nmadruga:
Fixes #362 

Branched off #361 please merge it before this one!

### Changes
- Create new type `Progress` used by `PageNEW` and `SectionStateNEW`. The new type has all properties as discussed in our meeting (24/02/2021) about how we want the new `ProgressBar` component to receive the structure to render.
- Removed not needed type `PageElementsNEW`...
- Create new file `generateProgress` with one exported function to add progress in each `Page` and `Section` based on current Responses. Filter out elements that aren't visible, editable or aren't questions (information elements).
- Fix on `useLoadApplicationNEW` to also get `category` on each element entered in `PageNEW.state` property.

### DIFF
--- a/src/utils/helpers/structure/generateProgress.ts
+++ b/src/utils/helpers/structure/generateProgress.ts
@@ -16,41 +16,30 @@ const calculateCompleted = (totalSum: number, doneRequired: number, doneNonRequi
   return totalSum === totalDone
 }
 
-const getSectionProgress = (pages: PageNEW[]): Progress => {
-  return pages.reduce((sectionProgress: Progress, { progress }) => {
-    if (!progress) return sectionProgress
-    let {
-      doneNonRequired,
-      doneRequired,
-      completed,
-      totalNonRequired,
-      totalRequired,
-      totalSum,
-      valid,
-    } = sectionProgress
-    doneNonRequired += progress.doneNonRequired || 0
-    totalNonRequired += progress.totalNonRequired || 0
-    doneRequired += progress.doneRequired || 0
-    totalRequired += progress.totalRequired || 0
-    totalSum = totalRequired + doneNonRequired
-    completed = calculateCompleted(totalSum, doneRequired, doneNonRequired)
-    if (!progress.valid) valid = false
-    return {
-      doneNonRequired,
-      doneRequired,
-      completed,
-      totalNonRequired,
-      totalRequired,
-      totalSum,
-      valid,
-    }
-  }, initialProgress)
-}
+const getSectionProgress = (pages: PageNEW[]): Progress =>
+  pages.reduce(
+    (sectionProgress: Progress, { progress }) => {
+      if (!progress) return sectionProgress
+      sectionProgress.doneNonRequired += progress.doneNonRequired || 0
+      sectionProgress.totalNonRequired += progress.totalNonRequired || 0
+      sectionProgress.doneRequired += progress.doneRequired || 0
+      sectionProgress.totalRequired += progress.totalRequired || 0
+      sectionProgress.totalSum = sectionProgress.totalRequired + sectionProgress.doneNonRequired
+      sectionProgress.completed = calculateCompleted(
+        sectionProgress.totalSum,
+        sectionProgress.doneRequired,
+        sectionProgress.doneNonRequired
+      )
+      if (!progress.valid) sectionProgress.valid = false
+      return sectionProgress
+    },
+    { ...initialProgress }
+  )
 
 export const generateResponsesProgress = (structure: FullStructure) => {
   Object.values(structure.sections).forEach((section) => {
     Object.values(section.pages).forEach((page) => {
-      page.progress = Object.assign({}, initialProgress)
+      page.progress = { ...initialProgress }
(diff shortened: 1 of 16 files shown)


src/utils/helpers/structure/generateProgress.ts
```diff
@@ -0,0 +1,82 @@
...
+      totalSum,
+      valid,
+    }
+  }, initialProgress)
+}
+
+export const generateResponsesProgress = (structure: FullStructure) => {
+  Object.values(structure.sections).forEach((section) => {
+    Object.values(section.pages).forEach((page) => {
+      const progress: Progress = initialProgress
```
andreievg, line 53: does this need to be spread ? I think it would store it as reference otherwise ? (and all pages will have the same progress ?)

src/utils/helpers/structure/generateProgress.ts
```diff
@@ -0,0 +1,82 @@
...
+        })
+        .forEach(({ element, response }) => {
+          const { isRequired } = element as ElementStateNEW
+          if (isRequired) progress.totalRequired++
+          else progress.totalNonRequired++
+          if (response?.text) {
+            if (isRequired) progress.doneRequired++
+            else progress.doneNonRequired++
+          }
+          if (response !== null && !response?.isValid) progress.valid = false
```
andreievg, line 67: Sorry I can't quite remember, i thought that progress.valid = false would only be set where validation caused an actual 'false' value ? (ie if response.isValid = null, it will still be valid ?). I think just doing `typeof response !== 'undefined' && response?.isValid == false` would work ?

Screen Shot 2021-02-24 at 5 16 40 PM

src/utils/helpers/structure/generateProgress.ts
```diff
@@ -0,0 +1,82 @@
...
+          progress.completed = calculateCompleted(
+            progress.totalSum,
+            progress.doneRequired,
+            progress.doneNonRequired
+          )
+        })
+      page.progress = progress
+    })
+    section.progress = getSectionProgress(Object.values(section.pages))
+    section.invalidPage =
```
andreievg, line 78: this is not strict right ?

maybe good idea to also do strict here (in another parameter ?

andreievg (changes requested):
Looks good, very easy to read !

A couple of minor requests

src/utils/helpers/structure/generateProgress.ts
nmadruga: Yes. This is happening! Oh, haven't noticed on this PR but on the following one.

src/utils/helpers/structure/generateProgress.ts
nmadruga: Yep. It should be only for actual `isValid` = `false`. Will update to use `typeof response !== 'undefined' && response?.isValid == false` now. Thanks : )

src/utils/helpers/structure/generateProgress.ts
nmadruga: I don't know if it would be needed. Do you think it would? When we are running the validation on `strict` mode the first page considered invalid is the first one without the `completed` field as true. Checking for incomplete as well and not only invalid

andreievg: > I don't know if it would be needed. Do you think it would? When we are running the validation on strict mode the first page considered invalid is the first one without the completed field as true. Checking for incomplete as well and not only invalid

Yeah, never mind, it would be super easy to calculate when we need it

andreievg (changes requested):
Another possible need to use ...spread or Object.assing

src/utils/helpers/structure/generateProgress.ts
```diff
@@ -0,0 +1,82 @@
...
+      totalSum,
+      valid,
+    }
+  }, initialProgress)
+}
+
+export const generateResponsesProgress = (structure: FullStructure) => {
+  Object.values(structure.sections).forEach((section) => {
+    Object.values(section.pages).forEach((page) => {
+      page.progress = Object.assign({}, initialProgress)
```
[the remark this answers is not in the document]
nmadruga: No problem. Changing to that now

src/utils/helpers/structure/generateProgress.ts
```diff
@@ -0,0 +1,82 @@
...
+    return {
+      doneNonRequired,
+      doneRequired,
+      completed,
+      totalNonRequired,
+      totalRequired,
+      totalSum,
+      valid,
+    }
+  }, initialProgress)
```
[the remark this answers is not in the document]
nmadruga: Cool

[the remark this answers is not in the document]
nmadruga: Done: acdd5b5 (with a small change to make the code maybe easier to read)

andreievg (approved):
Nice !

Good example where it's easier to write inline rather the destructuring.

Is the console log needed ?

PR #370, opened 2021-02-24
#348 PageElements component

CarlosNZ:
Implements #348.

A PageElements component for rendering a passed-in array of elements.

I've only implemented it to work with ApplicationPage for now, as it'll be easier to add the Summary and Review displays once we have their parent pages.

But it's working nicely with Application now, and I've implemented the plug-in Wrapper level validation state checks as per @andreievg 's latest diagram. If you want to see what a page looks like with "Strict" validation, just set `isStrictPage` to `true` temporarily

The ApplicationViewWrapperNEW file is not that different from the original ApplicationViewWrapper -- just updated validation functionality. You can compare the diffs between those two files to see changes if you want.

Note: for Summary and Review, there will need to be another "Sections" component inside their "Page" components to iterate over the Sections, as there needs to be a separate PageElements child per Section.

### DIFF
--- a/src/components/Application/PageElements.tsx
+++ b/src/components/Application/PageElements.tsx
@@ -2,7 +2,7 @@ import React from 'react'
 import { ElementStateNEW, ResponsesByCode } from '../../utils/types'
 import ApplicationViewWrapper from '../../formElementPlugins/ApplicationViewWrapperNEW'
 import SummaryViewWrapper from '../../formElementPlugins/SummaryViewWrapper'
-import { Button, Grid, Header, Message, Segment, Sticky, Form } from 'semantic-ui-react'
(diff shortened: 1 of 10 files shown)


andreievg (approved):
Can't see anything immediately wrong with it, i've PRed revalidation based on this one, and all seems to work fine. Only thing to check is that we are not regressing in terms of styles (@Nicole  can you have a quick look over this ?)

CarlosNZ:
> Only thing to check is that we are not regressing in terms of styles

Layout-wise, it's pretty much copy-paste from the previous component, so shouldn't be a problem. There's still a perSection Wrapper that needs to go in between SummaryPage and PageElements (probably in SummaryPage), and that should take care of remaning layout.

PR #371, opened 2021-02-24
347 revalidation logic

### LINKED ISSUE #347: Revalidation Logic

andreievg:
Fixes #347 (again)

### Prelude

Original logic didn't quite work as I hoped. When form element is updated i thought it was safe to run validation via useGetFullApplicationStructure, but it's not (responses are not refetched yet).

Will update diagram sometime

So added a few timestamp state variables to the hook.

Essentially the processing part of the hook will only re run:

* When data was changed (keeping track of last fetch and last process timestamps)
* When validation is required and last update of an element is < last fetch (so if it's not quite yet fetched, will revalidate next time it's fetched)

### Changes

Sorry for the extra bits that were change, I thought they were necessary

#### useGetFullApplicationStructure

* Consolidated fullStructure to also return responses and lastValidationTimestamp (helped with reducing amount of time ApplicationPage is "called")
*
* Removed all of the loading statuses (as per above)
* Now has two parameters or revalidation `revalidateAfterTimestamp` and `shouldRevalidate`, revalidateAfterTimestamps says validation should be executed if data is fetched after it and shouldRevalidate is true
* Added small use effect to watch for data changes and update lastRefetchedTimestamp
* Now main use effect doesn't watch data but watched above and comparese with lastProcessedTimestamp 
* Did a little bit of reshuffle (sorry it's stylistic I know, but my outdated branch that was tested had it that way)

### ApplicationPage

* Added types (changed strict indicator soz)
* Added RevalidationState (all in one to reducing amount of time ApplicationPage is "called")
* Added use effect and revalidation method
* Uses `useFormElementUpdateTracker` to set correct props for useGetFullApplicationStructure -> 
```
const shouldRevalidate = isLastElementUpdateProcessed && revalidationState.shouldProcessValidation
const revalidateAfterTimestamp = shouldRevalidate ? elementUpdatedTimestamp : 0
```

Above means, revalidate when data is refetcehed after element was updated.

### ApplicationViewWrapper

* Now uses `useFormElementUpdateTracker`

### useFormElementUpdateTracker

* Simple context state management for three fields:

`elementUpdatedTimestamp`

`elementEnteredTimestamp`

and `isLastElementUpdateProcessed` (deduce from above)

It's basically a coupy of ApplicationState (and trimmed)

### Testing

Using front end 347-testing and back end 347-(front-end)-test

`347-(front-end)-test`

Changes organisation application slightly and extends api end point for checking uniquenss. Organisation application first three questions
* First one same as before (validates uniquenss of company name)
* Second one, just a text fields
* Third one has validation that concats both first and second and checks uniequness

`347-testing`

Just added an onClick to a button to console log state of element 3 at the time of revalidation finishing

The idea behind this is, when we jump focus from first straight to submit button click, we wait for first field to be save (after it was validated internally through ApplicationViewWrapper). Full validation is queued to happen in `useGetFullApplicationStructure` after update to that field. Check out console log.

### To Test

Check out and init both branches.

Create new org application (might be a bit slow)

Change url to applicationNew

Enter something in all three fields.

Enter 'Us' in second field

Enter 'Drug-R-' in second field and without pressing tab or clicking anywhere press submit

See console output under `Re validated field -> `, should see that valid is false.

Can delete a character from first field and press submit, it should be true now


CarlosNZ:
Just a couple of quick comments/questions:

- I'm getting compile errors when trying to run front-end, regardless of the back-end I used, so haven't been able to test.
- Are the back-end changes just for testing, or to be kept? Will you be doing a separate PR that?

src/utils/helpers/structure/buildSectionsStructureNEW.ts
```diff
@@ -33,7 +33,7 @@ export const buildSectionsStructure = ({
       // Will build the array of elements
       const state = elements.map((element) => ({ element }))
 
-      const pageName = `Page ${pageNumber}`
+      const pageName = `Page${pageNumber}`
```
nmadruga, line 36: This is currently used to display the pages in the UI. To use for passing to props in to other functions you can use the Page property (in the object) `number` which has the page number to use in the URL.

andreievg: Ok I'll change

src/containers/Application/ApplicationPage.tsx
```diff
@@ -2,32 +2,96 @@ import React, { useEffect, useState } from 'react'
...
 import { Button, Grid, Header, Message, Segment, Sticky } from 'semantic-ui-react'
 import { PageElements } from '../../components/Application'
+import { useFormElementUpdateTracker } from '../../contexts/FormElementUpdateTrackerState'
 
 interface ApplicationProps {
   structure: FullStructure
   responses?: ResponsesByCode
 }
+// TODO move to shared types if used elsewhere
+type SectionAndPage = { sectionCode: string; pageName: string } | null
```
nmadruga, line 18: I'm using a very similar one in ProgressBar PR : )
https://github.com/openmsupply/application-manager-web-app/blob/b55cf0fd25c334ae09ac5207f34c53225b8a53e4/src/components/Application/ProgressBarNEW.tsx#L9-L12

src/containers/Application/ApplicationPage.tsx
```diff
@@ -2,32 +2,96 @@ import React, { useEffect, useState } from 'react'
...
+
+interface MethodToCallOnRevalidation {
+  (firstInvalidPage: SectionAndPage): void
+}
```
nmadruga, line 22: Also this one would be used by ProgressBar - See my TODO on PR #373

src/containers/Application/ApplicationPage.tsx
nmadruga: Also I don't think this needs to be a type and set to either value | null. You can just set the value here and in the state holding this `strictSectionPage` it is set to `undefined` until you have some `SectionAndPage` to define. The state can be either `SectionAndPage` | undefined - without having to define that

src/containers/Application/ApplicationPage.tsx
```diff
@@ -2,32 +2,96 @@ import React, { useEffect, useState } from 'react'
...
   const {
     userState: { currentUser },
   } = useUserState()
   const { push, query } = useRouter()
 
+  const {
+    state: { isLastElementUpdateProcessed, elementUpdatedTimestamp },
+  } = useFormElementUpdateTracker()
+
+  const [strictSectionPage, setStrictSectionPage] = useState<SectionAndPage>(null)
```
nmadruga, line 47:
```suggestion
  const [strictSectionPage, setStrictSectionPage] = useState<SectionAndPage>()
```

considering what was suggested in My first comment

src/containers/Application/ApplicationPage.tsx
```diff
@@ -2,32 +2,96 @@ import React, { useEffect, useState } from 'react'
...
+
+  const { error, fullStructure } = useGetFullApplicationStructure({
+    structure,
+    shouldRevalidate,
+    revalidateAfterTimestamp,
+  })
+
   const currentSection = query.sectionCode
-  const currentPage = `Page ${query.page}`
+  const currentPage = `Page${query.page}`
```
nmadruga: I suggest spreading the param from URL here `{ sectionCode, page }` and just using this page to pass to other function - although the page is set to string I would cast to number to match to the one in the `FullStructure` on the `pages` value object `Page.number`.

src/containers/Application/ApplicationPage.tsx
```diff
@@ -2,32 +2,96 @@ import React, { useEffect, useState } from 'react'
...
+// TODO move to shared types if used elsewhere
+type SectionAndPage = { sectionCode: string; pageName: string } | null
+
+interface MethodToCallOnRevalidation {
+  (firstInvalidPage: SectionAndPage): void
+}
+
+const getFirstInvalidPage: (
+  fullStructure: FullStructure | undefined
+) => { sectionCode: string; pageName: string } | null = (fullStructure) => {
```
nmadruga, line 26:
```suggestion
): SectionAndPage | undefined = (fullStructure) => {
```

nmadruga: Should add a TODO here?

src/containers/Application/ApplicationPage.tsx
```diff
@@ -2,32 +2,96 @@ import React, { useEffect, useState } from 'react'
...
+  }
+
+  // Revalidation Effect
+  useEffect(() => {
+    if (
+      fullStructure &&
+      revalidationState.methodToCallOnRevalidation &&
+      fullStructure.lastValidationTimestamp > revalidationState.lastRevalidationRequest
+    ) {
+      const lastValidPage = getFirstInvalidPage(fullStructure)
```
nmadruga, line 85:
```suggestion
      const firstInvalidPage = getFirstInvalidPage(fullStructure)
```

to keep consistency with method name

src/containers/Application/ApplicationPage.tsx
```diff
@@ -2,32 +2,96 @@ import React, { useEffect, useState } from 'react'
...
+      setRevalidationState({
+        ...revalidationState,
+        methodToCallOnRevalidation: null,
+        shouldProcessValidation: false,
+      })
+      if (lastValidPage != null) setStrictSectionPage(lastValidPage)
```
nmadruga, line 91:
```suggestion
      setStrictSectionPage(firstInvalidPage)
```

nmadruga: Which should be `undefined` in case there isn't one invalidPage...

nmadruga: One question: When running this for ProgressBar. For example you are currently on Page 2 and Page 3 is invalid, but if we click on a previousPage, then this method gets called and will set `firstInvalid` to Page 3, but we don't want to display any problems because current is Page 2 and the user clicked on Page 1. So maybe we need to receive here what is the page we are trying to do to?

src/utils/helpers/structure/buildSectionsStructureNEW.ts
andreievg: 8ff8640fef745374f34fcf13205a793558205b09

andreievg:
@CarlosNZ 

> I'm getting compile errors when trying to run front-end, regardless of the back-end I used, so haven't been able to test.

fixed: f10cee74e0644e27e60e6812d4b244682c402125

> Are the back-end changes just for testing, or to be kept? Will you be doing a separate PR that?

Just for testing

nmadruga (commented):
Partial review - a couple of points raised and one question on how to integrate with PR to use this validation on the ProgressBar

src/utils/types.ts
```diff
@@ -81,6 +81,7 @@ export {
   LoginPayload,
   BasicStringObject,
   SortQuery,
+  ContextFormElementUpdateTrackerState,
```
nmadruga, line 84: Can move this up on the alphabetical order please? Just so it's a little tidier. We will need to spread out this types in separated files at some point. It's getting too big and hard to see things anyway.

src/utils/hooks/useGetFullApplicationStructure.tsx
```diff
@@ -49,6 +50,13 @@ const useGetFullApplicationStructure = ({
...
 
+  // TODO - might need a use effect if serial is changes (navigated to another application from current page)
+
+  useEffect(() => {
+    if (!data) return
+    setLastRefetchedTimestamp(Date.now())
+  }, [data])
+
   useEffect(() => {
     if (loading) {
```
nmadruga, line 61: I don't think this one is needed anymore

src/containers/Application/ApplicationPage.tsx
andreievg: Doesn't it need to be specifically `null`, returning 'undefined' is pretty weird (an empty return statement), likewise setting things to 'undefined'

getFirstInvalidPage needs to return an indication that there are no invalid pages and pass that through to MethodToCallOnRevalidation, I thought null would be the way to go (and the type is sharable).

To use this type in other places, yeah it's better to just have it ` sectionCode: string; pageName: string }`
1b280a6189c93cfc82302916d5cfd682866f817c

src/containers/Application/ApplicationPage.tsx
andreievg: moved to types: 1b280a6189c93cfc82302916d5cfd682866f817c

nmadruga (changes requested):
Looks great. I'll have ago now testing out using the test branches.

I see the problem you described in the next `FormElementUpdateTrackerState`. Doyou think would be wise to add a issue to deal with that at some point? I think it instead of storing one if we would store one object with one per element, and that would be check as part of the `useGetFullApplicationStructure` when re-validating, it should be too big. But I'll leave to you to add the issue, so you can add more details to it.

Other than that I just made a few minor suggestions and a change on hot to pass the Page, I think we can keep using just the number and construct the URL with `Page${number}` as done in other places, or we can change where the pageName it's used in the UI to do `Page ${number}`, whatever works better

src/containers/Application/ApplicationPage.tsx
andreievg: Can have a chat about this if you want, I've added a comment above

src/containers/Application/ApplicationPage.tsx
andreievg: Hmm, I thought the idea of having page identifier in the key would be for easy of access ?

Otherwise have to wrap every iteration of search in Object.values. i.e. check out `getCurrentPageElements` file.

Btw, what's the difference between page key and seciton key (section key is a code but displayed through details and title) and look like page here is the page name for display

src/containers/Application/ApplicationPage.tsx
andreievg: added

src/containers/Application/ApplicationPage.tsx
andreievg: for sure, sorry, nice catch

src/containers/Application/ApplicationPage.tsx
andreievg: I had a few thought about this, it's probably not a good idea to set strict section page here automatically, I think it should be done by individual subscriber.

I think it might be best to pass through a method to set the strict section page to the subscriber:

how does that look ? 

5968da63adbdd1c34ea13e24dd73e85b928b744a

I.e. progress bar would get requestRevalidation through props, and use it on page/section navigation:
```
navigateToPage( () => {

  requestRevalidation( (firstInvalidPage, setStrictPage) => {

   // if firstInvalidPage === null, just navigate to that page
   // if firstInvalidPage is above or equal to page navigating to, just navigate to that page
   //  if firstInvalidPage is below page being navigated to, navigate to firstInvalidPage and setStrictPage(firstInvalidPage)
  })

})
```

andreievg:
> I see the problem you described in the next FormElementUpdateTrackerState. Do you think would be wise to add a issue to deal with that at some point? I think it instead of storing one if we would store one object with one per element, and that would be check as part of the useGetFullApplicationStructure when re-validating, it should be too big. But I'll leave to you to add the issue, so you can add more details to it.

I think we need to go through all of the TODO in the repo sometime and make issues (would assume it's ok to do that once we are close to MVP). 

Yeap agree with the solution you described

> Other than that I just made a few minor suggestions and a change on hot to pass the Page, I think we can keep using just the number and construct the URL with Page${number} as done in other places, or we can change where the pageName it's used in the UI to do Page ${number}, whatever works better

I made a comment above, I am happy with whatever way as long as we try and use the advantage of object (can easily access by key) and minimise doing more complex queries (where we need to iterate of values and keys etc..)

src/utils/hooks/useGetFullApplicationStructure.tsx
andreievg: so right 

c646ba662da02bf59768d000de2d11ce7f7f8b66

src/containers/Application/ApplicationPage.tsx
nmadruga: That's all good. As long as it makes sense to us, I feel that using `undefined` for that is also good. But we need to keep the same pattern everywhere. Otherwise is confusing.

src/utils/types.ts
andreievg: 6039e35c81dbcf572d7e639df31a68d514f59ead

src/containers/Application/ApplicationPage.tsx
```diff
@@ -1,34 +1,100 @@
...
   responses?: ResponsesByCode
 }
 
+const getFirstInvalidPage: (fullStructure: FullStructure) => SectionAndPage | null = (
+  fullStructure
+) => {
+  // TODO implement, should rely on .progress
+  // return { sectionCode: 'S1', pageName: 'Page 2' }
+  return null
+}
```
CarlosNZ, line 30: Suggestion: should this be calculated while generating Progress info, and stored in `fullStructure.info.firstInvalidPage`?

src/containers/Application/ApplicationPage.tsx
```diff
@@ -1,34 +1,100 @@
...
+import { useFormElementUpdateTracker } from '../../contexts/FormElementUpdateTrackerState'
 
 interface ApplicationProps {
   structure: FullStructure
   responses?: ResponsesByCode
 }
 
+const getFirstInvalidPage: (fullStructure: FullStructure) => SectionAndPage | null = (
+  fullStructure
+) => {
```
nmadruga, line 26: I think that using the typing here is confusing, maybe best to leave as:
```suggestion
const getFirstInvalidPage = ( fullStructure: FullStructure ):  SectionAndPage | null => {
```

nmadruga (commented):
Looks good. Thanks!

Few more comments - Also I tried the test, but couldn't work out how it should be displaying the result or the result was not successful idk. Can you have a quick look with me when you have a time?

src/containers/Application/ApplicationPage.tsx
andreievg: much better thanks

nmadruga:
Continuing on [this](https://github.com/openmsupply/application-manager-web-app/pull/371#issuecomment-785457663)

> I think we need to go through all of the TODO in the repo sometime and make issues (would assume it's ok to do that once we are close to MVP).

My thinking is -> Should add new issues as we think about then because we might forget about that later on and it gets hard to list what is missing when we get a chance to do more improvements - easier to have already the issues describing - and also because it's a boring job to do all at once : D

src/containers/Application/ApplicationPage.tsx
```diff
@@ -1,34 +1,98 @@
...
+        ...revalidationState,
+        methodToCallOnRevalidation: null,
+        shouldProcessValidation: false,
+      })
+      revalidationState.methodToCallOnRevalidation(firstInvalidPage, setStrictSectionPage)
+      // TODO hide loading modal
+    }
+  }, [revalidationState, fullStructure])
+
   console.log('Structure', fullStructure)
```
nmadruga, line 96: Please remove this one too?

andreievg: goneski: 72d29ae75fe9e477834f98cbdf47aebd9dab83b6

src/utils/hooks/useGetFullApplicationStructure.tsx
```diff
@@ -14,14 +14,16 @@ import { IQueryNode } from '@openmsupply/expression-evaluator/lib/types'
...
+  revalidateAfterTimestamp?: number
   firstRunValidation?: boolean
 }
 
 const useGetFullApplicationStructure = ({
   structure,
-  shouldProcessValidation,
-  firstRunValidation,
+  shouldRevalidate = false,
+  revalidateAfterTimestamp = 0,
```
CarlosNZ, line 25: I'm having trouble understanding this variable from its name. Is there something more specific or precise you could use?

src/containers/Application/ApplicationPage.tsx
nmadruga: 😜

nmadruga:
I would like to have a second run after all changes have been made. I'll wait for this to be up-to-date with master (and probably also merge with branch `347-testing` for testing), before I give my approval.

src/utils/hooks/useGetFullApplicationStructure.tsx
andreievg: hmm if i was to describe it in words: "revalidate after refetch that is later then this timestamp"

we could pass through an object, vs those two to identify that those two settings are grouped (and considered together, not individually), would that be better ?

andreievg:
> My thinking is -> Should add new issues as we think about then because we might forget about that later on and it gets hard to list what is missing when we get a chance to do more improvements - easier to have already the issues describing - and also because it's a boring job to do all at once : D

There are 35 or so TODO currently in this repo. 

Maybe an issue that says " deal with all TODO in the repo " would be enough at this stage ?

src/utils/hooks/useGetFullApplicationStructure.tsx
CarlosNZ: Okay, I get it now. I think my confusion came from the fact that it sounds like a verb ("revalidate..."), yet its value is a number. It's "timestamp to revalidate after". Is calling it `minRefetchTimestamp` any better? Or even `minRefetchTimestampForRevalidation` ? Too verbose? 🤷‍♂️

CarlosNZ:
> There are 35 or so TODO currently in this repo.
> 
> Maybe an issue that says " deal with all TODO in the repo " would be enough at this stage ?

Just to weigh in here, I'm using "TO-DO" to mean "something I would have done now but it relies on something else being done first so I can't". Ideally whoever then does the thing that it's relying on would come and finish the To-Do as part of that issue.

src/utils/hooks/useGetFullApplicationStructure.tsx
andreievg: 5af4dae0b6a9c6bec320ff3d0db34e392767c079

andreievg:
@nmadruga @CarlosNZ I think I've dealt with all of the comments, but might have missed something (please let me know)

nmadruga:
Cool. I'm going to have a look now

nmadruga (changes requested):
The test is running correctly. I'll continue testing now, just think there is still one missing thing to fix. I'm getting this error after opening the first page of newly created application:
URL: `http://localhost:3000/applicationNEW/7334/S1/Page1`
Screen Shot 2021-02-26 at 9 48 52 AM

andreievg:
> The test is running correctly. I'll continue testing now, just think there is still one missing thing to fix. I'm getting this error after opening the first page of newly created application:

i've updated, it's the testing branch not usiing methodtocallonvalidation property

nmadruga (approved):
Great

Issue #374, opened 2021-02-25
Check for allowed page on ApplicationPage load


CarlosNZ:
Based on discussion:

- Update progress calculation to set first invalid section/page in structure.info.firstInvalidPage
- Add check to ApplicationPage to ensure the currently loading page is less than firstInvalid, and redirect to firstInvalidPage if not.
- The above 2 points need to happen for sure. Slightly less certain is if we can rely on the above check for ALL page links (Next button/Progress bar) or whether they need to implement a bit of their own additional logic first. Try it and see.

andreievg:
> Add check to ApplicationPage to ensure the currently loading page is less than firstInvalid, and redirect to firstInvalidPage if not.

lest the or equal. isLinear

I think the name should be **firstStrictInvalidPage**

> The above 2 points need to happen for sure. Slightly less certain is if we can rely on the above check for ALL page links (Next button/Progress bar) or whether they need to implement a bit of their own additional logic first. Try it and see.

They need to set 'isStrictPage' (ifLinear etc..) and navigate to correct page (I don't think just having 'links' would do the job)

The first point about application page navigation, is a safe guard, but not complete solution.

Just remembered there is a point about this in the diagram:
![Screen Shot 2021-02-25 at 1 33 44 PM](https://user-images.githubusercontent.com/26194949/109085257-338e3e00-776e-11eb-8599-4e9dc275ea49.png)

PR #373, opened 2021-02-24
#351 new progress bar component

nmadruga:
Fixes #351 

~~Branched of #366 please merge it first before this one!~~ DONE

### Changes
- Use new `FullStructure` to display on the `ProgressBar` new component, now each Page has a progress and no longer needs to deduce the page progress on the go.
- Simplified the component (compared to old `ProgressBar`) to be easier to read, there is an alternative to using `Grid` commented, but that needs a bit tweaking to get rid of the background on `Label`...
- Change pages on **non linear** applications, ~~still has a TODO to deal with **linear** which is pending on checking with @andreievg what was the approach decided (it has been removed from Application Flow chart).~~
- [In progress] Adding validation to change to another page on **linear** applications

PR #376, opened 2021-02-25
#374 firstInvalidPage handling

CarlosNZ:
Implement #374 with re-direction on ApplicationPage.

Couple of comments:

- am calculating `firstInvalidPage` and `firstInvalidPageStrict`, but I think we really only want the "strict" value -- any re-directs are going to be based on strict, right? Let me know if you think we can remove the non-strict one
- Calulated these values during the iteration over the pages rather than a new function
- Removed the `invalidPage` field from .progress -- after discussion with Nicole, we don't really need to know the first invalid page for each section.
- Added a check on ApplicationPage -- will re-direct you if you attempt to load a page beyond your current valid one (in linear mode).
- It will work to have all buttons link to whatever page,then have this function re-direct, but it's a bit ugly. So the buttons will still need to have their own logic before pushing to the new URL

### DIFF
--- a/src/containers/Application/ApplicationPage.tsx
+++ b/src/containers/Application/ApplicationPage.tsx
@@ -30,10 +30,9 @@ const ApplicationPage: React.FC<ApplicationProps> = ({ structure }) => {
     push,
   } = useRouter()
 
-  const currentSection = sectionCode
+  const pageNum = Number(page)
   const currentPage = `Page ${page}`
-  const currentSectionIndex = structure.sections[currentSection].details.index
-  const currentPageNumber = structure.sections[currentSection].pages[currentPage].number
+  const currentSectionIndex = structure.sections[sectionCode].details.index
 
   console.log('Structure', fullStructure)
 
@@ -47,20 +46,20 @@ const ApplicationPage: React.FC<ApplicationProps> = ({ structure }) => {
       push(`/applicationNEW/${structure.info.serial}/summary`)
 
     // Re-direct if trying to access page higher than allowed
-    if (!fullStructure.info.isLinear) return
-    const firstInvalidSectionIndex = fullStructure.info?.firstInvalidPageStrict?.sectionIndex
-    const firstInvalidPageNum = fullStructure.info?.firstInvalidPageStrict?.pageNumber
-    if (
-      firstInvalidSectionIndex != null &&
-      firstInvalidPageNum != null &&
-      (currentSectionIndex > firstInvalidSectionIndex ||
-        (currentSectionIndex >= firstInvalidSectionIndex &&
-          currentPageNumber > firstInvalidPageNum))
-    ) {
-      push(
-        `/applicationNEW/${structure.info.serial}/${structure.info.firstInvalidPageStrict?.sectionCode}/Page${structure.info.firstInvalidPageStrict?.pageNumber}`
-      )
-    }
+    // if (!fullStructure.info.isLinear) return
+    // const firstInvalidSectionIndex = fullStructure.info?.firstInvalidPageStrict?.sectionIndex
+    // const firstInvalidPageNum = fullStructure.info?.firstInvalidPageStrict?.pageNumber
+    // if (
+    //   firstInvalidSectionIndex != null &&
+    //   firstInvalidPageNum != null &&
+    //   (currentSectionIndex > firstInvalidSectionIndex ||
+    //     (currentSectionIndex >= firstInvalidSectionIndex &&
+    //       currentPageNumber > firstInvalidPageNum))
+    // ) {
+    //   push(
+    //     `/applicationNEW/${structure.info.serial}/${structure.info.firstInvalidPageStrict?.sectionCode}/Page${structure.info.firstInvalidPageStrict?.pageNumber}`
+    //   )
+    // }
   }, [structure, fullStructure, sectionCode, page])
 
   const handleChangeToPage = (sectionCode: string, pageNumber: number) => {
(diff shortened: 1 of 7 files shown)


src/containers/Application/ApplicationPage.tsx
```diff
@@ -32,20 +32,36 @@ const ApplicationPage: React.FC<ApplicationProps> = ({ structure }) => {
...
   const currentSection = sectionCode
   const currentPage = `Page ${page}`
+  const currentSectionIndex = structure.sections[currentSection].details.index
+  const currentPageNumber = structure.sections[currentSection].pages[currentPage].number
 
   console.log('Structure', fullStructure)
 
   useEffect(() => {
-    if (!structure) return
+    if (!structure && !fullStructure) return
```
andreievg, line 41: i think !fullStructure is enough ?

src/containers/Application/ApplicationPage.tsx
```diff
@@ -32,20 +32,36 @@ const ApplicationPage: React.FC<ApplicationProps> = ({ structure }) => {
...
+    // Re-direct based on application status
     if (structure.info.current?.status === ApplicationStatus.ChangesRequired)
       push(`/applicationNEW/${structure.info.serial}`)
     if (structure.info.current?.status !== ApplicationStatus.Draft)
       push(`/applicationNEW/${structure.info.serial}/summary`)
 
-    // TO-DO: Redirect based on Progress (wait till Progress calculation is done)
-  }, [structure])
+    // Re-direct if trying to access page higher than allowed
+    if (!structure.info.isLinear) return
```
andreievg, line 50: shouldn't this be coming from fullStructure ?

src/containers/Application/ApplicationPage.tsx
nmadruga, line 41: I don't see why it's returning base on `fullStructure` not created but it only uses `structure` inside this effect... Should change to use `fullStructure`?

src/containers/Application/ApplicationPage.tsx
```diff
@@ -32,20 +32,36 @@ const ApplicationPage: React.FC<ApplicationProps> = ({ structure }) => {
...
+    const firstInvalidPageNum = structure.info?.firstInvalidPageStrict?.pageNumber
+    if (
+      firstInvalidSectionIndex != null &&
+      firstInvalidPageNum != null &&
+      (currentSectionIndex > firstInvalidSectionIndex ||
+        (currentSectionIndex >= firstInvalidSectionIndex &&
+          currentPageNumber > firstInvalidPageNum))
+    ) {
+      push(
+        `/applicationNEW/${structure.info.serial}/${structure.info.firstInvalidPageStrict?.sectionCode}/Page${structure.info.firstInvalidPageStrict?.pageNumber}`
```
nmadruga, line 61: Can you destruct this properties from `firstInvalidPageStrict` first? I think it would be cleaner this way.
Also do you think it's really needed to keep all 4 values in `firstInvalidPageStrict`? Wouldn't be enough just `sectionCode` and `pageName`? And just get in the structure `index` and `number` to compare with current?

andreievg (changes requested):

src/utils/helpers/structure/generateProgress.ts
```diff
@@ -62,10 +65,24 @@ export const generateResponsesProgress = (structure: FullStructure) => {
...
         })
+      if (
+        section.details.index <= firstInvalidSectionIndexStrict &&
+        page.number < firstInvalidPageInSectionStrict &&
+        (!page.progress.valid || page.progress.doneRequired < page.progress.totalRequired)
+      ) {
+        firstInvalidPageInSectionStrict = page.number
+        firstInvalidSectionIndexStrict = section.details.index
+        firstInvalidSectionCodeStrict = section.details.code
+      }
```
nmadruga, line 76: Better to use a separated method (in the same file) to do this

src/utils/helpers/structure/generateProgress.ts
```diff
@@ -62,10 +65,24 @@ export const generateResponsesProgress = (structure: FullStructure) => {
...
-    section.invalidPage =
-      Object.values(section.pages).find(({ progress }) => progress && !progress.valid)?.number ||
-      undefined
   })
+  structure.info.firstInvalidPageStrict = firstInvalidSectionCodeStrict
+    ? {
+        sectionCode: firstInvalidSectionCodeStrict,
+        pageName: `Page ${firstInvalidPageInSectionStrict}`,
+        sectionIndex: firstInvalidSectionIndexStrict,
+        pageNumber: firstInvalidPageInSectionStrict,
```
nmadruga, line 85: As mentioned before - don't think we need to store these 2

src/utils/types.ts
```diff
@@ -92,6 +93,7 @@ interface ApplicationDetails {
   outcome: string
   isLinear: boolean
   current?: StageAndStatus // TODO: Change to compulsory after re-strcture is finished
+  firstInvalidPageStrict?: SectionAndPage | null
```
nmadruga, line 96:
```suggestion
  firstInvalidPage?: SectionAndPage | null
```

nmadruga (changes requested):
Looks good!

I made a few suggestions please.
I guess when this one is approved it gets merged to #351 and from there I can continue the work to use this information (along with revalidation logic) to finish what is missing on PR #373. Is that alright?

src/containers/Application/ApplicationPage.tsx
CarlosNZ: Yes!

CarlosNZ: 66602ac

src/containers/Application/ApplicationPage.tsx
CarlosNZ: Yeah, but just FYI -- typescript won't let me destructure an object that might be `null`. (It would if it was `{sectionCode: null, pageName: null}` but not if the object itself can be `null`.

nmadruga: Maybe checking for `firstInvalidPage` !== null with `!fullStructure.info.isLinear` and then you could destruct it

CarlosNZ:
Okay, have made the discussed changes @nmadruga -- have another look when you get a chance.

Summary:

- each section returns a "firstIncompletePage" property as well as a global one.
- have made "pages" use the number as index rather than page name (will probably need to check if this breaks anything else)
- re-factored a few things to handle the aforementioned index change
- SectionAndPage object/type only has `sectionCode` and `pageNumber` fields (pageNumber cos that's now the index)
- tidied up the re-direct checking in ApplicationPage

nmadruga (approved):
Very nice to have the `firstIncomplete` in there. I will merge this one already so it can be used by me in the Progress bar (along with merged changes from #347) and we can continue any other suggestions in there - and keep it moving : )

PR #382, opened 2021-02-26
add review status to application list (to the view only)

andreievg:
Fixes ? (not sure which issue it fixes

needs backend branch:  https://github.com/openmsupply/application-manager-server/pull/227

## Changes

Added filter reviewerId = userId to all application list queries 
Added new fields form view:

```
        numberOfAssignedNotStartedReviews
        numberOfAssignedReviews
        numberOfChangesRequestedReviews
        numberOfDraftReviews
        numberOfReviewsSelfAssignedBySomeoneElse
        numberOfSelfAssignableReviews
        numberOfSubmittedReviews
```

### DIFF
--- a/src/utils/graphql/queries/getApplicationsList.query.ts
+++ b/src/utils/graphql/queries/getApplicationsList.query.ts
@@ -28,6 +28,13 @@ export default gql`
         status
         outcome
         lastActiveDate
+        numberOfAssignedNotStartedReviews
+        numberOfAssignedReviews
+        numberOfChangesRequestedReviews
+        numberOfDraftReviews
+        numberOfReviewsSelfAssignedBySomeoneElse
+        numberOfSelfAssignableReviews
+        numberOfSubmittedReviews
       }
       # Use the page and count info for rendering Pagination UI
       pageInfo {
--- a/src/utils/helpers/list/buildQueryFilters.ts
+++ b/src/utils/helpers/list/buildQueryFilters.ts
@@ -10,14 +10,14 @@ interface NamedDateMap {
   [key: string]: string[]
 }
 
-export default function buildQueryFilters(filters: BasicStringObject) {
+export default function buildQueryFilters(filters: BasicStringObject, userId: number) {
   const graphQLfilter = Object.entries(filters).reduce((filterObj, [key, value]) => {
     if (!mapQueryToFilterField[key]) return filterObj
     return { ...filterObj, ...mapQueryToFilterField[key](value) }
   }, {})
   // If no filters, return a dummy filter to prevent GraphQL empty object error
-  if (Object.keys(graphQLfilter).length === 0) return { templateCode: { isNull: false } }
-  return graphQLfilter
+  if (Object.keys(graphQLfilter).length === 0) return { templateCode: { isNull: true } }
+  return { reviewerId: { equalTo: userId }, ...graphQLfilter }
 }
 
 const mapQueryToFilterField: FilterMap = {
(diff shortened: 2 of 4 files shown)


nmadruga:
I think this would cover #227 partially - but might be good to break down this issue to have more granular actionable displayed after we get each specific parts of the review in place on the ReviewStructure

src/utils/graphql/queries/getApplicationsList.query.ts
```diff
@@ -28,6 +28,13 @@ export default gql`
         status
         outcome
         lastActiveDate
+        numberOfAssignedNotStartedReviews
+        numberOfAssignedReviews
+        numberOfChangesRequestedReviews
+        numberOfDraftReviews
+        numberOfReviewsSelfAssignedBySomeoneElse
```
nmadruga, line 35: I think this field name could be simplified with something as `numberOfAssignedToAnother`

nmadruga (approved):
Cool. Simple PR, I think it's ok, just make one suggestion (which involves change to the back-end). 

I haven't seen the `actionable` displayed in the list, I think that would be coming in a future PR (to complete what is required for #227). And also we need some solution to the problem - when review lvl > 1 needs to see how many reviews (from other users) have been submitted/re-submitted so they can take action to start a new consolidation.

CarlosNZ:
I'm just having a look at this now. I think I have a suggestion, but not sure if it'll work, so give me a few minutes before you merge?

src/utils/helpers/list/buildQueryFilters.ts
```diff
@@ -10,14 +10,14 @@ interface NamedDateMap {
...
-export default function buildQueryFilters(filters: BasicStringObject) {
+export default function buildQueryFilters(filters: BasicStringObject, userId: number) {
   const graphQLfilter = Object.entries(filters).reduce((filterObj, [key, value]) => {
     if (!mapQueryToFilterField[key]) return filterObj
     return { ...filterObj, ...mapQueryToFilterField[key](value) }
   }, {})
   // If no filters, return a dummy filter to prevent GraphQL empty object error
-  if (Object.keys(graphQLfilter).length === 0) return { templateCode: { isNull: false } }
-  return graphQLfilter
+  if (Object.keys(graphQLfilter).length === 0) return { templateCode: { isNull: true } }
```
CarlosNZ, line 19: This is just a error-protection because graphQL can't accept empty filter objects. It was meant to return everything if no filter object was supplied, but with `isNull: true` it'll return nothing. Since the next line adds a reviewerId to the filter object, we probably don't need this line (maybe comment it out for now?)

CarlosNZ (commented):
Just wondering -- why have the application_list view return a massive join that gives a full set of _all_ applications for _every_ reviewer? Wouldn't it be cleaner to just have a second query that fetches the review counts separately? Or, if the review counts are intended to be used for display in particular columns, perhaps those cell components can have their own internal query to fetch this info?

Not sure, just interested why you think this is the best way?

andreievg:
> Just wondering -- why have the application_list view return a massive join that gives a full set of all applications for every reviewer? Wouldn't it be cleaner to just have a second query that fetches the review counts separately? Or, if the review counts are intended to be used for display in particular columns, perhaps those cell components can have their own internal query to fetch this info?

How do you filter, pagenate etc with two queries ?

> Not sure, just interested why you think this is the best way?

As per above, so that we can still filter, pagenate and show counted filters in dashboard

I am also a bit worried about the big join, but i think this solutions should be way better: https://github.com/openmsupply/application-manager-server/pull/227#issuecomment-788670002

andreievg:
Please don't merge

andreievg:
comments will be transferred to this pr: https://github.com/openmsupply/application-manager-server/pull/266

new pr created for this : https://github.com/openmsupply/application-manager-web-app/pull/469

PR #384, opened 2021-02-26
polish of validation and progress bar logic

andreievg:
Branched from #351 

### Reason for PR

While reviewing #351, I've noticed a few issues:

* Is strict wasn't being applied to progress bar (i.e. when there is a strict validation fail progress bar should indicate which page and section)
* Empty field wasn't being shown as invalid in application view wrapper when it has an empty string
* When value was not change but field focused/unfocused, request validation was failing (it was hanging)
* When future page is failing strict validation but navigating to even further page from lower page, no navigation happens (i.e. I am on page 1, page 2 is invalid, i am trying to navigate to page 3)
* 

### Changes

* Renamed (globally) `firstIncompletePageInSection` to `firstStrictInvalidPageInSection` <- because it could be completed but not valid
* Changes FormElementUpdateTrackerState to keep track of previous and current text value, and used it in ApplicationViewWrapper and ApplicationPage (validation will re-run, not waiting for fetch if value wasn't changed)
* Passing strictSectionPage to progress bar now and if doing strict indication of page and section of not validity
* In progress bar when navigating to future page but a page in between current page and future page is not valid, navigate to it
* ApplicationViewWrapper watching isStrinctPage change to do revalidation (probably would do this differently ? Not sure if we want to revalidate at when isStrinctPage is set (just use existing value), but it's ok for now

### Further changes

* Added validation error to drop down
* Reset strict page when navigating away from error page to non error page (noticing section was being stuck on strict if page one was strict, but fixed, and navigated to page 2

@CarlosNZ I am a little bit confused, it looks like we are only passing text value to the ApplicationView for plugin, is that right ? Does this restrict us from dealing with more complex data types ?

### DIFF
--- a/src/components/Application/ProgressBarNEW.tsx
+++ b/src/components/Application/ProgressBarNEW.tsx
@@ -46,6 +46,7 @@ const ProgressBarNEW: React.FC<ProgressBarProps> = ({
     // display current page with strict validation
     requestRevalidation(({ firstStrictInvalidPage, setStrictSectionPage }: MethodToCallProps) => {
       if (!firstStrictInvalidPage) {
+        setStrictSectionPage(null)
         push(`/applicationNEW/${structure.info.serial}/${sectionCode}/Page${pageNumber}`)
         return
       }
@@ -55,9 +56,10 @@ const ProgressBarNEW: React.FC<ProgressBarProps> = ({
           firstIncomplete: firstStrictInvalidPage,
           current: { sectionCode, pageNumber },
         })
-      )
+      ) {
+        setStrictSectionPage(null)
         push(`/applicationNEW/${structure.info.serial}/${sectionCode}/Page${pageNumber}`)
-      else {
+      } else {
         setStrictSectionPage(firstStrictInvalidPage)
         push(
           `/applicationNEW/${structure.info.serial}/${firstStrictInvalidPage.sectionCode}/Page${firstStrictInvalidPage.pageNumber}`
--- a/src/components/Application/ProgressBarNEW.tsx
+++ b/src/components/Application/ProgressBarNEW.tsx
@@ -71,13 +71,13 @@ const ProgressBarNEW: React.FC<ProgressBarProps> = ({
   const getPageList = (
     sectionCode: string,
     pages: { [pageNumber: string]: PageNEW },
-    isIsStrictSection: boolean
+    isStrictSection: boolean
   ) => {
     const isActivePage = (sectionCode: string, pageNumber: number) =>
       currentSectionCode === sectionCode && Number(page) === pageNumber
 
     const checkPageIsStrict = (pageNumber: string) =>
-      isIsStrictSection && strictSectionPage?.pageNumber === Number(pageNumber)
+      isStrictSection && strictSectionPage?.pageNumber === Number(pageNumber)
 
     return (
       <List
@@ -117,7 +117,7 @@ const ProgressBarNEW: React.FC<ProgressBarProps> = ({
   }
 
   const sectionsList = Object.values(sections).map(({ details, progress, pages }, index) => {
-    const isIsStrictSection = !!strictSectionPage && strictSectionPage.sectionCode === details.code
+    const isStrictSection = !!strictSectionPage && strictSectionPage.sectionCode === details.code
 
     const stepNumber = index + 1
     const { code, title } = details
@@ -127,7 +127,7 @@ const ProgressBarNEW: React.FC<ProgressBarProps> = ({
         children: (
           <Grid>
             <Grid.Column width={4} textAlign="right" verticalAlign="middle">
-              {progress && getIndicator(progress, isIsStrictSection, stepNumber)}
+              {progress && getIndicator(progress, isStrictSection, stepNumber)}
             </Grid.Column>
             <Grid.Column width={12} textAlign="left" verticalAlign="middle">
               <Header as="h4">{title}</Header>
@@ -143,7 +143,7 @@ const ProgressBarNEW: React.FC<ProgressBarProps> = ({
       },
       onTitleClick: () => handleChangeToPage(code, 1),
       content: {
-        content: getPageList(code, pages, isIsStrictSection),
+        content: getPageList(code, pages, isStrictSection),
       },
     }
   })
--- a/src/components/Application/ProgressBarNEW.tsx
+++ b/src/components/Application/ProgressBarNEW.tsx
@@ -9,14 +9,20 @@ import {
   MethodToCallProps,
   PageNEW,
   Progress,
+  SectionAndPage,
 } from '../../utils/types'
 
 interface ProgressBarProps {
   structure: FullStructure
   requestRevalidation: MethodRevalidate
+  strictSectionPage: SectionAndPage | null
 }
 
-const ProgressBarNEW: React.FC<ProgressBarProps> = ({ structure, requestRevalidation }) => {
+const ProgressBarNEW: React.FC<ProgressBarProps> = ({
+  structure,
+  requestRevalidation,
+  strictSectionPage,
+}) => {
   const {
     info: { isLinear },
     sections,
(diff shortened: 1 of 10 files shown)


andreievg:
Also i notice a case where pageNumber was number in one case but string in another, would be good to align

CarlosNZ:
> @CarlosNZ I am a little bit confused, it looks like we are only passing text value to the ApplicationView for plugin, is that right ? Does this restrict us from dealing with more complex data types ?

Each plugin get the `currentResponse` (which is a full Response object) and a "value", which is a text value representing the response, and it's up to each plugin as to how to handle them. For example, shortText uses "value" as its UI component "value", and responds to changes in it, whereas checkbox uses the currentResponse and has more complex logic to turn that into interactive UI. Something like that, anyway.

andreievg:
> Each plugin get the currentResponse (which is a full Response object) and a "value", which is a text value representing the response, and it's up to each plugin as to how to handle them. For example, shortText uses "value" as its UI component "value", and responds to changes in it, whereas checkbox uses the currentResponse and has more complex logic to turn that into interactive UI. Something like that, anyway.

sorry, my bad, i missed this bit: https://github.com/openmsupply/application-manager-web-app/blob/26346d0efadcdd1d904a5655cc6e38f3e6632721/src/formElementPlugins/ApplicationViewWrapperNEW.tsx#L138

nmadruga:
While testing - with the `ReviewTest` application I noticed that on the first time I entered fields in the Page 1, even if the last field was invalid (as prompted by the ApplicationView) by loosing focus on that element is changing the Progress bar to **valid**, as in image below:
Screen Shot 2021-03-01 at 11 47 47 AM

Simply reloading the page will set the icon correctly to **invalid**

nmadruga:
> notice a case where pageNumber was number in one case but string in another, would be good to align

It might be because we are using the number as the key - which is converted to `string` and for that reason the object type is described as `{ [pageNumber: string]: Page }`

src/components/Application/ProgressBarNEW.tsx
```diff
@@ -31,29 +37,48 @@ const ProgressBarNEW: React.FC<ProgressBarProps> = ({ structure, requestRevalida
...
+        )
+      }
     })
   }
 
-  const getPageList = (sectionCode: string, pages: { [pageNumber: string]: PageNEW }) => {
+  const getPageList = (
+    sectionCode: string,
+    pages: { [pageNumber: string]: PageNEW },
+    isIsStrictSection: boolean
```
nmadruga, line 74:
```suggestion
    isStrictSection: boolean
```

src/components/Application/ProgressBarNEW.tsx
```diff
@@ -31,29 +37,48 @@ const ProgressBarNEW: React.FC<ProgressBarProps> = ({ structure, requestRevalida
...
+  ) => {
     const isActivePage = (sectionCode: string, pageNumber: number) =>
       currentSectionCode === sectionCode && Number(page) === pageNumber
 
+    const checkPageIsStrict = (pageNumber: string) =>
+      isIsStrictSection && strictSectionPage?.pageNumber === Number(pageNumber)
```
nmadruga, line 80:
```suggestion
      isStrictSection && strictSectionPage?.pageNumber === Number(pageNumber)
```

src/components/Application/ProgressBarNEW.tsx
```diff
@@ -82,14 +109,16 @@ const ProgressBarNEW: React.FC<ProgressBarProps> = ({ structure, requestRevalida
...
-        color={valid ? 'green' : 'red'}
+        name={isStrictlylInvalid ? 'exclamation circle' : 'check circle'}
+        color={isStrictlylInvalid ? 'red' : 'green'}
         size={step ? 'large' : 'small'}
       />
     )
   }
 
   const sectionsList = Object.values(sections).map(({ details, progress, pages }, index) => {
+    const isIsStrictSection = !!strictSectionPage && strictSectionPage.sectionCode === details.code
```
nmadruga, line 120:
```suggestion
    const isStrictSection = !!strictSectionPage && strictSectionPage.sectionCode === details.code
```

Was the `!!` intentional? What is it for?

src/components/Application/ProgressBarNEW.tsx
```diff
@@ -98,7 +127,7 @@ const ProgressBarNEW: React.FC<ProgressBarProps> = ({ structure, requestRevalida
         children: (
           <Grid>
             <Grid.Column width={4} textAlign="right" verticalAlign="middle">
-              {progress && getIndicator(progress, stepNumber)}
+              {progress && getIndicator(progress, isIsStrictSection, stepNumber)}
```
nmadruga, line 130:
```suggestion
              {progress && getIndicator(progress, isStrictSection, stepNumber)}
```

src/components/Application/ProgressBarNEW.tsx
```diff
@@ -114,7 +143,7 @@ const ProgressBarNEW: React.FC<ProgressBarProps> = ({ structure, requestRevalida
       },
       onTitleClick: () => handleChangeToPage(code, 1),
       content: {
-        content: getPageList(code, pages),
+        content: getPageList(code, pages, isIsStrictSection),
```
nmadruga, line 146:
```suggestion
        content: getPageList(code, pages, isIsStrictSection),
```
```suggestion
        content: getPageList(code, pages, isStrictSection),
```

src/formElementPlugins/ApplicationViewWrapperNEW.tsx
```diff
@@ -216,12 +216,12 @@ const calculateValidationState = async ({
   const validationResult = validationExpression
     ? await validate(validationExpression, validationMessage as string, evaluationParameters)
     : { isValid: true }
+  console.log(responses, isRequired && isStrictPage && !responses?.thisResponse, {
+    validationMessage: validationMessage || strings.VALIDATION_REQUIRED_ERROR,
+  })
```
nmadruga, line 221: Please remove console.log

src/formElementPlugins/ApplicationViewWrapperNEW.tsx
```diff
@@ -216,12 +216,12 @@ const calculateValidationState = async ({
...
   if (!validationResult.isValid) return validationResult
-  if (
-    isRequired &&
-    isStrictPage &&
-    (responses.thisResponse === undefined || responses.thisResponse === null)
-  )
+  // !responses.thisResponse, check for null, undefined, empty string
+  if (isRequired && isStrictPage && !responses?.thisResponse)
```
nmadruga, line 224: Maybe this was a safe-guard for elements that hold `thisResponse` == `false` ?

nmadruga (changes requested):
Very nice work! I was noticing some of the issues fixes here before but I wasn't sure exactly how they were happening - the order and it seems mostly all good now.

Just one test that I listed before that needs a fix and a few suggestion on the code

andreievg:
> While testing - with the ReviewTest application I noticed that on the first time I entered fields in the Page 1, even if the last field was invalid (as prompted by the ApplicationView) by loosing focus on that element is changing the Progress bar to valid, as in image below:

b74424dce5368fa215c080bc49ca0c1c88d1d126

@CarlosNZ sorry I might have been a little bit miss-leading, isValid was returning a function.

src/components/Application/ProgressBarNEW.tsx
andreievg: I did one global replace to make sure we capture all of the places this was mistyped: c7093b4381515e1a7d57a6201543e8a867bc4324

src/components/Application/ProgressBarNEW.tsx
andreievg: I did one global replace to make sure we capture all of the places this was mistyped: c7093b4381515e1a7d57a6201543e8a867bc4324

src/components/Application/ProgressBarNEW.tsx
andreievg: to turn it from null to false, otherwise type checking fails further down

andreievg: I did one global replace to make sure we capture all of the places this was mistyped: c7093b4

src/components/Application/ProgressBarNEW.tsx
andreievg: I did one global replace to make sure we capture all of the places this was mistyped: c7093b4

src/components/Application/ProgressBarNEW.tsx
andreievg: I did one global replace to make sure we capture all of the places this was mistyped: c7093b4

src/formElementPlugins/ApplicationViewWrapperNEW.tsx
andreievg: whoops, thanks a3412531168d0d3a3a64bb6a2863f359f98f4d83

src/formElementPlugins/ApplicationViewWrapperNEW.tsx
andreievg: hmm it's always been a text field (i changed it to this because empty string was failing to pass this test, and also checking for empty field makes the if statement extra large).

nmadruga (approved):
LGTM

PR #375, opened 2021-02-25
#345 Application home page

nmadruga:
Fixes #345 

### Branched from Issue's PR:
~~#371~~ MERGED 

### Changes
- New component `ApplicationHeader` to render the top part of the  container on Home page and **back to list** button. It also receives the content to be displayed inside its container
- New component `SectionsProgress`  to display  all sections in the application (using **fullStructure.sections** with each progress on sections and a **Resume** button on the first section with invalid/incomplete page
- Find the correct first section with invalid page to be done in issue #374 
- When all sections are complete show the **Review & Summary** button on the Home page
- When the application is not in `Draft` or `Changes Requested` status redirect to the Summary page (Issue #367 in progress)

### Unrelated changes
- Add `fluid` to main container to display the application a little better on mobile
- Still can't find a simple way to display the progress nicely on mobile. Probably something to do as part of [Epic#10 - App menu & layout](https://github.com/openmsupply/application-manager-web-app/issues/78)
Screen Shot 2021-02-25 at 10 39 56 PM

Issue #94, opened 2020-11-06
Epic #13 - Application Changes Requested


nmadruga:
## Description:
After the Application goes through a initial review and comes back to the Applicant OR comes back to the Reviewer (from the consolidator) to have another go - Edit application - the UI needs to display some new features and new records need to be created by mutations to allow the user to enter **new** responses in the application/review of application.

### Applicant
The user should be able to see the start page and click a button to **Make changes**.
The questions on application can now be edited (all questions are editable) with highlight for what was meant to be changed in the review. On the `Application page` the Applicant can see all fields and previous responses. The `Application start page` should take the user to the page where there is a **change required** which is where it can be edited.
After the user re-submits the Application it is only visible but not editable anymore. The new status is **Re-submitted**.
If the page is refreshed new responses should be kept. Closing the page also keeps the responses saved for the applicant.

### Reviewer
The user should be able to see the start page and click a button to **Continue** with changes requested.
All the review responses can now be edited with highlight on the ones required by the Consolidator (disagreement). On the `Review page` the Reviewer can see all fields, previous review responses and disagreements. The `Review start page` should take the user to the page where there is a **change requested** which is where it can be edited.
After the user re-submits the Review it is only visible but not editable anymore. The new status is **Submitted**.
If the page is refreshed new reviews should be kept. Closing the page also keeps the reviews saved for the reviewer.

### After changes are made
**TODO**> Write about Reviewer and Consolidator perspective after the changes are made by the solicited user...

## Documentation
- [ ] TODO: Add as part of [Epic #38](338) Public docs & tutorials

### Sprint 20:
- [x] [Back-end#137](https://github.com/openmsupply/application-manager-server/issues/137) Create Action to trim unchanged responses on submission - Effort 3

### Sprint 21:
- [x]  [Back-end#211](https://github.com/openmsupply/application-manager-server/issues/211) Design how to set review_responses to be visible to Applicant when LOQ is created - requires new field - Effort 2

### Sprint 23:
- [x] [Back-end#228](https://github.com/openmsupply/application-manager-server/issues/228) Update actions that trim applicationResponse and reviewResponses not changed during submission to update the time submitted of responses that have been changed - Effort AVG: 2 ~**In PR**

### Sprint 24:
- [x]  [Back-end#231](https://github.com/openmsupply/application-manager-server/issues/231) original response to be associated to every review_response on creation (if possible) - Effort: 1
[...]

PR #397, opened 2021-03-01
#318 Edit button on application home

### LINKED ISSUE #318: Show edit option on start page for other completed sections

nmadruga:
Fixes #318 

### Branched from
- ~~PR #384~~ MERGED
- ~~#375~~ MERGED 

### Changes
- Add edit button on `ApplicationHome` to display on application that editable, completed and before a strict section.
- Uses the `firstStrictSectionPage` to be displayed as the section with `Resume` button when user navigates to the home page.
- `SectionsProgress` component to be used by Review and Application to display the progress on each section - and options. The **badly** named received prop `changes` to be used by Review/Edit application and allow many different labels to the button that display on the same line as a section that needs attention. Thought about instead of doing passing a constructed a UI structure to this component that only renders cells... But just did this way now so we can discuss more the idea.

### DIFF
--- a/src/components/Sections/SectionsProgress.tsx
+++ b/src/components/Sections/SectionsProgress.tsx
@@ -4,11 +4,24 @@ import { Progress as ProgressType, SectionAndPage, SectionsStructureNEW } from '
 import strings from '../../utils/constants'
 
 interface SectionsProgressProps {
+  canEdit: boolean
+  changes?: { state: boolean; label: string }
   sections: SectionsStructureNEW
   resumeApplication: (location: SectionAndPage) => void
 }
 
-const SectionsProgress: React.FC<SectionsProgressProps> = ({ sections, resumeApplication }) => {
+interface SectionActionProps {
+  sectionCode: string
+  invalidPage?: number
+  progress?: ProgressType
+}
+
+const SectionsProgress: React.FC<SectionsProgressProps> = ({
+  canEdit,
+  changes = { state: false, label: '' },
+  sections,
+  resumeApplication,
+}) => {
   const getIndicator = ({ completed, valid }: ProgressType) => {
     return completed ? (
       <Icon name={valid ? 'check circle' : 'exclamation circle'} color={valid ? 'green' : 'red'} />
@@ -17,17 +30,47 @@ const SectionsProgress: React.FC<SectionsProgressProps> = ({ sections, resumeApp
     )
   }
 
-  const getProgress = ({ doneRequired, doneNonRequired, totalSum, valid }: ProgressType) => {
+  const SectionProgress: React.FC<ProgressType> = ({
+    doneRequired,
+    doneNonRequired,
+    totalSum,
+    valid,
+  }) => {
     const totalDone = doneRequired + doneNonRequired
     return totalDone > 0 && totalSum > 0 ? (
-      <Progress
-        style={{ width: 200 }}
-        percent={(100 * totalDone) / totalSum}
-        size="tiny"
-        success={valid}
-        error={!valid}
-      />
+      <Grid.Column style={{ width: 200, paddingRight: '5%' }} floated="right" width={4}>
+        <Progress
+          style={{ width: 200 }}
+          percent={(100 * totalDone) / totalSum}
+          size="tiny"
+          success={valid}
+          error={!valid}
+        />
+      </Grid.Column>
     ) : null
+  }
+
+  const SectionAction: React.FC<SectionActionProps> = ({ sectionCode, invalidPage, progress }) => {
+    return (
+      <Grid.Column style={{ minWidth: 100, padding: 0 }} width={2}>
+        {canEdit ? (
+          sectionCode === firstIncomplete?.sectionCode ? (
+            <Button color="blue" onClick={() => resumeApplication(firstIncomplete)}>
+              {strings.BUTTON_APPLICATION_RESUME}
+            </Button>
+          ) : progress?.completed ? (
+            <Icon name="pencil square" color="blue" style={{ minWidth: 100 }} />
+          ) : null
+        ) : changes.state && invalidPage ? (
+          <Button
+            color="blue"
+            onClick={() => resumeApplication({ sectionCode, pageName: `Page ${invalidPage}` })}
+          >
+            {changes.label}
+          </Button>
+        ) : null}
+      </Grid.Column>
+    )
   }
 
   // TODO: Use correct firstIncomplete sections
@@ -37,7 +80,7 @@ const SectionsProgress: React.FC<SectionsProgressProps> = ({ sections, resumeApp
     <List
       divided
       relaxed="very"
-      items={Object.entries(sections).map(([sectionCode, { details, progress }]) => ({
+      items={Object.entries(sections).map(([sectionCode, { details, invalidPage, progress }]) => ({
         key: `list-item-${sectionCode}`,
         icon: progress ? getIndicator(progress) : null,
         header: (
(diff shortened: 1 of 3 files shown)


andreievg:
Will check it out when #375 is merged and this one re-based

nmadruga:
@andreievg I have based on #375 now, if you would like to have a look on this it should be already now

src/components/Sections/SectionsProgress.tsx
```diff
@@ -22,19 +34,79 @@ const SectionsProgress: React.FC<SectionsProgressProps> = ({
     )
   }
 
-  const getProgress = ({ doneRequired, doneNonRequired, totalSum, valid }: ProgressType) => {
+  const getIsStrictSection = (sectionCode: string) =>
+    sectionCode === firstStrictInvalidPage?.sectionCode
+
+  const setIsAfterStrict = (sectionCode: string) => {
+    if (!isAfterStrict)
+      isAfterStrict = firstStrictInvalidPage === null ? true : getIsStrictSection(sectionCode)
```
andreievg, line 42: could just be `firstStrictInvalidPage === null || getIsStrictSection(sectionCode)`, but that's stylistic pref, not needed change

src/components/Sections/SectionsProgress.tsx
```diff
@@ -22,19 +34,79 @@ const SectionsProgress: React.FC<SectionsProgressProps> = ({
     )
   }
 
-  const getProgress = ({ doneRequired, doneNonRequired, totalSum, valid }: ProgressType) => {
+  const getIsStrictSection = (sectionCode: string) =>
+    sectionCode === firstStrictInvalidPage?.sectionCode
+
+  const setIsAfterStrict = (sectionCode: string) => {
```
andreievg, line 40: I am not a big fan of doing setIsAfterSrtict.

Would just do:
```
isStrictSectoin = ...
isAfterStrict = ...
```
 and use that scoped variable in `SectionAction`. 

Since the methods are scoped all of the props are available.

src/components/Sections/SectionsProgress.tsx
```diff
@@ -22,19 +34,79 @@ const SectionsProgress: React.FC<SectionsProgressProps> = ({
...
+              onClick={() =>
+                resumeApplication({
+                  sectionCode,
+                  pageNumber: firstStrictInvalidPage?.pageNumber || 1,
+                })
+              }
+            >
+              {strings.BUTTON_APPLICATION_RESUME}
+            </Button>
+          ) : progress?.completed && isAfterStrict ? (
```
andreievg, line 85: would this not be isBeforeStrict ?

src/components/Sections/SectionsProgress.tsx
```diff
@@ -22,19 +34,79 @@ const SectionsProgress: React.FC<SectionsProgressProps> = ({
...
+            <Icon
+              name="pencil square"
+              color="blue"
+              style={{ minWidth: 100 }}
+              onClick={() => resumeApplication({ sectionCode, pageNumber: 1 })}
+            />
+          ) : null
+        ) : changes.state && getIsStrictSection(sectionCode) ? (
```
andreievg, line 93: This condition is quite hard to follow.

canEdit ? 
 isStrictSection ? 
     'Resume' : 
     progress?.completed && isAfterStrict ? 
           'Edit' : 
           changes.state && getIsStrictSection(sectionCode) ? 
                changes.label : 
                null


wouldn't this be more clear ?

<Grid.Column style={{ minWidth: 100, padding: 0 }} verticalAlign="middle" width={2}>
  <SectionAction {...{ sectionCode, isBeforeStrict, progress }} />
</Grid.Column>

Inside SectionAction ->

if(!canEdit) return  null

if(isStrictSection) return 'Resume'

if(progress?.completed && isBeforeStrict ) return 'Edit'

if( changes.state) return  changes.label

By the way, I dont' think the 'update' one in the home page can be a button (can't navigate to future pates). Would just be an indicator of how many changes are requested right ?

src/components/Sections/SectionsProgress.tsx
```diff
@@ -47,17 +119,8 @@ const SectionsProgress: React.FC<SectionsProgressProps> = ({
...
-            </Grid.Column>
-            <Grid.Column style={{ minWidth: 100, padding: 0 }} width={2}>
-              {firstStrictInvalidPage !== null &&
-                sectionCode === firstStrictInvalidPage.sectionCode && (
-                  <Button color="blue" onClick={() => resumeApplication(firstStrictInvalidPage)}>
-                    {strings.BUTTON_APPLICATION_RESUME}
-                  </Button>
-                )}
-            </Grid.Column>
+            {canEdit && progress && <SectionProgress {...progress} />}
```
andreievg, line 122: Hmm, I think we want to show this even if can't edit (when LOQ is presented and before application is restarted

src/containers/Application/ApplicationHome.tsx
```diff
@@ -45,6 +45,12 @@ const ApplicationHome: React.FC<ApplicationProps> = ({ structure, template }) =>
 
   if (!fullStructure || !fullStructure.responsesByCode) return <Loading />
 
+
+  const canUserEdit = () => {
+    const { status } = fullStructure.info.current as StageAndStatus
```
andreievg, line 50: It seem easier to just do fullStructure.info?.current === ApplicationStatus.Draft (or is there a problem with types ?).

Anyways this is stylistic comment no need to change

andreievg (changes requested):
A couple of minor changes please

src/components/Sections/SectionsProgress.tsx
nmadruga: I don't know if that's the same... I needed to know if the section is **after** the strict not if it is the strict

nmadruga: Ah! Oops, I understood now. Sorry. Will change to that, it looks better indeed

src/components/Sections/SectionsProgress.tsx
nmadruga: No problem. That's tidier, will do that

nmadruga: Hmmm not sure how to do that Andrei, I was using that as a set since all sections are in order, all sections that are after the `strictInvalid` would be set to all sections after that one (without having to check the index - which isn't passed). So not sure how to do that without a set.

src/components/Sections/SectionsProgress.tsx
CarlosNZ: I agree, I generally prefer the "guard clause" pattern more than the complex nested ternary operators.

src/components/Sections/SectionsProgress.tsx
nmadruga: Thanks!
Done: 947ad03

src/components/Sections/SectionsProgress.tsx
nmadruga: Done: 947ad03

src/components/Sections/SectionsProgress.tsx
nmadruga: Nice. Thanks for having a thinking about that. Very helpful, will try to implement this solution now

src/components/Sections/SectionsProgress.tsx
nmadruga: Yes, I wasn't sure. But to be improved when we start working on issue [Epic#13 - Application Changes Required](https://github.com/openmsupply/application-manager-web-app/issues/94)

src/components/Sections/SectionsProgress.tsx
nmadruga: Done: de7aac7

src/containers/Application/ApplicationHome.tsx
nmadruga: Done: edc427a

src/components/Sections/SectionsProgress.tsx
andreievg: Oops sorry Nicole, that was a bad comment from me, the way you had it was good.

Your new changes should also work nicely

andreievg (approved):
LGTM, thanks for changes

Issue #383, opened 2021-02-26
Extract evaluator for responses from useGetReviewStructure to common helper


nmadruga:
Should only be basic structure in this issue! Only elements and responses with visibility!

PR #398, opened 2021-03-01
#383 extract evaluation of response from usetGetFullApplicationStructure

andreievg:
After a deeper look, it seems the functionality that needs to be shared is adding responses to structure (and evaluating them)

### Changes

* Extracted logic from useGetFullApplicationStructure to addEvaluatedResponsesToStructure
* Added conditional parameter to specify what needs to be evaluated (isVisiblity, isRequired, isEditable, isValid)

Most of the evaluator logic is the same as it was, apart from `evaluateSingleElement`.

Also had to fight type script a few times (when accessing properties of an object that are not indexable by string)

### DIFF
--- a/src/utils/hooks/useGetFullApplicationStructure.tsx
+++ b/src/utils/hooks/useGetFullApplicationStructure.tsx
@@ -1,17 +1,9 @@
 import { useState, useEffect } from 'react'
-import {
-  ElementStateNEW,
-  EvaluatorParameters,
-  FullStructure,
-  PageElement,
-  ResponsesByCode,
-  TemplateElementStateNEW,
-} from '../types'
+import { FullStructure } from '../types'
 import { ApplicationResponse, useGetAllResponsesQuery } from '../generated/graphql'
 import { useUserState } from '../../contexts/UserState'
-import evaluateExpression from '@openmsupply/expression-evaluator'
-import { IQueryNode } from '@openmsupply/expression-evaluator/lib/types'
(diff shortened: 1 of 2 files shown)


src/utils/hooks/useGetFullApplicationStructure.tsx
```diff
@@ -73,155 +63,37 @@ const useGetFullApplicationStructure = ({
...
       }
 
-      newStructure.responsesByCode = responseObject
-
       generateResponsesProgress(newStructure)
 
       setLastProcessedTimestamp(Date.now())
       setFirstRunProcessValidation(false)
       setFullStructure(newStructure)
+      console.log({ newStructure })
```
CarlosNZ, line 89: Remove?

CarlosNZ (approved):
LGTM! (Made a comment about the response object in `state`)

src/utils/hooks/useGetFullApplicationStructure.tsx
andreievg: d7a1b81b177b5fb4450493c16e041c135304b9da

andreievg:
> This is a bit confusing, but there are two fields on the state object -- "element" and "response". "element" is this evaluated element, and "response" is the response for that code from responseObject. If we were going to do your suggestion, it would probably make sense to go a step further and just combine these in the structure, so the state array just as "element" which also includes the response and isValid.

> Something to think about, as it would require a wee bit of re-factoring.

Noted

PR #399, opened 2021-03-02
#350 ApplicationPage with previous and next buttons

### LINKED ISSUE #350: Create NextButton and PreviousButton and use revalidation logic
Should we create a separated component to store both and one more specific to deal with validation logic for NextButton?

nmadruga:
Fixes #350 

### Branched from
- ~~#375~~ MERGED

### Changes
- Create new `Navigation` component to display **Previous** and **Next** button along with **Review & Summary** (to be implemented in #365) in a raised (using Sticky) component on the bottom of the current page.
- Uses `revalidateMethod` on the Next button handler to catch any problems in the current page (using the same approach from ProgressBar - that triggers a revalidation and calls the method received once it's done).

### Unrelated changes
- Add TODOs on various components/containers to be destroyed once Re-structure for Applications is completed - pretty soon.

### DIFF
--- a/src/components/Application/Navigation.tsx
+++ b/src/components/Application/Navigation.tsx
@@ -27,31 +27,30 @@ const Navigation: React.FC<NavigationProps> = ({
 
   const currentSectionDetails = sections[current.sectionCode].details
 
-  const hasNextSection = Object.values(sections).find(
-    ({ details: { index } }) => index > currentSectionDetails.index
-  )
-  const hasPreviousSection = Object.values(sections).find(
-    ({ details: { index } }) => index < currentSectionDetails.index
-  )
+  const nextSection = (): SectionDetails | null => {
+    const nextSections = Object.values(sections)
+      .filter(({ details: { index } }) => index > currentSectionDetails.index)
+      .sort(({ details: { index: aIndex } }, { details: { index: bIndex } }) => aIndex - bIndex)
+    return nextSections.length > 0 ? nextSections[0].details : null
+  }
 
-  const isFirstPage = current.pageNumber - 1 === 0 && !hasPreviousSection
-  const isLastPage = current.pageNumber + 1 > currentSectionDetails.totalPages && !hasNextSection
+  const previousSection = (): SectionDetails | null => {
+    const previousSections = Object.values(sections)
+      .filter(({ details: { index } }) => index < currentSectionDetails.index)
+      .sort(({ details: { index: aIndex } }, { details: { index: bIndex } }) => bIndex - aIndex)
+    return previousSections.length > 0 ? previousSections[0].details : null
+  }
+
+  const isFirstPage = current.pageNumber - 1 === 0 && previousSection() == null
+  const isLastPage =
+    current.pageNumber + 1 > currentSectionDetails.totalPages && nextSection() == null
 
   const getPreviousSectionPage = (): SectionAndPage => {
     const { sectionCode, pageNumber } = current
     if (pageNumber > 1) return { sectionCode, pageNumber: pageNumber - 1 }
-    else {
-      const previousSectionDetails = Object.values(sections).reduce(
-        (previousSection: SectionDetails, { details: { code, index } }) => {
-          if (index < previousSection.index) return sections[code].details
-          return previousSection
-        },
-        currentSectionDetails
-      )
-      return {
-        sectionCode: previousSectionDetails.code,
-        pageNumber: previousSectionDetails.totalPages,
-      }
+    return {
+      sectionCode: (previousSection() as SectionDetails).code,
+      pageNumber: (previousSection() as SectionDetails).totalPages,
     }
   }
 
@@ -59,18 +58,9 @@ const Navigation: React.FC<NavigationProps> = ({
     const { sectionCode, pageNumber } = current
     if (pageNumber < currentSectionDetails.totalPages)
       return { sectionCode, pageNumber: pageNumber + 1 }
-    else {
-      const nextSectionDetails = Object.values(sections).reduce(
-        (nextSection: SectionDetails, { details: { code, index } }) => {
-          if (index > nextSection.index) return sections[code].details
-          return nextSection
-        },
-        currentSectionDetails
-      )
-      return {
-        sectionCode: nextSectionDetails.code,
-        pageNumber: 1,
-      }
+    return {
+      sectionCode: (nextSection() as SectionDetails).code,
+      pageNumber: 1,
     }
   }
 
--- a/src/components/Application/Navigation.tsx
+++ b/src/components/Application/Navigation.tsx
@@ -1,5 +1,5 @@
 import React from 'react'
-import { Label, Segment } from 'semantic-ui-react'
+import { Button, Container, Label, Segment, Sticky } from 'semantic-ui-react'
 import { SectionAndPage, SectionDetails, SectionsStructureNEW } from '../../utils/types'
 import strings from '../../utils/constants'
 import { useRouter } from '../../utils/hooks/useRouter'
@@ -14,7 +14,6 @@ const Navigation: React.FC<NavigationProps> = ({ current, sections, serialNumber
   const { push } = useRouter()
 
   const currentSectionDetails = sections[current.sectionCode].details
-  console.log('currentSection', currentSectionDetails)
 
   const hasNextSection = Object.values(sections).find(
     ({ details: { index } }) => index > currentSectionDetails.index
@@ -79,22 +78,40 @@ const Navigation: React.FC<NavigationProps> = ({ current, sections, serialNumber
   }
 
   return (
-    <Segment.Group horizontal>
-      <Segment basic floated="left">
-        {!isFirstPage && (
-          <Label basic as="a" onClick={previousButtonHandler}>
-            {strings.BUTTON_PREVIOUS}
-          </Label>
-        )}
-      </Segment>
-      <Segment basic floated="right">
-        {!isLastPage && (
-          <Label basic as="a" onClick={nextPageButtonHandler}>
-            {strings.BUTTON_NEXT}
-          </Label>
-        )}
-      </Segment>
-    </Segment.Group>
+    <Sticky
+      pushing
+      style={{ backgroundColor: 'white', boxShadow: ' 0px -5px 8px 0px rgba(0,0,0,0.1)' }}
+    >
+      <Segment.Group horizontal>
+        <Segment style={{ minWidth: '50%' }}>
+          {!isFirstPage && (
+            <Button
+              basic
+              floated="left"
+              onClick={previousButtonHandler}
+              content={strings.BUTTON_PREVIOUS}
+            />
+          )}
+          {!isLastPage && (
+            <Button
+              basic
+              floated="right"
+              onClick={nextPageButtonHandler}
+              content={strings.BUTTON_NEXT}
+            />
+          )}
+        </Segment>
+        <Segment basic textAlign="center" clearing>
+          <Button
+            color="blue"
+            onClick={() => {
+              /* TO-DO */
+            }}
+            content={strings.BUTTON_SUMMARY}
+          />
+        </Segment>
+      </Segment.Group>
+    </Sticky>
   )
 }
 
--- a/src/components/Application/Navigation.tsx
+++ b/src/components/Application/Navigation.tsx
@@ -1,6 +1,12 @@
 import React from 'react'
-import { Button, Container, Label, Segment, Sticky } from 'semantic-ui-react'
-import { SectionAndPage, SectionDetails, SectionsStructureNEW } from '../../utils/types'
+import { Button, Segment, Sticky } from 'semantic-ui-react'
+import {
+  MethodRevalidate,
+  MethodToCallProps,
+  SectionAndPage,
+  SectionDetails,
+  SectionsStructureNEW,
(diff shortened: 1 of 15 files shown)


src/components/Application/Navigation.tsx
```diff
@@ -0,0 +1,143 @@
...
+  requestRevalidation,
+}) => {
+  const { push } = useRouter()
+
+  const currentSectionDetails = sections[current.sectionCode].details
+
+  const hasNextSection = Object.values(sections).find(
+    ({ details: { index } }) => index > currentSectionDetails.index
+  )
+  const hasPreviousSection = Object.values(sections).find(
```
andreievg, line 33: these ones return previous or next section || null, so they can be named as previousSeciton and nextSection and used in getPreviousSecitonPage or getNextSectionPage

src/components/Application/Navigation.tsx
```diff
@@ -0,0 +1,143 @@
...
+  )
+
+  const isFirstPage = current.pageNumber - 1 === 0 && !hasPreviousSection
+  const isLastPage = current.pageNumber + 1 > currentSectionDetails.totalPages && !hasNextSection
+
+  const getPreviousSectionPage = (): SectionAndPage => {
+    const { sectionCode, pageNumber } = current
+    if (pageNumber > 1) return { sectionCode, pageNumber: pageNumber - 1 }
+    else {
+      const previousSectionDetails = Object.values(sections).reduce(
```
andreievg, line 44: as per earlier comment hasPreviousSection can be renamed to previousSection and used here

andreievg (changes requested):
Logically look good, just a small simplifications suggestion.

Happy to adjust UI a bit later

src/components/Application/Navigation.tsx
nmadruga: Yep, good point.

nmadruga: Done: 7de0277

src/components/Application/Navigation.tsx
nmadruga: Done. Thanks : )

src/components/Application/Navigation.tsx
```diff
@@ -0,0 +1,133 @@
...
+  current,
+  sections,
+  serialNumber,
+  requestRevalidation,
+}) => {
+  const { push } = useRouter()
+
+  const currentSectionDetails = sections[current.sectionCode].details
+
+  const nextSection = (): SectionDetails | null => {
```
andreievg, line 30: hmm wonder why these need to be a method ?

andreievg (commented):
Looking pretty good, will approve once tested (would need home page PR merged first, and this one updated from master, I think the polish revalidation logic is not in the home page PR)

src/components/Application/Navigation.tsx
nmadruga: Just to check before accessing the first one from the returned array (of next sections). Do you think it could be done in one line?

andreievg: could just be 
```JS
const nextSections = Object.values(sections)
      .filter(({ details: { index } }) => index > currentSectionDetails.index)
      .sort(({ details: { index: aIndex } }, { details: { index: bIndex } }) => aIndex - bIndex)
const nextSection = nextSections[0].details
      
```

CarlosNZ:
One quick comment: in "non-linear" applications, pressing "Next" shouldn't do a re-validate, it should just go to the next page.

src/components/Application/Navigation.tsx
```diff
@@ -0,0 +1,133 @@
...
+  return (
+    <Sticky
+      pushing
+      style={{ backgroundColor: 'white', boxShadow: ' 0px -5px 8px 0px rgba(0,0,0,0.1)' }}
+    >
+      <Segment.Group horizontal>
+        <Segment style={{ minWidth: '50%' }}>
+          {!isFirstPage && (
+            <Button
+              basic
```
CarlosNZ, line 104: I think we should avoid using `float` for layout most of the time. It's an old-fashioned way of doing it before Flexbox was a thing, and things that are floated right don't wrap properly on mobile:
Screen Shot 2021-03-04 at 12 59 25 PM

I think you have to use Grid to lay this out properly, or maybe manually do flex box. Have a look at the Pagination component for one way of doing it (Yes, I know there is a `floated` in there, but it's actually no longer needed -- it looks the same if you remove the floats)

src/components/Application/Navigation.tsx
```diff
@@ -0,0 +1,133 @@
...
+      pageNumber: 1,
+    }
+  }
+
+  const sendToPage = (sectionPage: SectionAndPage) => {
+    const { sectionCode, pageNumber } = sectionPage
+    push(`/applicationNEW/${serialNumber}/${sectionCode}/Page${pageNumber}`)
+  }
+
+  const previousButtonHandler = (_: any) => {
```
CarlosNZ, line 72: You don't need any parameter here, I think. Can just do:
`const previousButtonHandler = () => {`

src/components/Application/Navigation.tsx
```diff
@@ -0,0 +1,133 @@
...
+    const previousSectionPage = getPreviousSectionPage()
+    sendToPage(previousSectionPage)
+  }
+
+  const nextPageButtonHandler = (_: any) => {
```
CarlosNZ, line 77: Same -- no param required

src/components/Application/Navigation.tsx
```diff
@@ -0,0 +1,133 @@
...
+
+  const getNextSectionPage = (): SectionAndPage => {
+    const { sectionCode, pageNumber } = current
+    if (pageNumber < currentSectionDetails.totalPages)
+      return { sectionCode, pageNumber: pageNumber + 1 }
+    return {
+      sectionCode: (nextSection() as SectionDetails).code,
+      pageNumber: 1,
+    }
+  }
```
CarlosNZ, line 65: I feel like these four methods could be combined into a single generic method that can go backwards and forwards and the input parameter specifies by how much. 

To go forward a page, you'd call: `const nextSectionPage = getSectionPage(1)`
And to go back you'd call: `const prevSectionPage = getSectionPage(-1)`

And the `getSectionPage` method would basically do:
- flatten all pages to a single, sorted array of SectionAndPage objects (i.e. has sectionCode and pageNum)
- infer the index in this array that the current page is
- Add the input param to the currentPageIndex -- if greater than .length, set newIndex = .length; if less than 0, newIndex = 0.
- return the sectionPage details of the object at the newIndex (code and page num)

(This would also work for jumping multiple pages, not that we need it, but: `getSectionPage(3)` for example

src/components/Application/Navigation.tsx
```diff
@@ -0,0 +1,133 @@
...
+  const previousButtonHandler = (_: any) => {
+    const previousSectionPage = getPreviousSectionPage()
+    sendToPage(previousSectionPage)
+  }
+
+  const nextPageButtonHandler = (_: any) => {
+    const nextSectionPage = getNextSectionPage()
+    // Use validationMethod to check if can change to page (on linear application) OR
+    // display current page with strict validation
+    requestRevalidation(({ firstStrictInvalidPage, setStrictSectionPage }: MethodToCallProps) => {
```
CarlosNZ, line 81: Need to check for `isLinear`, and only requestRevalidation if true.

CarlosNZ (changes requested):
Not sure about the layout for this -- what design are you basing it on?

Cheers.

nmadruga:
> One quick comment: in "non-linear" applications, pressing "Next" shouldn't do a re-validate, it should just go to the next page.

Nice catch! I thought about it the other day after finishing the work and forgot to update. Thank you!

src/components/Application/Navigation.tsx
nmadruga: Hmmm ok, there is also the check for null required there. But I'll do this change, it would be better to not have methods indeed. Thanks

nmadruga: Done: 4bd56a7

nmadruga:
> Not sure about the layout for this -- what design are you basing it on?

The idea to display the **Previous** & **Next** buttons in the same component as **Review & Submit** button was defined in one of our meeting last week - mentioned by @andreievg. Now those lines around the buttons was just some layout that I don't have a simple way to remove and continue using the Segment.Group... So I was going to leave for improvements in the App layout epic.

src/containers/Application/ApplicationPage.tsx
```diff
@@ -144,44 +145,31 @@ const ApplicationPage: React.FC<ApplicationProps> = ({ structure }) => {
...
-          </Button>
-        </Segment>
-      </Sticky>
+      <Navigation
+        current={{ sectionCode, pageNumber }}
+        sections={fullStructure.sections}
+        serialNumber={serialNumber}
+        requestRevalidation={requestRevalidation}
+      />
     </Segment.Group>
```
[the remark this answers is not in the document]
nmadruga: No, that's how it was before (but only for the **Review & Submit** button) to display on top of the current page - at all times. See how it looks when I remove this from the Segment.Group:
Screen Shot 2021-03-04 at 1 55 37 PM

Continue with the Navigation inside looks like this:
Screen Shot 2021-03-04 at 1 53 08 PM

src/containers/Application/ApplicationHome.tsx
```diff
@@ -39,7 +40,7 @@ const ApplicationHome: React.FC<ApplicationProps> = ({ structure, template }) =>
   }
 
   const handleSummaryClicked = () => {
-    push(`/applicationNEW/${serialNumber}/summary`)
+    replace(`/applicationNEW/${serialNumber}/summary`)
```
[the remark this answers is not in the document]
nmadruga: Alright. Good point, I think I have noticed this in other places - When moving from a page to the list. Not sure where though...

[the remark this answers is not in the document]
nmadruga: Done: cf1a450

src/components/Application/Navigation.tsx
nmadruga: But then instead of `onClick={previousButtonHandler}` I would need to do `onClick={() => previousButtonHandler}`. Preference?

src/components/Application/Navigation.tsx
nmadruga: Done: 3aac044

src/components/Application/Navigation.tsx
nmadruga: Hmmm I see your point, but really think it's ok since it's working correctly now and easy to read. And because no other use for this inside this component - jumping page for example. If required in the future we can change to that.

src/components/Application/Navigation.tsx
nmadruga: Yea, good point! Will have a look after the meeting : )

src/components/Application/Navigation.tsx
CarlosNZ: You don't have to do that, it works just fine with no parameter and your original `onClick`.

CarlosNZ: >  You only need to do the other type of `onClick` when you need to pass a parameter from the onClick

Speaking of which -- I think the nicest way would be to have a single handler function and you just send a parameter (the page offset -- see below comment) to tell it whether to go back or fowards.

So next button would have `onClick={() => navButtonHandler(1)}` and back button would have `onClick={() => navButtonHandler(-1)}`

src/components/Application/Navigation.tsx
nmadruga: Hi! I have looked in `Pagination` and changed to use Grid, but I can still see some not beautiful UI on mobile... I'm calling out for a good to have in App layout issue. But have changed to be using Grid, so hopefully better than before.

On mobile:
Screen Shot 2021-03-04 at 3 47 26 PM

On desktop:
Screen Shot 2021-03-04 at 3 47 38 PM

src/components/Application/Navigation.tsx
nmadruga: I think that's good for more generic buttons, but here the Previous is very simple compared with Next, and I choose to just do the special validation on the nextButtonHandler

CarlosNZ: Sure, but you still don't need the `_:any` parameters in `const previousButtonHandler = () => {...`

andreievg (approved):
LGTM

src/components/Application/Navigation.tsx
CarlosNZ: Sure, I'm not a fan of all the duplicated code -- but @andreievg 's approved and it works so carry on 👍

src/components/Application/Navigation.tsx
nmadruga: Ok. Great. I'll do this change now. : )

src/components/Application/Navigation.tsx
nmadruga: Yea, me neither, I would prefer to combine usually, and I see this solution as a possible one. I just think the very generic option wouldn't be as good since the Previous should get any validation at all

CarlosNZ: Sorry, couldn't resist having a quick play to see what the difference is. Check out branch `350-go-to-page-handler` (branched off this one) to see the alternative. Feel free to ignore, just wanted to see how it compared. 

Otherwise, merge away!

PR #400, opened 2021-03-02
#377 Create Review nested routes

### LINKED ISSUE #377: Create ReviewWrapper (container) with middleware router for Reviews

nmadruga:
Fixes #377 

### Changes
- Add `ReviewWrapper` with routes to new pages to be using new structure to render in UI
- Call `ReviewWrapper` from `ApplicationWrapper` to display nested routes (temporally) using paths:
  - Review Home [localhost:3000/applicationNEW/12345/review](http://localhost:3000/applicationNEW/12345/review)
  - Review Page [localhost:3000/applicationNEW/12345/review/1](http://ocalhost:3000/applicationNEW/12345/review/1)
  - Review Summary [localhost:3000/applicationNEW/12345/review/1/summary](http://localhost:3000/applicationNEW/12345/review/1/summary)

### Unrelated changes
- Add TODOs to remove container/components when re-structure is finished

### DIFF
--- a/src/containers/Application/ApplicationWrapper.tsx
+++ b/src/containers/Application/ApplicationWrapper.tsx
@@ -9,11 +9,13 @@ import { useRouter } from '../../utils/hooks/useRouter'
 import { FullStructure, User } from '../../utils/types'
 import { ApplicationHome, ApplicationPage } from './'
 import strings from '../../utils/constants'
+import { ReviewWrapper } from '../Review'
 
 const ApplicationWrapper: React.FC = () => {
-  const { match, query } = useRouter()
-  const { serialNumber } = query
-  const { path } = match
+  const {
+    match: { path },
+    query: { serialNumber },
+  } = useRouter()
   const {
     userState: { currentUser },
   } = useUserState()
@@ -41,6 +43,9 @@ const ApplicationWrapper: React.FC = () => {
       </Route>
       <Route exact path={`${path}/summary/submission`}>
         <ApplicationSubmissionNEW structure={structure} />
+      </Route>
+      <Route path={`${path}/review`}>
+        <ReviewWrapper />
       </Route>
       <Route>
         <NoMatch />
--- a/src/containers/Application/ApplicationWrapper.tsx
+++ b/src/containers/Application/ApplicationWrapper.tsx
@@ -45,7 +45,7 @@ const ApplicationWrapper: React.FC = () => {
         <ApplicationSubmissionNEW structure={structure} />
       </Route>
       <Route path={`${path}/review`}>
-        <ReviewWrapper />
+        <ReviewWrapper structure={structure} />
       </Route>
       <Route>
         <NoMatch />
--- a/src/containers/Review/ReviewWrapper.tsx
+++ b/src/containers/Review/ReviewWrapper.tsx
@@ -1,0 +1,53 @@
+import React from 'react'
+import { Route, Switch } from 'react-router'
+import { Header } from 'semantic-ui-react'
+import { NoMatch } from '../../components'
+import { useRouter } from '../../utils/hooks/useRouter'
+import { FullStructure } from '../../utils/types'
+
+const ReviewWrapper: React.FC = () => {
+  const {
+    match: { path },
+    query: { reviewId },
+  } = useRouter()
+
+  return (
+    <Switch>
+      <Route exact path={path}>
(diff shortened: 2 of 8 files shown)


src/containers/Application/ApplicationWrapper.tsx
```diff
@@ -42,6 +44,9 @@ const ApplicationWrapper: React.FC = () => {
       <Route exact path={`${path}/summary/submission`}>
         <ApplicationSubmissionNEW structure={structure} />
       </Route>
+      <Route path={`${path}/review`}>
+        <ReviewWrapper />
```
andreievg, line 48: Would need structure passed through

src/containers/Review/ReviewWrapper.tsx
```diff
@@ -0,0 +1,53 @@
...
+    query: { reviewId },
+  } = useRouter()
+
+  return (
+    <Switch>
+      <Route exact path={path}>
+        <ReviewHomeNEW />
+      </Route>
+      <Route exact path={`${path}/:reviewId`}>
+        <ReviewPageNEW reviewId={Number(reviewId)} />
```
andreievg, line 20: reviewId would be extracted further on, we would just need to pass though structure and reviewAssignments

src/containers/Review/ReviewWrapper.tsx
```diff
@@ -0,0 +1,53 @@
...
+      </Route>
+      <Route exact path={`${path}/:reviewId/summary`}>
+        <ReviewSummaryNEW reviewId={Number(reviewId)} />
```
andreievg, line 23: I need to refine the diagram a little bit, I am still thinking that ReviewPageWrapper is needed (which will display ReviewPage or ReviewSummary (but will filter reviewAssignment from the passed list to match reviewId (and if not then 404)

andreievg (changes requested):
Looks good, for now the only request is passing through structure

src/containers/Application/ApplicationWrapper.tsx
nmadruga: Oh! I was thinking we would rebuild the structure on the ReviewWrapper. Wasn't that the idea?

src/containers/Review/ReviewWrapper.tsx
nmadruga: should be done as part of issue #378, but I can remove this reviewId from the props now.

src/containers/Review/ReviewWrapper.tsx
nmadruga: I think that is a bit confusing, because it seems like this ReviewPagePageWrapper would be pointing to 2 routes of the router that would otherwise be in ReviewWrapper? I'm confused. After you refine this let me know if we  need to create new issues to do that change. 

I was going to start working on issue #378 now, so should be able to help testing how it works after passing the reviewAssignment to ReviewPage route and on #380 see if can compare assignment with review without problems.

src/containers/Application/ApplicationWrapper.tsx
andreievg: The idea was for the BaseStructure to still come from ApplicationWrapper, it will get populated with other bits by useGetReviewStructure

src/containers/Review/ReviewWrapper.tsx
andreievg: it's ok, we can do it in another issue

src/containers/Review/ReviewWrapper.tsx
andreievg: I've adjusted application flow diagram just now.

The reason for ReviewPageWrapper:

useGetReviewStructure takes reviewAssignment (not reviewId), because it's used to calculate progress even when review does not exist. So we need to use reviewAssignments from ReviewWrapper (passed as prop to ReviewPageWrapper), and try to find matching reviewAssignment, if found we can render Summary and ReviewPage, and pass through reviewAssignment.

src/containers/Application/ApplicationWrapper.tsx
nmadruga: Ah, yea. I see. Will do this change in the evening.

nmadruga: Done: 37155ad

andreievg (approved):

PR #402, opened 2021-03-03
#385 update use get review structure to get review responses for lvl 1 review (first review)

andreievg:
Fixes: #385

### Changes

* added reviewResponseFragment
* getAllResponses also gets a fragment of review response (to show for applicant during LOQ, still needs visibility filter), this was done because of changes to getEvaluatedResponses
* new query of getReviewNew (needs reviewAssignment and userid), review response that we get for applicationResponse are filtered such that only submitted or draft ones belonging to existing user are returned. Also we do orgerBy TIME_CREATED for reviewResponses
* add helper to add elementsById to structure (see it as useful field, it is held by reference so when it changed, corresponding elements will change in structure)
* added getGroupedReviewResponses, returns key templateElementId, and values = reviewResponses
* added addOwnedReviewResponse helper, will add latest owned review response to structure
* changed addEvaluatedResponseToStructure to all add review responses to response
* useGetFullReviewStructure hook adds owned review response to the structure

### Testing

* use 385-testing branch
* use this query in PG to see what review_assignments are there (for what user and what status)
```
select application.id as appid, application.serial, review_assignment.id as raid, review.reviewer_id as reviewer_id, review_response.time_created, review_response.status from application join review_assignment on review_assignment.application_id = application.id join review on review.application_id = application.id join review_response on review_response.review_id = review.id
```

See output of new structure in `applicationNew/${serial}/reviewTest` and adjust user (switch user) and reviewAssignment in `ReviewPageWrapperTestNew` to see different results

### DIFF
--- a/src/utils/helpers/structure/addEvaluatedResponsesToStructure.ts
+++ b/src/utils/helpers/structure/addEvaluatedResponsesToStructure.ts
@@ -7,7 +7,6 @@ import {
   FullStructure,
   PageElement,
   ResponsesByCode,
-  TemplateElementStateNEW,
   User,
 } from '../../types'
 
@@ -54,7 +53,9 @@ const addEvaluatedResponsesToStructure = async ({
       }
 
       if (response.reviewResponses.nodes.length === 0) return
-      reviewResponses[templateElementId || ''] = response.reviewResponses.nodes[0] as ReviewResponse
+      if (!templateElementId) return
+      // Responses should be orderd by timestamp in GraphQL query
+      reviewResponses[templateElementId] = response.reviewResponses.nodes[0] as ReviewResponse
     }
   })
 
@@ -64,7 +65,7 @@ const addEvaluatedResponsesToStructure = async ({
   // updated with evaluated elements and responses without re-building
   // structure
   const results = await evaluateAndValidateElements(
-    flattenedElements.map((elem: PageElement) => elem.element),
+    flattenedElements.map((elem: PageElement) => elem.element as ElementStateNEW),
     responseObject,
     currentUser,
     evaluationOptions
@@ -81,7 +82,7 @@ const addEvaluatedResponsesToStructure = async ({
 }
 
 async function evaluateAndValidateElements(
-  elements: TemplateElementStateNEW[],
+  elements: ElementStateNEW[],
   responseObject: ResponsesByCode,
   currentUser: User | null,
   evaluationOptions: EvaluationOptions
@@ -110,7 +111,7 @@ const evaluateExpressionWithFallBack = (
   })
 
 async function evaluateSingleElement(
-  element: TemplateElementStateNEW,
+  element: ElementStateNEW,
   responseObject: ResponsesByCode,
   currentUser: User | null,
   evaluationOptions: EvaluationOptions
@@ -131,7 +132,7 @@ async function evaluateSingleElement(
 
   const evaluations = evaluationKeys.map((evaluationKey) => {
     const elementKey = evaluationMapping[evaluationKey as keyof EvaluationOptions]
-    const evaluationExpression = element[elementKey as keyof TemplateElementStateNEW]
+    const evaluationExpression = element[elementKey as keyof ElementStateNEW]
 
     return evaluateExpressionWithFallBack(evaluationExpression, evaluationParameters, true)
   })
@@ -157,7 +158,7 @@ async function evaluateSingleElement(
 }
 
 const flattenStructureElements = (structure: FullStructure) => {
-  const flattened: any = []
+  const flattened: PageElement[] = []
   Object.keys(structure.sections).forEach((section) => {
     Object.keys(structure.sections[section].pages).forEach((page) => {
       flattened.push(...structure.sections[section].pages[Number(page)].state)
--- a/src/utils/helpers/structure/addEvaluatedResponsesToStructure.ts
+++ b/src/utils/helpers/structure/addEvaluatedResponsesToStructure.ts
@@ -1,6 +1,6 @@
 import evaluateExpression from '@openmsupply/expression-evaluator'
 import { IQueryNode } from '@openmsupply/expression-evaluator/lib/types'
-import { ApplicationResponse } from '../../generated/graphql'
+import { ApplicationResponse, ReviewResponse } from '../../generated/graphql'
 import {
   ElementStateNEW,
   EvaluatorParameters,
@@ -40,17 +40,22 @@ const addEvaluatedResponsesToStructure = async ({
 
   // Build responses by code (and only keep latest)
   const responseObject: any = {}
+  const reviewResponses: { [templateElementId: string]: ReviewResponse } = {}
 
   applicationResponses?.forEach((response) => {
-    const { id, isValid, value, templateElement, timeCreated } = response
+    const { id, isValid, value, templateElement, templateElementId, timeCreated } = response
     const code = templateElement?.code as string
-    if (!(code in responseObject) || timeCreated > responseObject[code].timeCreated)
+    if (!(code in responseObject) || timeCreated > responseObject[code].timeCreated) {
       responseObject[code] = {
         id,
         isValid,
         timeCreated,
         ...value,
       }
+
+      if (response.reviewResponses.nodes.length === 0) return
+      reviewResponses[templateElementId || ''] = response.reviewResponses.nodes[0] as ReviewResponse
+    }
   })
 
   const flattenedElements = flattenStructureElements(newStructure)
@@ -65,8 +70,11 @@ const addEvaluatedResponsesToStructure = async ({
     evaluationOptions
   )
   results.forEach((evaluatedElement, index) => {
-    flattenedElements[index].element = evaluatedElement
-    flattenedElements[index].response = responseObject[evaluatedElement.code]
+    const flattenedElement = flattenedElements[index]
+    flattenedElement.element = evaluatedElement
+    flattenedElement.response = responseObject[evaluatedElement.code]
+    if (flattenedElement.response)
+      flattenedElement.response.reviewResponse = reviewResponses[flattenedElement.element.id]
   })
   newStructure.responsesByCode = responseObject
   return newStructure
--- a/src/utils/helpers/structure/addOwnedReviewResponse.ts
+++ b/src/utils/helpers/structure/addOwnedReviewResponse.ts
@@ -10,8 +10,9 @@ const addOwnedReviewResponse = ({
   structure: FullStructure
   sortedReviewResponses: ReviewResponse[]
 }) => {
-  const newStructure = addElementsById(structure)
-  const { elementsById } = newStructure
+  const { elementsById } = structure
+
+  if (!elementsById) return structure
 
   const groupedReviewResponses = getGroupedReviewResponses(sortedReviewResponses)
 
@@ -23,7 +24,7 @@ const addOwnedReviewResponse = ({
     if (responseGroup.length === 1) return
   })
 
-  return newStructure
+  return structure
 }
 
 export default addOwnedReviewResponse
--- a/src/utils/helpers/structure/addOwnedReviewResponse.ts
+++ b/src/utils/helpers/structure/addOwnedReviewResponse.ts
@@ -1,9 +1,9 @@
 import { ReviewResponse } from '../../generated/graphql'
 import { FullStructure } from '../../types'
 import addElementsById from './addElementsById'
-import getSortedGroupedReviewResponses from './getGroupedReviewResponses'
+import getGroupedReviewResponses from './getGroupedReviewResponses'
 
-const addLatestOwnedtReviewResponse = ({
+const addOwnedReviewResponse = ({
   structure,
   sortedReviewResponses,
 }: {
@@ -13,19 +13,17 @@ const addLatestOwnedtReviewResponse = ({
   const newStructure = addElementsById(structure)
   const { elementsById } = newStructure
 
-  const groupedReviewResponses = getSortedGroupedReviewResponses(sortedReviewResponses)
+  const groupedReviewResponses = getGroupedReviewResponses(sortedReviewResponses)
 
   Object.entries(groupedReviewResponses).forEach(([templateElementId, responseGroup]) => {
     const element = elementsById[templateElementId]
     if (!element) return
 
-    element.latestOwnedtReviewResponse = responseGroup[0]
+    element.latestOwnedReviewResponse = responseGroup[0]
     if (responseGroup.length === 1) return
-
-    element.previousReviewResponse = responseGroup[1]
   })
 
   return newStructure
 }
 
-export default addLatestOwnedtReviewResponse
+export default addOwnedReviewResponse
--- a/src/utils/hooks/useGetFullReviewStructure.tsx
+++ b/src/utils/hooks/useGetFullReviewStructure.tsx
@@ -9,7 +9,7 @@ import {
 } from '../generated/graphql'
 import addEvaluatedResponsesToStructure from '../helpers/structure/addEvaluatedResponsesToStructure'
 import { useUserState } from '../../contexts/UserState'
-import addOwnedReviewResponse from '../helpers/structure/addOwnedReviewResponse'
+import addOwnReviewResponse from '../helpers/structure/addOwnReviewResponse'
 import addElementsById from '../helpers/structure/addElementsById'
 
 interface useGetFullReviewStructureProps {
@@ -64,7 +64,7 @@ const useGetFullReviewStructure = ({
           data.reviewAssignment?.id
         )
       if (review) {
-        newStructure = addOwnedReviewResponse({
+        newStructure = addOwnReviewResponse({
           structure: newStructure,
           sortedReviewResponses: review?.reviewResponses.nodes as ReviewResponse[], // Sorted in useGetReviewNewQuery
         })
--- a/src/utils/hooks/useGetFullReviewStructure.tsx
+++ b/src/utils/hooks/useGetFullReviewStructure.tsx
@@ -53,12 +53,11 @@ const useGetFullReviewStructure = ({
     }).then((evaluatedStructure: FullStructure) => {
       let newStructure: FullStructure = addElementsById(evaluatedStructure)
 
-      // here we add reviewless review repsonses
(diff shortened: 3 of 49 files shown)


src/utils/helpers/structure/addEvaluatedResponsesToStructure.ts
```diff
@@ -65,8 +70,11 @@ const addEvaluatedResponsesToStructure = async ({
     evaluationOptions
   )
   results.forEach((evaluatedElement, index) => {
-    flattenedElements[index].element = evaluatedElement
-    flattenedElements[index].response = responseObject[evaluatedElement.code]
+    const flattenedElement = flattenedElements[index]
+    flattenedElement.element = evaluatedElement
+    flattenedElement.response = responseObject[evaluatedElement.code]
+    if (flattenedElement.response)
+      flattenedElement.response.reviewResponse = reviewResponses[flattenedElement.element.id]
```
nmadruga, line 77: I didn't understood why you are setting an reviewResponse inside the response here? The type `ResponseFull` in the state doesn't have a `review`. Are you already doing the change to use the new structure envisioned in [LOQ Structure Diagram](https://app.diagrams.net/#G1gzlvn8D9ulQP8baqo93qspjsQqEDlxOa)? If that's the case then I suggest copying the state to a new `PageStateNEW`, rename the field to `latestApplicationResponse` and also add the type to consider new fields? I think that would be clearer.

src/utils/hooks/useGetFullReviewStructure.tsx
```diff
@@ -0,0 +1,83 @@
...
+      evaluationOptions: {
+        isEditable: false,
+        isVisible: true,
+        isRequired: false,
+        isValid: false,
+      },
+    }).then((evaluatedStructure: FullStructure) => {
+      let newStructure: FullStructure = addElementsById(evaluatedStructure)
+
+      // here we add reviewless review repsonses
```
nmadruga, line 60: What would these be?

src/utils/hooks/useGetFullReviewStructure.tsx
```diff
@@ -0,0 +1,83 @@
...
+      // generate indications of reviewless review response (numbe of non conform etc.)
+
+      const review = data?.reviewAssignment?.reviews?.nodes[0] as Review
```
nmadruga, line 63: Good to add a comment here why is getting the first review. Also good to check (and log error) in case there is more than one review associated to one reviewAssignment?

nmadruga (changes requested):
Haven't tested yet. Code looks good, although I wonder if it is the best option to keep `getGroupedReviewResponses` and `getOwnedReviewResponses` separated since they are so much connected. I would think it might be easier to build the grouped object and find the latestReviewResponse on the same block.

Also suggested changes to the type so can reflect changes made on how we store responses (similar to `latestOwnedReviewResponse` - which I haven't added a comment).

src/utils/hooks/useGetFullReviewStructure.tsx
```diff
@@ -0,0 +1,83 @@
...
+  ReviewAssignment,
+  ReviewResponse,
+  useGetReviewNewQuery,
+} from '../generated/graphql'
+import addEvaluatedResponsesToStructure from '../helpers/structure/addEvaluatedResponsesToStructure'
+import { useUserState } from '../../contexts/UserState'
+import addOwnedReviewResponse from '../helpers/structure/addOwnedReviewResponse'
+import addElementsById from '../helpers/structure/addElementsById'
+
+interface useGetFullApplicationStructureProps {
```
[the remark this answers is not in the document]
andreievg: changed here: 9e6db37b5002f55a4d9a2238e40f9f298d56c885

src/utils/hooks/useGetFullReviewStructure.tsx
```diff
@@ -0,0 +1,83 @@
...
+  structure: FullStructure
+  reviewAssignment: ReviewAssignment
+}
+
+const useGetFullReviewStructure = ({
+  structure,
+  reviewAssignment,
+}: useGetFullApplicationStructureProps) => {
```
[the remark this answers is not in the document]
andreievg: changed here: 9e6db37b5002f55a4d9a2238e40f9f298d56c885

src/utils/hooks/useGetFullReviewStructure.tsx
```diff
@@ -0,0 +1,83 @@
...
+  useGetReviewNewQuery,
+} from '../generated/graphql'
+import addEvaluatedResponsesToStructure from '../helpers/structure/addEvaluatedResponsesToStructure'
+import { useUserState } from '../../contexts/UserState'
+import addOwnedReviewResponse from '../helpers/structure/addOwnedReviewResponse'
+import addElementsById from '../helpers/structure/addElementsById'
+
+interface useGetFullApplicationStructureProps {
+  structure: FullStructure
+  reviewAssignment: ReviewAssignment
```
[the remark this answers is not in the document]
andreievg: changed here: 9e6db37b5002f55a4d9a2238e40f9f298d56c885

src/utils/hooks/useGetFullReviewStructure.tsx
```diff
@@ -0,0 +1,83 @@
...
+  const [isError, setIsError] = useState(false)
+
+  const {
+    userState: { currentUser },
+  } = useUserState()
+
+  const { data, error } = useGetReviewNewQuery({
+    variables: {
+      reviewAssignmentId: reviewAssignment.id,
+      userId: currentUser?.userId || 0,
```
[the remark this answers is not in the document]
andreievg: > Other hooks are receiving this by props, so it can be checked outside if currentUser is undefined and only receive here an object that is a User type

Hmm, we kind of check that already at the very top route. (so should never get here without userId.

I've cast it to number, that should be better for debugging. Btw I like the idea of using context info in the places where they are needed (vs passing as props):

changed here: 9e6db37b5002f55a4d9a2238e40f9f298d56c885

src/utils/hooks/useGetFullReviewStructure.tsx
```diff
@@ -0,0 +1,83 @@
...
+  structure: FullStructure
+  reviewAssignment: ReviewAssignment
+}
+
+const useGetFullReviewStructure = ({
+  structure,
+  reviewAssignment,
+}: useGetFullApplicationStructureProps) => {
+  const [fullStructure, setFullStructure] = useState<FullStructure>()
+  const [isError, setIsError] = useState(false)
```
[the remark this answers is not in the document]
andreievg: I kind of used useGetFullApplicationStructureProps as a starting point for this hook. 

changed both here: 9e6db37b5002f55a4d9a2238e40f9f298d56c885

src/utils/hooks/useGetFullReviewStructure.tsx
```diff
@@ -0,0 +1,83 @@
...
+      const review = data?.reviewAssignment?.reviews?.nodes[0] as Review
+      if (review) {
+        newStructure = addOwnedReviewResponse({
+          structure: evaluatedStructure,
+          sortedReviewResponses: review?.reviewResponses.nodes as ReviewResponse[], // Sorted in useGetReviewNewQuery
+        })
+      }
+
+      // Generate review progress
+
```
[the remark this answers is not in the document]
andreievg: both this one and useGetFullApplicationStructure, just return structure (we check if it's not null as loading check)

src/utils/helpers/structure/addEvaluatedResponsesToStructure.ts
```diff
@@ -40,17 +40,22 @@ const addEvaluatedResponsesToStructure = async ({
...
+    if (!(code in responseObject) || timeCreated > responseObject[code].timeCreated) {
       responseObject[code] = {
         id,
         isValid,
         timeCreated,
         ...value,
       }
+
+      if (response.reviewResponses.nodes.length === 0) return
+      reviewResponses[templateElementId || ''] = response.reviewResponses.nodes[0] as ReviewResponse
```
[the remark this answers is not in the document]
andreievg: changed both: 9e6db37b5002f55a4d9a2238e40f9f298d56c885

but not sure about checking on the caller, templateElementId would always exists btw, it's just the generated graphql types make it optional

src/utils/helpers/structure/addEvaluatedResponsesToStructure.ts
andreievg: Kind of need to link response to reviewResponse to know the progress (i.e. in the review progress PR I check against response having reviewResponse for assigned elements).

not 100% sure what you mean, but added changes to types. it looks like i didn't notice any type errors because previous flattenStructureElements function was returning `any[]`

9e6db37b5002f55a4d9a2238e40f9f298d56c885

src/utils/hooks/useGetFullReviewStructure.tsx
andreievg: like when consolidation is not started etc..

src/utils/hooks/useGetFullReviewStructure.tsx
andreievg: I've added a comment. There are other places where we have a relationship that looks like one to many, but is really one to one, like this one:
https://github.com/openmsupply/application-manager-web-app/blob/a3961e915dccac92e82731ee67399e57ea4999ee/src/utils/hooks/useLoadApplicationNEW.ts#L100

9e6db37b5002f55a4d9a2238e40f9f298d56c885

andreievg:
> Haven't tested yet. Code looks good, although I wonder if it is the best option to keep getGroupedReviewResponses and getOwnedReviewResponses separated since they are so much connected. I would think it might be easier to build the grouped object and find the latestReviewResponse on the same block.

I was wondering if groupedReviewResponses will be needed for other operations within that hook, thus extracted it. Can consolidate. 

By the way, in terms of doing things in the same block, although it does look optimised, I quite often prefer separating them to avoid having a big block of clever logic, that might be less readable.

> Also suggested changes to the type so can reflect changes made on how we store responses (similar to latestOwnedReviewResponse - which I haven't added a comment).

Can you elaborate further please.

src/utils/helpers/structure/addOwnedReviewResponse.ts
```diff
@@ -0,0 +1,30 @@
+import { ReviewResponse } from '../../generated/graphql'
+import { FullStructure } from '../../types'
+import addElementsById from './addElementsById'
+import getGroupedReviewResponses from './getGroupedReviewResponses'
+
+const addOwnedReviewResponse = ({
```
CarlosNZ, line 6: `addOwnReviewResponse` (rather than `addOwnedReviewResponse`)? (Filename too.)

andreievg: 90a82229a21c53edd7baaaedd7e6ad30a63d3f7f

CarlosNZ:
Just a couple of questions:

- the `response` object has a `reviewResponse` field inside it. Is this the same object as the `latestOwnReviewResponse`? I see that it is looking at local examples, but will it always be? (And if so, why have both?) I'm guessing that the `reviewResponse` is the immediate one connected to the applicationResponse that it's inside of, which might not be the current `latestOwnReviewResponse`, right?
- What's the purpose of the `orderBy` in reviewResponse query? Is it necessary?
- Shouldn't the `latestOwnReviewResponse` belong to the currently logged-in user? I'm seeing the same thing in that field regardless of which user I'm logged in as. Is that correct?

src/utils/types.ts
```diff
@@ -282,6 +292,7 @@ interface PageNEW {
 type PageElement = {
   element: ElementBaseNEW | ElementStateNEW
   response: ResponseFull | null
```
nmadruga, line 294: Can you add a // TODO: Remove this here  please so we don't forget about it?

src/utils/hooks/useGetFullReviewStructure.tsx
nmadruga: Ahhh now I understood `reviewless` as without an associated review. But could it be that you get review_responses that doesn't have any review associated? Not right, they have some review it might not be yours...

src/utils/hooks/useGetFullReviewStructure.tsx
nmadruga: Yep. I know, that's the expected case, but sometimes (for bugs) that rule might be broken, that's why adding the check before it like for this example, the previous 2 lines are: 
https://github.com/openmsupply/application-manager-web-app/blob/a3961e915dccac92e82731ee67399e57ea4999ee/src/utils/hooks/useLoadApplicationNEW.ts#L98-L99

nmadruga:
>> Also suggested changes to the type so can reflect changes made on how we store responses (similar to latestOwnedReviewResponse - which I haven't added a comment).

> Can you elaborate further please.

I was only suggesting that `latestOwnedReviewResponse` gets added to the PageState and `reviewResponse` to ResponseFull types, which you have done in the latest commit : )

nmadruga:
>> Haven't tested yet. Code looks good, although I wonder if it is the best option to keep getGroupedReviewResponses and getOwnedReviewResponses separated since they are so much connected. I would think it might be easier to build the grouped object and find the latestReviewResponse on the same block.

> I was wondering if groupedReviewResponses will be needed for other operations within that hook, thus extracted it. Can consolidate.

> By the way, in terms of doing things in the same block, although it does look optimised, I quite often prefer separating them to avoid having a big block of clever logic, that might be less readable.

No problem, I agree with breaking down, very good for reusability. I just though that this case might not be used in separated places. I get your point that it should be better to see if the logic break-down in separated methods though, no problem to leave like it is for now.

nmadruga:
@andreievg I'm not sure about the tests... I was running the `#385-testing` updated with this PR latest version and when I check this page: `http://localhost:3000/applicationNEW/34567/reviewTest` I get the same result for Reviewer 1 and Reviewer 2, focusing on the first Question (template id 4001, Response id 4020) the Reviewer 1 has **Declined** which is showing correctly on their ReviewTest page, but when I change to Reviewer 2 I get the same result, although this user does't have a review for this question as I checked in the database.

Screen Shot 2021-03-05 at 10 36 09 AM

src/utils/hooks/useGetFullReviewStructure.tsx
andreievg: yep that's correct, my comment wasn't though 9dfd303d37e3b41e56cf11d32ec9de3f4897cee4

src/utils/hooks/useGetFullReviewStructure.tsx
andreievg: Cools, the log was added, and updated cause it wasn't accurate: 9dfd303d37e3b41e56cf11d32ec9de3f4897cee4

src/utils/types.ts
andreievg: Do you mean rename to latestApplicationResponse ?

CarlosNZ: I think she just means as TO-DO comment as a marker for things we need to delete later.

andreievg:
@CarlosNZ 
> the response object has a reviewResponse field inside it. Is this the same object as the latestOwnReviewResponse? I see that it is looking at local examples, but will it always be? (And if so, why have both?) I'm guessing that the reviewResponse is the immediate one connected to the applicationResponse that it's inside of, which might not be the current latestOwnReviewResponse, right?

Yep, lastestOwneReviewResponse will be different further up the chain

> What's the purpose of the orderBy in reviewResponse query? Is it necessary?

To make sure we always get the latest reviewResponse for application response

> Shouldn't the latestOwnReviewResponse belong to the currently logged-in user? I'm seeing the same thing in that field regardless of which user I'm logged in as. Is that correct?

Was thinking it should belong to the review from the review_assignment, maybe it should be renamed to be `latestThisReviewResponse` ?  I wasn't describing it accurately previously.

andreievg:
@nmadruga 

> @andreievg I'm not sure about the tests... I was running the #385-testing updated with this PR latest version and when I check this page: http://localhost:3000/applicationNEW/34567/reviewTest I get the same result for Reviewer 1 and Reviewer 2, focusing on the first Question (template id 4001, Response id 4020) the Reviewer 1 has Declined which is showing correctly on their ReviewTest page, but when I change to Reviewer 2 I get the same result, although this user does't have a review for this question as I checked in the database.

Sorry I should be explain this stuff a big better. It's based on review_assignment. The latest review_response from review of that review assignment (that is either submitted or draft if for current user), is attached to response object  (you can check this by using branched from #386 test for back end and similar instructions of testing as #386, you should see draft responses appear for only the users they belong to

> I even tried to logout and login with Reviewer 2 again to check if it wasn't a cache problem, but it got the same thing. Maybe I'm just confused, but I thought the latestOwnedReviewResponse should be only the ones assigned to you.

latestOwnReviewResponse should really be renamed (again) to indicate it's latestReviewReviewResponse (or something like that, I had it as Current previously but that wasn't quite explanatory enough. This will change if you change reviewAssignment in the test page

andreievg:
@nmadruga @CarlosNZ
I've updated naming: https://github.com/openmsupply/application-manager-web-app/pull/402/commits/129d5dfdca6942c23559d739fc75e797e42bc966

src/utils/types.ts
andreievg: Sure but I am not sure what to delete later based on the code this comment is pointing to

andreievg:
@nmadruga @CarlosNZ, not sure if there is anything else that needs to be done for this PR

src/utils/hooks/useGetFullReviewStructure.tsx
nmadruga: Oopsy I have missed that too!

src/utils/types.ts
nmadruga: Hmm I think it was just a confusion. I though the new field would be a replacement fro `response` but it isn't , since `thisReviewLatestResponse` is a review instead. 😜

nmadruga:
No. I think it's good now. Great to see these improvements in place. Naming can be hard and silly at the same time, but hopefully will help us with finding what each fields are just by quickly looking at them. Should I update in the Structure Diagrams too?

nmadruga (approved):

andreievg:
> Should I update in the Structure Diagrams too? 

I'll update those on the weekend

andreievg:
Thanks for reviews !

Issue #379, opened 2021-02-25
Create ReviewHome (container) with not expandable reviews - should list each review for the User


nmadruga:

andreievg:
@nmadruga do you mind if someone takes this issue over while you are away ?

nmadruga:
No, all good by me. I was keen on working on it but really shouldn't so I'm happy to work on what's available when I'm ready to do more coding - probably next week.

For now I'll focus in reviews and meetings ☺️

andreievg:
There is a slight refinement (maybe not so slight).
https://app.diagrams.net/#G1-kKRiGtnZ8__cL82EOdpYap7Rv79GBvX

But the idea is it makes a lot more sense to display concurrent tasks for the same section group by section.

I am going to have a go at this one now

PR #406, opened 2021-03-03
#365 New submission page

### LINKED ISSUE #365: Add SubmissionNEW page

nmadruga:
Fixes #365 

### Changes
- Create new `ApplicationSubmission` page container that receives the structure and only display the page - I wasn't sure if that might be just a component - since there isn't much logic going on here. But I guess we might have later on.
- Change type **FullStructure.info** to also keep the `submissionMessage` used in the Submission page for an application
- Change type **FullStructure.stages** to be replaced with what was agreed on the new [Naming Changes Structure Diagram](https://app.diagrams.net/#G1wEk_eH1uWd8LC0JeKAM58bUkGncxKDgW). 
- Disclaimer: Although changes were done to same structure in use by old Application pages, I didn't have to change anything since the previous `stages` with `submissionMessage` was stored in a different object with type `ApplicationStages` that hasn't changes. And the addition to `submissionMessage` in the `ApplicationDetails` is using the temporary set to be optional field, so not required for previous use.
- Set redirect to the `ApplicationSummary` page when the application latest status isn't SUBMITTED/RE-SUBMITTED
- Links to Application list (based on **application.type**) and home page from the Submission page.

### DIFF
--- a/src/containers/Application/ApplicationSubmissionNEW.tsx
+++ b/src/containers/Application/ApplicationSubmissionNEW.tsx
@@ -1,9 +1,92 @@
-import React from 'react'
-import { Header } from 'semantic-ui-react'
-import { ApplicationProps } from '../../utils/types'
+import React, { useState } from 'react'
+import { Button, Header, Icon, Label, List, Message, Segment } from 'semantic-ui-react'
+import evaluate from '@openmsupply/expression-evaluator'
+import Markdown from '../../utils/helpers/semanticReactMarkdown'
+import { ApplicationProps, EvaluatorParameters } from '../../utils/types'
+import { useUserState } from '../../contexts/UserState'
+import { useRouter } from '../../utils/hooks/useRouter'
+import strings from '../../utils/constants'
+import { Link } from 'react-router-dom'
 
 const ApplicationSubmission: React.FC<ApplicationProps> = ({ structure }) => {
-  return <Header>SUBMISSION PAGE</Header>
+  const [submissionMessageEvaluated, setSubmissionMessageEvaluated] = useState('')
+  const {
+    userState: { currentUser },
+  } = useUserState()
+
+  const {
+    query: { serialNumber },
+    push,
+  } = useRouter()
+  const {
+    info: { submissionMessage, type },
+    stages,
+  } = structure
+
+  const evaluatorParams: EvaluatorParameters = {
+    objects: { currentUser },
+    APIfetch: fetch,
+  }
+  evaluate(submissionMessage || '', evaluatorParams).then((result: any) =>
+    setSubmissionMessageEvaluated(result)
+  )
+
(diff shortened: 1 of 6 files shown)


src/containers/Application/ApplicationSubmissionNEW.tsx
```diff
@@ -0,0 +1,91 @@
...
+  )
+    push(`/applicationNEW/${serialNumber}/summary`)
+
+  const evaluatorParams: EvaluatorParameters = {
+    objects: { currentUser },
+    APIfetch: fetch,
+  }
+  evaluate(submissionMessage || '', evaluatorParams).then((result: any) =>
+    setSubmissionMessageEvaluated(result)
+  )
```
CarlosNZ, line 41: Do you think we should evaluate this in `useLoadApplication`, to keep more logic out of this display component?

src/containers/Application/ApplicationSubmissionNEW.tsx
```diff
@@ -0,0 +1,91 @@
...
+            to={`/applicationNEW/${serialNumber}/summary`}
+            style={{ minWidth: 200 }}
+            content={`${strings.BUTTON_BACK_TO} ${type}`}
+          />
+          <Label as={Link} to={'/'} content={strings.BUTTON_BACK_DASHBOARD} />
+        </Segment>
+      </Segment>
+    </Segment.Group>
+  )
+}
```
CarlosNZ, line 89: I'm going to assume this UI is basically identical to the old one?

CarlosNZ (commented):
Great! One small suggestion about moving the evaluation, see if you think it's a good idea.

src/containers/Application/ApplicationSubmissionNEW.tsx
nmadruga: Yep. It is : )

src/containers/Application/ApplicationSubmissionNEW.tsx
nmadruga: I'll have a thought tomorrow, I think it's a good idea, but wonder if should do the same for `startMessage` in `useLoadTemplate` => In PR #404 - XD

CarlosNZ: Yes, I actually just made the same comment on that one too -- haven't submitted review yet though.

CarlosNZ (approved):
LGTM -- once you merge this I'll pull it into my #409 PR and make sure that's working properly with it.

You want to do the change to move the Start/Submit message evaluation to hooks?

nmadruga:
> You want to do the change to move the Start/Submit message evaluation to hooks?

I was planning to do it today - although also would like to start #379, so if you are keen to do it it's fine by me 👍

PR #409, opened 2021-03-04
#367 part 2 - Submit button functionality

CarlosNZ:
Don't think this had a dedicated issue, consider it Part 2 of Summary page issue (part 1 #407  - please review first)

UPDATE: Added re-validation and re-direct functionality for "Review & Submit" button

The logic for re-validating and then submitting is a bit simpler than with the Prev/Next buttons, hopefully it makes sense.

Basic outline:
- When user clicks "Submit", the `shouldRevalidate` value is set to `true`, which forces a re-validation in the `useGetFullStructure` hook
- UseEffect on FullStructure runs, re-directs if validation errors are found, and sets `isRevalidated` to `true`
- UseEffect watches `isRevalidated` and runs "Submit()" method (from `useSubmitApplication` hook) when true.
- One more useEffect watches the `submitted` value from `useSubmitApplication` hook and re-directs when true.

One issue: this is not the only place I've seen this, but console reports the following error after submission:
``` 
Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in a useEffect cleanup function.
```
I guess this is because there are still async functions running in `useGetFullStructure` hook at the time the page re-direct happens (which unmounts the components that are calling this hook). I've no idea what we should do about it though -- any ideas?

### DIFF
--- a/src/components/Application/Navigation.tsx
+++ b/src/components/Application/Navigation.tsx
@@ -95,6 +95,15 @@ const Navigation: React.FC<NavigationProps> = ({
     })
   }
 
+  const summaryButtonHandler = () => {
+    requestRevalidation(({ firstStrictInvalidPage, setStrictSectionPage }: MethodToCallProps) => {
+      if (firstStrictInvalidPage) {
+        setStrictSectionPage(firstStrictInvalidPage)
+        sendToPage(firstStrictInvalidPage)
+      } else push(`/applicationNEW/${serialNumber}/summary`)
+    })
+  }
+
   return (
     <Sticky
       pushing
@@ -119,13 +128,7 @@ const Navigation: React.FC<NavigationProps> = ({
             />
           </Grid.Column>
           <Grid.Column width={5}>
-            <Button
-              color="blue"
-              onClick={() => {
-                /* TO-DO */
-              }}
-              content={strings.BUTTON_SUMMARY}
-            />
+            <Button color="blue" onClick={summaryButtonHandler} content={strings.BUTTON_SUMMARY} />
           </Grid.Column>
         </Grid.Row>
       </Grid>
--- a/src/containers/Application/ApplicationSummary.tsx
+++ b/src/containers/Application/ApplicationSummary.tsx
@@ -66,7 +66,7 @@ const ApplicationSummary: React.FC<ApplicationProps> = ({ structure }) => {
       <Header as="h1" content={strings.TITLE_APPLICATION_SUBMIT} />
       {Object.values(sections).map((section) => (
         <SectionWrapper
-          key={`ApplicationSection_${section.details.code}`}
+          key={`ApplicationSection_${section.details.id}`}
           section={section}
           responsesByCode={responsesByCode as ResponsesByCode}
           serial={info.serial}
--- a/src/containers/Application/ApplicationSummary.tsx
+++ b/src/containers/Application/ApplicationSummary.tsx
@@ -53,11 +53,8 @@ const ApplicationSummary: React.FC<ApplicationProps> = ({ structure }) => {
   }, [isValidated])
 
   useEffect(() => {
-    console.log('Submitted', submitted)
-    if (submitted) push(`/application/${fullStructure?.info.serial}/submission`)
+    if (submitted) push(`/applicationNEW/${fullStructure?.info.serial}/submission`)
   }, [submitted])
-
-  console.log('fullStructure', fullStructure)
 
   if (error) return <Message error header={strings.ERROR_APPLICATION_PAGE} list={[error]} />
   if (submitError)
--- a/src/containers/Application/ApplicationSummary.tsx
+++ b/src/containers/Application/ApplicationSummary.tsx
@@ -17,7 +17,7 @@ interface ApplicationProps {
 const ApplicationSummary: React.FC<ApplicationProps> = ({ structure }) => {
   const { replace, push } = useRouter()
   const [shouldRevalidate, setShouldRevalidate] = useState(false)
-  const [isValidated, setIsValidated] = useState(false)
+  const [isRevalidated, setIsRevalidated] = useState(false)
   const {
     userState: { currentUser },
   } = useUserState()
@@ -38,7 +38,7 @@ const ApplicationSummary: React.FC<ApplicationProps> = ({ structure }) => {
       const { sectionCode, pageNumber } = fullStructure.info.firstStrictInvalidPage
       replace(`/applicationNEW/${fullStructure.info.serial}/${sectionCode}/Page${pageNumber}`)
     }
-    if (shouldRevalidate) setIsValidated(true)
+    if (shouldRevalidate) setIsRevalidated(true)
   }, [fullStructure])
 
   const handleSubmit = () => {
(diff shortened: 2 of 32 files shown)


src/containers/Application/ApplicationSummary.tsx
```diff
@@ -13,24 +15,51 @@ interface ApplicationProps {
...
-  const { replace } = useRouter()
+  const { replace, push } = useRouter()
+  const [shouldRevalidate, setShouldRevalidate] = useState(false)
+  const [isRevalidated, setIsRevalidated] = useState(false)
+  const {
+    userState: { currentUser },
+  } = useUserState()
   const { error, fullStructure } = useGetFullApplicationStructure({
     structure,
+    shouldRevalidate,
```
andreievg, line 26: I don't think we need to revalidate ? 

useGetFullApplicationStructure will do full validation first time it fires right ?

andreievg (changes requested):
A small comment about revalidation.

Also, I've noticed in the parent branch.  There was a change to types, that should propagate to this branch: `MethodRevalidate`  vs `MethodToCallOnRevalidation` and might be some others: (if you go to the type file diff here https://github.com/openmsupply/application-manager-web-app/pull/407/files)

src/containers/Application/ApplicationSummary.tsx
CarlosNZ: Yeah, but the intention was to run validation both on page load, and on Submit. You could argue that's overkill, but that's the way it was working in the current version, and there is an outside chance that someone could leave their Summary page open for a long time before submitting, in which time something became invalid.

andreievg: hmm, this would require a full refetch vs just setting shouldRevalidate, right ?

I am ok with approving this one for now, and we can polish of this use case closer to MVP

andreievg:
> I guess this is because there are still async functions running in useGetFullStructure hook at the time the page re-direct happens (which unmounts the components that are calling this hook). I've no idea what we should do about it though -- any ideas?

Yeah it would be something along those lines, use effect method can return a clean up function btw. We can deal with this error close to mvp time, thanks for noticing it

src/containers/Application/ApplicationSummary.tsx
CarlosNZ: > hmm, this would require a full refetch vs just setting shouldRevalidate, right ?

Setting `shouldValidate` from `false` to `true` forces the hook to re-run, so you get a new fullStructure with validation, which then triggers the rest of the process.

andreievg: but if you had the summary page open in a different tab, you wouldn't get the refetch of current data

CarlosNZ: So you mean, if someone has their Summary page open, and then changes some entries on another device or different tab, then presses "Submit" on the original Summary page, the validation wouldn't be from the new data?

Hmm, seems like a pretty edge case. I guess you could also have a similar problem if someone left their Summary page open and then Submitted from another device -- they'd be able to submit again from the original Summary page. 

The only way round that would be to force the original `useLoadApplication` hook to re-run, not just the `useGetFullStructure` hook (so it updates status, etc.) -- but doing that seems beyond the scope of this issue, don't you think?

andreievg: Yeah, we would need to think this one through (but I think it can wait for now).

Btw I thought my example above was the same as yours below (if now what was the use case for your one?):

> You could argue that's overkill, but that's the way it was working in the current version, and there is an outside chance that someone could leave their Summary page open for a long time before submitting

CarlosNZ: > Btw I thought my example above was the same as yours below (if now what was the use case for your one?):

It's not so much about the application responses changing, more for the case when the existing responses are no longer valid (e.g. name is no longer unique).

andreievg (approved):

src/containers/Application/ApplicationSummary.tsx
andreievg: ahh ok, this case what you've done works nicely

CarlosNZ:
> Also, I've noticed in the parent branch. There was a change to types, that should propagate to this branch: `MethodRevalidate` vs `MethodToCallOnRevalidation` and might be some others: (if you go to the type file diff here https://github.com/openmsupply/application-manager-web-app/pull/407/files)

I think we're all good on this front now, right? (Have merged master into here, and seems cool)

andreievg:
> I think we're all good on this front now, right? (Have merged master into here, and seems cool)

There are duplicate types still i think

src/containers/Application/ApplicationSummary.tsx
```diff
@@ -13,24 +15,51 @@ interface ApplicationProps {
...
+
+  useEffect(() => {
+    // This should only run after user has clicked Submit
+    // and revalidation has completed successfully
+    if (!fullStructure?.responsesByCode) return
+    submit(Object.values(fullStructure?.responsesByCode))
+  }, [isRevalidated])
+
+  useEffect(() => {
+    if (submitted) push(`/applicationNEW/${fullStructure?.info.serial}/submission`)
```
andreievg, line 56: if we wait for submission, what could happen is this:
* mutation fires up to change application trigger
* the first getLoadApllicationNEW re-fires because getApplicationNEW query refetched (hopefully)
* then loading is set to true
* the component tree is now gone this component doesn't exist anymore
* when loading is finished components are rebuild and this component has new state (so newer really => submits).

I think you should call navigation to submission straight after the submit or await for submit and call navigation (can await apolo mutations btw, not sure that useSubmitApplication.submit is though)

src/components/Application/Navigation.tsx
```diff
@@ -95,6 +95,15 @@ const Navigation: React.FC<NavigationProps> = ({
     })
   }
 
+  const summaryButtonHandler = () => {
+    requestRevalidation(({ firstStrictInvalidPage, setStrictSectionPage }: MethodToCallProps) => {
+      if (firstStrictInvalidPage) {
+        setStrictSectionPage(firstStrictInvalidPage)
+        sendToPage(firstStrictInvalidPage)
+      } else push(`/applicationNEW/${serialNumber}/summary`)
```
andreievg, line 103: if we restructure routing slightly, home, applicationPage and application summary, to have a wrapper that calls useGetFullApplicationStructure, then we can put the re-validation useEffect and the method there and pass down to summary, home and page. Then the logic in summary submit would be much simpler. What do you think ?

andreievg: ApplicationWrapper->
  ApplicationPageWrapper ->
      ApplicationPage
      ApplicationHome
      ApplicationSummary
Application Submit
...Review

andreievg (changes requested):
i thought of something that may be causing the issue you experience + a suggestion.

src/components/Application/Navigation.tsx
CarlosNZ: Righto, made a new branch based off this one with these suggested/discussed improvements:
https://github.com/openmsupply/application-manager-web-app/pull/424

PR #411, opened 2021-03-05
#378 New hook useGetReviewInfo

### LINKED ISSUE #378: Create hook useReviewInfo (rename if needed) to get all review_assignments for the current user and all associated reviews

nmadruga:
Fixes #378 

### Back-end branch
Use new back-end in [PR#229](https://github.com/openmsupply/application-manager-server/pull/229) to get `timeCreated` field for the current `status` in each review.

### Summary
To be continue in #379 (Hopefully PR today) and display all ReviewAssignment/Reviews associated to current user in the ReviewHome page. There will be no sections displayed yet, and the reason why the `getReviewInfo` query only return the number of ReviewQuestionAssignment so far (out of curiosity).  

### Changes
- Add query `getReviewInfo.query` to simply get all reviewAssignments based on current reviewer and application id
- Add hook `useGetReviewInfo` which will create the array of AssignmentDetails (new type based on existing one), with fields to display on #379 for each row of Review
- Passed on the `assignments` array to new `ReviewHome` page just logging the result for now
- Returns **NoMatch** in `ReviewWrapper` if no ReviewAssignment is found for current user
- Show error page if something goes wrong with query or loading while loading

### Unrelated changes
- Continue to add some "TODO: Remove this" to help us when removing old methods/hook/queries now replace to use the new structure.

### DIFF
--- a/src/containers/Review/ReviewWrapper.tsx
+++ b/src/containers/Review/ReviewWrapper.tsx
@@ -1,9 +1,11 @@
 import React from 'react'
 import { Route, Switch } from 'react-router'
-import { Header } from 'semantic-ui-react'
-import { NoMatch } from '../../components'
+import { Header, Message } from 'semantic-ui-react'
+import { Loading, NoMatch } from '../../components'
+import { useUserState } from '../../contexts/UserState'
+import useGetReviewInfo from '../../utils/hooks/useGetReviewInfo'
 import { useRouter } from '../../utils/hooks/useRouter'
-import { FullStructure } from '../../utils/types'
+import { AssignmentDetailsNEW, FullStructure } from '../../utils/types'
 
 interface ReviewWrapperProps {
   structure: FullStructure
@@ -13,13 +15,25 @@ const ReviewWrapper: React.FC<ReviewWrapperProps> = ({ structure }) => {
   const {
     match: { path },
   } = useRouter()
+  const {
+    userState: { currentUser },
+  } = useUserState()
 
-  console.log('Strcture', structure)
+  const { error, loading, assignments } = useGetReviewInfo({
+    applicationId: structure.info.id,
+    userId: currentUser?.userId as number,
+  })
+
+  if (error) return <Message error />
+
+  if (loading) return <Loading />
+
+  if (!assignments || assignments.length === 0) return <NoMatch />
 
   return (
     <Switch>
       <Route exact path={path}>
-        <ReviewHomeNEW />
+        <ReviewHomeNEW assignments={assignments} />
       </Route>
       <Route exact path={`${path}/:reviewId`}>
         <ReviewPageNEW />
@@ -34,7 +48,12 @@ const ReviewWrapper: React.FC<ReviewWrapperProps> = ({ structure }) => {
   )
 }
 
-const ReviewHomeNEW: React.FC = () => {
+interface ReviewHomeProps {
+  assignments: AssignmentDetailsNEW[]
+}
+
+const ReviewHomeNEW: React.FC<ReviewHomeProps> = ({ assignments }) => {
+  console.log(assignments) // To be continued in #379
   return <Header>REVIEW HOME PAGE</Header>
 }
 
--- a/src/containers/Review/ReviewWrapper.tsx
+++ b/src/containers/Review/ReviewWrapper.tsx
@@ -33,7 +33,7 @@ const ReviewWrapper: React.FC<ReviewWrapperProps> = ({ structure }) => {
   return (
     <Switch>
       <Route exact path={path}>
-        <ReviewHomeNEW assignments={assignments} />
+        <ReviewHomeNEW assignments={assignments} structure={structure} />
       </Route>
       <Route exact path={`${path}/:reviewId`}>
         <ReviewPageNEW />
@@ -50,9 +50,10 @@ const ReviewWrapper: React.FC<ReviewWrapperProps> = ({ structure }) => {
 
 interface ReviewHomeProps {
   assignments: AssignmentDetailsNEW[]
+  structure: FullStructure
 }
 
-const ReviewHomeNEW: React.FC<ReviewHomeProps> = ({ assignments }) => {
+const ReviewHomeNEW: React.FC<ReviewHomeProps> = ({ assignments, structure }) => {
   console.log(assignments) // To be continued in #379
   return <Header>REVIEW HOME PAGE</Header>
 }
--- a/src/containers/Review/ReviewWrapper.tsx
+++ b/src/containers/Review/ReviewWrapper.tsx
@@ -6,6 +6,7 @@ import { useUserState } from '../../contexts/UserState'
 import useGetReviewInfo from '../../utils/hooks/useGetReviewInfo'
 import { useRouter } from '../../utils/hooks/useRouter'
 import { AssignmentDetailsNEW, FullStructure } from '../../utils/types'
+import strings from '../../utils/constants'
 
 interface ReviewWrapperProps {
   structure: FullStructure
@@ -24,7 +25,7 @@ const ReviewWrapper: React.FC<ReviewWrapperProps> = ({ structure }) => {
     userId: currentUser?.userId as number,
   })
 
-  if (error) return <Message error />
+  if (error) return <Message error header={strings.ERROR_REVIEW_PAGE} list={[error]} />
 
   if (loading) return <Loading />
 
--- a/src/utils/graphql/queries/getReviewInfo.query.ts
+++ b/src/utils/graphql/queries/getReviewInfo.query.ts
@@ -6,12 +6,6 @@ export default gql`
       nodes {
         id
         timeCreated
-        reviewer {
-          id
-          username
-          firstName
-          lastName
-        }
         reviews {
           nodes {
             id
--- a/src/utils/graphql/queries/getReviewInfo.query.ts
+++ b/src/utils/graphql/queries/getReviewInfo.query.ts
@@ -13,9 +13,10 @@ export default gql`
         reviews {
           nodes {
             id
+            level
             status
             timeCreated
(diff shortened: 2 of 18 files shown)


src/utils/graphql/queries/getReviewInfo.query.ts
```diff
@@ -0,0 +1,26 @@
+import { gql } from '@apollo/client'
+
+export default gql`
+  query getReviewInfo($reviewerId: Int!, $applicationId: Int) {
+    reviewAssignments(condition: { reviewerId: $reviewerId, applicationId: $applicationId }) {
+      nodes {
+        id
```
andreievg, line 7: i think we also want level for display

andreievg: but that can be added in the issue of displaying the review assignments

src/utils/hooks/useGetReviewInfo.tsx
```diff
@@ -0,0 +1,63 @@
...
+          reviewAssignment.id
+        )
+
+      const stage = reviewAssignment.stage
+
+      // Extra field just to use in initial example - might conflict with future queries
+      // to get reviewQuestionAssignment
+      const totalAssignedQuestions = reviewAssignment.reviewQuestionAssignments.totalCount
+
+      const { id, status, timeCreated } = review
```
andreievg, line 39: it's quite possible that review will not exist

src/utils/hooks/useGetReviewInfo.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+      return {
+        id: reviewAssignment.id,
```
andreievg, line 42: we would want reviewAssignmentStatus

andreievg: we can also get timeCreated for reviewAssignment (to order them by that, or can order them by stage then level in the home page)

src/containers/Review/ReviewWrapper.tsx
```diff
@@ -13,13 +15,25 @@ const ReviewWrapper: React.FC<ReviewWrapperProps> = ({ structure }) => {
...
-  console.log('Strcture', structure)
+  if (loading) return <Loading />
+
+  if (!assignments || assignments.length === 0) return <NoMatch />
 
   return (
     <Switch>
       <Route exact path={path}>
-        <ReviewHomeNEW />
+        <ReviewHomeNEW assignments={assignments} />
```
andreievg, line 36: and structure

andreievg (changes requested):
A couple of minor comments. Looking good.

We would also want to have similar mechanism to the one in application load to make sure triggers on reviewAssignments and associated review are null (and refetch if not)

andreievg:
> To be continue in #379 (Hopefully PR today) 

Would be great to merge as much existing work as we can today before starting further work (i.e. deal with all of the outstanding reviews etc..)

nmadruga:
> make sure triggers on reviewAssignments and associated review are null (and refetch if not)

Yes, good point! I'll do that in the end of this review

src/containers/Review/ReviewWrapper.tsx
nmadruga: Done:  d70298b

src/utils/graphql/queries/getReviewInfo.query.ts
nmadruga: No problem. Added: 5620162

src/utils/hooks/useGetReviewInfo.tsx
nmadruga: Yes, very good point. I'll set it to `null` in case there is no review associated yet. I think that's how we have done in most places.

nmadruga: Done: b68e7f2 & bbde101

src/utils/hooks/useGetReviewInfo.tsx
nmadruga: Good call! Done: 9e6468d & f93bff3

src/utils/graphql/queries/getReviewInfo.query.ts
andreievg: we would want level from reviewAssignment since review might not exist

nmadruga:
Ok. It's ready now. Have included the trigger checking for reviews as well.

andreievg (approved):
Very good and precise. Thanks Nicole !

Can I ask you to note a couple of TODO (can do it in the other PR, happy to merge this one now). 
* Get level from review_assignment and add to result in hook (review might not exist, but we might want to display level in row headers before they are expanded.
* add check review_assignment trigger as well as review trigger in refetch logic

src/utils/graphql/queries/getReviewInfo.query.ts
```diff
@@ -0,0 +1,32 @@
...
+        status
+        timeCreated
+        reviews {
+          nodes {
+            id
+            level
+            status
+            timeCreated
```
andreievg, line 18: this was causing an issue

src/utils/graphql/queries/getReviewInfo.query.ts
```diff
@@ -0,0 +1,32 @@
...
+      orderBy: TIME_CREATED_DESC
+    ) {
+      nodes {
+        id
+        status
+        timeCreated
+        reviews {
+          nodes {
+            id
+            level
```
andreievg, line 16: i think it's better to get level from reviewAssignment

src/utils/hooks/useGetReviewInfo.tsx
```diff
@@ -0,0 +1,95 @@
...
+    if (!reviewAssigments) {
+      setIsFetching(false)
+      return
+    }
+
+    const reviews: Review[] = reviewAssigments.map(({ reviews }) => reviews.nodes[0] as Review)
+
+    // Checking if any of reviews trigger is running before refetching assignments
+    // This is done to get latest status for reviews (afte trigger finishes to run)
+    if (reviews.every(({ trigger }) => trigger === null)) {
```
andreievg, line 40: ` if (reviews.every((review) => !review || review?.trigger === null)) {` (otherwise forever loading if no reviews)

src/utils/graphql/queries/getReviewInfo.query.ts
andreievg: we don't seem to set this yet in schema, but definitely should

andreievg (changes requested):
Hi Nicole, sorry, while working on something with this branch notices a couple of errors

src/utils/hooks/useGetReviewInfo.tsx
```diff
@@ -0,0 +1,95 @@
...
+  const [isFetching, setIsFetching] = useState(true)
+  const [fetchingError, setFetchingError] = useState('')
+  const [refetchAttempts, setRefetchAttempts] = useState(0)
+
+  const { data, loading, error, refetch } = useGetReviewInfoQuery({
+    variables: {
+      reviewerId: userId,
+      applicationId,
+    },
+    notifyOnNetworkStatusChange: true,
```
andreievg, line 23: why do we need this btw ? I've noticed hook being called twice due to this one and i get ` Can't perform a React state update on an unmounted component.` warning with it (same on the applicationNew page).

I've removed it in my branch for now and doing this:
```
if(loading) return setIsFetching(true)
if (!data) return
```

src/utils/graphql/queries/getReviewInfo.query.ts
nmadruga: Hi, have you used the back-end branch (as described at the top) as well? And ran `yarn database_init`?

src/utils/graphql/queries/getReviewInfo.query.ts
nmadruga: Yea, good point

nmadruga: 532de2d

src/utils/hooks/useGetReviewInfo.tsx
nmadruga: Ooh, no! :D Nice catch! Done: a391b36

src/utils/hooks/useGetReviewInfo.tsx
nmadruga: I have replace the check inside the useEffect with this suggested, although can't really remove the `notifyOnNetworkStatusChange` since that's used to keep the **useEffect** getting called after a `refetch` finishes. Otherwise it doesn't get called since the `loading` status doesn't change on the second (and next) calls. See more here: https://www.apollographql.com/docs/react/data/queries/#inspecting-loading-states

src/utils/graphql/queries/getReviewInfo.query.ts
andreievg: oops i didn't sorry

src/utils/hooks/useGetReviewInfo.tsx
andreievg: Still not quite right, but almost there ( i think my suggestion works nicely)
* if no reviews but review_assignments, there is an error that trigger is not defined
* if that error is fixed, but we are checking for .length of review_assignments, not review (since mapping over review assignments)

src/utils/hooks/useGetReviewInfo.tsx
andreievg: Thanks for clarifying

src/utils/hooks/useGetReviewInfo.tsx
andreievg: 8a870d0702f4065cb8faddbf016a946136f6cc34

andreievg (approved):

PR #416, opened 2021-03-06
#412 Text field error state fix

### LINKED ISSUE #412: Bug: Error state showing in form field even on first entry
Should only update on keystrokes after first loseFocus

CarlosNZ:
Fixes #412

Not sure why I changed this in the first place -- I think I thought the re-factored `isValid` field would be "true" when the input was still `null` if it wasn't strictly validated. Anyway, it's back to how it was now.

### DIFF
--- a/src/formElementPlugins/ApplicationViewWrapperNEW.tsx
+++ b/src/formElementPlugins/ApplicationViewWrapperNEW.tsx
@@ -80,6 +80,7 @@ const ApplicationViewWrapper: React.FC<ApplicationViewWrapperPropsNEW> = (props)
       isStrictPage,
       responses,
       evaluationParameters: { objects: { responses, currentUser }, APIfetch: fetch },
+      currentResponse,
     })
     setValidationState(newValidationState)
     return newValidationState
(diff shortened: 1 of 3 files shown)


andreievg (changes requested):
hmm, this makes the required field not appear with a is required error (old and new way)

would suggest something like this: https://github.com/openmsupply/application-manager-web-app/compare/suggested-fix

CarlosNZ:
> hmm, this makes the required field not appear with a is required error (old and new way)

Yeah, good point!

> would suggest something like this: https://github.com/openmsupply/application-manager-web-app/compare/suggested-fix

I think it can be done without creating a new State variable -- just check `currentResponse.txt` in the place where you were checking the `isValueEmpty` value. Seems to work as expected -- can you see any issues?

143ad29

andreievg:
>I think it can be done without creating a new State variable -- just check currentResponse.txt in the place where you were checking the isValueEmpty value. Seems to work as expected -- can you see any issues?

This looks way cleaner, thanks !

Just wondering if there is ever a need to check for null ? But otherwise approved

andreievg (approved):

andreievg:
btw that approval assumes it's tested now and working (i haven't checked it myself)

CarlosNZ:
> Just wondering if there is ever a need to check for null ? But otherwise approved

I don't think the .text field can ever be null. It gets a string assigned to it when first initiated (which is when user starts entering info)

> btw that approval assumes it's tested now and working (i haven't checked it myself)

Looking correct to me, strict and normal.

Issue #277, opened 2021-01-19
Consolidation decision step


nmadruga:
Logic to check if can go to the decision step for `is_last_level` reviewers.
Should show in the same page (good for when the Reviewer is also the Consolidator) or in a separated page.
Depending if all is Agreed or one is reviewResponse is Disagreed would display different Consolidation conclusion step - which would be implemented in next issues.

nmadruga:
We are calling this the "Decision box" which should be displayed at the bottom of the review page. When tehre are questions with disagreement (with reviewer or Applicant) and the available decision(s) displayed together before submssion.

The decision should be selected by default - some times only one option available. 
Have a look on Response_flow diagram (wiki) for what is to be selected in the decision box.

PR #415, opened 2021-03-06
#380 review page part 2

andreievg:
Branched from #380 
Also has #378 merged into it
Fixes: #405

~~**DIFF** would be **TOO BIG** until #378 and #386 are merged~~
**DIFF FIXED*

### Changes

* added activeSection hook and it's use in Application Summary and Review pages -> `useQuerySectionActivation` describes how it's implemented in the comments
* added scrollable hook (to add scrollable element and scroll to them when needed) -> `useScrollableAttachments` describes how it works and it's used in Review Page
* review submission button now should show a modal if at least on invalid page, and on OK modal will navigate to that page
* otherwise review submission button will ask for ok to submit (this will evolve further, in different PR)
* slight fixes to `useGetReviewInfo`
* added `ReviewSubmitButton` -> does all of the stuff related to submitting or trying to submit`

### Testing

* best to use #386 branch from back end
* Just create a new review using the old way for ABC123 application either with reviewer 1 or two
* Change to applicationNew and play around : )

PR #420, opened 2021-03-07
#277 Add decision area

andreievg:
Fixes: #277 

Branched from #415 
Requires back end: https://github.com/openmsupply/application-manager-server/pull/230

As per diagram: https://app.diagrams.net/#G1-Prk8w8lJRBg__EfK2TkfgFSQJDUmKhu
`NOTE` -> i didn't spend any time worrying about styles, this is just functionality PR

### Changes

**PageElements**
* stop rendering edit of review when review is submitted

**ReviewSubmit**
* new component that displays `ReviewSubmitButton`, `ReviewComment` and `ReviewDecision`
* uses hook `useGetDecisionOptions` for most of state control
* `ReviewSubmitButton` handles validation (with help from `useGetDecisionOptions`)

**useGetDecisionOptions
* added a hook to help with review submission and review option
* see diagram for explanation and type definition for interface

`note` I've added hook and modal folder to review component, this might be tidier or more confusing, what do you think ?

**useSubmitReviewHook**
* a hook that returns just one method to submit review, uses `useUpdateReviewMutation` (new) mutation, that takes 'patch' parameter as data. Most of the hook is constructing the patch (typescript validates that patch shape is correct0

**useUpdateReviewDecisionComment**
* simple hook that return update method, uses useUpdateReviewDecisionMutation 

**ReviewPage**
* stop render of ApproveAll button when not draft

**constants (strings)/message**
* added some strings

**createReview.mutation**
* now adds decision (no decision by default)

**sumitReview.mutation**
* changed mutation name (for the old version of it)

**updateReview/updateReviewDecisionMutation**
* added two new mutations, they just accept $data for patch

**useGetFullReviewInfo**
* now returns review information

### Testing

* should be able to create review with old and check out this functionality in new way
* use this query to get LOQ decision option `update review_decision set is_last_level = true`

~~TODO add diagram~~
https://app.diagrams.net/#G1-Prk8w8lJRBg__EfK2TkfgFSQJDUmKhu


nmadruga:
A few conflicts to be fixed here, I'm planning to review this one today. After finished to review PR b#415

andreievg:
> A few conflicts to be fixed here, I'm planning to review this one today. After finished to review PR #415
@nmadruga I've updated from parent branch, resolved conflicts and re-tested

andreievg:
I've updated from master, and resolved conflicts. seems to be working ok

nmadruga:
> I've added hook and modal folder to review component, this might be tidier or more confusing, what do you think ?

Idk I think it's not needed to have the `modal` folder because there aren't many modals for the review. Maybe batter to rename to `DecisionAreaModal` or `ModalDecisionArea` if you want to.

The hooks inside the components folder I also think it's a problem. 1st because it's logic related so should really be inside the `containers` if we agreed. But also because that's a NEW way of doing this project folder structure. I get the idea that you think it would be closer to where it is used, but to me it's so less confusing when things are separated by type (that's why I started with the hooks folder). If you think that's not working well it's ok to change but probably better to do for everything instead of just one and now have 2 ways of doing this which leads to chaos.

andreievg:
> Idk I think it's not needed to have the modal folder because there aren't many modals for the review. Maybe batter to rename to DecisionAreaModal or ModalDecisionArea if you want to.

> The hooks inside the components folder I also think it's a problem. 1st because it's logic related so should really be inside the containers if we agreed. But also because that's a NEW way of doing this project folder structure. I get the idea that you think it would be closer to where it is used, but to me it's so less confusing when things are separated by type (that's why I started with the hooks folder). If you think that's not working well it's ok to change but probably better to do for everything instead of just one and now have 2 ways of doing things which is how we get a messy project.

@nmadruga 
Sorry Nicole, didn't intend this causing you to type all of this up. I'll revert, but just to confirm:

-- components 
---- Review
------ ReviewSubmit
------ ModalDecisionArea (rename and move)
-- containers
---- Review
------ ReviewPage
-- utils
----  hooks
------ useGetCecisionOptions (move)
------ useSubmitReviewNew (move/rename)
------ useUpdateReviewDecisionComment (move)

src/components/Review/hooks/useGetDecisionOptions.ts
```diff
@@ -0,0 +1,111 @@
...
+
+        if (!isDraft) {
+          isVisible = value = option.code === decisionInStructure
+        } else if (canSubmitReviewAs !== Decision.NonConform || !reviewerNeedsToMakeDecision) {
+          isVisible = value = option.code === canSubmitReviewAs
+        } else {
+          isVisible =
+            option.code === Decision.NonConform || option.code === Decision.ListOfQuestions
+          value = false
+        }
```
nmadruga, line 75: I think I know what this block is doing but it's a bit unclear. I think it might be easier to read if the order of things is changed

src/utils/graphql/queries/getReviewInfo.query.ts
```diff
@@ -12,6 +12,7 @@ export default gql`
         status
         timeCreated
         level
+        isLastLevel
```
nmadruga, line 15: Why this one needed to be returned here?

src/utils/helpers/structure/generateReviewProgress.ts
```diff
@@ -1,4 +1,4 @@
-import { ReviewResponseDecision } from '../../generated/graphql'
+import { Decision, ReviewResponseDecision } from '../../generated/graphql'
```
nmadruga, line 1: `ReviewResponseDecision` not in use?

nmadruga (commented):
Half reviewed. Needs to understand how things are working first.

I agree with the proposed changes for files in your previous message, plus have added suggestion to move things to some new components and `ReviewSubmit` to the containers folder

andreievg:
@nmadruga -> https://app.diagrams.net/#G1-Prk8w8lJRBg__EfK2TkfgFSQJDUmKhu, this should help with logic.

I found a bug (i remember removing that bit of code : /), if LOQ or nonConform was chose, and then comment edited, it's reset

src/components/Review/ReviewSubmit.tsx
```diff
@@ -0,0 +1,206 @@
...
+  return (
+    <Form>
+      <TextArea
+        defaultValue={initialComment}
+        onChange={(_, { value }) => setComment(String(value))}
+        onBlur={() => submitComment(comment)}
+      />
+    </Form>
+  )
+}
```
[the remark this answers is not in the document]
andreievg: i've moved to a component: 55c3b4f96de84ca7ecb6e23b33c61bca534b791a, but not sure about passing through submitComment, is that something that we do with all components (no hooks in components?)

src/components/Review/ReviewSubmit.tsx
```diff
@@ -0,0 +1,206 @@
...
+              name={option.title}
+              value={option.code}
+              checked={option.value}
+              onChange={() => setDecision(option.code)}
+            />
+          </Form.Field>
+        ))}
+    </Form>
+  )
+}
```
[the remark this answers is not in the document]
andreievg: 55c3b4f96de84ca7ecb6e23b33c61bca534b791a

src/components/Review/ReviewSubmit.tsx
```diff
@@ -0,0 +1,206 @@
...
+
+const ReviewSubmit: React.FC<ReviewSubmitProps> = (props) => {
+  const { structure } = props
+  const thisReview = structure?.thisReview
+  const reviewDecision = thisReview?.reviewDecision
+  const {
+    decisionOptions,
+    getDecision,
+    setDecision,
+    getAndSetDecisionError,
```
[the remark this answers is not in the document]
andreievg: I thought to name it in such a way where it's obvious that this will not only get the decision error state, but set it (it does a check if it's ok to submit in terms of decision state, and if it's not ok it sets the state to show as error in the ReviewDecision component)

src/components/Review/hooks/useGetDecisionOptions.ts
```diff
@@ -0,0 +1,111 @@
...
+import { Decision, ReviewStatus } from '../../../utils/generated/graphql'
+import { ReviewState } from '../../../utils/types'
+import strings from '../../../utils/constants'
+
+type DecisionOption = {
+  code: Decision
+  title: string
+  isVisible: boolean
+  value: boolean
+}
```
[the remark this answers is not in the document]
andreievg: hmm, not sure what the difference is to any other type that's defined and used locally, happy to move it (but it's only used here)

src/components/Review/hooks/useGetDecisionOptions.ts
andreievg: 4f7295b948dad54ce8560d01506182ef161107a3

src/utils/helpers/structure/generateReviewProgress.ts
andreievg: it looks like it might be used

src/utils/graphql/queries/getReviewInfo.query.ts
andreievg: https://github.com/openmsupply/application-manager-web-app/blob/4f7295b948dad54ce8560d01506182ef161107a3/src/utils/hooks/useGetFullReviewStructure.tsx#L81-L87

do you think I should change this to use reviewAssignment passed through from previous query ?

src/utils/hooks/useGetReviewAssignment.tsx
```diff
@@ -87,7 +87,13 @@ const useGetReviewAssignment = ({ reviewerId, serialNumber }: UseGetReviewAssign
       setAssignment({
         id: currentAssignment.id,
         review: review
-          ? { id: review.id, status: review.status as ReviewStatus, stage: { id: 0, name: '' } }
+          ? {
+              id: review.id,
+              status: review.status as ReviewStatus,
+              stage: { id: 0, name: '' },
+              isLastLevel: false,
+              level: 0,
```
nmadruga, line 95: I'm still confused about this, why are you adding those dummy values here? If the ReviewAssigned has review. Shouldn't get this info from the review?

src/utils/hooks/useGetReviewInfo.tsx
```diff
@@ -23,6 +23,7 @@ const useGetReviewInfo = ({ applicationId, userId }: UseGetReviewInfoProps) => {
     notifyOnNetworkStatusChange: true,
     // if this is removed, there might be an infinite loading when looking at a review for the frist time, after clearing cache
     // it's either this or removing 'totalCount' in `reviewQuestionAssignments` from this query
+    // ended up removing totalCount from query and keeping this as nextFetchPolicy (was still seeing glitched with totalCount and had "can't update unmounted component error")
     fetchPolicy: 'network-only',
```
nmadruga, line 27: Maybe can be removed now?

nmadruga (commented):
Partial second-review.

Just made a few comments - it's way easier to follow changes now. I'll test and add review to remaining files (ReviewSubmit and useSubmitReviewNEW) if needed.

src/utils/hooks/useSubmitReviewNEW.ts
```diff
@@ -0,0 +1,69 @@
...
+      },
+    }
+
+    return {
+      trigger: Trigger.OnReviewSubmit,
+      reviewResponsesUsingId: {
+        updateById: reviewResponsesPatch,
+      },
+      reviewDecisionsUsingId: {
+        updateById: [reviewDecisionPatch],
```
nmadruga, line 53: Why the `reviewDecisionPatch` needs to go with brackets and the `reviewResponsesPatch` doesn't?

src/utils/hooks/useSubmitReviewNEW.ts
```diff
@@ -0,0 +1,69 @@
...
+      variables: {
+        reviewId: reviewId,
+        reviewPatch: constructReviewPatch(structure, decision),
+      },
+    })
+
+  return submitReview
+}
+
+export default useSubmitReview
```
nmadruga, line 69: Very nice and easy to follow : )

src/utils/graphql/queries/getReviewInfo.query.ts
nmadruga: I think this have been changed to get `isLastLevel` from review, so can be removed from reviewAssignment query?

src/containers/Review/ReviewPage.tsx
```diff
@@ -127,6 +128,8 @@ const ApproveAllButton: React.FC<{ page: PageNEW }> = ({ page }) => {
   }
 
   if (responsesToReview.length === 0) return null
+  if (responsesToReview.some((response) => response?.status !== ReviewResponseStatus.Draft))
```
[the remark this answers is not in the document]
andreievg: I think we do approve all if some response are draft. the number next to approve all would decrement until all are not draft

src/utils/graphql/queries/getReviewNew.query.ts
```diff
@@ -3,15 +3,9 @@ import { gql } from '@apollo/client'
 // For review responses linked to applicationResponses, we want to get either submitted responses
 // Or draft response that belong to current user
 export default gql`
-  query getReviewNew($reviewAssignmentId: Int!, $userId: Int!, $sectionIds: [Int!]) {
+  query getReviewResponses($reviewAssignmentId: Int!, $userId: Int!, $sectionIds: [Int!]) {
```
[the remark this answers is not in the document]
andreievg: ohh I've change the getReviewNEW to only return review response, and made the getReviewInfo to return all the stuff related to review (assignment, review status, etc..).  Otherwise we seemed to be getting information for review in both queries

src/utils/graphql/queries/getReviewResponses.query.ts
```diff
@@ -0,0 +1,35 @@
+import { gql } from '@apollo/client'
```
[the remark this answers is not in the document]
andreievg: I am not sure what's happening with the git, in the commit it was 'moved' (getReviewNEW -> getReviewResponses): Anyways i've redone it: https://github.com/openmsupply/application-manager-web-app/pull/420/commits/bfd340cc1efebc745103137e51b58055c025261b

src/utils/hooks/useGetDecisionOptions.ts
```diff
@@ -0,0 +1,116 @@
...
+
+    setIsDecisionError(isDecisionError)
+    return isDecisionError
+  }
+
+  return { decisionOptions, getDecision, setDecision, getAndSetDecisionError, isDecisionError }
+}
+
+export default useGetDecisionOptions
+export { DecisionOption }
```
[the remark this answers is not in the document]
andreievg: ohh true, sorry, my bad: aa2dc22d93368d9332abd707492b5128d10f7e15

src/utils/hooks/useGetFullReviewStructure.tsx
```diff
@@ -75,6 +62,9 @@ const useGetFullReviewStructure = ({
         structure: newStructure,
         sortedReviewResponses: review?.reviewResponses.nodes as ReviewResponse[], // Sorted in useGetReviewNewQuery
       })
+
+      // TODO talk to team and decide if we should put everything in full structure (join all of required info from reviewAssignment ?)
+      newStructure.thisReview = reviewAssignment.review
```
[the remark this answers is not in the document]
andreievg: sure, i ended up just using everything from reviewAssignment passed throug

src/utils/hooks/useGetReviewAssignment.tsx
andreievg: this is the old one, I didn't want to touch the query it was using. can do though

src/utils/hooks/useGetReviewInfo.tsx
andreievg: i think it needs to be network only, since it needs to wait for the trigger. plus i had constant issues with cache after database re-initilisation

andreievg: i can remove the comment, but thought it was useful

src/utils/hooks/useSubmitReviewNEW.ts
andreievg: yeah that's a bit confusing, i've changed here: d6e976a74133b2ffefd5fb48cf1a7b562c095e54, renamed to be reviewResponsesPatches

src/utils/hooks/useSubmitReviewNEW.ts
andreievg: phew. thanks Nicole, I was worried about this a little bit, it seemed good to me but didn't know if it would be clear to others

andreievg:
> I made a comment before about components not making logic decisions therefor not having calls to hooks - but on a second though I think this looks pretty good and we can still deal as a component. The division when using this design pattern is very difficult to get and I don't think we need to be so dramatic about what goes where - as long as it makes sense to us.

I agree, about logic decision etc within components.

I think if the logic decision are done within the hook, then it should be ok.

> I really liked how ReviewComment component this is encapsulating everything in it That's good.

I agree thanks ! I wished we can do this more often in the app, but it seems that a lot of state is dependent on computing something considering ALL of the data (like review responses change progress, and change decision). And even harder, application responses are changing each other ! Thus our top down structure approach.

You will note that I had to be careful not to include comment in reviewDecision part of getReviewInfo (to make sure the whole thing doesn't reload when comment changes)

src/containers/Review/ReviewPage.tsx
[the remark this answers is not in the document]
nmadruga: All good! It makes sense to me now

nmadruga (approved):
Thanks for all the changes and making so good work here!
 APPROVED \o/\o/\o/

andreievg:
> Thanks for all the changes and making so good work here!
APPROVED \o/\o/\o/

yay, thanks for reviewing

PR #424, opened 2021-03-08
#367 - part 3 - Refactor ApplicationPageWrapper

CarlosNZ:
Made this into a seperate PR since most of this is only tangentially related to the original issue, but feel free to merge into #409 if you'd rather look at it all at once (although I think most of the rest of that one is pretty much sorted).

As suggested by @andreievg in [this comment](https://github.com/openmsupply/application-manager-web-app/pull/409#discussion_r588142880), I've made an intermediate Wrapper between ApplicationWrapper and ApplicationPage/ApplicationSummary so they can share the re-validation logic. A large chunk of this PR is just moving logic from ApplicationPage to ApplicationPageWrapper (the new one).

Also, did some small logic improvements to useLoadApplication and useSubmitApplication to ensure that they work as expected. You should now be able to submit an application (with throttled Network speed, just to be sure) and the submission page should correctly load.

### DIFF
--- a/src/containers/Application/ApplicationPage.tsx
+++ b/src/containers/Application/ApplicationPage.tsx
@@ -6,6 +6,7 @@ import {
   SectionAndPage,
   MethodRevalidate,
   MethodToCallProps,
+  ApplicationProps,
 } from '../../utils/types'
 import useGetFullApplicationStructure from '../../utils/hooks/useGetFullApplicationStructure'
 import { ApplicationStatus } from '../../utils/generated/graphql'
@@ -19,11 +20,6 @@ import { PageElements } from '../../components/Application'
 import { useFormElementUpdateTracker } from '../../contexts/FormElementUpdateTrackerState'
 import checkPageIsAccessible from '../../utils/helpers/structure/checkPageIsAccessible'
 import { Navigation } from '../../components'
-
-interface ApplicationProps {
-  structure: FullStructure
-  responses?: ResponsesByCode
-}
 
 interface MethodToCall {
   (props: MethodToCallProps): void
--- a/src/containers/Application/ApplicationPage.tsx
+++ b/src/containers/Application/ApplicationPage.tsx
@@ -21,19 +21,9 @@ import { useFormElementUpdateTracker } from '../../contexts/FormElementUpdateTra
 import checkPageIsAccessible from '../../utils/helpers/structure/checkPageIsAccessible'
 import { Navigation } from '../../components'
 
-interface MethodToCall {
-  (props: MethodToCallProps): void
-}
-
-interface RevalidationState {
-  methodToCallOnRevalidation: MethodToCall | null
-  shouldProcessValidation: boolean
-  lastRevalidationRequest: number
-}
-
 const ApplicationPage: React.FC<ApplicationProps> = ({
   structure: fullStructure,
-  revalidate,
+  requestRevalidation,
   strictSectionPage,
 }) => {
   const {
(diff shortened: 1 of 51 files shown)


src/containers/Application/ApplicationPage.tsx
```diff
@@ -140,8 +80,8 @@ const ApplicationPage: React.FC<ApplicationProps> = ({ structure }) => {
         <Grid.Column width={4}>
           <ProgressBarNEW
             structure={fullStructure}
-            requestRevalidation={requestRevalidation}
-            strictSectionPage={strictSectionPage}
+            requestRevalidation={requestRevalidation as MethodRevalidate}
```
andreievg, line 83: hmm it looks like types can be improved, but that's being picky. 
not required change, but if we weren't so pressed for time would make props for ApplicationPage props requestRevalidation not optional and strictSection as SectionAndPage| null

then don't have to cast `as` in the page

src/containers/Application/ApplicationSummary.tsx
```diff
@@ -10,54 +9,53 @@ import { SectionWrapper } from '../../components/Application'
...
+    requestRevalidation &&
+      requestRevalidation(
+        async ({ firstStrictInvalidPage, setStrictSectionPage }: MethodToCallProps) => {
+          if (firstStrictInvalidPage) {
+            const { sectionCode, pageNumber } = firstStrictInvalidPage
+            setStrictSectionPage(firstStrictInvalidPage)
+            replace(`/applicationNEW/${fullStructure.info.serial}/${sectionCode}/Page${pageNumber}`)
+          } else {
+            try {
+              fullStructure?.responsesByCode &&
```
andreievg, line 47: hmm any reason to check responsesByCode ?

src/containers/Application/ApplicationSummary.tsx
```diff
@@ -10,54 +9,53 @@ import { SectionWrapper } from '../../components/Application'
...
-
-const ApplicationSummary: React.FC<ApplicationProps> = ({ structure }) => {
+const ApplicationSummary: React.FC<ApplicationProps> = ({
+  structure: fullStructure,
+  requestRevalidation,
+}) => {
   const { replace, push } = useRouter()
-  const [shouldRevalidate, setShouldRevalidate] = useState(false)
-  const [isRevalidated, setIsRevalidated] = useState(false)
+  const [error, setError] = useState(false)
```
andreievg, line 17: not seeing this being set

andreievg (approved):
Approving this one to go straight into master as I've reviewed #367-2-submit-functionality, and this was the only remaining change request

src/containers/Application/ApplicationSummary.tsx
CarlosNZ: Cos Typescript complains that responsesByCode could be undefined in the next statement.

src/containers/Application/ApplicationPage.tsx
CarlosNZ: Yeah, the problem is we're using the same type (ApplicationProps) for both Wrappers, and the top-most one doesn't have those last two properties to pass down.

src/containers/Application/ApplicationSummary.tsx
CarlosNZ: Line 51 -- in the `catch` block ;)

andreievg: oops, my bad

src/containers/Application/ApplicationSummary.tsx
andreievg: sweet

Issue #421, opened 2021-03-08
is visible to applicant logic on front end


andreievg:
Part of epic https://github.com/openmsupply/application-manager-web-app/issues/94 

as per diagram `LOQ proposal diagram`.

Back End: is_visible_to_applicant (boolean) and recommend_applicant_visibility (enum, `Original Response Visible to Applicant`, `Original Response Not Visible to Applicant` <- default)

Front End (on final submission)

* lvl 1 ?
   *  is non_conform ? Original Response Visible to Applicant : Original Response Not Visible to Applicant
 * lvl > 1
   * is agree ?
       * is original response non_conform ? Original Response Visible to Applicant : Original Response Not Visible to Applicant
    * is no agree then ` Original Response Not Visible to Applicant`

PR #425, opened 2021-03-08
#421 applicant review response visibility

andreievg:
branched from #277 
Fixes: #421 

### Changes

**useSubmitReview**
* now has `recommendedApplicationVisibility` 
**updateReviewResponse**
* new hook, uses `useUpdateReviewResponseMutation`
* computes `recommendedApplicationVisibility`

**DecisionAreaNew**
* now uses `recommendedApplicationVisibility`

**getReviewNew** 
*now has originalReponseId and originalResponse (for `updateReviewResponse` to compute `recommendedApplicationVisibility` for consolidator

**updateReviewResponse.mutation**
* now also saves `recommendedApplicantVisibility`

While testing noticed that reviewQuestionAssignment isn't set on reviewResponse during reviewResponse creation in creating review, this will need to be done (awaiting PR for new review home page)

### Testing
* create new review for ABC123 (using old way)
* change to new
* make some approvals and declines
* run `select * from review_response`, notice recommended_applicant_visibility


andreievg:
@CarlosNZ do you mind having a look at this one, it's somewhat related to backed changes you've reviewed

CarlosNZ:
Just one quick comment from testing: When a question is "Approved", but they leave a comment as well, shouldn't that also go to "Visible to Applicant"?

src/utils/hooks/useUpdateReviewResponseNEW.ts
```diff
@@ -0,0 +1,57 @@
...
+  ReviewResponseDecision,
+  ReviewResponseRecommendedApplicantVisibility,
+  useUpdateReviewResponseMutation,
+} from '../generated/graphql'
+
+// below lines are used to get return type of the function that is returned by useUpdateReviewResponseMutation
+type UseUpdateReviewMutationReturnType = ReturnType<typeof useUpdateReviewResponseMutation>
+type promiseReturnType = ReturnType<UseUpdateReviewMutationReturnType[0]>
+
+// hook used to update review response, comment and decision, use as per type definition belo
```
CarlosNZ, line 12:
```suggestion
// hook used to update review response, comment and decision, use as per type definition below
```
Typo

src/utils/hooks/useUpdateReviewResponseNEW.ts
```diff
@@ -0,0 +1,57 @@
+import {
+  ReviewResponse,
+  ReviewResponseDecision,
+  ReviewResponseRecommendedApplicantVisibility,
+  useUpdateReviewResponseMutation,
+} from '../generated/graphql'
+
+// below lines are used to get return type of the function that is returned by useUpdateReviewResponseMutation
+type UseUpdateReviewMutationReturnType = ReturnType<typeof useUpdateReviewResponseMutation>
+type promiseReturnType = ReturnType<UseUpdateReviewMutationReturnType[0]>
```
CarlosNZ, line 10: Should we keep consistency with the standard that types and interfaces are written with capitals? So: `PromiseReturnType`?

src/utils/hooks/useUpdateReviewResponseNEW.ts
```diff
@@ -0,0 +1,57 @@
...
+    return ReviewResponseRecommendedApplicantVisibility.OriginalResponseNotVisibleToApplicant
+
+  const isLevelOne = reviewResponse.originalReviewResponseId === reviewResponse.id
+
+  if (isLevelOne) {
+    if (reviewResponse.decision === ReviewResponseDecision.Decline)
+      return ReviewResponseRecommendedApplicantVisibility.OriginalResponseVisibleToApplicant
+
+    return ReviewResponseRecommendedApplicantVisibility.OriginalResponseNotVisibleToApplicant
+  }
```
CarlosNZ, line 47: What about if there is only one level? The responses should become visible to applicant in that case, shouldn't they?

src/utils/hooks/useUpdateReviewResponseNEW.ts
```diff
@@ -0,0 +1,57 @@
...
+  // consolidator
+  if (
+    reviewResponse.decision === ReviewResponseDecision.Agree &&
+    reviewResponse.originalReviewResponse?.decision === ReviewResponseDecision.Decline
+  )
+    return ReviewResponseRecommendedApplicantVisibility.OriginalResponseVisibleToApplicant
```
CarlosNZ, line 53: Just re-iterating the comment I made elsewhere. If the original reviewer has made a comment, but Approved, shouldn't this be visible to applicant as well? So maybe this should check for a non-empty comment on originalReviewResponse rather than "Declined" decision?

src/utils/hooks/useUpdateReviewResponseNEW.ts
```diff
@@ -0,0 +1,57 @@
...
+  useUpdateReviewResponseMutation,
+} from '../generated/graphql'
+
+// below lines are used to get return type of the function that is returned by useUpdateReviewResponseMutation
+type UseUpdateReviewMutationReturnType = ReturnType<typeof useUpdateReviewResponseMutation>
+type promiseReturnType = ReturnType<UseUpdateReviewMutationReturnType[0]>
+
+// hook used to update review response, comment and decision, use as per type definition belo
+// also computes and updates recommendedApplicantVisibility
+type useUpdateReviewResponse = (
```
CarlosNZ, line 14: Same here -- capitalise type to `UseUpdateReviewResponse` ?

CarlosNZ (changes requested):
So this is just for generating the recommended_applicant_visibility? Changing "is_visible_to_applicant" will be done elsewhere, right?

andreievg:
> So this is just for generating the recommended_applicant_visibility? Changing "is_visible_to_applicant" will be done elsewhere, right?

Back end for this already done and merged and in core actions :)

andreievg:
> Just one quick comment from testing: When a question is "Approved", but they leave a comment as well, shouldn't that also go to "Visible to Applicant"?

From my understanding, no, but we have an option of adding checkbox in the future, mutation is done within the component that will have the checkbox

src/utils/hooks/useUpdateReviewResponseNEW.ts
andreievg: indeed !
bca100f36430f3126044b34e9d2b4676f5b83bdb

src/utils/hooks/useUpdateReviewResponseNEW.ts
andreievg: Yeah that isLevelOne clause should deal with it

src/utils/hooks/useUpdateReviewResponseNEW.ts
andreievg: I've made a seperate comment about it, hope it makes sense? https://github.com/openmsupply/application-manager-web-app/pull/425#issuecomment-799824628

src/utils/hooks/useUpdateReviewResponseNEW.ts
andreievg: 7979039a198be6cc7dfb1299e935770add5a1bd0, how did that even work lol, looks like type and variables can share names

CarlosNZ:
> > Just one quick comment from testing: When a question is "Approved", but they leave a comment as well, shouldn't that also go to "Visible to Applicant"?
> 
> From my understanding, no, but we have an option of adding checkbox in the future, mutation is done within the component that will have the checkbox

But why would you ever write a comment for an approved response if it wasn't going to be visible to the applicant? If we're not going to allow the applicant to see it, then shouldn't we disable the comment input form while the decision is "Approved"?

src/utils/hooks/useUpdateReviewResponseNEW.ts
CarlosNZ: Okay, sure, I think I was confused for the same reason as the other comment -- I thought comments for Approved responses should also be visible to applicant.

andreievg:
> But why would you ever write a comment for an approved response if it wasn't going to be visible to the applicant? If we're not going to allow the applicant to see it, then shouldn't we disable the comment input form while the decision is "Approved"?

good point. as far as i am aware reviewers are not going to write comments for approved things. But this gives them an option, plus we can add 'is visible to applicant' checkbox, where this can be controlled (default false for approvals). I remember Valerio mentioning that LOQ is directed at non conformity, not approvals. 

I am not seeing any harm of having comment on approval though. Plus ability to do so was there from the start (in the decision modal), and I followed the flowchart that we've discussed previously https://app.diagrams.net/#G1fre--mkvNKg2bDWDITEmwMNWiggCAEjj, under rules section

andreievg:
@CarlosNZ we have a few branches that are testing LOQ functionality, and they require changing to this branch to produce LOQ (and the back to the branch being tested). Would be good to merge this one. I can make an issue to discuss the point about comment on approvals (so it's flagged), if you feel strongly about it.

CarlosNZ (approved):
> @CarlosNZ we have a few branches that are testing LOQ functionality, and they require changing to this branch to produce LOQ (and the back to the branch being tested). Would be good to merge this one. I can make an issue to discuss the point about comment on approvals (so it's flagged), if you feel strongly about it.

Cool, that's fine for now. I'd suggest making an issue so we at least flag that this needs discussing at some point. 

Just out of interest, why the reluctance to make approved comments non visible? I'd have thought it would make sense to allow it at this end, and we just restrict it in the UI (i.e. can't comment if approved) if we wanted to prevent it. To me that makes more sense than allowing it but not showing it.

andreievg:
> Just out of interest, why the reluctance to make approved comments non visible? I'd have thought it would make sense to allow it at this end, and we just restrict it in the UI (i.e. can't comment if approved) if we wanted to prevent it. To me that makes more sense than allowing it but not showing it.

The reluctance is that it's not breaking anything and shouldn't cause any UI confusion. And there is a small use case where comment could be made (approved -> 'pending check that documents match'). It would still be visible to consolidator etc.. 

> Cool, that's fine for now. I'd suggest making an issue so we at least flag that this needs discussing at some point.

https://github.com/openmsupply/application-manager-web-app/issues/451

Thanks for review !

Issue #369, opened 2021-02-24
Remove old router and files used for Application


nmadruga:
After all done in Sprint 22 for [Epic#21](https://github.com/openmsupply/application-manager-web-app/issues/101) we can remove previous router and Containers/Components of Application

CarlosNZ:
And rename all the NEW ones to the original names (files and components and routes)

nmadruga:
Need to update most components to be using ERROR_GENERIC (as started in PR #444)

New:
- ApplicationCreateNEW
- ApplicationWrapperNEW
- ApplicationSummary
- ApplicationWrapper

Continued:
- ListWrapper

Old (will get removed fully):
- ApplicationCreate
- ElementsBox
- ApplicationOverview
- ApplicationSubmission

Strings to be removed (from `constants` file):
- ERROR_APPLICATION_CREATE
- ERROR_APPLICATION_ELEMENTS
- ERROR_APPLICATION_OVERVIEW
- ERROR_APPLICATION_PAGE
- ERROR_APPLICATION_SECTION
- ERROR_APPLICATION_SUBMISSION
- ERROR_APPLICATION_SUBMIT
- ERROR_APPLICATIONS_LIST

andreievg:
@nmadruga. The reason I've changed to generic was because there were quite a few using application_overview error, I think ideally we should change from ERROR_GENERIC to more specific (i just didn't think it was part of the PR was working on, and wanted to do a quick fix)

nmadruga:
Ok. Regarding what is described in [comment](https://github.com/openmsupply/application-manager-web-app/issues/369#issuecomment-800693334) should only really remove `strings` that are not in use (after removing the old container/components) and verify if any remaining ERROR should be using the Generic one instead of a specific one.

nmadruga:
Partially done by PR #470

PR #444, opened 2021-03-15
#379 review home part 3

andreievg:
Fixes; #379 
Branched from: #379-review-home-part-2

### Changes

* noticed that the same error was being use (application overview) in a lot of pages, changed it to generic
* useGetFullReviewStructure to also store reviewQuestionAssignmentId in pageElement
* added method to useCreateReview `createReviewFromStructure` to do all of the logic of creating a review
* added StartReviewButton to `ReviewSectionRowAction`


src/utils/constants.ts
```diff
@@ -38,6 +38,7 @@ export default {
   ERROR_APPLICATION_SUBMIT: 'Problem submitting application',
   ERROR_APPLICATIONS_LIST: 'Problem loading applications list',
   ERROR_REVIEW_APPLICATION: 'Problem loading application for review',
+  ERROR_GENERIC: 'Something went wrong',
   ERROR_REVIEW_PAGE: 'Problem loading review',
   ERROR_REVIEW_OVERVIEW: 'Problem loading review homepage',
   ERROR_REVIEW_SECTION_ROW: 'Problem loading review section row',
```
nmadruga, line 44:
```suggestion
  ERROR_GENERIC: 'Something went wrong',
  ERROR_REVIEW_PAGE: 'Problem loading review',
  ERROR_REVIEW_OVERVIEW: 'Problem loading review homepage',
```

Other error strings to be replaced by Generic error (and removed from here) as part of changes done in #369 - Already added a comment to remember about doing this in there)

src/utils/hooks/useCreateReview.tsx
```diff
@@ -40,7 +46,24 @@ const useCreateReview = ({ reviewerId, serialNumber, onCompleted }: UseCreateRev
...
+
+    return await createReviewMutation({
+      variables: {
+        reviewAssigmentId,
+        applicationResponses,
+      },
+    })
+  }
+
+  const createReview = ({ applicationResponses }: CreateReviewProps) => {
```
nmadruga, line 66: This is here only for existing use? Good to leave a comment for this to be removed. I suggest:
`// TODO: Remove this`

If you do a global search there are many places with this comment to help us when deleting all the old stuff

nmadruga (commented):
Looks good. Just a few comments to have a look please

src/utils/constants.ts
andreievg: I've adde a comment, here; https://github.com/openmsupply/application-manager-web-app/issues/369#issuecomment-800704948, basically i noticed wrong error string was being used, and changed to generic (didn't really want to change all of them to correct ones, i think we should still use specific errors)

src/components/Review/ReviewSectionRowAction.tsx
```diff
@@ -45,4 +50,45 @@ const ReviewSectionRowAction: React.FC<ReviewSectionComponentProps> = ({
...
+    push,
+  } = useRouter()
+
+  const [startReviewError, setStartReviewError] = useState(false)
+
+  const { createReviewFromStructure } = useCreateReview({
+    reviewAssigmentId: assignment.id,
+    reviewerId: currentUser?.userId as number,
+    serialNumber: fullStructure.info.serial,
+    onCompleted: () => {},
```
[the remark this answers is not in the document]
andreievg: i think it's still used (in previous way), i've added a todo cd40cf8f3145b29aa37e8877578b52d05bd361d8

src/utils/hooks/useCreateReview.tsx
andreievg: cd40cf8f3145b29aa37e8877578b52d05bd361d8

src/utils/constants.ts
nmadruga: Oh! I think I haven't get the idea correctly sorry! I'll ament to my comment on the other issue

nmadruga: You can already remove not in use strings `ERROR_REVIEW_APPLICATION` and `ERROR_REVIEW_SECTION_ROW`. But not a big requirement anyway

nmadruga (approved):
LGTM : )

src/utils/constants.ts
andreievg: I was wondering if we can continue that discussion after lunch, and sum it up here: https://github.com/openmsupply/application-manager-web-app/issues/369

andreievg:
Thanks for review Nicole

Issue #441, opened 2021-03-15
UI to show applicant which questions need updating (Application Page & Application Summary)


nmadruga:
`Andrei comments`
I've been working on duplicating application responses and progress of changes : https://github.com/openmsupply/application-manager-server/pull/254 and https://github.com/openmsupply/application-manager-web-app/pull/445

This one would need both. 

I think it should be broken down as follows:

- [x] part 1, slight changes to ProgressBarNew, with test branch demonstrating how application response are duplicated and what can be used to display changes requested
- [x] part 2 - making Application Page UI as designs
- [x] part 3 - making ApplicationSummary UI as designs  ~ **In PR**

`end of Andrei comments`

Changes to the Application Page to display icons for "Changes requested" questions (from LOQ) and "Changed" questions by Applicant choice.

To be done (in parallel) or after #437.

**To be decided:**
* Should also cover applicant making changes to each question ?
* What to do about bottom progress, what progress is it showing ?

As new designs show:
* Progress bar to show "Invalid" in pages where there are Changes requested in LOQ
* ![Screen Shot 2021-03-15 at 2 50 17 PM](https://user-images.githubusercontent.com/16461988/111093450-d1bb3a00-859d-11eb-9685-bedd6a551106.png) icon showing below the "Changes required" before a new response is added.
* ![Screen Shot 2021-03-15 at 2 50 27 PM](https://user-images.githubusercontent.com/16461988/111093474-e7c8fa80-859d-11eb-87cd-20740df4acdd.png) icon showing below the "Changes required" after a new response is added.
* ![Screen Shot 2021-03-15 at 2 50 36 PM](https://user-images.githubusercontent.com/16461988/111093498-f8797080-859d-11eb-9bcc-acf034e35c10.png) icon showing below a response without changes required after a new response is added.
* ![Screen Shot 2021-03-15 at 2 55 02 PM](https://user-images.githubusercontent.com/16461988/111093696-7178c800-859e-11eb-9016-7646d0b96cf4.png) label + icon + date besides new responses added (in the Summary page).

### Designs
![Screen Shot 2021-03-15 at 2 47 44 PM](https://user-images.githubusercontent.com/16461988/111093597-3b3b4880-859e-11eb-819d-1d5291b9da79.png)
![Screen Shot 2021-03-15 at 2 47 10 PM](https://user-images.githubusercontent.com/16461988/111093598-3c6c7580-859e-11eb-86ae-cac5488155ee.png)
[...]

Issue #437, opened 2021-03-14
UI to show to applicant that changes are requested (Application home page)


nmadruga:
Changes to the Application home page to display buttons to "Update" sections with "Changes required" (from LOQ) and display progress on the Application after the Applicant clicks on the Update button to start to **make changes**.

## Designs
![Screen Shot 2021-03-15 at 2 48 08 PM](https://user-images.githubusercontent.com/16461988/111093974-15fb0a00-859f-11eb-8c37-45cbb959012a.png)
![Screen Shot 2021-03-15 at 2 46 48 PM](https://user-images.githubusercontent.com/16461988/111093978-17c4cd80-859f-11eb-9a05-c7b95277c4f3.png)

andreievg:
@nmadruga awesome thanks.

- [ ] Part 1 - Add helpers to structure, as per : https://app.diagrams.net/#G1gzlvn8D9ulQP8baqo93qspjsQqEDlxOa
- [ ] Part 2 - Add UI elements

For part 1:

![Screen Shot 2021-03-15 at 5 53 42 PM](https://user-images.githubusercontent.com/26194949/111105270-7dbd4f00-85b7-11eb-8e26-bf733af0d4ba.png)

![Screen Shot 2021-03-15 at 5 54 05 PM](https://user-images.githubusercontent.com/26194949/111105283-82820300-85b7-11eb-9844-182087599ea9.png)

andreievg:
@nmadruga I haven't assigned myself to this one, but did the first part (second part is open if you want to do it, i am going to do the https://github.com/openmsupply/application-manager-web-app/issues/314, as a prerequisite to https://github.com/openmsupply/application-manager-web-app/issues/441)

andreievg:
this PR: https://github.com/openmsupply/application-manager-web-app/pull/447, under testing, will give an idea what needs to be done for part 2

nmadruga:
I'll assign this to me then, should start right after looking at some review today.

PR #445, opened 2021-03-15
#437 part 1 add change request progress to applicant

andreievg:
Branched from: #379-review-home-part-3

### Changes

The aim was to add the following to structure:

```
// to pages and sections
changeRequestsProgress {
  totalChangeRequests: number
  doneChangeRequests: number
}

// to element
isChangeRequest
isChanged

// to full structure info
isChangeRequest
```

A lot of the changes are from GraphQL types (i think it's the first time the changes are generated for latest back end changes)

* added deep equal for previous and latest application review comparision
* relocated addSortedSectionsAndPages from getFullReviewStructure to be sharable
* adding latestApplicationResponse and previousApplicationResponse to application structure. Why adding latestApplicationResponse ? Well the response object type in element.response, doesn't actually have type, and in evaluation, where it's added, `any` type is used, it is then converted to FullResponse type. To change this properly quite a bit of refactor needs to be done, so I thought to add latestApplicationResponse to element, just for progress calculation
* add isChangeRequest and isChanged to element
* calculate progress for pages and sections
* getAllResponses changes are from https://github.com/openmsupply/application-manager-web-app/pull/425, needed here to restrict visibility of review responses
* generateResponse progress is altered, to consider change requests as not valid and incomplete (this should show them correctly in home progress)
* isLinear is changed to only apply when isChangeRequest for structure is false

### Testing

This is best demonstrated here (https://github.com/openmsupply/application-manager-web-app/pull/447) in the testing bit

Create a review for ABC123, submit as LOQ

Add console log to line 109 of useGetFullApplicationStructure (console.log(newStructure)). You should see changes requested progress in section/pages and if you look at elementsById should see isChangeRequest isChanged

Got to application home

PR #447, opened 2021-03-16
#441 part 1 application page LOQ

andreievg:
branched from: #314-duplicate-application-responses
requires back end: #314fe-requires-restart-status

the purpose of this branch was to give an introduction (via the Testing branch) to how work up to this branch can be used to full-fill #441 (application page and application summary for LOQ) and #437 (application home for change requested)

### Changes

* slight change to progress bar, something I've noticed when testing, when anything on a page is not valid we should show it as such (rather then blue circle)

### Testing

use branch #441-TESTING-part-1-application-page-LOQ
and back end branch above

create a review for "ABC123" with LOQ

the go to application home 

* notice that header is showing that 'there are issues'
* notice button you can click to start change requested (this should only be clicked once, after that use resume button)
* press CheckChangesRequested
* Notice progress bar statuses and json under the elements
* play around with changing things and trying to submit (logic should work just like it did previously in non linear, will consider change request that are not done as invalid and incomplete)

Here is the diff for reference: `https://github.com/openmsupply/application-manager-web-app/compare/%23441-part-1-application-page-LOQ...%23441-TESTING-part-1-application-page-LOQ`


nmadruga:
Wow! Tested this one briefly and it looks great.

I'll have a look on the code now, I think the 3 PRs are very dependent, but ideally we need the fix for at least the last one I noticed and added this [comment]() to stop the loop when loading the page with Changes requested so we can add one by one. Or if you think it's quicker to just make changes to this PR and we can do any more refinements only on this one and merge the others as is, will be good for me too - I think the later solution will make this easier to follow.

nmadruga (approved):
Looks good to me. I think once you have addressed comments done on #445 these is good to go. I'll have a quick look on the one before this as well.

andreievg:
> Wow! Tested this one briefly and it looks great.

thanks

> I'll have a look on the code now, I think the 3 PRs are very dependent, but ideally we need the fix for at least the last one I noticed and added this comment to stop the loop when loading the page with Changes requested so we can add one by one. Or if you think it's quicker to just make changes to this PR and we can do any more refinements only on this one and merge the others as is, will be good for me too - I think the later solution will make this easier to follow.

nice find, i fixed it there
https://github.com/openmsupply/application-manager-web-app/pull/445#issuecomment-800823866

PR #452, opened 2021-03-16
#381 review page after submitted review

### LINKED ISSUE #381: Update ReviewPage (container) to deal with Review submitted

nmadruga:
Fixes #381 

### Branched of PR
#444

### Back-end branch
Requires back-end [PR#255](https://github.com/openmsupply/application-manager-server/pull/255)
- Needed to fix some review assignment for testing
- Also added one Review as **Locked** for testing purpose (and found one not-resolved bug) - more details bellow

### Changes
- Fixed 2 small bugs in `useGetDecisionOptions` hook:
  - Options returned with wrong values when looking at submitted review (replaced `if` with `else if`)
  - Error message when only one `Conform` option was selected (`getAndSetDecisionError` should return `false`)
- Removed not needed path to `ReviewSummary` since we decided to show decision/List of questions in `ReviewPage`
- Layout for Submitted of Review comment and decision **components**

### Testing
- [ ] Use (old Review home page) with Reviewer 1 to start review of Application serial 23456 - Approve all and select **Conform**
- [ ] Refresh database and re-start review of Application serial 23456 - Decline one and select **Non-Conform**
- [ ] Refresh database and re-start review of Application serial 23456 - Decline one and select **LOQ** -> We still need a separated issue to display the List of Question here too (not implemented as part of this PR)
- [ ] Change to Reviewer 2 and view Review page of Application serial 34567 - it is **Locked**
  - Should show review/edit buttons and not allow to Submit (there is a bug which is not showing the Review buttons - will create a separated issue)


src/components/Review/ReviewDecision.tsx
```diff
@@ -1,36 +1,55 @@
...
+                checked={option.value}
+                onChange={() => setDecision(option.code)}
+              />
+            </Form.Field>
+          ))}
+      </Form>
+    )
+  // If review has been submitted/locked:
+  // Check if has one selected option
+  const visibleOptions = decisionOptions.filter((option) => option.isVisible)
```
andreievg, line 41: I wonder if we need the three checks here ?

what if we just do `decionOptions.find`, and if null or `no decision` don't display ?

andreievg (commented):
Look solid, just one comment. I want to do a quick test before approving.

thanks for fixing my bugs

src/components/Review/ReviewComment.tsx
```diff
@@ -8,22 +8,36 @@ import {
...
+  if (!reviewDecisionId) return null
+
   if (error) return <Message error title={strings.ERROR_GENERIC} list={[error]} />
 
   if (!data) return <Loading />
 
   const initialComment = data?.reviewDecision?.comment || ''
 
-  if (!isEditable) return <p>{initialComment}</p>
+  if (!isEditable)
```
andreievg, line 31: do we want to hide the comment ? (if it doesn't exist ?)

andreievg (commented):
looks to good to me, and ready to approve (did some testing). Just making a comment for now though, was to see what you think about the comments i've made

src/components/Review/ReviewDecision.tsx
nmadruga: Yes that's ok. I was kind of looking for an invalid case, but really unlikely. And in that case it would simply find the first option, which is fine too.

src/components/Review/ReviewComment.tsx
nmadruga: Tried something. But haven't tested. In a rush : P

nmadruga: Testing now

nmadruga: Looks good to me, going to merge this one now

Issue #433, opened 2021-03-11
File upload plugin - thumbnail previews


CarlosNZ:
From File upload management [Epic #41](#430)

Retrieve thumbnails of uploaded files and display as icons in ApplicationView and SummaryView. Clicking them opens the file in a new tab.

Issue #432, opened 2021-03-11
File upload plugin - basic functionality


CarlosNZ:
From File upload management [Epic #41](https://github.com/openmsupply/application-manager-web-app/issues/430)

- click to upload button
- send to endpoint, show Loading spinner
- delete icon
- button to "Add another" after first
- clickable download links for uploaded file(s)

Based on [spec](https://github.com/openmsupply/application-manager-web-app/issues/301) and [UI mockup](https://github.com/openmsupply/application-manager-web-app/issues/431)

Issue #431, opened 2021-03-11
File upload UI mockup


CarlosNZ:
From File upload management [Epic #41](https://github.com/openmsupply/application-manager-web-app/issues/430)

Create wireframe for Form element plugin

andreievg:
from @nmadruga on another issue: 

> Just one idea here from a website that O was using to add some docs uploaded (using mobile web site), could be used for reference...

> Before uploading file

![110858854-767c1400-831f-11eb-9874-7144886d943d (1)](https://user-images.githubusercontent.com/26194949/110861916-b34a0a00-8323-11eb-81db-1adaa1c2b7a0.png)
![110859002-a7f4df80-831f-11eb-946c-d07ede94a230](https://user-images.githubusercontent.com/26194949/110861928-b644fa80-8323-11eb-9d96-bdbfe99ac7f3.png)
![110859664-a2e46000-8320-11eb-8d34-613cc4e3d8d5](https://user-images.githubusercontent.com/26194949/110861946-ba711800-8323-11eb-84fd-67f7032fb72b.png)

CarlosNZ:
Here is Deb's latest mockup:
![Screen Shot 2021-03-15 at 12 28 58 PM](https://user-images.githubusercontent.com/5456533/111089443-cad9fa80-8590-11eb-960e-0831643da5af.png)

Couple quick comments:
- I think we agreed not do do drag-n-drop for MVP, right?
- I don't know how we do progress bar. The endpoint only returns a response once upload is complete. From my basic research, it seems like tracking the amount of data sent so you can calculate progress is not something that can easily be done in a Fetch POST request. There's packages like axios that make this easier, but probably best not to worry about this for now, right? My suggestion for now would be to show the standard "Loading" spinner in the place that the thumbnails will appear.
- Really like the delete-on-hover icon, seems like a good way to do it without cluttering the UI.

With that in mind, would something like this do for first demo: ?

(Original at: https://balsamiq.cloud/scs7giw/ponj59g/r93F3)

**Initial view:**
![Screen Shot 2021-03-15 at 1 26 08 PM](https://user-images.githubusercontent.com/5456533/111089805-4daf8500-8592-11eb-89ad-f907d5122c1b.png)

**2 files uploaded, one more currently uploading:**
![Screen Shot 2021-03-15 at 1 27 52 PM](https://user-images.githubusercontent.com/5456533/111089823-599b4700-8592-11eb-967a-8f705c9d8eff.png)

[...]

PR #454, opened 2021-03-17
#432 File Upload plugin

CarlosNZ:
First iteration of file upload plugin #432 (also includes #433). 

Requires back-end branch from [PR #258](https://github.com/openmsupply/application-manager-server/pull/258).

Implements spec from #431 as best I can for a first iteration:

Screen Shot 2021-03-17 at 3 08 17 PM

To test, start a **FeatureShowcase** application -- the file upload elements are in **Section 2 - Documents**.

Fixed up a few other little bugs along the way that cropped up:

- Elements weren't being sorted correctly overall (only within page). Added "OrderBy" to getApplicationNEW query to handle this.
- PageElements -- Summary block needed to filter out invisible items (they were invisible in their Plugin, but the PageElements wrapper was still creating a block for them)
- Description display was missing from DefaultSummaryView
- Checkbox plugin was missing `key`s for one of the conditional blocks
- The "Show Password" toggle in Password plugin was displaying the "required" asterisk *
- Added generic GET method to `fetchMethods` -- though didn't end up using this


src/components/Application/PageElements.tsx
```diff
@@ -64,27 +64,29 @@ const PageElements: React.FC<PageElementProps> = ({
...
-                  {element.category === TemplateElementCategory.Question && canEdit && (
-                    <Grid.Column floated="right" textAlign="right">
-                      <Button
-                        content={strings.BUTTON_SUMMARY_EDIT}
-                        size="small"
-                        as={Link}
-                        to={`/application/${serial}/${sectionCode}/Page${pageNumber}`}
-                      />
+          {elements
+            .filter(({ element }) => element.isVisible)
```
andreievg, line 68: i think this would be moved to happen before return (see this being needed for reviews also).

not required change

src/formElementPlugins/fileUpload/constants.ts
```diff
@@ -0,0 +1,9 @@
+export default {
+  ERROR_FILE_ALREADY_UPLOADED: 'File already uploaded',
```
andreievg, line 2: hmm not a required change but I think this would need to be parameters of the plugin (localised when configuring plugin)

src/utils/helpers/fetchMethods.ts
```diff
@@ -17,6 +17,18 @@ export async function postRequest(body: object, endpointUrl: string, headers: ob
   }
 }
 
-export async function getRequest(body: object, endpointUrl: string) {
-  // TO-DO: Generic GET request
+export async function getRequest(endpointUrl: string, headers: object = {}) {
```
andreievg, line 20: i saw your comment in the PR (as in it's not being used), but was wondering, what would be the use case for this ? Wouldn't 'fetch' be supplied to the plugins ?

src/utils/types.ts
```diff
@@ -365,7 +365,8 @@ interface ResponseFull {
   optionIndex?: number
   reference?: any // Not yet decided how to represent
   isValid?: boolean | null
-  hash?: string
+  hash?: string // Used in Password plugin
+  files?: any[] // Used in FileUpload plugin
```
andreievg, line 369: hmm, this is a bit wired, i just noticed the ResponseFull having `hash` and `optionIndex` and `files` etc.. I thought as far as the 'core' code base is concerned, only thing that plugin updates is `value` and maybe `textValue` ? Otherwise they are not really `plugins` if they add types to core

andreievg: I think it's fine if we make a seperate issue to fix this for all of the plugins

andreievg (commented):
just a a couple of comments so far, I haven't look as ApplicationView or SummaryView yet. Will try to get through the review within the next couple of hours

src/components/Application/PageElements.tsx
CarlosNZ: Yeah, good point. So put this at the top of the component:
`visibleElements = elements.filter(({ element }) => element.isVisible)`
And replace `elements` with `visibleElements` throughout PageElements?

src/formElementPlugins/fileUpload/constants.ts
CarlosNZ: Do you mean for all strings associated with a plugin? I did it like this also for Checkbox, etc. Thought this would do until we decide an "official" localisation strategy for plugins. (FWIW, I was imagining each plugin would have their own localisation files, but they'd "inherit" the language setting from the main app, however we do that.)

src/utils/helpers/fetchMethods.ts
CarlosNZ: The generic "postRequest" is used in the Login module. I thought I was going to use a GET request here, but didn't end up needing it. Not sure if it'll ever be needed -- we can delete later, perhaps?

src/utils/types.ts
CarlosNZ: Yes I agree. So would you suggest that each plugin imports the basic "ResponseFull" and extends it with custom fields in their own file?

src/components/Application/PageElements.tsx
andreievg: that would be great

src/formElementPlugins/fileUpload/constants.ts
andreievg: yeah let's flag it as discussion: https://github.com/openmsupply/application-manager-web-app/issues/455

src/utils/helpers/fetchMethods.ts
andreievg: sure, happy to keep it. My comment was, why we would abstract it further than passing through 'fetch' ? Was it so that authorisation can be set ? (in that case i would assume any type of wrapper around fetch would make fetch work as is but with extra bits added to it, i.e. plugin can still use fetch and set method, headers, etc.. but extra header is added in the wrapper)

src/utils/types.ts
andreievg: I am not sure how to best deal with it atm, would be good to have a think about it next week, made an issue: https://github.com/openmsupply/application-manager-web-app/issues/456

andreievg (changes requested):
### Requests

`As per comments`

For now the main comment is about UUI and types. And I think it would be better to have just one type instead of three.

i.e. 

```JS
const [fileData, setFileData] = useState<onType)[]>(   (initialValue?.file || []).map(constructFileData)

constructFileData = (fileInfo) => ({ 
   key: generateUUIDKey(),
  isLoading: false,
  isError: false,
     etc..,
   ...fileinfo,
})
}
```

And use constructFileData each to you add to newFileData

(but on save only save the juice bits)

Also when deleting, delete by UUID

### Nice to haves

`Not required but nice to have (can be done in another PR if recorded in issue), below are my "suggestions"`

* Can reformat to make things more readable (i.e. extract <RenderError>, <RenderFileThumbnail>, <UploadButton> <- including input field)
* Extract error checking and adding to a method, something like this (with better method/variable naming that example:
```JS
const errors = [
 {
 condition: (fileInfo) => fileData.find(file => fileInfo.name === file.filename),
 error: 'ERROR_FILE_ALREADY_UPLOADED'
}
...
]
const generate = (file) => {
 const error = errors.find( error => error.condition(file))
 // technically speaking error: true is not needed if errorMessage is present
 if(error) return constructFileData({ ...file, error: error.error, error:true })
 return constructFileData(file)
}
...
newFileData.unshift( generate(file))
```

pass through everything (apart from 'react', 'semantic-ui-react')

src/components/Application/PageElements.tsx
CarlosNZ: Done d5c49d4

src/utils/helpers/fetchMethods.ts
CarlosNZ: The reason for abstracting the Login request was so that the large fetch request, with all it's error handling and response processing, etc, could be reduced to just:
```
const loginResult: LoginPayload = await postRequest({ username, password }, loginURL)
```

And I figured it was generic enough that it might be useful elsewhere.

src/formElementPlugins/fileUpload/src/ApplicationView.tsx
```diff
@@ -0,0 +1,243 @@
...
+  // These values required for file upload query parameters:
+  const {
+    userState: { currentUser },
+  } = useUserState()
+  const {
+    query: { serialNumber },
+  } = useRouter()
+  const application_response_id = initialValue.id
+
+  const [fileData, setFileData] = useState<(FileInfo | FileLoading | FileError)[]>(
```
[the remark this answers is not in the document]
CarlosNZ: Yeah, makes sense. The only reason I did it like this is that the return fields (fileURL, etc.) are not available until after uploading is complete, so they'd have to be optional fields in the combined Type.

I think I'll try nest the "Response" data, something like this:

```
interface FileResponseData {  // This info stored in DB
  uniqueId: string
  filename: string
  fileUrl: string
  thumbnailUrl: string
  mimetype: string
}

interface FileInfo {
  key: string
  loading: boolean
  error: boolean
  errorMessage: string | null
  fileData: FileResponseData | null
}
```

And then I can just save fileInfo.fileData to the Response to save.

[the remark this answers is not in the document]
CarlosNZ: Done. 3c2ab40

Object shape not _exactly_ as above, but pretty close.

Couple of comments:
- `uniqueID` is now included in the `/upload` endpoint response (in another branch). It's needed for Action processing, and it seemed a bit hacky to try and re-parse it from the URL.
- `filename` is in both the Parent "FileInfo" object and the nested "FileResponseData", though technically one is from the user selection and one is returned from the upload endpoint. We need the filename while loading, but I really wanted to make the "FileResponseData" inner object an exact copy of the endpoint response and the saved database response.

src/formElementPlugins/fileUpload/src/ApplicationView.tsx
```diff
@@ -0,0 +1,243 @@
...
+          error: true,
+          errorMessage: strings.ERROR_UPLOAD_PROBLEM,
+        }
+      }
+      setFileData([...newFileData])
+    })
+  }
+
+  const handleDelete = async (filename: string) => {
+    setFileData(fileData.filter((file) => file.filename !== filename))
```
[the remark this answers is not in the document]
CarlosNZ: Done -- see [below comment](https://github.com/openmsupply/application-manager-web-app/pull/454#discussion_r595919084)

src/formElementPlugins/fileUpload/src/ApplicationView.tsx
```diff
@@ -0,0 +1,243 @@
...
+        />
+        <Segment basic textAlign="center">
+          <Button primary disabled={!isEditable} onClick={() => fileInputRef?.current?.click()}>
+            <Icon name="upload" />
+            {fileData.length === 0 ? strings.BUTTON_CLICK_TO_UPLOAD : strings.BUTTON_UPLOAD_ANOTHER}
+          </Button>
+        </Segment>
+        <List horizontal verticalAlign="top">
+          {fileData.map((file) => (
+            <List.Item key={file.filename} style={{ maxWidth: 150 }}>
```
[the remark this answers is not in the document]
CarlosNZ: > Done -- see [below comment](https://github.com/openmsupply/application-manager-web-app/pull/454#discussion_r595919084)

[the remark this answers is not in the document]
CarlosNZ: Do you think you should be allowed to upload two files with the same name? I had an explicit error check for this, as I thought that would be a no-no. How do we tell if they've added a different file with the same name (from another location) as opposed to just adding the exact same file twice? We don't want to allow the latter case, surely?

(Only way I can see to do it is to make the "key" a MD5 checksum or something, but that seems a bit overkill for a pretty edge case)

CarlosNZ:
> ### Nice to haves
> 
> `Not required but nice to have (can be done in another PR if recorded in issue), below are my "suggestions"`
> 
>     * Can reformat to make things more readable (i.e. extract `<RenderError>`, `<RenderFileThumbnail>`, `<UploadButton>` <- including input field)
> 
>     * Extract error checking and adding to a method, something like this (with better method/variable naming that example:
> 
> 
> ```js
> const errors = [
>  {
>  condition: (fileInfo) => fileData.find(file => fileInfo.name === file.filename),
>  error: 'ERROR_FILE_ALREADY_UPLOADED'
> }
> ...
> ]
> const generate = (file) => {
>  const error = errors.find( error => error.condition(file))
>  // technically speaking error: true is not needed if errorMessage is present
>  if(error) return constructFileData({ ...file, error: error.error, error:true })
>  return constructFileData(file)
> }
> ...
> newFileData.unshift( generate(file))
> ```
> 
> pass through everything (apart from 'react', 'semantic-ui-react')

Okay, will address the rest of these in the morn, they shouldn't take too long. Good suggestions, thanks.

src/formElementPlugins/fileUpload/src/ApplicationView.tsx
```diff
@@ -0,0 +1,243 @@
...
+  // currentResponse,
+  validationState,
+  onSave,
+  Markdown,
+  initialValue,
+}) => {
+  const { label, description, fileCountLimit, fileExtensions, fileSizeLimit } = parameters
+
+  // These values required for file upload query parameters:
+  const {
```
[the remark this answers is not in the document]
CarlosNZ: We should probably clarify exactly what the front-end plugins have access to, and how they get it. It's pretty clear in the back-end Actions now, but front-end is a bit vague. For example, I've used the global "config.json" file to get the server host details, and there isn't really any way to evaluate this as a parameter expression. It would need to passed in separately from the "parameters", which are evaluator expressions.

Perhaps we need a "globalData" object that gets passed down into all plugins, similar to the "applicationData" object in the back-end? That seems like the most straightforward solution.

It would contain (just off the top of my head):
```
{
 user: (id, username, etc)
 application: (id, serial, name, etc.)
 config: (whole config.json object)
 ... (more as we think of them/need them)
}
```
Point is, these values would need to be "static" and don't really change depending on the application state or plugin type, as opposed to "parameters" which are inherently dynamic.

CarlosNZ:
FYI -- have used nanoid package to generate the keys, so you'll need to run `yarn install` _within the plugin folder_ to get it working.

CarlosNZ:
Refactored error detection (mostly) as per suggestion:
19b8e4c

src/formElementPlugins/fileUpload/src/ApplicationView.tsx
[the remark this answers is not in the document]
CarlosNZ: Made an issue to discuss this: #458 
For now we'll leave as is?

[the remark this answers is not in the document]
andreievg: Thanks for the summary, will leave it for next week

andreievg (approved):
Thanks Carl, looks great!

src/formElementPlugins/fileUpload/src/ApplicationView.tsx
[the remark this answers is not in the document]
andreievg: Hmm, i think the check you have currently is enough. From there on it's up to the reviewer

PR #461, opened 2021-03-18
#437 part 2 Applicant home show changes requested

### LINKED ISSUE #437: UI to show to applicant that changes are requested (Application home page)

nmadruga:
Fixes #437 

### Back-end requirement
requires back end: `#314fe-requires-restart-status`

### Changes
- Basically moved things around in `SectionWrapper`, since that's only used for Application I decided it would be simpler to just use 2 different components to render each Section based on: isChangesRequest or not
- Each Section action show very differently depending if
1. Application has **ChangesRequested**:
   - If there is changes in section show **Update** button action with number of question to update (remaining)
   - If no remaining updates to do OR none requested - Don't show anything (if not restarted)
   - If no remaining updates to do OR none requested - And restarted (Draft) - Show edit icon
2. Application is **Draft** (before first submission) - that's the `ActionGeneral` (couldn't think of any better naming):
 Same as before...
   - If section is not completed **Continue** button
   - If no remaining required question (and before some invalid section) - Show edit icon
- Now the time we run the `restartApplication` that will change the status from **ChangesRequested** to **Draft** is the first time the use clicks on one of the **Update** buttons. After that it would only load the page with invalid questions.
- Everything else is kept almost the same - `Progress` would show only when Application is in **Draft** of **ChangesRequested**. Didn't seem likely we would want to display progress in any other situation (to the Applicant).

### Remaining TODO
~~- I'm still trying to figure out problem with mutation, but thought to already post this PR so you could have a look~~ FIXED


andreievg:
Will check it out after lunch, thanks Nicole, btw in the re-review, this is what i did for mutation and navigation: 

https://github.com/openmsupply/application-manager-web-app/blob/a3a8aa7140ebe013b0153d5f6b09e9f1a585d1d9/src/components/Review/ReviewSectionRowAction.tsx#L93-L96

https://github.com/openmsupply/application-manager-web-app/blob/a3a8aa7140ebe013b0153d5f6b09e9f1a585d1d9/src/components/Review/ReviewSectionRowAction.tsx#L76-L88

src/containers/Application/ApplicationHome.tsx
```diff
@@ -45,25 +49,32 @@ const ApplicationHome: React.FC<ApplicationProps> = ({ structure, template }) =>
...
       <>
+        <ChangesRequestedTitle status={current?.status} isChangeRequest={isChangeRequest} />
         <Segment>
           <Header as="h5">{strings.SUBTITLE_APPLICATION_STEPS}</Header>
           <Header as="h5">{strings.TITLE_STEPS.toUpperCase()}</Header>
           <SectionsProgress
-            canEdit={canUserEdit}
-            sections={fullStructure.sections}
+            changesRequested={isChangeRequest}
+            draftStatus={current?.status === ApplicationStatus.Draft}
```
andreievg, line 66: i think we can extract this to a variable (looks like it's used in a few places)

src/containers/Application/ApplicationHome.tsx
```diff
@@ -45,25 +49,32 @@ const ApplicationHome: React.FC<ApplicationProps> = ({ structure, template }) =>
...
+            sections={sections}
             firstStrictInvalidPage={firstStrictInvalidPage}
+            restartApplication={async ({ sectionCode, pageNumber }) => {
+              await restartApplication(fullStructure)
```
andreievg, line 70: would be good to do some error checking (as per example in the first comment in this PR)

andreievg (changes requested):
Looks very good, thanks Nicole, I have a couple of tiny suggestions

andreievg:
I think PR is pretty solid, there are some comment that were made, i'll transfer them to an issue. This PR would be very useful to be merged in some of the parallel branches, thus approving and merging : )

https://github.com/openmsupply/application-manager-web-app/issues/471

andreievg (approved):

Issue #463, opened 2021-03-18
Show Stage in ReviewSectionRow and filter by Stage


nmadruga:

andreievg:
The idea is to align it to theses designs: 

Note: `yourself` and `stage`

![Screen Shot 2021-03-18 at 6 17 06 PM](https://user-images.githubusercontent.com/26194949/111576865-2d8ff800-8816-11eb-979a-c7179603f738.png)

PR #467, opened 2021-03-18
#463 filter review home by stage and reviewer

andreievg:
Fixes: #463 

### Changes

* make sure stage id is available in full structure stage list
* add `< ReviewerAndStageSelection>` to show two drop downs, one for stage and one for reviewer
* add filters to `<ReviewHome>` to be set by `<ReviewerAndStageSelection>`

Was trying to not use useEffect, but it wold have required making `ReviewerAndStageSelection` a wrapper, since fullStructure is not available at the start

### Testing

Try with 34567/review


src/containers/Review/ReviewHome.tsx
```diff
@@ -55,10 +66,96 @@ const ReviewHome: React.FC<ReviewHomeProps> = ({ assignments, structure }) => {
...
+  const changeFilters = (filterType: keyof Filters) => (_: any, { value }: any) => {
+    if (filters) setFilters({ ...filters, [filterType]: value })
+  }
+
+  if (!filters) return null
+
+  return (
+    <Grid columns="equal">
+      <Grid.Column floated="left">
+        Show tasks for{' '}
```
nmadruga, line 108: Plase localise : )

src/containers/Review/ReviewHome.tsx
```diff
@@ -55,10 +66,96 @@ const ReviewHome: React.FC<ReviewHomeProps> = ({ assignments, structure }) => {
...
+    if (id === currentUserId || !id || !firstName || !lastName) return
+    reviewerOptions.push({
+      value: id,
+      key: id,
+      text: `${firstName} ${lastName}`,
+    })
+  })
+
+  return reviewerOptions
+}
```
nmadruga, line 159: I think those 2 should be moved to some helper function - for better readability of this. The name of the function already tells the story

src/containers/Review/ReviewHome.tsx
```diff
@@ -55,10 +66,96 @@ const ReviewHome: React.FC<ReviewHomeProps> = ({ assignments, structure }) => {
...
+      value: id,
+      text: title,
+    }))
+
+const getReviewerOptions = (assignments: AssignmentDetailsNEW[], currentUserId: number) => {
+  const reviewerOptions: { value: number; key: number; text: string }[] = [
+    {
+      value: ALL_REVIEWERS,
+      key: ALL_REVIEWERS,
+      text: 'All',
```
nmadruga, line 141: To localise 2

src/containers/Review/ReviewHome.tsx
```diff
@@ -55,10 +66,96 @@ const ReviewHome: React.FC<ReviewHomeProps> = ({ assignments, structure }) => {
...
+    },
+    {
+      value: currentUserId,
+      key: currentUserId,
+      text: 'Yourself',
```
nmadruga, line 146: To localise 3

src/containers/Review/ReviewHome.tsx
```diff
@@ -55,10 +66,96 @@ const ReviewHome: React.FC<ReviewHomeProps> = ({ assignments, structure }) => {
...
+      <Grid.Column floated="left">
+        Show tasks for{' '}
+        <Dropdown
+          options={getReviewerOptions(assignments, currentUser?.userId as number)}
+          value={filters?.selectedReviewer}
+          onChange={changeFilters('selectedReviewer')}
+        />
+      </Grid.Column>
+      <Grid.Column floated="right" textAlign="right">
+        STAGE{' '}
```
nmadruga, line 116: To localise 4

nmadruga (commented):
Just a few comments. Testing now

nmadruga:
Hmm testing the UI it has some non-expected behaviours. Selecting something on either one of the dropdown not always show the correct result... I wonder if this one is going to have a Part 2. Also should we use URL filter to display, what is the benefit in this case - I think the bugs might be related to page reload and selection of item in dropdown. Not sure, just made a quick test.

src/containers/Review/ReviewHome.tsx
andreievg: sorry, missed it: 92a4c9992b32a2d8370573f6d842417c595e4c94

src/containers/Review/ReviewHome.tsx
andreievg: sorry, missed it: 92a4c99

src/containers/Review/ReviewHome.tsx
andreievg: sorry, missed it: 92a4c99

src/containers/Review/ReviewHome.tsx
andreievg: sorry, missed it: 92a4c99

src/containers/Review/ReviewHome.tsx
andreievg: 75ee2014fa6a0379e4fc44773462dbf921da3088

src/containers/Review/ReviewHome.tsx
andreievg: Hmm, I was also thinking it's getting a bit much for ReviewHome page. What do you think, if i move the whole component `ReviewerAndStageSelection` along with it's helper methods, would that be ok ?

andreievg:
> Hmm testing the UI it has some non-expected behaviours. Selecting something on either one of the dropdown not always show the correct result... I wonder if this one is going to have a Part 2. Also should we use URL filter to display, what is the benefit in this case - I think the bugs might be related to page reload and selection of item in dropdown. Not sure, just made a quick test.

I am glad you pointed this out. Even thought my testing was not showing any bugs, but I've tried a bit harder:

After clearing cache, when selecting all there is a spasm (then the spasm goes away).

I was looking at network request (btw added this to see what operations are firing): 0ce66d56e9fb601346d1138fe2c1ecaf69776682

And every time after cache is reset, reviewGetInfo is re-fired (in the middle of getting reviewResponses), this was very similar to what we were seeing with totalCount on questionAssignments (except it was just causing spasms, not UI freezes).

I ended up figuring it out, for some reason, filtering of reviews in reviewResponse query, causes reviewGetInfo to re-fire (it must notice that different review.nodes are returned for reviewAssignment, so decided getReviewInfo needs to be re-fried). I think it's all fixed now (by filtering at reviewResponse not review end, as it should be anyway, to allow other reviewers looking at previously submitted reviewResponse):

15f85f84945fda9b0a12a21dda01bb4ca9db611a

src/containers/Review/ReviewHome.tsx
nmadruga: Yes, that will be nice

nmadruga:
Awesome! The feature seems to work very nicely now.

nmadruga (approved):
👍 Thanks for all changes and new feature.

Have just one minor comment to make - but something that can be changed afterwards (part of #356 maybe). It's not important for now. I noticed the page doesn't show 404 when Reviewer lvl 1 isn't assigned to something. I think it's because they can still see Assigned to All options. But I wonder if this shouldn't be visible (or if permissions would prevent this).

The page showing when loaded (no reviews assigned to Reviewer2):
Screen Shot 2021-03-21 at 5 54 32 PM


And when filtering assigned to ALL:
Screen Shot 2021-03-21 at 5 54 21 PM

andreievg:
Thanks fore review @nmadruga

andreievg:
> 👍 Thanks for all changes and new feature.

> Have just one minor comment to make - but something that can be changed afterwards (part of #356 maybe). It's not important for now. I noticed the page doesn't show 404 when Reviewer lvl 1 isn't assigned to something. I think it's because they can still see Assigned to All options. But I wonder if this shouldn't be visible (or if permissions would prevent this).

> The page showing when loaded (no reviews assigned to Reviewer2):

Hmm, at the very least it would be good to parameterise stage and user select (navigate to 'all', if user has no tasks for the application, and perhaps change stage to latest stage to latest stage of assignments they can see)

andreievg:
https://github.com/openmsupply/application-manager-web-app/issues/485

PR #468, opened 2021-03-18
#441 ApplicationPage with changes requested & other changes made warnings

### LINKED ISSUE #441: UI to show applicant which questions need updating (Application Page & Application Summary)

nmadruga:
Fixes #441 

### Branched of PR
#461 

### Changes
- Display in `ApplicationViewWrapper` a warning to either one of the possible icons (as described in issue #441):
  * Screen Shot 2021-03-15 at 2 50 17 PM icon showing below the "**Changes required**" before a new response is added.
  * Screen Shot 2021-03-15 at 2 50 27 PM icon showing below the "**Changes required**" after a new response is added.
  * Screen Shot 2021-03-15 at 2 50 36 PM icon showing below a response without changes required after a new response is added.
- Some styling added to allow icon to display closer to Response field and keep same place if not displaying

### Questions - might need changes in this PR:
~~- Border of field is not following pattern: 'red' for **Changes required** or 'blue' for changes done (required or not). I'm thinking to use the less styles for that. Might be tidier than what I have done for icons...~~ SOLVED using a div around question (when we display the warning)
- Couldn't find a proper semantic component to display the Icon + message. Thought abut using `Comment` and `Feed` but ended up with simpler solution of `<div><Icon/>text</div>` let me know if there is one that I could replace this with.
- When comments are too long would be nice to used a trimmed version and display full comment in pop up? Maybe a tidy up issue for later to add in App layout changes.


src/containers/Application/ApplicationPage.tsx
```diff
@@ -89,7 +89,7 @@ const ApplicationPage: React.FC<ApplicationProps> = ({
                 sectionCode === strictSectionPage?.sectionCode &&
                 pageNumber === strictSectionPage?.pageNumber
               }
-              canEdit
+              canEdit={current?.status === ApplicationStatus.Draft}
```
CarlosNZ, line 92: This was previously being done higher up and passed down in props. Any particular reason for putting it here?

src/formElementPlugins/ApplicationViewWrapperNEW.tsx
```diff
@@ -149,15 +149,72 @@ const ApplicationViewWrapper: React.FC<ApplicationViewWrapperPropsNEW> = (props)
...
+  const displayResponseWarning = !!currentReview || (isChanged && validationState.isValid)
+
   return (
     <ErrorBoundary pluginCode={pluginCode}>
       <React.Suspense fallback="Loading Plugin">
-        {parametersReady && <Form.Field required={isRequired}>{PluginComponent}</Form.Field>}
+        <div
+          style={{
+            border: displayResponseWarning ? 'solid 1px' : 'transparent',
+            borderColor: isChanged ? 'blue' : !!currentReview ? 'red' : 'black',
```
CarlosNZ, line 166: Tiny addition to make it look a little more like the designs:
`borderRadius: 7` ;)

src/formElementPlugins/ApplicationViewWrapperNEW.tsx
```diff
@@ -149,15 +149,72 @@ const ApplicationViewWrapper: React.FC<ApplicationViewWrapperPropsNEW> = (props)
...
+            padding: '5px',
+            margin: '5px',
```
CarlosNZ, line 168: FYI -- 'px' is the default, so you can just have the number unquoted:
```
padding: 5,
margin: 5,
```
Only need quoted strings if using other units (%, rem, etc.)

src/formElementPlugins/ApplicationViewWrapperNEW.tsx
```diff
@@ -149,15 +149,72 @@ const ApplicationViewWrapper: React.FC<ApplicationViewWrapperPropsNEW> = (props)
...
+      }}
+    >
+      <Icon
+        style={{
+          display: displayResponseWarning ? 'inline-block' : 'none',
+        }}
+        name={
+          currentReview
+            ? isChanged
+              ? 'comment alternate outline'
```
CarlosNZ, line 206: Don't love this icon, doesn't seem to convey anything. How does no icon look for this. The other two are error and info icons, which make sense, but once it's correct, there's nothing really to inform the user of, so maybe doesn't need any? (You've still got the border to show it's changed)

src/formElementPlugins/ApplicationViewWrapperNEW.tsx
```diff
@@ -149,15 +149,72 @@ const ApplicationViewWrapper: React.FC<ApplicationViewWrapperPropsNEW> = (props)
...
+  isChanged,
+  isValid,
+}) => {
+  const displayResponseWarning = currentReview || (isChanged && isValid)
+
+  return (
+    <div
+      style={{
+        color: isChanged ? 'grey' : 'red',
+        minHeight: 18, // Used to keep space for warning when not displaying
```
CarlosNZ, line 196: Setting a minHeight is probably not the best way to do this, as there's no way to guarantee pixel heights of text and icons between browsers. For example, in Firefox I still see a slight jump.
I'm guessing you used `display:none` and it didn't achieve what you wanted? I think I told you wrong -- you should use `visibility: hidden` instead. That keeps the space but just hides the element, whereas `display:none` removes it from rendering completely.

CarlosNZ (changes requested):
Great, looks good and simple to follow logic. Just a few little comments.

I did think it might be nicer to have the border/icon change on keystroke, but we'd have to change the response to update on every keystroke too, so probably not worth doing.

andreievg:
@nmadruga I just checked out conflicts (just out of interest, since one of my branches had changes to PageElements), looks pretty easy, just to be aware of `visibleElements.map` instead of elements.map

src/containers/Application/ApplicationPage.tsx
nmadruga: Done: 3055394

src/formElementPlugins/ApplicationViewWrapperNEW.tsx
nmadruga: Looks great. Thanks!

src/formElementPlugins/ApplicationViewWrapperNEW.tsx
nmadruga: Yes. Good point. Done: 9f16d99

src/formElementPlugins/ApplicationViewWrapperNEW.tsx
nmadruga: I'm not sure which one you're talking abut? This one: 
Screen Shot 2021-03-19 at 5 27 34 PM

In this case - that's what are in the designs. I was just following orders : P

src/formElementPlugins/ApplicationViewWrapperNEW.tsx
nmadruga: > Setting a minHeight is probably not the best way to do this

Oops. I wasn't sure what was the best approach. 

> I'm guessing you used display:none and it didn't achieve what you wanted?

I didn't : ( That's exact why I did that. But it's not looking good indeed. After I added the border it started to jump in Chrome too.

> you should use visibility: hidden instead. That keeps the space but just hides the element, whereas display:none removes it from rendering completely.

Will try that. Thanks!

nmadruga: It works well. : ) Done: 7c5859b

nmadruga:
> @nmadruga I just checked out conflicts (just out of interest, since one of my branches had changes to PageElements), looks pretty easy, just to be aware of visibleElements.map instead of elements.map

Thanks @andreievg. Doing the merge now

src/containers/Application/ApplicationPage.tsx
CarlosNZ: Okay, I think that works. I'm slightly wary about the default being `true` though, as you really wouldn't want something to be editable when it shouldn't be (whereas the reverse would be not as bad).

Just have to ensure it's **always** passed down from Summary and Review. Can you not make it default to `false` and then pass it from ApplicationPage? I don't know, up to you. Moving on...

src/formElementPlugins/ApplicationViewWrapperNEW.tsx
CarlosNZ: Oh, really? Okay. Save it for another day discussion then :)

src/formElementPlugins/ApplicationViewWrapperNEW.tsx
CarlosNZ: Yeah great. Tiny picky thing -- there's still the smallest jump due to the 1px border adding to the padding. Can fix with:
`padding: displayResponseWarning ? 4 : 5,`
😉

nmadruga: Cool doing that now : )

src/containers/Application/ApplicationPage.tsx
nmadruga: Yes, that makes sense, we don't want that. Well. In this case I will revert this to only pass the canEdit flag and remove the default

andreievg:
@CarlosNZ looks like Nicole dealt with all of the requested changes ?

CarlosNZ (approved):

CarlosNZ:
> @CarlosNZ looks like Nicole dealt with all of the requested changes ?

Sorry, I thought this one was already out the door. Cheers.

PR #470, opened 2021-03-18
Replace NEW routes

CarlosNZ:
Quick update to just replace the NEW routes with the "standard" routes, and have changed the previous "standard" to have OLD in the URL, so we can still get to them if we want.

Replaced all the links throughout that were pointing to a NEW url to now point to standard urls

No changes to file names or component names, that can be done later. This is just to make our life a bit easier in the meantime.


andreievg (approved):
cheers !

Issue #443, opened 2021-03-15
UI to show to reviewer that changes were made (Review home page)


nmadruga:
- [ ] Part 1 - add progress to review structure for thing needing re-review
- [ ] Part 2 - add re-review button

Changes to the Review home page to display buttons to "Re-review" sections with "Changes requested" questions updated (from LOQ) or "Changed" questions updated by Applicant's choice and display progress on the Review after the Reviewer clicks on the Re-review button to start making the second review.

## Designs
**Home page**
![Screen Shot 2021-03-15 at 3 01 40 PM](https://user-images.githubusercontent.com/16461988/111094196-9c175080-859f-11eb-98bc-ed02e5ce18a9.png)

nmadruga:
Planning to work on this one after our meeting

nmadruga:
Ok. Change to work on #441 next - @andreievg is working on it

Issue #442, opened 2021-03-15
UI to show reviewer which questions were updated (Review page)


nmadruga:
### Parts

- [x] Part1 - show re-review (https://github.com/openmsupply/application-manager-web-app/pull/460)
- [x] Part2 - Fix bug with review summary changing when application re-submission is started, looks like draft application responses are mixing things up [PR](https://github.com/openmsupply/application-manager-web-app/pull/493)
- [x] Part3 - show previous review-response & styling [PR](https://github.com/openmsupply/application-manager-web-app/pull/565)

Now I am not sure if we want to show previous review response, there would be re-review or review on all of the questions that were changed, and history when they press the re-review/review that will show previous review response. If application response is not changed, new review response will be duplicated from previous one and visible with edit button.

Changes to the Review Page to display icons for "Changes requested" questions updated (from LOQ) or "Changed" questions updated by Applicant choice. Should also show the total number of changes by Applicant per section within progress bars. 

## Designs
![Screen Shot 2021-03-15 at 3 02 15 PM](https://user-images.githubusercontent.com/16461988/111094189-991c6000-859f-11eb-9af7-4757e038168f.png)
![Screen Shot 2021-03-15 at 3 01 04 PM](https://user-images.githubusercontent.com/16461988/111094247-c0732d00-859f-11eb-9c60-7f5b8b3713cb.png)

andreievg:
This one does a lot of the grunt work: https://github.com/openmsupply/application-manager-web-app/pull/460 for this issue, I am going to wait and see how far Nicole gets with styling application summary before formatting this page further

PR #472, opened 2021-03-19
#442-part-2-re-review-in-review-page

andreievg:
Needs back end: #442f-add-status-to-application-response, PR: https://github.com/openmsupply/application-manager-server/pull/267

This PR fixes and issue described in part of of https://github.com/openmsupply/application-manager-web-app/issues/442

See testing

### Changes

* application response to have status (it's default as draft on creation on back end), and changed to submitted on submission of application

### Testing

To replicate the bug, switch to #443-show-in-review-home-re-submission-of-application.

* create LOQ review for ABC123
* keep the review summary page open after re-submission
* start re-submission for applicant (dont' submit)
* refresh the review summary page (you will see the review responses disapearing)

Do the same in this branch and you should see review summary page not affected by start of applicant's resubmission

### Notes

This problem could have been fixed by row level permissions, but it seems to be better to build some resilience in the system (even if row level permission are incorrect), similar to isVisibleToApplicant being a filter on application response in getAllReasponses (could be done with row level permission, but it's also good to add it to the query)


nmadruga (approved):
LGTM

Issue #480, opened 2021-03-22
Bug: RadioChoice element with default value isn't passing the validation - it requires a value change to allow moving to next page


CarlosNZ:

CarlosNZ:
Is the same thing happening for Dropdown selector?

CarlosNZ:
Suggested fix: in the plugin ApplicationView component, if a default is specified, run the `onSave` method at first render.

PR #481, opened 2021-03-22
#480 Fix default value not being saved

CarlosNZ:
Simple fix for #480. Did the same thing to Dropdown selector as well, as it works exactly the same way.

Just fixed the `useEffect` that runs on first load to always save the response if there is a default specified.

You can see it working with the "Feature Showcase" template. Page 2 has a Radio button whose selection affects the visibility of the following Dropdown. Previously, no additional dropdown appeared until a selection was made in the radio button, even though there was a default selection. Now you'll see the Dropdown appear immediately before any user interaction with the radio buttons. You can also inspect the database to see responses for Radio buttons and Dropdowns being written immediately on load, not just when selected by a user. 

If no default is specified in parameters, this does not run, so response will remain NULL until selected.

### DIFF
--- a/src/formElementPlugins/dropdownChoice/src/ApplicationView.tsx
+++ b/src/formElementPlugins/dropdownChoice/src/ApplicationView.tsx
@@ -18,9 +18,16 @@ const ApplicationView: React.FC<ApplicationViewProps> = ({
 
   useEffect(() => {
     onUpdate(value)
-    if (!value && defaultOption !== undefined)
-      onSave({ text: value, optionIndex: options.indexOf(value) })
-    setValue(options[getDefaultIndex(defaultOption, options)])
+    // This ensures that, if a default is specified, it gets saved on first load
+    if (!value && defaultOption !== undefined) {
+      const defaultIndex = getDefaultIndex(defaultOption, options)
+      const defaultValue = options?.[defaultIndex]
+      onSave({
+        text: defaultValue,
+        optionIndex: defaultIndex,
+      })
+      setValue(defaultValue)
+    }
   }, [])
 
   function handleChange(e: any, data: any) {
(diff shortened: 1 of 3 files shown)


src/formElementPlugins/dropdownChoice/src/ApplicationView.tsx
```diff
@@ -18,9 +18,16 @@ const ApplicationView: React.FC<ApplicationViewProps> = ({
 
   useEffect(() => {
     onUpdate(value)
-    if (!value && defaultOption !== undefined)
-      onSave({ text: value, optionIndex: options.indexOf(value) })
-    setValue(options[getDefaultIndex(defaultOption, options)])
```
andreievg, line 23: there is a slight difference in logic, previously setValue would fire even when if condition wasn't met (haven't tested thought)

andreievg (changes requested):

src/formElementPlugins/dropdownChoice/src/ApplicationView.tsx
CarlosNZ: Yeah, I think the previous logic was actually wrong. It was saving "value", instead of the default option. And value will be NULL to start with. Have replaced it so that it saves the default option rather than value. Condition has not changed (i.e. there's no current value, and there IS a default option specified)

andreievg: ok, so for this fix, was it necessary to only `setValue` if the condition is met ?

CarlosNZ: Yeah, cos if there's already a value, we don't need to set it. (And if there isn't we set one from from the default, but if there's no default, then there's nothing to set)

andreievg (approved):

PR #491, opened 2021-03-23
#484 GraphQL connection fix

### LINKED ISSUE #484: Troubleshoot GraphQL expressions in Templates from front-end
For some reason, GraphQL expressions are not working in the front-end. The exact same expressions work fine in the GUI builder. Probably a network config problem -- will investigate

CarlosNZ:
Fixes #484 
Adds GraphQL connection object to front-end for dynamic parameter lookups.

Requires back-end branch from [this PR](https://github.com/openmsupply/application-manager-server/pull/276) to see it in action -- Company lookup on Page 3 (Section1)

### DIFF
--- a/src/formElementPlugins/ApplicationViewWrapperNEW.tsx
+++ b/src/formElementPlugins/ApplicationViewWrapperNEW.tsx
@@ -18,6 +18,9 @@ import Markdown from '../utils/helpers/semanticReactMarkdown'
 import { IQueryNode } from '@openmsupply/expression-evaluator/lib/types'
 import strings from '../utils/constants'
 import { useFormElementUpdateTracker } from '../contexts/FormElementUpdateTrackerState'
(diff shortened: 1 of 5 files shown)


andreievg (changes requested):
should this alsoi be added to: `addEvaluatedREsponsesToStructure`.
It almost sounds like a small hook would do the trick:

i.e. 
```
const evaluateExpression = useEvaluateExpression()
...
const objects = { responses } // all other things are added by the hook
const result = await evaluateExpression(expression, object, /*optional fallback value*/)
```

CarlosNZ:
Okay, have added it as suggested and to a couple of other evaluation instances. a6c3d0e6

Will have a think about abstracting to its own method/hook sometime.

Cheers.

andreievg (approved):

Issue #465, opened 2021-03-18
Mutation on drop down selection


nmadruga:
Show modal with confirmation before running the mutation - after mutation runs should disable the dropbox selection for now.

nmadruga:
Sorry, not sure if this one is done yet. I'll move back to `In progress` column.

Issue #464, opened 2021-03-18
ReviewWrapper to also get reviewAssignments where current user can be assigner and display in row


nmadruga:

andreievg:
Current query `getReviewInfo`, get's all of the reviewAssignments for current application. And we iterate over them for every section to render (or not render) ReviewSectionRow.

There are a couple of options to implementing this issue:

* Show seperate row for assigner task
* Include assignment task in current row

Second option seems to be what the designer intended in the latest designs, but I think it's a lot easier to do first option (since there could be multiple review_assignment that are not assigned, I am not sure how to group them at this stage), so would suggest going with first option

![Screen Shot 2021-03-18 at 4 53 18 PM](https://user-images.githubusercontent.com/26194949/111570823-7b066800-880a-11eb-8c5a-52e7ca4830e1.png)

Suggested implementation:

 `getReviewInfo` to also return assigners, and filter by ->

```
status = isAssigned or currentUser is one of the Assigners
```

`ReviewHome` ->

When mapping to render `ReviewSectionRow`, filter by `assigner` = me and render `<AssignmentSectionRow>` (passing through all of the reviewAssignment that were filtered), filter by assignment status = 'assigned` -> map and render `<ReviewSectionRow>`

 `<AssignmentSectionRow>`  can use reviewAssignments passed to it to render drop down

PR #497, opened 2021-03-25
#464 #465 add assignments logic

andreievg:
Requires back end  #464F-#465F-add-self-assignment-option-on-templatePermission
Fixes: #464
Fixes: #465

changes in this PR will make assignment and self assignment work (not re-assignment or actionables in application list)

### Changes

(not targeted)

* update error checking for async mutations (noticed we were doing some extra bits inside the component, outsourced to hooks

(targeted)

* added mutation and helper hook, `useUpdateReviewAssignmet`
* changes `getReviewInfoQuery` to also get assigners through join and `templateSectionRestrictions`
* `useGetReview` hook adds, `iscurrentUserReviewer`, isCurrentUserAssigner` and `assibnableSectionRestrictions`
*  add `AssignmentSectionRow`
* changed `generateReviewSectionActions` to only return `canSelfAssign ` option if reviewer is current user (would there be a need to show all users that can self assign in review home /)

(notes)

* maybe we should do all of the calculation inside the hook vs in `AssignmentSectionRow` (but it would require combining multiple assignments into one structure)

### Testing

* using back end branch above
* create new ReviewTest application and submit
* reviewer options (assigner options, self assignment etc) should comply with the digram below
* One thing to note, self assignment and manual assignment are not meant to be working on the same application/stage/level (it's either one or the other). In this PR the only interchangeable thing is that assigner will see self assigned user and wont' be able to assign another one. 
* Self when one user self assigns, the other should't see the option anymore (via action)

Screen Shot 2021-03-25 at 12 37 42 PM


CarlosNZ:
Couple of initial comments from testing:

- I can see the "Self-Assign" button for the two Sections. When I click one of them, they both get assigned to me. Is this intended behaviour?
- After assigning, when I click "Start", it just gives an error and review not started (after a refresh) -- looks like this:
Screen Shot 2021-03-25 at 2 02 16 PM

andreievg:
> I can see the "Self-Assign" button for the two Sections. When I click one of them, they both get assigned to me. Is this intended behaviour?

Yep - I think we need a modal to say all sections will be assigned? Or only self assign that section ?

> After assigning, when I click "Start", it just gives an error and review not started (after a refresh) -- looks like this:

I am not getting this error, hmm, can you give a replication example (which back end branch, is it new application or  an application from insertData) ?

src/containers/Review/ReviewHome.tsx
```diff
@@ -28,13 +29,17 @@ const ReviewHome: React.FC<ReviewHomeProps> = ({ assignments, structure }) => {
 
   const [filters, setFilters] = useState<Filters | null>(null)
 
+  const getFilteredByStage = (assignments: AssignmentDetailsNEW[]) => {
+    if (!filters) return []
```
CarlosNZ, line 33: Don't you think that no filters should return ALL, not NONE?

CarlosNZ (approved):
All looks pretty solid and tests okay (not getting that weird but any more, don't really know why).

Will probably notice other bugs once it's properly integrated with more functionality, but seems good to merge now. Just one small question, nothing major.

nmadruga:
Having a look on this one today. I'm willing to merge if all good in test so we can remove some of the old PRs from the list of PRs to review - It's growing to big and starting to get messy to test without other things in place.

src/utils/hooks/useUpdateReviewAssignment.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+    const createReviewQuestionAssignments = assignableElements.map((element) => ({
+      templateElementId: element.element.id,
+    }))
+
+    return {
+      status: ReviewAssignmentStatus.Assigned,
+      // onReviewSelfAssign will trigger core action to disallow other reviewers to self assign
+      trigger: isSelfAssignment ? Trigger.OnReviewSelfAssign : Trigger.OnReviewAssign,
+      // TODO: change to timeUpdated
```
nmadruga, line 57: Adding a placeholder issue for that

nmadruga (approved):
Looks pretty solid to me after testing for a little while. 
I have made just a few commits:
- Remove console.log 
- Rename UN_ASSIGN to UNASSIGN constant
- Remove `templateElementId` from queries (and replaced where used with `templateElement.id`)

Hope that's all good with you @andreievg. I'm merging this one and will create separated issues for things that might need a fix/improvement as part of ReviewAssignment Epic.

src/utils/hooks/useUpdateReviewAssignment.tsx
nmadruga: Created a placeholder note for this issue [here](https://github.com/orgs/openmsupply/projects/6#card-58402523)

Issue #190, opened 2020-12-07
Responses need a way to map selection to actual entity in the database. Query syntax for GraphQL should have that returned as part of the result (not only list of strings)


nmadruga:
## Suggested solution:

- Change the query that fetches the list (of organisations, for example) to get the whole object, not just `name`.
- Add an additional parameter to plugin (Dropdown, RadioChoice (maybe checkbox?)) to specify the name of the display field.
- We save _all_ the fields in the Response, but we only display the "display field" list to the user in the list

PR #498, opened 2021-03-25
#190 Dropdown plugin changes to handle ALL entity data

CarlosNZ:
Implements #190 as described in the initial post on that issue.

This is the front-end changes. You can test it using back-end branch `joinCompanyTemplate` and use the template with code "OrgJoin". Note, I'm not ready to PR the back-end yet as I want to get everything in the template working first. But the changes in this PR shouldn't affect any existing functionality.

Main change here was to use the "selectedIndex" as the primary local state. The plugin can now handle the options list being either an array of strings (as it was before), or an array of objects. If it's objects, there needs to be an additional parameter supplied "optionsDisplayProperty" with the name of the field in each object to present to the UI.

Test using new OrgJoin template: http://localhost:3000/application/new?type=OrgJoin The list of Orgs on the first page uses the new parameter. Take note of the response stored in the database -- the whole original object is saved under the `selection` field. (`optionIndex` and `text` same as before)

This should be enough to enable proper handling of entities in the back-end, since the "id" of the selected item is readily available from responses.


src/formElementPlugins/types.ts
```diff
@@ -56,7 +56,7 @@ type ValidationState = {
 interface ApplicationViewProps extends ApplicationViewWrapperProps {
```
nmadruga, line 56: I'm confused why are we still using extends `ApplicationViewWrapperProps` instead of ApplicationViewWrapperPropsNEW here? I was going to point out to the `parameters` prop not defining all these possible arguments received. I guess we might want to 1 - change to use `ApplicationViewWrapperPropsNEW` and 2 - list all these new types in a new type for `parameters` since now this is well defined.

I have started here:

```
interface QuestionParameters {
    label: string
    description?: string
    placeholder?: string
    search?: boolean
    options: string[]
    optionsDisplayProperty?: string
    default: string | number
}
```

src/formElementPlugins/types.ts
```diff
@@ -56,7 +56,7 @@ type ValidationState = {
...
   onUpdate: Function
   onSave: Function
-  value: string // TODO: Change to allow object with any shape
+  value: any
   setValue: (text: string) => void // TO update the value on the ApplicationViewWrapper
```
nmadruga, line 60: Isn't the `value` not used for theDropdown? Why are we changing here. And if the case is to change, should also change the `setValue` to accept any value too?

src/formElementPlugins/dropdownChoice/src/ApplicationView.tsx
```diff
@@ -1,47 +1,67 @@
...
+    })
   }
 
   const dropdownOptions = options.map((option: any, index: number) => {
     return {
       key: `${index}_${option}`,
-      text: option,
-      value: option,
-      index,
+      text: optionsDisplayProperty ? option[optionsDisplayProperty] : option,
```
nmadruga, line 63: I think there might be a problem with this if  `optionsDisplayProperty` comes with the wrong string for a field.

src/formElementPlugins/dropdownChoice/src/ApplicationView.tsx
```diff
@@ -1,47 +1,67 @@
...
 
   function handleChange(e: any, data: any) {
-    const { value } = data
-    setValue(value)
-    onSave({ text: value, optionIndex: options.indexOf(value) })
+    const { value: optionIndex } = data
+    setSelectedIndex(optionIndex)
+    onSave({
+      text: optionsDisplayProperty
+        ? options[optionIndex][optionsDisplayProperty]
```
nmadruga, line 53: Same problem here. I think you need something like 
```
const fullObject = options[optionIndex]
return fullObject[optionsDisplayProperty] || fullObject
```

src/formElementPlugins/dropdownChoice/src/ApplicationView.tsx
```diff
@@ -1,47 +1,67 @@
...
-      const defaultIndex = getDefaultIndex(defaultOption, options)
-      const defaultValue = options?.[defaultIndex]
+    if (!currentResponse?.text && defaultOption !== undefined) {
+      const optionIndex = getDefaultIndex(defaultOption, options)
       onSave({
-        text: defaultValue,
-        optionIndex: defaultIndex,
+        text: optionsDisplayProperty
+          ? options[optionIndex][optionsDisplayProperty]
+          : options[optionIndex],
```
nmadruga, line 36: Same here

documentation/internal/Element-Type-Specs.md
```diff
@@ -159,18 +159,19 @@ _Multi-choice question, with one allowed option, displayed as Drop-down list (Co
...
+- **default**: `string`/`number` -- the value initially selected before user input. If `number`, refers to the index of the options array. If not provided, no option will be pre-selected.
 - **search**: `boolean` (default: `false`) -- if `true`, the list of options can be searched and filtered by user
-- **hasOther**: `boolean` -- if `true`, an additional text-entry field is provided so the user can add their own alternative option _(not yet implemented)_
+- **optionsDisplayProperty**: If `options` (above) consists of an array of objects, this parameter specifies the field of each object to be displayed in the options list. For example, if `options` was a list of companies, you'd probably specify `name` as the `optionsDisplayProperty`. Note that even though one field is displayed to the user in the Dropdown list, the _entire_ selected object is saved as the selection.
+- ~~**hasOther**: `boolean` -- if `true`, an additional text-entry field is provided so the user can add their own alternative option _(not yet implemented)_~~
```
nmadruga, line 166: I think this can be removed

nmadruga (commented):
Just a few comments - will start testing now

nmadruga:
Ok. Have tested and seen the result on the `db`. All good, although I would like to see the joinOrg functioning I think I might prefer to wait for the back-end PR to be included, so I can have one user to do the Review. I check that the branch `joinCompany-template` it's using templateName id 8000 for both Apply and Review, but no user is linked to it yet. I guess that in your PR you have users linked to it, so I can test this?

Here is how the user/permission is showing on the current branch for me:
10	Assigner1	9000
10	Assigner1	10100
11	Assigner2	9000
12	Screener1	10000
12	Screener1	10300
13	Screener2	10000
13	Screener2	10300
14	assessor1	10000
14	assessor1	10400
15	assessor2	10000
15	assessor2	10400
16		1000
1	Nicole	10100
2	Carl	10100
3	Andrei	10100
4	Valerio	10100
1	Nicole	2000
2	Carl	2000
3	Andrei	2000
4	Valerio	2000
6	Reviewer1	5000
7	Reviewer2	5000
7	Reviewer2	5001
8	Consolidator1	5001
6	Reviewer1	6000
7	Reviewer2	6000
8	Consolidator1	6001
9	Consolidator2	6001
8	Consolidator1	7000
17	Admin	1000
17	Admin	2000
17	Admin	4000

nmadruga:
Looking at the back-end branch (now it has a PR - so should be ready)

src/formElementPlugins/types.ts
CarlosNZ: Yeah, I'll need to update this once you've merged #568 I think. I'll try and review that this morn so we can merge it then I'll pull it into here and tidy up.

src/formElementPlugins/types.ts
CarlosNZ: Yeah same, related to the last comment I think.

src/formElementPlugins/dropdownChoice/src/ApplicationView.tsx
CarlosNZ: Same comment as earlier -- we don't want to be saving a "fullObject" to a text value -- that would cause worse problems elsewhere as "text" is expected to be a string throughout the app. Better to just have the plugin fail and show error, don't you think?

src/formElementPlugins/dropdownChoice/src/ApplicationView.tsx
CarlosNZ: I'm getting the feeling I might be missing what you're referring to. Just to clarify:
- if `optionsDisplayProperty` is not defined, it'll assume the options is an array of text values
- if `optionsDisplayProperty` IS defined, it'll display that field from each object in the array
- if `optionsDisplayProperty` is defined but that field doesn't exist, it'll throw an error and not display the plugin

Is this not what we want (Or is it not actually doing what I think it's doing?)

documentation/internal/Element-Type-Specs.md
CarlosNZ: I've tidied up this documentation substantially in some of my later PRs, so probably just ignore this for now.

src/formElementPlugins/types.ts
CarlosNZ: Okay, I've merged from the new master now, so this should be fine I think.

src/formElementPlugins/dropdownChoice/src/ApplicationView.tsx
nmadruga: That's ok to throw the error for 3rd options "if `optionsDisplayProperty` is defined but that field doesn't exist, it'll throw an error and not display the plugin", I was just checking what would happen in this case.

src/formElementPlugins/dropdownChoice/src/ApplicationView.tsx
nmadruga: Yes, I agree. I was just trying to see, now I got it. Maybe a good idea to flag that on the documentation (if you haven't already).

nmadruga (approved):
I just checked what is saved in the database and it's all good. I couldn't test with the back-end though, I'm not sure how to do this. Maybe we can sit together for 5 minutes and you can show to me?

src/formElementPlugins/dropdownChoice/src/ApplicationView.tsx
CarlosNZ: Sure, I'll add a quick note to the docs before I merge this. Thanks for review

PR #500, opened 2021-03-25
#499 - Tiny bug fix

### LINKED ISSUE #499: Bug: displaying non-required response changed for Applicant

nmadruga:
Fixes #499 

### Changes
- Fix problem not showing warning on Application changes required - for responses without changes required.
**Edited on 30/03**
- Bug fix to stop displaying the "This is a new submission and will require a review." warning on a new Application.
- To fix unwanted warning I removed `isChanged` and `isChangeRequired` fields getting set on `addApplicantChangeRequestStatusToElement`, so it only get set to each `state` element in the fullStructure when there is at least one review done in the full structure. This prevents a new application to display warnings - since they use the same component `PageElements` and `ApplicationViewWrapper`.

### DIFF
--- a/src/utils/helpers/structure/addApplicantChangeRequestStatusToElement.ts
+++ b/src/utils/helpers/structure/addApplicantChangeRequestStatusToElement.ts
@@ -7,6 +7,19 @@ import {
 import { FullStructure } from '../../types'
 
 const addApplicantChangeRequestStatusToElement = (structure: FullStructure) => {
+  const checkHasAnyReviewResponses = Object.values(structure?.sortedPages || {}).flatMap(
+    ({ state }) =>
+      state.filter(({ latestApplicationResponse, previousApplicationResponse }) =>
+        // For application not in draft - check if any latestApplicationResponse has been reviewed
+        // For applications in draft - check if any previousApplicationResponse has been reviewed
+        structure?.info?.current?.status !== ApplicationStatus.Draft
+          ? latestApplicationResponse?.reviewResponses.nodes[0] !== undefined
+          : previousApplicationResponse?.reviewResponses.nodes[0]
+      )
+  )
+  // No reviews, then don't set isChanged and isChangeRequested fields
+  if (checkHasAnyReviewResponses.length === 0) return
+
   const questionElements = Object.values(structure?.elementsById || {}).filter(
     ({ element }) => element?.category === TemplateElementCategory.Question
   )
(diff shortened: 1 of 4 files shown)


andreievg:
Changes looks good, although when i was looking at the code i noticed use of 'currentReview' -> i was hoping that isChangeRequest and isChanged would be enough to determine the state of the change for display, but realised isChanged would be set to 'true' even on the first start of application. (I think the best place to fix it is in `addApplicantChangeRequestStatusToElement`) <- only run it if there is at least one reviewResponse (in the whole application)

nmadruga:
`currentReview` is required to display the comment from the reviewer. I have tried changing to not set values of `isChanged` and `isChangeRequest` on any element if not a application coming back from a review (from status ChangesRequired) on the `addApplicantChangeRequestStatusToElement` but I'm not entirely sure I understood the idea you gave here.

nmadruga:
Hi @andreievg I have made the change requested. I seems to work well now. Ideally the code would be better structured, but I didn't want to use to time time on this, I think we should improve it at some point though.

src/utils/helpers/structure/addApplicantChangeRequestStatusToElement.ts
```diff
@@ -7,6 +7,19 @@ import {
 import { FullStructure } from '../../types'
 
 const addApplicantChangeRequestStatusToElement = (structure: FullStructure) => {
+  const checkHasAnyReviewResponses = Object.values(structure?.sortedPages || {}).flatMap(
```
andreievg, line 10: This looks good although could be simplified to something like this : 
``` 
const hasReviewResponse = (applicationResponse: ApplicationResponse) => 
applicationResponse?.reviewResponse..nodes?.length > 0

const hasReviewResponses = questionElements.some(({latestApplicationResponse, previousApplicationResponse}) => 
hasReviewResponse(latestApplicationResponse) || hasReviewResponse(previousApplicationResponse)
```

not required change

andreievg (approved):
> currentReview is required to display the comment from the reviewer. I have tried changing to not set values of isChanged and isChangeRequest on any element if not a application coming back from a review (from status ChangesRequired) on the addApplicantChangeRequestStatusToElement but I'm not entirely sure I understood the idea you gave here.

ok, i was hoping we could use previousApplicationResponse in the element. if you remember this PR (https://github.com/openmsupply/application-manager-web-app/pull/447), it came with a testing branch with decision and comment in JSON form under each question (https://github.com/openmsupply/application-manager-web-app/blob/%23441-TESTING-part-1-application-page-LOQ/src/components/Application/PageElements.tsx)

src/utils/helpers/structure/addApplicantChangeRequestStatusToElement.ts
nmadruga: Updating to this now. Will test and merge after : )

Issue #494, opened 2021-03-24
Bug fix -- reject login if can't parse JWT


CarlosNZ:
There's an annoying bug that sometimes occurs when the front-end gets an empty JWT stuck in localStorage. The back-end gets stuck in a perpetual loop, saying "Cannot parse JWT". 

We'll need to do a fix on both back and front-end probably:
- Back-end: if "cannot parse JWT" error, return something that the front-end can recognise as an authentication fail
- Front-end: when confirming stored details in JWT, if error message returned, force a log-out (which should purge the local storage)

Issue #298, opened 2021-01-27
Progress bar: UI when page as no required elements -- to consider


CarlosNZ:
This is not a big deal right now, but just something to think about. I noticed when starting a new application, if a later page has no required elements, then it immediately shows up as Valid in the progress bar:
![Screen Shot 2021-01-27 at 2 14 51 PM](https://user-images.githubusercontent.com/5456533/105928112-32ec8400-60aa-11eb-938c-14bd3a87aa06.png)

This is technically correct as per our current rules, but it seems a bit weird to me and potentially confusing for a user. Wonder what the best option is here. Maybe it should stay "neutral" until you've had focus on at least one element on that page? I'm not sure what the best thing to do is here, probably a question for Debbie.

PR #553, opened 2021-03-31
#494 Bug fix for "Can't parse JWT" error

CarlosNZ:
(Hopefully 🤞) this very small fix will prevent the annoying bug reported in #494.

Requires branch from [back-end PR #298](https://github.com/openmsupply/application-manager-server/pull/298).

Obviously we'll need some kind of server-side flood control or something for production, cos you shouldn't be able to tie up the server in an infinite loop like this, but I don't know much about that kind of thing.


nmadruga (changes requested):
This is still not working for me - I have Opera browser stuck in one of these JWT == `undefined`. Now I think if something get's changed on method `isLoggedIn` this might get solved. Although I tried and this lines didn't work, so I'm clueless too:
```
const isLoggedIn = () => {
  const JWT = localStorage.getItem('persistJWT')
  if (JWT === undefined) return false
  console.log(JWT)
  return JWT !== null
}
```
It keeps logging `undefined` : (

CarlosNZ:
Yeah, it's really hard to test because I don't know how to create the conditions to cause the problem.

Since you've got a broken one, can you put a console.log() in UserState just before line 90:
```
    if (JWT == undefined) logout()
```

I can't see why it could be anything other than `undefined` at that point, unless your browser is not actually reading it correctly when it's in that weird state? Or maybe this line is not being run, in which case it's something else.

CarlosNZ:
But yeah, your additional check is a good idea anyway, so I've added it to `isLoggedIn()`: 57d2d5d3

nmadruga:
> Since you've got a broken one, can you put a console.log() in UserState just before line 90:
>    if (JWT == undefined) logout()

Result:
Screen Shot 2021-04-08 at 1 15 44 PM

As you can see - many many times (and keeps going).

> I can't see why it could be anything other than undefined at that point, unless your browser is not actually reading it correctly when it's in that weird state? Or maybe this line is not being run, in which case it's something else.

I have seem this same thing - it's `undefined` but the if statement in the next line doesn't run how it should - in the other place I told you we should have this check. Although it doesn't stop this process... Probably since it's running synchronously it doesn't wait for it, and that's why it keeps trying to reach the server? Anyway I think this change doesn't fix this problem. And I'm not sure how to do this either (I would, but we have limited time - so haven't spent to long looking).

CarlosNZ:
> I have seem this same thing - it's `undefined` but the if statement in the next line doesn't run how it should - in the other place I told you we should have this check.

@nmadruga -- I have solved this mystery! Turns out that the local storage is storing the *string* "undefined", so it will match correctly with `if (JWT == 'undefined') logout()` 🤦‍♂️

So I've fixed that, but I've also made it bit more robust against other related errors --  basically, any unexpected problems result in a `logout*()` call. I added an additional back-end PR (#298) as I realised that the "success" field was missing from the user-info endpoint when it was successful. This PR will need that back-end branch now, otherwise it won't log in at all.

Pretty sure this should log you out correctly to get rid of your Opera fault state now. 🥳

nmadruga:
> @nmadruga -- I have solved this mystery! Turns out that the local storage is storing the string "undefined", so it will match correctly with if (JWT == 'undefined') logout() 🤦‍♂️

Great finding!!

nmadruga:
> So I've fixed that, but I've also made it bit more robust against other related errors -- basically, any unexpected problems result in a logout() call. I added an additional back-end PR (#298) as I realised that the "success" field was missing from the user-info endpoint when it was successful. This PR will need that back-end branch now, otherwise it won't log in at all.

Cool! Having a look now on both! Seems to be working all good for me now 😄

nmadruga (approved):
It's working now! Thanks for fixing this : )

Issue #488, opened 2021-03-22
Update Radio plugin: option for "Other" field


nmadruga:

CarlosNZ:
Question: do you think we should build this functionality into the RadioChoice plugin, or just have template add an additional "shortText" whose visibility is set to show if "Other" is selected in previous RadioChoice?

(i.e. we can already do this with existing functionality by configuring the template -- so is it worth adding more complexity to plugin?)

nmadruga:
Using the solution that you proposed is already implemented on Drug Registration latest PR (back-end#270 link), and I agree it's not a priority. But it seems to me that having an optional (for common use of types) would be a nice to have. Considering how the config is done to add a separate `shortText`  with visibility criteria it seemed reasonable to have a an extra configuration for the `radioChoice` plugin. Also for layout purpose.

Another idea is to have extension of an elementPlugin and we could also make the complexity of adding something like that separated from the basic element.

Issue #475, opened 2021-03-21
Create template for join company and make it functional


CarlosNZ:

CarlosNZ:
> ### Remaining issues to get this all up to MVP:
> - [ ] Schema: add "logo_url" field to "organisation" table (do with DisplayImage plugin below)
> - [ ] "ApplicationData" for front-end evaluator -- put into FullStructure (initial issue #458 -- will update to discussed spec)
> - [ ] LongText element plugin: Issue #479 (needed for Address field, until we get a proper Address autocomplete plugin)
> - [ ] DisplayImage element plugin: Issue #549
> - [ ] "Other" option for RadioButton: Issue #488 (almost done)
> - [ ] Update templates for CreateOrg and JoinOrg (will do along with above issues)

I'm planning on knocking the rest of these out in the next few days, mostly while we're away, so we can crack on with the "mini-epics" once we're back.

PR #554, opened 2021-03-31
#488 "Other" option for radio button selector plugin

CarlosNZ:
Implements #488. Requires [back-end PR branch #289](https://github.com/openmsupply/application-manager-server/pull/289) to see it in action.

~~_Branched from #498 -- please review that one first_~~  Merged

---

Part of [CreateCompany](https://github.com/openmsupply/application-manager-web-app/issues/96) and [JoinCompany](https://github.com/openmsupply/application-manager-web-app/issues/475) epics

To get this working smoothly, I re-factored similarly to the Drop-down changes in #498 so that it uses the "index" as the primary local "value", rather than the passed in "value" which is just the `text` field. It seems to be much easier for most element plugins to handle their "value" locally like this, so the `value/setValue` parameters passed down aren't terribly useful for most plugins.

Also like Drop-down, it can also handle a whole-object array for the options list, as long as a `optionsDisplayProperty` is specified (see #498). 

Anyway, added "hasOther" and "otherPlaceholder" parameters (see docs), and changed the saved Response object as follows (also in docs):

```
{
  optionIndex: <integer> (index from the options array)
  text: <string> (actual text from options array, will be "Other: <value>" if "Other" is selected)
  selection: <string | object> (entire object or string from the supplied options list)
  other: <string> (text of the "Other" selection, only present if "Other" is selected)
}

```

### To test:
Load a Feature Showcase application (using the above mentioned branch) and navigate to Section 1 -> Page 3. The first element on the page should have the "Other" option. 

### Notes

- It would look a lot better if the "Other" text field was inline with the "Other" button label. However, this it not trivial to do, since both `<Radio>` and `<Input>` are block-level elements, and the Radios form a repeating list. Something to think about...
- Let me know if you think the saved Response shape is okay. Another possibility is to not have the "other" field and just set "selection" to the Other value. But then you don't have a record of whether or not "Other" was selected.
- Obviously it'd be advisable to only provide an array of strings as options if you're going to use the `hasOther` parameter, otherwise you'd be mixing-and-matching objects with strings (the "Other" option is always just a string). I haven't tested what happens if you try, but I imagine the results are not pretty.
- Have not yet tested Validation for the "Other" text field. In theory it should work fine, but I'll check it properly when I update the actual "createCompany" template.


nmadruga:
> so the value/setValue parameters passed down aren't terribly useful for most plugins.

I wonder if that can be completely replaced by the ones received from `data` so we avoid confusions in the future.

nmadruga:
A few comments about the response:

`optionIndex: <integer> (index from the options array)` => maybe optionIndex: <number> so it's more Typescript like?
`text: <string> (actual text from options array, will be "Other: <value>" if "Other" is selected)` => Could we change so when other is selected the value will be only what is included in there? I just saw one case (in IRIMS) that what is entered in the other box would be used in another question.

src/formElementPlugins/radioChoice/src/ApplicationView.tsx
```diff
@@ -1,59 +1,123 @@
...
+              label={option.text}
               name={`${code}_radio_${index}`} // This is GROUP name
-              value={option}
-              checked={option === value}
+              value={selectedIndex}
+              checked={index === selectedIndex}
               onChange={handleChange}
               index={index}
             />
+            {/* TO-DO: Need to find a way to make this input field inline with  the last Other radio button*/}
```
nmadruga, line 110: Screen Shot 2021-04-06 at 4 12 05 PM

: ) Just need to do the same check on `<Form.Field (...) inline={hasOther && index === allOptions.length - 1}>` prob better to keep the same check in a variable.

src/utils/types.ts
```diff
@@ -377,6 +377,7 @@ interface ResponseFull {
   isValid?: boolean | null
   hash?: string // Used in Password plugin
   files?: any[] // Used in FileUpload plugin
+  other?: string | undefined // Used in RadioChoice plugin
```
nmadruga, line 380:
```suggestion
  other?: string // Used in RadioChoice plugin
```

When using `?` you already count for `undefined`

nmadruga (changes requested):
Thanks for working on this and have a solution already : D

I really think we can drop one of the result fields, and only keep the other value in one of the existing ones. I don't really like to have a separated field `other` to store what was entered in the other input field. But I'll leave this decision with you.

Other than that it looks great. : )

CarlosNZ:
> `optionIndex: <integer> (index from the options array)` => maybe optionIndex: so it's more Typescript like?

Not quite sure what you mean here. ?

> `text: <string> (actual text from options array, will be "Other: <value>" if "Other" is selected)` => Could we change so when other is selected the value will be only what is included in there? I just saw one case (in IRIMS) that what is entered in the other box would be used in another question.

Yeah, I guess that's a good case for keeping the custom value in "text". Perhaps then the `other` field (below) could be a boolean to indicate that "Other" was selected? Otherwise we have no way of knowing that. What do you think?

> `other: <string> (text of the "Other" selection, only present if "Other" is selected)` => We might not need this one if `text` already holds this? Would this be too confusion, do you think?

See comment for "text" above.

src/formElementPlugins/radioChoice/src/ApplicationView.tsx
CarlosNZ: Oh that's great -- much nicer. Thanks. :)

CarlosNZ: Done 3a152ee7

src/utils/types.ts
CarlosNZ: Okay, that seems to work, thanks. I could have sworn I was getting a typescript error when I did a `if (other===undefined)` at one point, as though it made a distinction between having the value `undefined` and just being absent. 

I guess not. 🤷‍♂️

082f8c4d

src/formElementPlugins/radioChoice/pluginConfig.json
```diff
@@ -2,7 +2,7 @@
   "isCore": true,
   "code": "radioChoice",
   "displayName": "Radio Button Selector",
-  "dynamicParameters": ["label", "description", "options"],
+  "dynamicParameters": ["label", "description", "options", "hasOther", "otherPlaceholder"],
```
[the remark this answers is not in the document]
CarlosNZ: Yeah, good call. 4cbaf6d7

documentation/internal/Element-Type-Specs.md
```diff
@@ -118,60 +118,63 @@ _For displaying blocks of text in the application_
...
-- **default**: `string`/`number` -- the value initially selected before user input. If `number`, refers to the index of the options array. If not provided, no option will be pre-selected.
-- **search**: `boolean` (default: `false`) -- if `true`, the list of options can be searched and filtered by user
-- **optionsDisplayProperty**: If `options` (above) consists of an array of objects, this parameter specifies the field of each object to be displayed in the options list. For example, if `options` was a list of companies, you'd probably specify `name` as the `optionsDisplayProperty`. Note that even though one field is displayed to the user in the Dropdown list, the _entire_ selected object is saved as the selection.
-- ~~**hasOther**: `boolean` -- if `true`, an additional text-entry field is provided so the user can add their own alternative option _(not yet implemented)_~~
+- **options\***: `array[string | object]` -- as above (in [Drop-down](#dropdown))
```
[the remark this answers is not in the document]
CarlosNZ: f5ec90a5

CarlosNZ:
Okay, have addressed the discussed comments, and merged from the new, tidied up master. Should be ready for another look.

documentation/internal/Element-Type-Specs.md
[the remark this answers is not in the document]
nmadruga: Thanks

src/formElementPlugins/radioChoice/src/ApplicationView.tsx
```diff
@@ -1,57 +1,128 @@
...
+      const { optionIndex } = currentResponse
+      setSelectedIndex(optionIndex)
     }
   }, [])
 
   function handleChange(e: any, data: any) {
-    const { value, index } = data
-    setValue(value)
-    onSave({ text: value, optionIndex: index })
+    console.log('CHanged..')
```
nmadruga, line 60: Please remove :)

src/formElementPlugins/radioChoice/src/ApplicationView.tsx
```diff
@@ -1,57 +1,128 @@
...
+    const { index: optionIndex } = data
+    setSelectedIndex(optionIndex)
+    console.log('Selected:', optionIndex)
```
nmadruga, line 63: And this one

src/formElementPlugins/radioChoice/src/ApplicationView.tsx
```diff
@@ -1,57 +1,128 @@
...
       <label>
         <Markdown text={label} semanticComponent="noParagraph" />
       </label>
       <Markdown text={description} />
-      {options.map((option: string, index: number) => {
+      {radioButtonOptions.map((option: any, index: number) => {
+        const showOther = hasOther && index === allOptions.length - 1
         return (
-          <Form.Field key={`${index}_${option}`} disabled={!isEditable}>
+          <Form.Field key={`${index}_${option}`} disabled={!isEditable} inline={showOther}>
```
nmadruga, line 107:
```suggestion
          <Form.Field key={option.key} disabled={!isEditable} inline={showOther}>
```

nmadruga (changes requested):
Just a few more things

src/formElementPlugins/radioChoice/src/ApplicationView.tsx
CarlosNZ: Oops. 👍

src/formElementPlugins/radioChoice/src/ApplicationView.tsx
CarlosNZ: Cool

src/formElementPlugins/radioChoice/src/ApplicationView.tsx
CarlosNZ: 44218fe4

nmadruga (approved):

Issue #479, opened 2021-03-22
Template Element plugin: LongText


CarlosNZ:
A text input for longer answers. ~~Should be able to enter in Markdown format and see a live (on ButtonCLick?) preview~~

Parameters:
- lines: number of lines for the textArea input (default 5?)
- maxLength (in characters) (default no limit)

Issue #290, opened 2021-01-21
Bug: Loading the Submission page after application is submitted


nmadruga:
Not all the times the submission page is loading, this is to do with the order of things. Some times, when trying to load the submission page the application is still not submitted! It's easy to see that by logging on `useSubmitApplication` method: `onCompleted` and when the push to Submission page happens on `ApplicationOverview`. Even though there is a await on the submit process, this is returned before the process of submitting finishes. We need to change to use the onCompleted call back or other mechanism to prevent some failures to load Submission page.

CarlosNZ:
Not sure if this is related, but I just noticed this error go by in the back-end when submitting a User registration.
![Screen Shot 2021-01-22 at 11 47 30 AM](https://user-images.githubusercontent.com/5456533/105424604-80cd4a80-5cac-11eb-9340-7a434da3063f.png)

No idea what the problem is, but might help in tracking down the root of this issue.

PR #555, opened 2021-03-31
#479 Long text element plugin

CarlosNZ:
Implements #479. 

_Branched from #554 -- please review that one first._

---

Use [back-end PR #290](https://github.com/openmsupply/application-manager-server/pull/290)  to see it in action.

This is _very_ similar to the "shortText" plugin. It's almost a straight copy, but have changed `<Form.Input>` to `<Form.TextArea>`, and added a "lines" parameter to specify the height of the text area (default: 5)

Was going to add a "maxLength" parameter, but it's actually easy just to add a validation expression to check for text length if required (See example in Feature Showcase).

To test: fire up a Feature Showcase application. You'll find LongText elements on:
- Section 1, Page 3 (Need to select "Other" to enable it) -- just has default values
- Section 2, Page 4 -- lines: 8, validation expression to limit response to 100 characters


nmadruga:
> Was going to add a "maxLength" parameter, but it's actually easy just to add a validation expression to check for text length if required (See example in Feature Showcase).

That called my attention when I was checking the back-end PR, I wonder if it would be nice to have something as a `maxLength` variable, just because - for now - it's such a hard thing to add a simple `validation` using REGEX.

nmadruga:
Something weird happening when testing the Long text on Section 2 Page 4:

If I try to enter a very long response, it allows me. Even if I click out of the box - nothing changes:
Screen Shot 2021-04-12 at 1 00 37 PM

Then going to another page (previous in this case) and back to Page 4 - it's gone 😮 :
Screen Shot 2021-04-12 at 1 00 42 PM

src/formElementPlugins/longText/pluginConfig.json
```diff
@@ -0,0 +1,8 @@
+{
+  "isCore": true,
+  "code": "longText",
+  "displayName": "Multi-line Text Input",
+  "dynamicParameters": ["label", "description", "placeholder"],
```
nmadruga, line 5:
```suggestion
  "dynamicParameters": ["label", "description", "placeholder", "lines"],
```

src/formElementPlugins/longText/src/TemplateView.tsx
```diff
@@ -0,0 +1,16 @@
+import React, { useState } from 'react'
+import { Input } from 'semantic-ui-react'
+import { TemplateViewProps } from '../../types'
+
+// TODO type should be global
```
nmadruga, line 5: Is this just copied from `shortText`? Not sure if that's required anymore (TODO's 👮‍♀️)

CarlosNZ:
> Something weird happening when testing the Long text on Section 2 Page 4:
> 
> If I try to enter a very long response, it allows me. Even if I click out of the box - nothing changes:
> Screen Shot 2021-04-12 at 1 00 37 PM

It's not saving because it exceeds the maximum length. You should have seen the error state once you exceeded it, not quite sure why you didn't.
Screen Shot 2021-04-12 at 1 12 18 PM

src/formElementPlugins/longText/pluginConfig.json
CarlosNZ: Do you really think "lines" needs to be dynamic? I'm struggling to think of a use case where the number of lines available in a text box would be conditional on something else.

src/formElementPlugins/longText/src/TemplateView.tsx
CarlosNZ: Not touching "Template View" at the moment -- please just ignore the whole file.

CarlosNZ:
> That called my attention when I was checking the back-end PR, I wonder if it would be nice to have something as a `maxLength` variable, just because - for now - it's such a hard thing to add a simple `validation` using REGEX.

Okay, have implemented a basic `maxLength` function and updated docs as we discussed. Have also applied it to the "shortText" plugin.

src/formElementPlugins/longText/src/TemplateView.tsx
nmadruga: ok

nmadruga (approved):
Very nice. Thank you

Issue #549, opened 2021-03-30
Template Element plugin -- Image Display


CarlosNZ:
Parameters:

- URL (of image file)
- ~~maxWidth (default 100%) -- can be "percent" or "pixels"~~
- ~~maxHeight (default 200px) -- can be "percent" or "pixels"~~
- size -- as per Semantic options: "mini", "tiny", "small", "medium", "large", "big" etc.
- alignment: "left" | "center" | "right"
- altText: for tool-tips and screenreaders

Issue #458, opened 2021-03-17
ApplicationData object for front-end evaluator


CarlosNZ:
This came up in discussion for #454, the FileUpload plugin, which needed access to user and application info, as well as global config.json data.

With back-end Actions, we have been consistent with everything that the plugin needs being passed in (rather than imported). Do we need to do the same for front-end? (Probably).

### Suggested Solution:

(After discussion with @andreievg )
Create an "applicationData" prop in FullStructure, with the following fields:
```
{
 user: (id, username, etc)
 organisation: (id, name, licence, logoURL, etc.)
 application: (id, serial, name, etc.)
 config: (whole config.json object)
 ... (more as we think of them/need them)
}
```
Note that user/org values are the user/org  associated with the application (i.e. the applicant), not the currently logged-in user. This means that any dynamic expressions in the application that refer to user/org/application should be extracted from here, so they show the _Applicant_ info rather than the Reviewer user info.

This "applicationData" object is then passed (along with Responses, currentUser, etc.) into all front-end evaluator calls.

CarlosNZ:
> Create an "applicationData" prop in FullStructure, with the following fields:
> 
> ```
> {
>  user: (id, username, etc)
>  organisation: (id, name, licence, logoURL, etc.)
>  application: (id, serial, name, etc.)
>  config: (whole config.json object)
>  ... (more as we think of them/need them)
> }
> ```

Since a lot of this info is already available in `fullStructure.info`, I think it'll be best to just add an additional `user` and `config` fields to this "info" object.

PR #556, opened 2021-03-31
#549 Image Display element plugin

CarlosNZ:
Implements #549. A fairly basic version of an Image Display plugin.

_Branched from #555 -- please review that one first._

---

Rather than get too carried away with sizing styling, I've just restricted the "size" parameter to the same as Semantic-UI image "size" options:, i.e. `mini`, `tiny`, `small`, `medium`, `large`, `big`, `huge`, `massive`

### To Test:
- Use branch from [back-end PR #291](https://github.com/openmsupply/application-manager-server/pull/291).
- Fire up a **Feature Showcase** application and navigate to **Section2**, **Page 3**.
- Upload an image file in the File Upload element, and an ImageDisplay element will become visible, displaying this image. 
- Dynamic Size and Alignment controls for the Image Display are also available as Dropdown selectors below the image

**Note**: there is a hard-coded host value ("http://localhost:8080") in one of the image template expressions. We'll be able to make this dynamic once we implement "ApplicationData" ([#458](https://github.com/openmsupply/application-manager-web-app/issues/458)) in front-end.


documentation/internal/Element-Type-Specs.md
```diff
@@ -157,7 +177,7 @@ _Multi-choice question, with one allowed option, displayed as Drop-down list (Co
...
 - **default**: `string`/`number` -- if not provided, defaults to index 0.
 - **search**: `boolean` (default: `false`) -- if `true`, the list of options can be searched and filtered by user
-- **optionsDisplayProperty**: If `options` (above) consists of an array of objects, this parameter specifies the field of each object to be displayed in the options list. For example, if `options` was a list of organisation objects (i.e. `{orgId, name, licenceNumber}`), you'd probably specify `name` as the `optionsDisplayProperty`. Note that even though one field is displayed to the user in the Dropdown list, the _entire_ selected object is saved as the selection.
+- **optionsDisplayProperty**: If `options` (above) consists of an array of objects, this parameter specifies the field of each object to be displayed in the options list. For example, if `options` was a list of organisation objects (i.e. `{orgId, name, licenceNumber}`), you`d probably specify `name`as the`optionsDisplayProperty`. Note that even though one field is displayed to the user in the Dropdown list, the _entire_ selected object is saved as the selection.
```
nmadruga, line 180: Unintended?

src/formElementPlugins/pluginProvider.tsx
```diff
@@ -45,6 +45,13 @@ class pluginProvider {
         folderName: 'textInfo',
         category: 'Informative',
       },
+      imageDisplay: {
+        code: 'imageDisplay',
+        isCore: true,
+        displayName: 'Image Display',
```
nmadruga, line 51: Could be called just "Image"

src/formElementPlugins/imageDisplay/pluginConfig.json
```diff
@@ -0,0 +1,8 @@
+{
+  "isCore": true,
+  "code": "imageDisplay",
+  "displayName": "Image Display",
+  "dynamicParameters": ["label", "description", "url", "size", "alignment"],
```
nmadruga, line 5:
```suggestion
  "dynamicParameters": ["label", "description", "url", "size", "alignment", "altText"],
```

src/formElementPlugins/imageDisplay/pluginConfig.json
```diff
@@ -0,0 +1,8 @@
+{
+  "isCore": true,
+  "code": "imageDisplay",
+  "displayName": "Image Display",
```
nmadruga, line 4:
```suggestion
  "displayName": "Image",
```

nmadruga (changes requested):
That's nice to have.

Just the test isn't the easiest. I think it would be good to add a simpler version of image displayed that we can easy change the size, alignment. It's not working on the go (which the provided test) on Section 2 Page 3

CarlosNZ:
> Just the test isn't the easiest. I think it would be good to add a simpler version of image displayed that we can easy change the size, alignment. It's not working on the go (which the provided test) on Section 2 Page 3

The dynamic switching of the image should totally be working, but you might have to do a full reset (cache flush etc.) and database_init. I had to sometimes, but it always worked for me after that.

nmadruga:
> The dynamic switching of the image should totally be working, but you might have to do a full reset (cache flush etc.) and database_init. I had to sometimes, but it always worked for me after that.

src/formElementPlugins/imageDisplay/pluginConfig.json
CarlosNZ: Again -- do you really think something like "altText" needs to be dynamic?

nmadruga (approved):

nmadruga:
Just made some comments for you and you will need to resolve the conflicts on Element-Type-Specs

documentation/internal/Element-Type-Specs.md
CarlosNZ: Good catch 😉 -- Fixed in latest commit (merged from below)

src/formElementPlugins/imageDisplay/pluginConfig.json
nmadruga: Hmm maybe I didn't get this field correctly, I though it was listing fields that could be passed in to this plugin question

CarlosNZ: No, this is just the list of fields that the Wrapper will assume could be evaluator expressions, so will run evaluator on them. Fields not listed here will be passed through to plugin as is.

src/formElementPlugins/pluginProvider.tsx
CarlosNZ: I dunno, it kind of seems like a parallel name to "Text Info". Just "Image" as a name is a bit too ambiguous, don't you think? Could be image upload, etc.

PR #557, opened 2021-04-01
#458 applicationData for front end

CarlosNZ:
Implements #458.

Requires branch from [back-end PR #292](https://github.com/openmsupply/application-manager-server/pull/292)

_Branched from #556 -- please review that one first._

---

~~There are more changed files than expected here, as I had to do quite a bit of tinkering with the types. I discovered a few errors where some of the NEW components were still using OLD types (e.g. ApplicationViewProps). In order to make this work, I had to break some of the old components, so I've replaced a couple of blocks with just placeholder JSX so they don't give compilation errors. We'll be deleting all these soon, anyway.~~ Deleted and fixed

### In Summary:
- getApplication.query updated to return the `user` and `org` associated with the application
- the above `user` and `org` objects added to `fullStructure.info` in useLoadApplication.
- system "config" data also added to `fullStructure.info` in useLoadApplication.
- This `fullStructure.info` is passed down to the plugin Wrappers (Application & Summary) as `applicationData`, and, in turn, to all element plugins
- Evaluations for Elements (visibility etc.) and Dynamic parameters in plugins now have "applicationData" as one of the objects available to query
- fileUpload plugin now gets its config info (host, etc.) from applicationData rather than importing from main project

**One question**: Do you think `applicationData` is an appropriate name for this object? It's the same name as the back-end one and I just wondered if it might be confusing that items (elements and actions) in the template both refer to "applicationData", but are referring to different objects with different data available.


nmadruga (approved):
Maybe that could have been a good use of Context, so you wouldn't need to pass `ApplicationData` all down the chain of components, but I'm not sure if we want to go this way or not. Using the nested Routes could make this context more complicated to use.

nmadruga:
> **One question**: Do you think applicationData is an appropriate name for this object? It's the same name as the back-end one and I just wondered if it might be confusing that items (elements and actions) in the template both refer to "applicationData", but are referring to different objects with different data available.

It could be named applicationDetails, so it isn't the same. But so far it doesn't seem confusing

CarlosNZ:
> Maybe that could have been a good use of Context, so you wouldn't need to pass `ApplicationData` all down the chain of components, but I'm not sure if we want to go this way or not. Using the nested Routes could make this context more complicated to use.

Andrei said not to use Context for the application stuff, since it's mostly all in fullStructure now. As far as I was aware, the "Application" context is essentially deprecated, no?

nmadruga:
We didn't want to go very much on that path, indeed. But we are still using ApplicationState and UserState context in the same places as we started. The `UserState` has been used in other places too. The idea of the Context is exactly that of not having to pass things along the call chain, that's why I'm saying.

PR #568, opened 2021-04-08
#369 Remove unused files and rename all types

### LINKED ISSUE #369: Remove old router and files used for Application

nmadruga:
Fixes #369 

### Linked to back-end PR:
[PR#225](https://github.com/openmsupply/application-manager-server/pull/296) remove application trigger states view 

### Changes
- Removed all component, containers, hooks (query and mutations) unused
- Removed old types and renamed all files/types with NEW to the correct one
- Tried to change the minimum possible to avoid conflicts, but just too many files. Prob not a good idea to do a massive revew. Just TEST EVERYTHING : D

### Pending
~~- Remove unused `constants` - will add a commit for this tomorrow~~ DONE


src/utils/hooks/useSubmitApplication.tsx
```diff
@@ -4,22 +4,13 @@ import {
...
-      setSubmitted(true)
-    },
-    onError: (submissionError) => {
-      setProcessing(false)
-      setError(submissionError.message)
-    },
+    onError: (submissionError) => setError(submissionError.message),
   })
 
   const submitFromStructure = async (structure: FullStructure) => {
```
CarlosNZ, line 16: Yeah, this is the one -- should probably rename it to just `submit` as the only reason it had this name was so as not to clash with the old `submit` method. (Should have been called submitNEW really).

CarlosNZ (changes requested):
One immediate problem -- I don't think it's just me, I tried in two different browsers:
The main application page is not showing any elements. The progress bar and Sections headings are there, but it's just blank where the form elements should be.

CarlosNZ:
> One immediate problem -- I don't think it's just me, I tried in two different browsers:
> The main application page is not showing any elements. The progress bar and Sections headings are there, but it's just blank where the form elements should be.

Just from a quick look in PageElements -- ELements are present but the `visibleElements` array is empty. Inspecting the structure, I can see that `isVisible`, `isEditable` and `isRequired` are all undefined, so probably something going wrong with the evaluation.

nmadruga:
> One immediate problem -- I don't think it's just me, I tried in two different browsers:
> The main application page is not showing any elements. The progress bar and Sections headings are there, but it's just blank where the form elements should be.

Investigating this now. I think it has to do with changing to use the NEW `ApplicationViewWrapperProps` I have removed the 3 fields from it since the NEW `ElementState` had them defined in it, but I can see they are arriving undefined in the ApplicationPage - to be passed through to `PageElements` component:
```
isEditable: undefined
isRequired: undefined
isVisible: undefined
```

src/utils/hooks/useSubmitApplication.tsx
nmadruga: Done: 6269273

CarlosNZ (approved):
Hi @nmadruga -- I have fixed the 2 issues we discussed. 
I think this is all working nicely now, I ran through all the basic functionality (apply, review) and everything seems to be working properly. Maybe you want to have another quick look at it (and review the changes I made)? But otherwise seems good to go I think.

nmadruga:
Thank you so much Carl! I'm having a quick look now and should be merging in 10 minutes

Issue #311, opened 2021-02-05
Deal with users that have more than on roles: Assigner and others


nmadruga:

nmadruga:
PR with partially work to deal with different roles. There is still some discussions about how we defined the user as `Consolidator` and if we not allow the user to be an `Applicant` and something else, how do we prevent a wrong permission setup to display on the UI?

PR #572, opened 2021-04-11
#227 & #189 Actionable in application list

### LINKED ISSUE #189: Display 'changes requested' on application list next to applications

### LINKED ISSUE #227: Display actionable in cell on Status column for Reviewer and Consolidator

nmadruga:
Fixes #227 & fixes #189

~~### Based on PR
#601~~ MERGED

### Changes
- Copying changes made for `demo-22-03` regarding list and actions for Applicant and Reviewer
  - Remove unused expansion on `ApplicationList` and `ApplicationWrapper`
  - Couple more changes to styles on `ReviewHome`, `ReviewSectionRowAction`, `ReviewSectionRowProgress` & `ReviewSectionRow`
  - Added `ApplicantActionCell` and `ReviewActionCell` to show the available actions on the list
  - Update `SectionProgress`, `ApplicationCreate` & `ApplicationPage` styles
  - Added global colours starting global layout
  - Added a few extra `constants` and changes to hooks `useCreateReview` & `useRestartReview` (Not sure what it was fixing) 
- Update other cells styles as designs `ApplicationNameCell`, `SerialNumberCell`, `StageCell` - Using new `Stage` component to temporary set specific colour to stages (hardcoded for Assignment : P future issue)
- Update `ReviewHeader` to also use new `Stage` temporary component
- Using inlineStyles in constants to help when we move to use .less files only

### Pending:
- Update Home to be using styles from demo - PR #601 
- Fix problem showing `Assign` action for users not assigned to a review ~**In Progress - as part of #311**


nmadruga:
**Warning**

About to rebase this PR to another specific to changes related to ReviewHome styles. Please hold on to review.

nmadruga:
Ready for Review now : )

nmadruga:
Hi @CarlosNZ and @andreievg can some of you have a look on this PR please? It's 7 days old now.

CarlosNZ (commented):
One initial comment, not sure if it needs to be part of this issue: when I click an item in the list, I go to the Application Home page. Then, when I click "Back", it doesn't go back to the list. I think this is because there is a redirect on the "HOme" page, so "Back" just goes to the original application Home page then re-directs again. So we probably need to make the redirect on the Home page do a `.replace()` rather than a `.push`.

src/components/List/ApplicationsList.tsx
```diff
@@ -81,45 +77,26 @@ interface ApplicationRowProps {
...
-    <Table.Row key={`ApplicationList-application-${index}-sections`} colSpan={colSpan}>
-      <Table.Cell colSpan={colSpan}>
-        <Segment color="grey">TODO: SECTIONS</Segment>
-      </Table.Cell>
+    <Table.Row
+      key={`ApplicationList-application-${application.id}`}
+      onClick={() => {
+        console.log(query)
+        if (query.userRole === 'applicant') push(`/application/${application.serial}`)
+        else push(`/application/${application.serial}/review`)
```
CarlosNZ, line 91: I don't really like how the whole row is clickable, especially since there is no cursor change to indicate that it will be. It's too easy to click and then get whisked away without wanting to. 
Possible solutions:
- make sure the cursor turns into the "clicking" icon on hover (usually achieved with `as="link"`)
- Only make the Title clickable, and make it a proper link (this would be my preferred option)
- Only make the little arrow on the right clickable

src/components/List/Cells/ApplicationNameCell.tsx
```diff
@@ -1,11 +1,6 @@
 import React from 'react'
-import { Link } from 'react-router-dom'
 import { CellProps } from '../../../utils/types'
 
-const ApplicationNameCell: React.FC<CellProps> = ({ application }) => (
-  <p>
-    <Link to={`/application/${application.serial}`}>{application.name}</Link>
-  </p>
-)
+const ApplicationNameCell: React.FC<CellProps> = ({ application }) => <p>{application.name}</p>
```
CarlosNZ, line 4: Regarding my earlier comment, I think you would put the Link on this.

src/components/List/Cells/ReviewerActionCell.tsx
```diff
@@ -0,0 +1,74 @@
...
+    reviewDraftCount,
+    reviewPendingCount,
+    reviewSubmittedCount,
+    reviewAvailableForSelfAssignmentCount,
+    serial,
+    isFullyAssignedLevel1,
+  },
+}) => {
+  const actions = []
+  const not0 = (num: number | undefined) => num && Number(num) !== 0
```
CarlosNZ, line 20: Personal preference: `notZero`, but up to you.

CarlosNZ (changes requested):
I'm getting an error when trying to review the list as a reviewer. I'm logged in as (either) Reviewer, and can see the list fine with `user-role=applicant`. But for this URL:
`http://localhost:3000/applications?type=ReviewTest&user-role=reviewer`
I get a blank page. 

Console error is `Uncaught TypeError: columnsPerUserRole_1.default[userRoles] is undefined`

nmadruga:
> One initial comment, not sure if it needs to be part of this issue: when I click an item in the list, I go to the Application Home page. Then, when I click "Back", it doesn't go back to the list. I think this is because there is a redirect on the "HOme" page, so "Back" just goes to the original application Home page then re-directs again. So we probably need to make the redirect on the Home page do a `.replace()` rather than a `.push`.

Good idea. I'll try that now : )

nmadruga:
> I'm getting an error when trying to review the list as a reviewer. I'm logged in as (either) Reviewer, and can see the list fine with `user-role=applicant`. But for this URL:
> `http://localhost:3000/applications?type=ReviewTest&user-role=reviewer`
> I get a blank page.
> 
> Console error is `Uncaught TypeError: columnsPerUserRole_1.default[userRoles] is undefined`

So, the problem is that, only on the next PR I'm actually updating the user-roles to replace `reviewer1` and `reviewer2` with `reviewer`. This needs to be improved, and also the problem with entering an un-existing user-role should be set to show the list with some error - unexpected user-role or simply reload with the first user-role available for that user.

Short answer is: we need to fix the problem with URL user-roles and to see the correct result use now `reviewer1`.

src/components/List/ApplicationsList.tsx
nmadruga: Cool. I'll go with solution 2 (and 3) so we have the title clickable but also the right arrow that kind of indicate that will take you somewhere.

src/components/List/Cells/ApplicationNameCell.tsx
nmadruga: Thanks

src/components/List/ApplicationsList.tsx
nmadruga: I'm not happy about the looks of the right arrow - but now it's showing as clickable. I wonder how this will lock with actions in place. I'm thinking we might not want this anymore...

src/components/List/Cells/ReviewerActionCell.tsx
nmadruga: Can change this now, not problem.

nmadruga:
Oh, ok. I didn't realise the actions were in the same PR... The arrow + the actions don't look good to me:
Screen Shot 2021-04-20 at 12 35 38 PM

I'm just keeping the arrow now as a arrow down, but not clickable anymore.

src/components/List/Cells/ReviewerActionCell.tsx
nmadruga: Done: 7312770

src/components/List/ApplicationsList.tsx
CarlosNZ: Yeah, it seems to be a bit redundant now.

src/components/List/Cells/ApplicationNameCell.tsx
```diff
@@ -1,6 +1,15 @@
...
 import { CellProps } from '../../../utils/types'
 
-const ApplicationNameCell: React.FC<CellProps> = ({ application }) => <p>{application.name}</p>
+const ApplicationNameCell: React.FC<CellProps> = ({ application }) => (
+  <Header
+    size="small"
+    as={Link}
+    to={`/application/${application.serial || 0}`}
+    content={application.name as string}
+  />
```
CarlosNZ, line 12: I wonder why there is no "hover" style for this link. Something to look into with styling...

CarlosNZ (commented):
Looking good, except, as you pointed out, there doesn't seem to be a way to see the "ReviewerActionCell", is there?

I'm happy to merge this if you'd rather fix it in the other issue, but is there any way to see the ReviewerActionCell" working currently?

src/components/List/Cells/ApplicationNameCell.tsx
nmadruga: Sure. Can this be flagged as something to be done as part of #592?

nmadruga:
Overall result for this PR is mergable?

CarlosNZ:
> Overall result for this PR is mergable @CarlosNZ ?

Sure, if there's no way to currently see the ReviewerActionCell in action.

CarlosNZ (approved):

nmadruga:
> Sure, if there's no way to currently see the ReviewerActionCell in action.

There is, you just need to use `reviewer1` instead of `reviewer` on the URL, like: `http://localhost:3000/applications?type=ReviewTest&user-role=reviewer1`

CarlosNZ:
> > Sure, if there's no way to currently see the ReviewerActionCell in action.
> 
> There is, you just need to use `reviewer1` instead of `reviewer` on the URL, like: `http://localhost:3000/applications?type=ReviewTest&user-role=reviewer1`

Nice -- looks great! :) Merge away.

Issue #586, opened 2021-04-12
Global font and colours


nmadruga:
Update colour variables to be used for all semantic component that we have (considering button, borders, validation, background etc). Update `site.variables` with all font styles to use ones defined in vision. Add font definition from google to `index.html` (have a look in demo-22-03) - and choose the weight based on what is in inVision: https://debs181572.invisionapp.com/overview/Drug-registration-app-V2-ckmfpptbx008401xqfx1j17ha/screens

Add media query for mobile vs desktop on each style added globally

![image](https://user-images.githubusercontent.com/16461988/114343114-53a87e00-9bb1-11eb-9450-f176bcdc40ef.png)
![image](https://user-images.githubusercontent.com/16461988/114343115-560ad800-9bb1-11eb-888f-00120c92cc21.png)

CarlosNZ:
I will start doing this, but I'm happy to pass it over or split it out if it's taking too long.

PR #612, opened 2021-04-14
#586  #191 Global fonts and colours

CarlosNZ:
Implements #586, #191 and sets a basic starting point for further styling.

#### #191
- Removed (zipped) all unused themes
- Removed (commented-out) unused components in `semantic.less` (This brings the CSS compile time from ~18s to ~12s, so would be good to get some further optimisations here. Would need the hot-reload module to be able to recompile only new changes, not the entire CSS, but it's beyond me to do that at this point)

#### #586 
- Started **documentation**: a guide to using semantic styles and over-rides -- please read (`Styles-Guide.md`)
- Design colours added as variables in `site.variables`
- Font/Heading over-rides to match designs
- Mobile/Desktop heading sizes
- A few relevant custom classes in `site.overrides`
- Showcase all heading/paragraph/form input fonts on the Home page, including separate mobile fonts.


semantic/src/site/globals/site.variables
```diff
@@ -1,24 +1,106 @@
...
+@h4Mob: 12px;
+@h5Mob: 15px;
+@h6Mob: 11px;
+
+// New variables for font-weights
+@regularWeight: 400;
+@semiboldWeight: 600;
+
+// Text colours
+@darkGrey: #4a4a4a;
```
nmadruga, line 71:
```suggestion
@darkGrey: @headersInteractiveLow;
```

semantic/src/site/elements/header.overrides
```diff
@@ -1,3 +1,78 @@
...
+  h4.ui.header {
+    font-size: @h4Mob;
+    font-weight: @semiboldWeight;
+    text-transform: uppercase;
+    color: @darkGrey;
+  }
+  h5.ui.header {
+    font-size: @h5Mob;
+    font-weight: @semiboldWeight;
+    text-transform: uppercase;
```
nmadruga, line 69: I think this one shouldn't be **uppercase**

nmadruga (approved):
Perfect! Made a quick one suggestions - maybe 2, but that's it. It's wonderful!

I have added issue #613 about improvement described in the description to improve the build time to <12s for building CSS, I think we need a little research to find how to exactly do that, but shouldn't be a big - or impediment for us to continue with the style issues next.

semantic/src/site/elements/header.overrides
CarlosNZ: Yes, good catch. :)

97d65405

I also moved the hide-on-mobile classes to the `overrides` file rather than `variables`, where they should have been. 😊

PR #616, opened 2021-04-15
#593 application create & home styles

### LINKED ISSUE #593: Update layout and use global styles: Application create page
Remember: Before changing styles take a screenshot on current page and add which style to use on each element - post to the issue so others can have a look and agree.

Just change elements with existing style. Anything not defined should be done in another (to be created) issue.

nmadruga:
Fixes #593

### Changes
- Styles of `ApplicationHeader` which wraps either `ApplicationCreate` or `ApplicationHome` (another PR)
- Still trying to mock what is in the invision, so far only part of ApplicationHeader... missing everything else : P
- Found one problem when displaying the back button to go to the application list, when the name of the application is too long, will probably need a trim:
Screen Shot 2021-04-15 at 5 30 55 PM

### DIFF
--- a/semantic/src/site/elements/container.overrides
+++ b/semantic/src/site/elements/container.overrides
@@ -1,3 +1,9 @@
 /*******************************
          Site Overrides
 *******************************/
+
+@media only screen and (min-width: 1200px) {
+  .ui.container {
+    width: 1024px;
+  }
+}
--- a/semantic/src/site/elements/container.overrides
+++ b/semantic/src/site/elements/container.overrides
@@ -7,3 +7,10 @@
     width: 1024px;
   }
 }
+
+@media only screen and (max-width: 1200px) {
+  .ui.container {
+    margin-left: 0 !important;
+    margin-right: 0 !important;
+  }
+}
--- a/semantic/src/site/elements/container.overrides
+++ b/semantic/src/site/elements/container.overrides
@@ -2,15 +2,8 @@
          Site Overrides
 *******************************/
 
-@media only screen and (min-width: 1200px) {
-  .ui.container {
-    width: 1024px;
-  }
-}
+// Overrides container width
+@maxWidth: 1024px;
 
-@media only screen and (max-width: 1200px) {
-  .ui.container {
-    margin-left: 0 !important;
-    margin-right: 0 !important;
-  }
-}
+// Overrides to remove left and right margin on mobile
+@mobileGutter: 0;
--- a/semantic/src/site/elements/label.overrides
+++ b/semantic/src/site/elements/label.overrides
@@ -1,3 +1,8 @@
 /*******************************
          Site Overrides
 *******************************/
+
+.ui.label {
+  line-height: @lineHeightSmall;
+  font-weight: @semiboldWeight;
+}
--- a/semantic/src/site/elements/label.overrides
+++ b/semantic/src/site/elements/label.overrides
@@ -3,6 +3,5 @@
 *******************************/
 
 .ui.label {
-  line-height: @lineHeightSmall;
   font-weight: @semiboldWeight;
 }
--- a/semantic/src/site/globals/site.overrides
+++ b/semantic/src/site/globals/site.overrides
@@ -134,6 +134,7 @@ p strong {
   display: flex;
   flex-wrap: wrap;
   justify-content: space-between;
+  padding-bottom: 10px;
 }
 
 // Back to application list label
@@ -158,6 +159,7 @@ p strong {
 
 #application-area .ui.header .sub.header {
   margin-top: 11px;
+  margin-bottom: 30px;
   font-size: large;
   letter-spacing: @letterSpacingSmall;
 }
@@ -166,7 +168,17 @@ p strong {
   padding: 0;
   margin-left: 0;
   background-color: transparent;
+  color: black;
   font-size: small;
+}
+
+#application-area .ui.list > .item .header {
+  color: @darkGrey;
+  font-weight: @regularWeight;
+  font-size: medium;
+}
+#application-area .ui.list > .item > i.icon {
+  color: @darkGrey;
 }
 
 #application-area .application-segment {
--- a/semantic/src/site/globals/site.overrides
+++ b/semantic/src/site/globals/site.overrides
@@ -157,7 +157,7 @@ p strong {
 
 #application-area .ui.header .sub.header {
   margin-top: 11px;
-  margin-bottom: 30px;
+  margin-bottom: 16px;
   font-size: large;
   letter-spacing: @letterSpacingSmall;
 }
--- a/semantic/src/site/globals/site.overrides
+++ b/semantic/src/site/globals/site.overrides
@@ -87,9 +87,41 @@ p strong {
 }
 
 #content-area {
-  padding-top: @headerHeight + 25px;
-  padding-right: 25px;
-  padding-left: 25px;
+  padding-top: @headerHeight + 12px;
+  padding-right: 12px;
+  padding-left: 12px;
+}
+
+/* Application home */
+#application-area {
+  font-family: @userAreaFont;
+  display: flex;
+  flex-direction: column;
+}
+
+.company-title .h4 {
+  height: 45px;
+  width: 185px;
+  letter-spacing: @letterSpacingRegular;
+  line-height: @lineHeightBig;
+}
+
+/* TODO: Add issue to trim if app name is too big*/
+.back {
+  align-self: flex-start;
+  width: 132px;
+  height: 20px;
+  font-weight: @regularWeight;
+  font-size: @h5;
+  line-height: @fontHeightRegular;
+}
+
+.rectangle {
+  height: 814px;
+  width: 800px;
+  background-color: @surfacesWhite;
+  padding: 30px 32px 50px;
+  flex: 1;
 }
 
 .sup .ui.tiny.progress {
--- a/semantic/src/site/globals/site.overrides
+++ b/semantic/src/site/globals/site.overrides
@@ -16,7 +16,7 @@ p strong {
 }
 
 /*--------------------------------
-   Main area, User are & Footer
+   Main area, User area & Footer
 ---------------------------------*/
 #main-container {
   padding: 0;
@@ -148,7 +148,6 @@ p strong {
   letter-spacing: @letterSpacingRegular;
   font-family: @alternativeFont;
   font-weight: @mildboldWeight;
-  line-height: @lineHeightBigger;
 }
 
 @media only screen and (min-width: @tabletBreakpoint) {
@@ -242,7 +241,6 @@ h1 {
 }
 h2 {
   font-size: @h2;
-  line-height: @lineHeightBig;
   font-weight: @regularWeight;
 }
 h3 {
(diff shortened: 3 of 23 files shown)


CarlosNZ:
>     * Found one problem when displaying the back button to go to the application list, when the name of the application is too long, will probably need a trim:

Can't you just make the label wider to accommodate?

semantic/src/site/globals/site.overrides
```diff
@@ -123,6 +123,38 @@ p strong {
...
+  height: 45px;
+  width: 185px;
+  letter-spacing: @letterSpacingRegular;
+  line-height: @lineHeightBig;
+}
+
+/* TODO: Add issue to trim if app name is too big*/
+.back {
+  align-self: flex-start;
+  width: 132px;
```
CarlosNZ, line 143: Just set a max-width so it can accommodate longer text?

semantic/src/site/globals/site.overrides
```diff
@@ -123,6 +123,38 @@ p strong {
...
 }
 
+/* Application home */
+#application-area {
+  font-family: @userAreaFont;
+  display: flex;
+  flex-direction: column;
+}
+
+.company-title .h4 {
```
CarlosNZ, line 133: You probably mean `h4` not `.h4`, right?
But why have the h4 tag at all?

semantic/src/site/globals/site.overrides
```diff
@@ -123,6 +123,38 @@ p strong {
...
+.back {
+  align-self: flex-start;
+  width: 132px;
+  height: 20px;
+  font-weight: @regularWeight;
+  font-size: @h5;
+  line-height: @fontHeightRegular;
+}
+
+.rectangle {
```
CarlosNZ, line 150: I think more specific class names would be useful, so it's clear what it will be/is being used for (I had to do a "find" on this to know) -- either "what it is" (e.g. `.application-content`), or "what it does" (e.g. `.alert`, `.extra-large`).

src/components/Application/ApplicationHeader.tsx
```diff
@@ -13,40 +13,31 @@ export interface AppHeaderProps {
...
   const { code, name } = template
   return (
-    <>
-      <Button as={Link} to={`/applications?type=${code}`} style={inlineStyles} icon>
-        <Icon name="angle left" />
-        {`${name} ${strings.LABEL_APPLICATIONS}`}
-      </Button>
-      <Header textAlign="center" style={inlineStyles}>
+    <Container id="application-area">
+      <Button
```
CarlosNZ, line 17: You might find this is easier as a `<Label>` rather than a `<Button>` (looks more like a label in the designs to me)

semantic/src/site/globals/site.overrides
```diff
@@ -123,6 +123,38 @@ p strong {
   margin-left: 10px;
 }
 
+/* Application home */
+#application-area {
+  font-family: @userAreaFont;
```
CarlosNZ, line 128: FYI -- @userAreaFont is not the OpenSans one, it's the one used in the Header area. So you probably don't want to set this for the whole `#application-area`. 

I'd say make a class called `.alt-font`, with just `font-family: @userAreaFont`, then apply it in conjuction with the appropriate `<h>` tags, e.g. for the company title, you'd have:
`<Header as="h2" className=".alt-font">{company.name...}`

semantic/src/site/globals/site.overrides
```diff
@@ -123,6 +123,38 @@ p strong {
...
+  align-self: flex-start;
+  width: 132px;
+  height: 20px;
+  font-weight: @regularWeight;
+  font-size: @h5;
+  line-height: @fontHeightRegular;
+}
+
+.rectangle {
+  height: 814px;
```
CarlosNZ, line 151: Shouldn't need to specify a height.

CarlosNZ (changes requested):
Sorry, probably should have waited till this is no longer a "Draft" PR, but just a couple of initial comments. Look forward to seeing it properly when it's ready :)

semantic/src/site/globals/site.overrides
nmadruga: Ok. Changing that along with the change from **Button** to **Label** : )

semantic/src/site/globals/site.overrides
nmadruga: Just my attempt to change the h4 definition. Will do as you suggested in another comment

src/components/Application/ApplicationHeader.tsx
nmadruga: Done: a58df28

semantic/src/site/globals/site.overrides
nmadruga: Done: a58df28

semantic/src/site/globals/site.overrides
nmadruga: Done: a58df28

semantic/src/site/globals/site.overrides
nmadruga: Done: a58df28

semantic/src/site/elements/label.overrides
```diff
@@ -1,3 +1,8 @@
 /*******************************
          Site Overrides
 *******************************/
+
+.ui.label {
+  line-height: @lineHeightSmall;
```
CarlosNZ, line 6: Just wondering -- I noticed you've used line-height in a few places. Do we really need it? In most cases the automatic line height is fine, so just wondering if you're going for something in particular?

semantic/src/site/globals/site.overrides
```diff
@@ -15,6 +15,9 @@ p strong {
   font-weight: @semiboldWeight;
 }
 
+/*--------------------------------
+   Main area, User are & Footer
```
CarlosNZ, line 19:
```suggestion
   Main area, User area & Footer
```

semantic/src/site/globals/site.overrides
```diff
@@ -123,6 +120,48 @@ p strong {
...
+#application-area {
+  display: flex;
+  flex-direction: column;
+  padding-left: 0;
+  padding-right: 0;
+}
+
+.company-title {
+  letter-spacing: @letterSpacingRegular;
+  line-height: @lineHeightBig;
```
CarlosNZ: Same comment here about line height. If you're trying to make more space below, then `margin-bottom` is probably better. You only need to consider line height when considering multiple lines of the same text (and this company title should never span multiple lines).

semantic/src/site/globals/site.overrides
```diff
@@ -123,6 +120,48 @@ p strong {
...
+  height: 25px;
+  background-color: transparent;
+}
+
+/*--------------------------------
+      General used classes
+---------------------------------*/
+
+// In use for ApplicationCreate and ApplicationHome
+#application-content {
```
CarlosNZ, line 169: IDs should only appear ONCE, they're meant to be unique identifiers. If it's going in multiple places, then just make it a `.class`.

semantic/src/site/elements/container.overrides
```diff
@@ -1,3 +1,9 @@
 /*******************************
          Site Overrides
 *******************************/
+
+@media only screen and (min-width: 1200px) {
+  .ui.container {
+    width: 1024px;
+  }
+}
```
CarlosNZ, line 9: Why do we need this? Semantic-UI already has good default width settings for Container, I think.
Or, if the default is too wide (I think it's actually 1027px), then just over-ride the variable that has this value.

semantic/src/site/globals/site.overrides
```diff
@@ -123,6 +120,66 @@ p strong {
...
+}
+
+#application-area .top-container {
+  display: flex;
+  flex-wrap: wrap;
+  justify-content: space-between;
+}
+
+#application-area .company-title {
+  width: 100%;
```
CarlosNZ, line 140: I don't think the .top-container needs to have flex definitions. The natural vertical flow of a straight "div" is enough. Then you won't need the "width" property in .company-title.

semantic/src/site/globals/site.overrides
```diff
@@ -123,6 +120,66 @@ p strong {
...
+  letter-spacing: @letterSpacingRegular;
+  font-family: @alternativeFont;
+  font-weight: @mildboldWeight;
+  line-height: @lineHeightBigger;
```
CarlosNZ, line 144: Same comment about line height here. If you're just wanting space below, then `margin-bottom` is better. Line height only becomes relevant for multiple lines within a text block. And .company-title should never be more than one line.

CarlosNZ (changes requested):
Starting to look good! :) 

I just made a few more comments, mainly ways to simplify further.

semantic/src/site/globals/site.overrides
nmadruga: Cool, good point. I was just trying to bring what was in the invision to it, but I agree that for single lines that's probably not done using lineHeights, I'll remove those settings now.

semantic/src/site/globals/site.overrides
nmadruga: Thanks

semantic/src/site/globals/site.variables
```diff
@@ -85,9 +87,28 @@
...
+}
+.large {
+  font-size: 20px;
+}
+.x-large {
+  font-size: 27px;
+}
+.xx-large {
+  font-size: 36px;
+}
```
CarlosNZ, line 111: Why classes not @variables?

semantic/src/site/globals/site.overrides
nmadruga: Oh, All good, thanks for pointing that out

semantic/src/site/elements/container.overrides
nmadruga: Hmmm Ok, I can do that, but have also defined something extra for mobile - although we aren't  doing mobile for now, the container was showing all the content of application very narrow. So I added this:

semantic/src/site/globals/site.overrides
nmadruga: Done... althought instead of using the class .alt-font I added as one of the properties on `.company-title`

semantic/src/site/globals/site.variables
nmadruga: To extend how they are used (on sematic-ui elements). Do you think this wont work?

CarlosNZ: Are you saying that they're over-riding existing Semantic classes? That's cool, but then they should probably be in .overrides, not .variables.

Also -- I'm not sure you can use them the way you have, just putting them inside other definitions. I dunno, maybe its a LESS feature, but it's not valid CSS.

semantic/src/site/globals/site.overrides
nmadruga: If I change this to be a class then it is not finding other styles (defined used this class included). Can you give me a hand with that please?

semantic/src/site/globals/site.overrides
nmadruga: Done: 466b283

nmadruga:
Updated PR with changes to `ApplicationCreate` page:
Screen Shot 2021-04-19 at 3 33 40 PM

nmadruga:
Created second PR just with changes related to #593

Issue #576, opened 2021-04-11
File upload fix --  Hide upload button once fileCountLimit has been reached.


nmadruga:

CarlosNZ:
Just going to re-open this as I discovered a case where it causes a problem. When there is NO file limit specified (in the template), the button doesn't show up at all. 🙄

Will update with a new PR shortly.

PR #617, opened 2021-04-15
#576 Hide Upload button if fileCountLimit reached

CarlosNZ:
#576 
Small tweak to only show the button if file count is less than the limit.


nmadruga (approved):
LGTM - tested and even when I reload the application the button is missing. All good!

Issue #295, opened 2021-01-25
Create review_responses on Consolidation start/continue


nmadruga:
_Andrei Notes_

Structure created here : [#390](https://github.com/openmsupply/application-manager-web-app/issues/390), should provide enough info to streamline this issue:

Whenever 'continue' or 'start' is pressed from within the review home page.
* either run useRestartReview or useCreateReview  (if review doesn't exist) 
* both take assignment and structure as prop at some stage, and both can check for level > 1, then use:
   * isPendingReview to create review responses (they should also be linked to review_responses, from structure `previousLevelReviewResponse`)
  
> Whenever 'continue' or 'start' is pressed from within the review home page. 

This means that 'start' button on section with pending consolidation (when consolidation is in progress for another section), is only indicative and 'continue' button will also start any consolidations that are pending

_End Of Andrei Notes_

Continue work done in #423
Level 2 reviewer ->

* look at all of the latest review_response from review that have status == submitted (not review_responses that are status submitted)
* same logic as level 1 reviewer but for review_response

----

There is a diagram to be followed to be merged with [pull](https://github.com/openmsupply/application-manager-web-app/pull/309)

And a few issues should be done before this to help what needs to be done in this one: 
- ~~[Back-end#207](https://github.com/openmsupply/application-manager-server/issues/207)~~ Done
- ~~[Back-end#195](https://github.com/openmsupply/application-manager-server/issues/195)~~ Done
- ~~#274~~ Done
- #273 
- #423

CarlosNZ:
Hey, this says that #273 should be done before this one? Do you think that's still the case @nmadruga ? You want me to take that one, or were you planning on doing it next?

nmadruga:
The work to be done in issue #273 was just to add what could be missing on the review **Start page** which should already display the option to start a review for the consolidator (it might need some tweaks though).
You can try our the scenario using `http://localhost:3000/application/12345/review` while loged with **testConsolidator1** user. The current review status is: Reviewers lvl 1 have submitted 2 reviews (of each section) and the Review lvl 2 is already assigned to the Consolidator.

Issue #537, opened 2021-03-26
Epic #79 - Add new review to consolidation


nmadruga:
## Description

Consolidations can start when one review is submitted (for example for one section only) then new review submitted should set the review (consolidation) as status "pending", show a prompt to the user to restart the consolidation and go to homepage. After re-starting the consolidation should get new questions (from new review) and display them along the existing ones. We are using review_response.status to track what was submitted (to create duplicates) and what wasn't yet (draft) and don't duplicate. This way we keep only one review_response  for consolidator to act on.

## Documentation
- [ ] TODO: Add as part of [Epic #38](338) Public docs & tutorials

### Sprint 33:
- [x] When consolidation is in progress and new review is submitted needs to either show in the UI that the consolidation should be re-started or reload the ReviewHome #675

Issue #390, opened 2021-02-28
Update useGetReviewStructure to add reviewResponses for lvl 2 consolidation first review


nmadruga:

andreievg:
Would need to add `LatestPreviousLevelResponse`, `ConsolidationProgress`, `firstNotDoneColsolidationPage`,`canSubmitAs` calculations.

During review work, we've added some helpers for `actions` on review home, these will also need to be updated for consolidator.

Some diagrams for reference:
https://app.diagrams.net/#G1RRcvqdU2ZXR0F_gVAWknO_94kLR2pbah
https://app.diagrams.net/#G12_fvjjn_dr7YbwR-z-2_g_QI0tnS13nf
https://app.diagrams.net/#G1gAJrF1S2xrg-tZrzKxJNqV9Mtul096zZ

andreievg:
this seems like a more natural transition:

https://app.diagrams.net/#G10qb2_JN9SeL6JkNBjk7bOR3ond76NFUn

![Level 2 Review Structure Changes (1)](https://user-images.githubusercontent.com/26194949/115547077-f1aaef80-a2f9-11eb-8385-e215ad15169e.png)

Issue #276, opened 2021-01-19
Agree/disagree Consolidation modal


nmadruga:
Should make the mutation with the consolidator decision and then show Consolidation decision in the page.

Designs (as discussed in meeting on 06/05/2021) is a mix of some of existing designs in Miro. The modal should show always wide - not to worry about Mobile for now, it would have a special version for it later...
- Show on the right side as in the latest designs plus 
- **Note**: History area at the bottom to be implemented in #477
<img width="498" alt="Screen Shot 2021-05-06 at 8 38 29 PM" src="https://user-images.githubusercontent.com/16461988/117269309-49c22400-aeac-11eb-831a-85ef05be95c8.png">

- On the left display the current ApplicationResponse (SummaryView)
- **Note**: Don't need to have the preview for PDFs for now
<img width="578" alt="Screen Shot 2021-05-06 at 8 39 04 PM" src="https://user-images.githubusercontent.com/16461988/117269287-4333ac80-aeac-11eb-8d8a-413336b45a76.png">

- For Re-review or Consolidation show the previous Reviewer/Consolidator comments bellow the ApplicationResponse, similar to how is in the ReviewPage:
<img width="151" alt="Screen Shot 2021-05-06 at 8 46 14 PM" src="https://user-images.githubusercontent.com/16461988/117269131-213a2a00-aeac-11eb-80c6-b822b3f9994b.png">

PR #632, opened 2021-04-23
295 create review response on consolidation start/continue

andreievg:
Fixes: #295 

Branched from 390

Require back end: 390F-Consolidation-Structure-Changes-Test
**Note**: This is a branch for testing this PR (not a PRd one)

### Changes

* Changed create review mutation and useCreateReview to allow it to be used for both consolidation and normal review creation
* useRestartReview should also work for consolidator and isPendingReview 
* updated assignmentId on formElement to be reviewQuestionAssignmentId, to be more specific
* combined logic for start, continue, restart and re-review (seemed easier to follow and dryer)

### Testing

Check snapshots from 390 PR for tests

* Try different combinations of submitting review, starting consolidation, submitting another review and seeing what the progress, actions and review page shows.
* Can also check inspector for graphQL queries (i.e. see which responses are created when consolidation is restarted


nmadruga:
11 days ago! Probably time to get this one in : D

src/components/Review/ReviewSectionRowAction.tsx
```diff
@@ -13,33 +13,36 @@ const ReviewSectionRowAction: React.FC<ReviewSectionComponentProps> = (props) =>
     action,
     section: { reviewProgress },
     isAssignedToCurrentUser,
+    thisReview,
```
nmadruga, line 16:
```suggestion
    isAssignedToCurrentUser,
```

Not used

src/components/Review/ReviewSectionRowAction.tsx
```diff
@@ -60,57 +63,59 @@ const ReviewSectionRowAction: React.FC<ReviewSectionComponentProps> = (props) =>
 const reReviewableCount = (reviewProgress?: ReviewProgress) =>
   (reviewProgress?.totalNewReviewable || 0) - (reviewProgress?.doneNewReviewable || 0)
 
-// RE-REVIEW button
-const ReReviewButton: React.FC<ReviewSectionComponentProps> = ({
+// START CONTINUE OR RESTART BUTTON
```
nmadruga, line 66:
```suggestion
// START REVIEW, CONTINUE REVIEW OR RE-REVIEW BUTTON
```

src/utils/hooks/useCreateReview.tsx
```diff
@@ -1,45 +1,67 @@
...
+      }
     )
 
-    const applicationResponses = reviewableElements.map((element) => ({
-      applicationResponseId: element.response?.id || 0,
-      reviewQuestionAssignmentId: element.assignmentId,
-    }))
-    const result = await createReviewMutation({
+    return {
+      trigger: Trigger.OnReviewCreate,
```
nmadruga, line 44: We though it was better to set this on the mutation to make sure it is passed, would this way of doing this (using `ReviewInput`) have this as a compulsory input field?

src/utils/hooks/useCreateReview.tsx
```diff
@@ -1,45 +1,67 @@
...
+      reviewAssignmentId: assignment.id,
+      reviewResponsesUsingId: {
+        create: reviewResponseCreate,
+      },
+      // create new empty decision
+      reviewDecisionsUsingId: {
+        create: [{ decision: Decision.NoDecision }],
+      },
```
nmadruga, line 52: Also another one that would be preferable set as a compulsory part of this `mutation` in my opinion.

src/utils/hooks/useRestartReview.ts
```diff
@@ -25,23 +31,36 @@ const useRestartReview: UseRestartReview = ({ reviewId, structure, assignment })
 
   const constructReviewPatch: ConstructReviewPatch = (structure) => {
     const elements = Object.values(structure?.elementsById || {})
-
+    // Check for draft review response in case consolidation is ongoing and new review response is added
```
nmadruga, line 34: Hmmm nice! This should probably part of [Epic #79 - Add new review to consolidation](https://github.com/openmsupply/application-manager-web-app/issues/537) already implemented here : D

src/utils/hooks/useRestartReview.ts
```diff
@@ -25,23 +31,36 @@ const useRestartReview: UseRestartReview = ({ reviewId, structure, assignment })
...
+        thisReviewLatestResponse,
+        response,
+        reviewQuestionAssignmentId,
+        latestPreviousLevelReviewResponse,
+      }) => {
+        const applicationResponseId = assignment.level > 1 ? undefined : response?.id
+        const reviewResponseLinkId =
+          assignment.level === 1 ? undefined : latestPreviousLevelReviewResponse?.id
+        // create new if element is awaiting review
+        const shouldCreateNew = isPendingReview
```
nmadruga, line 57: Hmm I'm not sure if I got this straight, isn't all elements in `reviewableElements` already have `isPendingReview` = true. So why do this check `shouldCreateNew ? null : thisReviewLatestResponse?.decision` ?

nmadruga (changes requested):
Few comments, overall easy to follow the idea. And hopefully this is already the logic behind what is needed to have Epic#79 on its way : )

Will be testing now and come back with results - I also will be checking PR for #390 after this one, just started with this since it was a smaller overhead for me and also because I'm using this one to test both - as suggested.

src/utils/hooks/useCreateReview.tsx
andreievg: hmm, i see your point, was wondering what the difference is between setting it in mutation or here ? is it to protect someone using that mutation outside of this hook ?  (can we say only usable with this hook in the query comment, would that be sufficient ?

I would have to change to another type of mutation to set it explicitly in mutation, i was tying to avoid `ReviewResponseReviewIdFkeyReviewResponseCreateInput` etc.. which would need to be used

src/utils/hooks/useCreateReview.tsx
andreievg: hmm, i see your point, was wondering what the difference is between setting it in mutation or here ? is it to protect someone using that mutation outside of this hook ?  (can we say only usable with this hook in the query comment, would that be sufficient ?

I would have to change to another type of mutation to set it explicitly in mutation, i was tying to avoid `ReviewResponseReviewIdFkeyReviewResponseCreateInput` etc.. which would need to be used

We can make it not null in the db schema, that would protect it from having a null value.

src/utils/hooks/useCreateReview.tsx
andreievg: As per comment above, I would need to re-write it, still thought quite interested to know why you inclined to have it set in mutation vs here (would this be enough, a comment: 173085881e1c39f7c16e5e960a1f4424cfa00067)

src/utils/hooks/useRestartReview.ts
andreievg: I think this is an issue, thanks for pointing it out

andreievg: I had to create some diagrams to make sense of it all again:

it's this one that address this comment: https://app.diagrams.net/#G1ttYeAHmdxWCnxyTY5UX1VSRDFPe-5Nvd

And here are the changes:

f72b56f838d337fd9e714cf67cbed499edf22eb0

More diagrams:

https://app.diagrams.net/#G1345qSOpVZ3PIw-_YXsZg_qgvm8CmIMDy
https://app.diagrams.net/#G1XyiT_SSpjy521KJA-hJ8WjGgdqj8m5Ru
https://app.diagrams.net/#G1JmZc2daHZswUyjOFWZicJy8sttpS_-eZ

src/utils/hooks/useCreateReview.tsx
nmadruga: Well, it needs to be set to make the action work. I guess it's fine to have it included as part of the hook, since we are the ones using it and we know this is required. A comment will be enough, but it can't be `null` since that's how we use to set a already processed trigger - changing it to null.

src/utils/hooks/useCreateReview.tsx
nmadruga: Yes. That's fine, my worrying was to prevent bugs happening from changing this triggers. In this case it's pretty obvious, but some other cases (other tables that uses triggers) it might be not that obvious.

nmadruga:
Cute message on commit 1730858 : )

src/utils/hooks/useRestartReview.ts
nmadruga: This is great. I'm trying to read those, ideally next step would be to have it updated (on top of existing PRs) to have it on our wiki for future reference. Thanks for doing this, it will probably take me half-hour to read this and make an exact idea of how this is working, but that's for a good sake of understanding and making sure there isn't any gaps to this logic. I'll make any comments and questions direct to you in Telegram so this can be more straight-forward.

src/utils/hooks/useCreateReview.tsx
andreievg: ohh true, yeah it can't be null (the more we use nulls btw, the more I think we should avoid using nulls for logic : )

nmadruga (approved):

nmadruga:
I'm approving this now. It seems robust to me, and it will be easier to test it once we have my PR of #275 & #276 today. So any adjustments required can be done in new issues.

Just a request - please add an issue to update this documentation to the wiki, so when you have some time available you cna add that in and we keep it updated with all this material - I'm still having a look on it. It's so useful!

andreievg:
> Just a request - please add an issue to update this documentation to the wiki, so when you have some time available you cna add that in and we keep it updated with all this material - I'm still having a look on it. It's so useful!

https://github.com/openmsupply/application-manager-web-app/issues/694

I was getting lost without these diagrams

andreievg:
thanks for review !

Issue #561, opened 2021-04-06
Setup to use pre-defined colour of stage label 


nmadruga:
Now the colour of the stage label is getting set on the ReviewPage hardcoded checking if label == **Assessment** or else. The colour to display on the label (ReviewPage and List) needs to be  setup with a colour for each Stage name (which is swet dynamically per template). How we will do this needs to be defined by this issue.

<img width="933" alt="Screen Shot 2021-04-07 at 11 30 55 AM" src="https://user-images.githubusercontent.com/16461988/113790147-c6ba8a80-9794-11eb-89b3-0d7ba93db22b.png">

CarlosNZ:
I'll do this as part of List styles #592.

I'll add a "colour" field to the stage table, which stores a hex value (will have default blue perhaps) and this will be read by front-end and styled in-line.

PR #633, opened 2021-04-23
#592 #561 Application List styling (& #629)

### LINKED ISSUE #629: Bug: List loading problem
List is not reliably loading when refreshing the page.

Quite a lot of the time I get:
Screen Shot 2021-04-22 at 2 03 02 PM

Something must be not waiting properly.

CarlosNZ:
Styles and layout settings for the Application List (and related).

Requires [back-end PR branch #319](https://github.com/openmsupply/application-manager-server/pull/319) (for schema changes to show stage label colour)

Should also cover #561, but let me know if you think that issue needs to stay open if this doesn't do enough.

**Update**: this also fixes #629. There was a problem of no userId being specified in the getApplicationList query if the user info hadn't loaded from the server yet, so I've just added a default of 0 to the $userId variable in that query.

### Global layout:
- `#list-container` for wrapping entire list
- `#list-top` for the components above the list (with `.ui.header` override)

### Global over-rides (affect all components of this type)

- Link colours (interactiveHigh -blue)
- Table over-rides to match design (table.overrides)
- Small over-ride for icons in (all) buttons
- border-radius for (all) buttons (8px)
- Search box color overrides (search/input)

### Custom classes
- `.stage-label` for Labels showing stage (except colour, see below)
- `.list-row` - padding for each row
- `flex-grow-1` - just adds `flex-grow: 1` property to any element -- used for forcing an element in a flexbox to consume maximum width of remaining
- `user-action` - font-styles for clickable "Actions" in list

### Other
- modified `@fontXXsmall` and `@fontXsmall` to be smaller  (so we have a wider range of options)
- replaced all inline styles (related to the list) with appropriate class names
- updated most List Cell components to be more accurate and consistent (some of these will need further work once we have them working properly)
- "Stage" labels get their colour from the database and set it using in-line style (since it's dynamic)


semantic/src/site/globals/site.overrides
```diff
@@ -226,6 +238,38 @@ p strong {
   padding: 0;
 }
 
+/*--------------------------------
+      Application List
+---------------------------------*/
+
+#list-container {
+  padding: 25px 0;
+  max-width: 900px; // Maybe this could be wider if table is getting cluttered?
```
andreievg, line 247: Hmm, i think this should be ok for now, but was wondering why not 'width' 100%

src/utils/helpers/list/mapColumnsByRole.ts
```diff
@@ -29,7 +29,7 @@ import { ColumnDetails } from '../../types'
 
 const allColumns: { [key in APPLICATION_COLUMNS]: ColumnDetails } = {
   SERIAL_NUMBER: {
-    headerName: '',
+    headerName: 'Serial',
```
andreievg, line 32: i think this is headerless in designs ?

andreievg (approved):
looks good to me tbh, just a couple of questions

semantic/src/site/globals/site.overrides
CarlosNZ: Deb's design is a 900px container within the larger container, so it has some margin at the sides:
Screen Shot 2021-04-28 at 3 54 05 PM

src/utils/helpers/list/mapColumnsByRole.ts
CarlosNZ: Yes, I thought those design headers were kind of "provisional". 

I dont' really mind either way. In one sense it looks better without it, but if you sort by that column, the sort arrow without a word next to it looks kind of silly, so... 🤷‍♂️

PR #635, opened 2021-04-27
#620 Fix bug user role based on user

### LINKED ISSUE #620: Bug: when entering an unexisting user-role on the URL throws an error and display black page

nmadruga:
Fixes #620 

### Changes
- Quick fix to remove 2 problems:
 1. When entering an invalid userRole on the URL (it will replace with first existing for current user)
 2. When entering a userRole that the current user doesn't have permission (also replaces with first one user has permission)
- Changed the order of userRoles to always show list based on `applicant` user when user can act as applicant or reviewer.
- We don't display on the UI how to swap between different user, but you can enter another (valid) userRole to see the list from another perspective.

### DIFF
--- a/src/utils/helpers/list/findUserRole.ts
+++ b/src/utils/helpers/list/findUserRole.ts
@@ -15,10 +15,10 @@ type UserRoles = {
  */
 
 const userRoles: UserRoles = {
+  applicant: [PermissionPolicyType.Apply],
   consolidator: [PermissionPolicyType.Assign, PermissionPolicyType.Review],
   reviewer: [PermissionPolicyType.Review],
   assigner: [PermissionPolicyType.Assign],
-  applicant: [PermissionPolicyType.Apply],
 }
 
 // permissions: Array<PermissionPolicyType>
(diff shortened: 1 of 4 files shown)


src/utils/helpers/list/findUserRole.ts
```diff
@@ -15,26 +15,49 @@ type UserRoles = {
...
+  type: string,
+  userRole: string
+) => {
+  const list = Object.values(USER_ROLES)
+  const existing = list.includes(userRole as USER_ROLES)
+  if (!existing) return false
+
+  // If userRole correspond to one existing, check if user has permission
+  const userRoles = getUserRolesForType(templatePermissions, type)
+  return userRoles?.includes(userRole) || false
```
CarlosNZ, line 60: Is the `|| false` necessary?

CarlosNZ (approved):
Looks good, seems to work well. Thanks. :)

src/utils/helpers/list/findUserRole.ts
nmadruga: I'm not sure what does `userRoles?includes` returns in case `userRoles` is undefined. I'll check now : )

nmadruga: Thanks, that was def not needed : )

Issue #591, opened 2021-04-12
Update layout and use global styles: Login page


nmadruga:
Remember: Before changing styles take a screenshot on current page and add which style to use on each element - post to the issue so others can have a look and agree.

Just change elements with existing style. Anything not defined should be done in another (to be created) issue.

CarlosNZ:
Screenshots from Invision board:

<img width="1155" alt="Screen Shot 2021-04-28 at 10 08 40 PM" src="https://user-images.githubusercontent.com/5456533/116387022-78b61580-a86e-11eb-80ad-ea722a9cddd7.png">
<img width="701" alt="Screen Shot 2021-04-28 at 10 08 54 PM" src="https://user-images.githubusercontent.com/5456533/116387040-7eabf680-a86e-11eb-843c-4656671c15f1.png">
<img width="778" alt="Screen Shot 2021-04-28 at 10 09 05 PM" src="https://user-images.githubusercontent.com/5456533/116387049-81a6e700-a86e-11eb-895f-bf391aaf1299.png">

Issue #548, opened 2021-03-30
Allow user to login without an organisation


CarlosNZ:
So there would be a "None" option in the "Select Org" dialog on login.

We would (at some point) make it so a user could only apply to join an organisation if logged in with no current org (so they're not applying AS and existing user-org)

PR #644, opened 2021-04-29
#591 #548 Login UI & Log in without org

CarlosNZ:
Applied designer styles to Login interface (#591), and since I had to refactor the org login a bit, it was straightforward to also implement #548 (Login without org)

### Summary of changes

-- set default colours for form error conditions (should also affect application forms, etc.)
-- some additional generic "header" styles (for margins essentially)
-- an additional "list" class to prevent bottom-most divider appearing (in list of orgs) -- `.no-bottom-divider`
-- generic classes for `.center-text`, `.alert` (makes text pink), `.clickable` (shows "click" cursor when hovering, so we can see "onClick" links)
-- login page layout CSS
-- additional global variable for `@hoverOpacity` (for link hovering)
-- updated login logic to handle new layout, including a "no organisation" option for org selection

### Still to-do:

- Need a higher res version of the mSupply man logo. I can't remember where my master file is for the one I made, will have to find it.
- Probably should include the logos in the list of orgs, right?

### Testing

Try and log in!

Be sure to check the "error" state when you log in incorrectly:
Screen Shot 2021-04-29 at 2 54 00 PM


src/containers/User/Login.tsx
```diff
@@ -15,10 +15,16 @@ const Login: React.FC = () => {
   const [isError, setIsError] = useState(false)
   const [networkError, setNetworkError] = useState('')
   const [loginPayload, setLoginPayload] = useState<LoginPayload>()
-  const [selectedOrgIndex, setSelectedOrgIndex] = useState(0)
+  const [selectedOrgId, setSelectedOrgId] = useState<number>(-1)
   const { push, history } = useRouter()
   const { onLogin } = useUserState()
 
+  const noOrgOption: OrganisationSimple = {
+    orgId: 0,
```
andreievg, line 23: can 0 and -1 be constants (LOGIN_AS_NO_ORG, NO_ORG_SELECTED)

andreievg (changes requested):
looks good, but a couple of comments (constants and hover).

also, it would be more readable to have JSX like this:

`https://github.com/openmsupply/application-manager-web-app/compare/%23591-login-page...%23591-login-page-jsx-order`

src/containers/User/Login.tsx
```diff
@@ -70,77 +64,121 @@ const Login: React.FC = () => {
...
+          {loginPayload && (
+            <p>
+              <strong>{messages.LOGIN_ORG_SELECT}</strong>
+            </p>
+          )}
+          {loginPayload && loginPayload?.orgList && (
+            <List
+              celled
+              relaxed="very"
+              className="clickable no-bottom-border"
```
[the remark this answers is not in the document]
CarlosNZ: Yeah, that makes more sense. 98540422

src/containers/User/Login.tsx
CarlosNZ: a36683d8

CarlosNZ:
> also, it would be more readable to have JSX like this:
> 
> `https://github.com/openmsupply/application-manager-web-app/compare/%23591-login-page...%23591-login-page-jsx-order`

Sweet. 9cc305ac

CarlosNZ:
Have also replaced icon/logo with higher quality version, and made a new folder for image_resources (i.e. photoshop edits, original sources, etc.)

semantic/src/site/globals/site.overrides
```diff
@@ -22,7 +22,72 @@ a {
...
+  flex-direction: row;
+  align-items: center;
+  justify-content: center;
+}
+
+.login-header {
+  margin: 0 !important;
+  text-transform: uppercase !important;
+  letter-spacing: @letterSpacingRegular !important;
+}
```
nmadruga, line 78: I wonder if this needs to be `!important` because it's done here and not on the `headers.overrides` file?

semantic/src/site/globals/site.overrides
```diff
@@ -22,7 +22,72 @@ a {
...
+
+.image-icon {
+  height: 45px;
+  margin-right: 12px;
+}
+
+.form-extra-spacing {
+  margin-bottom: 25px !important;
+}
```
nmadruga, line 87: Same thing just mentioned but for `form.overrides`

src/containers/User/Login.tsx
```diff
@@ -65,82 +62,122 @@ const Login: React.FC = () => {
...
+    attemptLoginOrg({
+      orgId,
+      JWT: loginPayload.JWT,
+      onLoginOrgSuccess: (loginOrgResult: LoginPayload) => {
+        finishLogin(loginOrgResult)
+      },
+    }).catch((error) => {
+      setNetworkError(error.message)
+    })
+  }, [selectedOrgId])
```
nmadruga, line 90: Can't be done included in the previous `useEffect`, remember we were trying to avoiding having more than one of these per module

nmadruga (commented):
Very nice. I have just a few questions. I'm testing now and checking the styles visually.

The styles look very neat. Thanks for that, just one tiny thing is it's an easy fix:

The username on input is showing yellow - maybe can be white as is the design:
Screen Shot 2021-04-30 at 10 25 28 AM
Screen Shot 2021-04-30 at 10 25 35 AM

CarlosNZ:
> The username on input is showing yellow - maybe can be white as is the design.

That's weird, it should be like this:
Screen Shot 2021-04-30 at 10 30 56 AM

I think the yellow you're seeing might be coming from your browser's password manager over-riding the CSS

nmadruga:
> think the yellow you're seeing might be coming from your browser's password manager over-riding the CSS

Oh, that makes sense! It's the reusable fields extension I think.

CarlosNZ:
> Also the text inside the Input seems slight different, a bit bigger and other font maybe, is this another case of using the alternative font?

The text looks the same size on mine (perhaps a slightly different shade, but black is probably fine for user text):
Screen Shot 2021-04-30 at 10 35 06 AM
Screen Shot 2021-04-30 at 10 35 00 AM

semantic/src/site/globals/site.overrides
CarlosNZ: Yeah, it is required, as the component files take precedence over the ".site" properties, so these don't work without it.

semantic/src/site/globals/site.overrides
CarlosNZ: Same

src/containers/User/Login.tsx
CarlosNZ: You mean with both `loginPayload` and `selectedOrgId` in the same dependency array?

How would I then check for which one changed?

andreievg (approved):

src/containers/User/Login.tsx
nmadruga: I think it is anything that isn't dealt as a loginPayload change? I mean, you add the code for `selectedOrgId` after the code done for `loginPayload`. Just so it's concise in the same `useEffect`, like the portal to any changes on the current state of the login.

semantic/src/site/globals/site.overrides
nmadruga: Why not move to `headers.overrides` then?

Issue #594, opened 2021-04-12
Update layout and use global styles: Application page + Progress Bar


nmadruga:
Remember: Before changing styles take a screenshot on current page and add which style to use on each element - post to the issue so others can have a look and agree.

Just change elements with existing style. Anything not defined should be done in another (to be created) issue.

CarlosNZ:
FYI -- I'm doing this, but I'll need to split it into two PRs. I'm nearly done with Application page, but Progress bar (sidebar) will be quite a bit more fiddling around, so I'll PR the first part shortly

PR #647, opened 2021-04-30
#594 Application form styles (Part 1)

CarlosNZ:
I just want to get this PR out before I go, but it doesn't include styling for the Progress Bar or Summary Page (~~or Nav Bar~~), so I will do those in a separate PR(s).

But this tidies up the main form, so filling in applications will be cleaner for the demo, so thought it best to get it out like this.

**UPDATE:** Have updated the Nav Bar area to the interim solution we discussed

Note, the branch name is incorrect -- I thought I was working on a different issue.

Will revise over the weekend (maybe), but this should be okay for the demo.

I've restructured the main divs/containers for the Application Area, etc. and simplified them, set reasonable column widths for the main form area, set appropriate headers/classes for form headers, etc, and matched a few more colours.

Also updated z-index of things that should be on top of others.

For Nav bar I've just made it the height of the main Footer (so it completely obscures it), and full width of screen.


src/containers/Application/ApplicationHome.tsx
```diff
@@ -49,19 +49,19 @@ const ApplicationHome: React.FC<ApplicationProps> = ({ structure, template }) =>
...
       <ChangesRequestedTitle status={current?.status} isChangeRequest={isChangeRequest} />
-      <ApplicationHomeWrapper startMessage={template.startMessage}>
+      <ApplicationHomeWrapper startMessage={template.startMessage} name={template.name}>
         <ApplicationSections fullStructure={structure} />
+        {current?.status === ApplicationStatus.Draft && !firstStrictInvalidPage && (
+          <Segment basic className="padding-zero" textAlign="right">
+            <Button as={Link} color="blue" onClick={handleSummaryClicked}>
+              {strings.BUTTON_SUMMARY}
+            </Button>
+          </Segment>
```
nmadruga, line 61: For consistency it might be better to pass this button the same way you did for ApplicationCreate, BTW this button isn't included in the original designs, I just though that, if all sections are completed and valid, we should allow the user to go straight to the `ApplicationSummary`  from `ApplicationHome`, peek how it display now:
Screen Shot 2021-05-04 at 10 05 11 AM

src/containers/Application/ApplicationPage.tsx
```diff
@@ -54,36 +60,35 @@ const ApplicationPage: React.FC<ApplicationProps> = ({
...
-              canEdit={current?.status === ApplicationStatus.Draft}
-              elements={getCurrentPageElements(fullStructure, sectionCode, pageNumber)}
-              responsesByCode={fullStructure.responsesByCode}
-              applicationData={fullStructure.info}
-              isStrictPage={
-                sectionCode === strictSectionPage?.sectionCode &&
-                pageNumber === strictSectionPage?.pageNumber
-              }
+      <Container id="application-form">
+        <Grid stackable className="dev-border">
```
nmadruga, line 64: Why is this css called `dev-border`?

src/containers/Application/ApplicationPage.tsx
```diff
@@ -1,12 +1,18 @@
 import React, { CSSProperties, useEffect } from 'react'
```
nmadruga, line 1: `CSSProperties` can be removed ;)

src/containers/Application/ApplicationWrapper.tsx
```diff
@@ -25,6 +25,7 @@ const ApplicationWrapper: React.FC = () => {
     currentUser: currentUser as User,
     networkFetch: true,
   })
+  console.log('template', template)
```
nmadruga, line 28: Please remove

src/formElementPlugins/ApplicationViewWrapper.tsx
```diff
@@ -179,7 +179,11 @@ const ApplicationViewWrapper: React.FC<ApplicationViewWrapperProps> = (props) =>
           margin: 5,
         }}
       >
-        {parametersReady && <Form.Field required={isRequired}>{PluginComponent}</Form.Field>}
+        {parametersReady ? (
+          <Form.Field required={isRequired}>{PluginComponent}</Form.Field>
+        ) : (
+          <Loader active inline />
```
nmadruga, line 185: Thanks! That'a already a bit better than before

semantic/src/site/globals/site.overrides
```diff
@@ -306,23 +325,89 @@ a:hover {
...
+  padding-right: 20px;
+}
+
+/*--------------------------------
+      General used classes
+---------------------------------*/
+
+// In use for ApplicationHeader around all Application pages
+
+.application-form-width {
```
nmadruga, line 409: Couldn't find where this is used

nmadruga (changes requested):
👍 Good job! I prefer using more specific ids for defining the layout on each page - as you did here than combining in more general classes as we were thinking before, it seems easier to read and to work with. I hope we can keep doing that for all pages as a pattern.

Make just minor comments, please have a look. I'll be testing this after "Changes requested" with the LOQ (with the bug fix) to see how this is looking. And that should be ready to merge : )

nmadruga:
Tested. Seems to work well after adding the bug fix - which add in a separated one line PR.
The looks of it seems ok, we will probably want to improve that after talking to Debbie, I'm not sure if we have other designs for this considering these borders...
1. On Changes requested:
Screen Shot 2021-05-04 at 10 25 20 AM
2. After changed by user:
Screen Shot 2021-05-04 at 10 25 31 AM
3. On other changes made:
Screen Shot 2021-05-04 at 10 20 49 AM

In case you want to test here just update line 55 to the following in `ApplicationHome`:
`        <ApplicationSections fullStructure={fullStructure} />      `

nmadruga:
Other comments, I personally don't like the current design for TextInfo, we probably need to agree on how we would like to display since there is no specific designs for this, but the current grey background seems ugly:
Screen Shot 2021-05-04 at 10 29 00 AM

Probably not for this PR though : P

nmadruga:
And the last, but not least... The nav buttons look bad on mobile. Not a biggie - not for now but might be good to left a note!
Screen Shot 2021-05-04 at 10 08 27 AM

nmadruga:
Oops, found another style while testing the bug fix - The validation message should be red:
Screen Shot 2021-05-04 at 10 34 45 AM

nmadruga:
Err, found 2 other things... Not related to this PR. Sorry, but if you can fix that here it would be great!
In the changes requested `ApplicationHome` page:
1. The Changes requested title looks too big
2. When the Update button shows the progress is sliding to the left (should be aligned with the one below)
Screen Shot 2021-05-04 at 10 37 11 AM


I'll take a snapshot and send to you to help replicate this same page (remember just to update the `ApplicationHome` line 55 please)

CarlosNZ:
> I prefer using more specific ids for defining the layout on each page

Yeah, I think a good general principle is to use `#id`s for targeting layout things (since they're specific to one-off elements) and try and use generic classes for formatting (as they're used a lot throughout).

semantic/src/site/modules/dropdown.variables
```diff
@@ -1,3 +1,5 @@
 /*******************************
     User Variable Overrides
 *******************************/
+
+@selectionBorder: 1px solid #777777;
```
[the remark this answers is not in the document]
CarlosNZ: Yeah, I tried using `@headersInteractiveLow` but it seemed too dark to me in this context. Should we make a new custom colour for "outlines" (since I think you'll need the same one for the Segment borders in Review Home page)?

src/components/PageElements/PageElements.tsx
```diff
@@ -45,7 +45,7 @@ const PageElements: React.FC<PageElementProps> = ({
   // Editable Application page
   if (canEdit && !isReview && !isSummary)
     return (
-      <Form>
+      <Form className="form-area">
         {visibleElements.map(
```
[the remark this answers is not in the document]
CarlosNZ: I'll leave this one till another issue, as I'm trying to keep this issue to just styles relating to initial application. Won't forget about it though ;)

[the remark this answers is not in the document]
CarlosNZ: Have started an issue for "moppping" up all remaining small styles: #662

src/containers/Application/ApplicationHome.tsx
CarlosNZ: The only reason I did it like that in ApplicationCreate was because the Button was conditional, whereas this one is always present on ApplicationHome (isn't it?), so should we just leave it as is? (I think that seems better than polluting ApplicationSections with an additional button, as it's already pretty full, wouldn't you say?)

src/containers/Application/ApplicationPage.tsx
CarlosNZ: Thanks, that was just for testing so I could see the div boundaries but I've removed the class now. FYI, I've put a more thorough "dev CSS" solution into #656 for doing this kind of thing going forwards.

src/formElementPlugins/ApplicationViewWrapper.tsx
CarlosNZ: Yeah, I'd like to implement a more thorough solution to incrementally show evaluated elements, but this'll do for now :)

semantic/src/site/globals/site.overrides
CarlosNZ: Yeah, good catch. I've replaced this with the variable `@applicationFormWidth`

CarlosNZ: Addressed this and most of the above inline comments: 830e3c13

CarlosNZ:
> Tested. Seems to work well after adding the bug fix - which add in a separated one line PR.
> The looks of it seems ok, we will probably want to improve that after talking to Debbie, I'm not sure if we have other designs for this considering these borders...

Yeah, this needs a little thinking about. Will address in #662 (unless we get to it beforehand)

CarlosNZ:
> Other comments, I personally don't like the current design for TextInfo, we probably need to agree on how we would like to display since there is no specific designs for this, but the current grey background seems ugly:

I completely agree. Seems like the most obvious thing is to remove TextInfo from being inside a `<Message>` component. Have made a small change to TextInfo formatting in #656, so I'll probably make an extension to that on that PR. Let me know if you have any other specific ideas for TextInfo formatting.

semantic/src/site/modules/dropdown.variables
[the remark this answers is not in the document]
nmadruga: Yep. That's a good idea.

src/containers/Application/ApplicationHome.tsx
nmadruga: > The only reason I did it like that in ApplicationCreate was because the Button was conditional, whereas this one is always present on ApplicationHome (isn't it?)

No - only when ALL responses are completed and valid. Otherwise it would show a `Resume` button like this:
Screen Shot 2021-05-04 at 12 17 28 PM

src/containers/Application/ApplicationPage.tsx
nmadruga: Nice! Will have a look after reviewing this one.

src/containers/Application/ApplicationPage.tsx
nmadruga: Please remove this.

semantic/src/site/modules/dropdown.variables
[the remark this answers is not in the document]
CarlosNZ: Cool, since you've also used this for the Segment borders in ReviewHome, I'll wait till you've merged that into here, then I'll replace them all with the variable value.

src/containers/Application/ApplicationHome.tsx
CarlosNZ: Ah, okay, I'll update this shortly.

src/containers/Application/ApplicationPage.tsx
nmadruga: Oops, sorry. It's removed, was seeing an older version

semantic/src/site/globals/site.overrides
nmadruga: Thanks

nmadruga (approved):

semantic/src/site/modules/dropdown.variables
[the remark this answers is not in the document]
nmadruga: Nice. Will do that now.

CarlosNZ:
> Oops, found another style while testing the bug fix - The validation message should be red:
> Screen Shot 2021-05-04 at 10 34 45 AM

This whole "pointer" element is not quite right (it should just be a label below the input field), but Andrei said this is fine for now. I'll add a tidy-up to #662

src/containers/Application/ApplicationHome.tsx
CarlosNZ: Cool, this makes more sense now, thanks for picking it up. deb0e3d8

CarlosNZ:
> And the last, but not least... The nav buttons look bad on mobile. Not a biggie - not for now but might be good to left a note!
> Screen Shot 2021-05-04 at 10 08 27 AM

Cool, added `flex-wrap` to this now, which works better (still needs to tweak spacing a bit on mobile, but will save that till later) 10b32f56

Issue #599, opened 2021-04-12
Update layout and use global styles: Review  home page


nmadruga:
Remember: Before changing styles take a screenshot on current page and add which style to use on each element - post to the issue so others can have a look and agree.

Just change elements with existing style. Anything not defined should be done in another (to be created) issue.

Issue #600, opened 2021-04-12
Update layout and use global styles: Application + Summary + Template question plugins


nmadruga:

CarlosNZ:
Any ideas how to format the textInfo element? The current `<Message>` component doesn't seem right:
<img width="543" alt="Screen Shot 2021-05-06 at 10 22 13 PM" src="https://user-images.githubusercontent.com/5456533/117283044-83e5f280-aeb9-11eb-8c43-1d6ff53d90d3.png">

And yet with nothing around the page looks bland and the info doesn't stand out:
<img width="512" alt="Screen Shot 2021-05-06 at 10 22 59 PM" src="https://user-images.githubusercontent.com/5456533/117283159-9fe99400-aeb9-11eb-9525-f24eac182011.png">

@andreievg maybe you could also ask Deb for some ideas for this?

CarlosNZ:
Thinking of just adding a "style" parameter to textInfo, which selects which type of "Message" box to show it in (See https://react.semantic-ui.com/collections/message/#variations-info)

"style" can take the following values:

- `none` (no Message box, just markdown formatting)
- `basic` (grey)
- `info` (blue)
- `warning` (yellow)
- `success`/`positive` (green)
- `negative`/`error`

CarlosNZ:
Implemented with #680

PR #650, opened 2021-04-30
#599 review home

nmadruga:
Still working on this...

There were more than I expected to be worked with, Row for assignment, Consolidation, reviews. 
I have looked into most of it I believe, that's why making this Draft PR, should be continuing to work a little more today to get all actions styled (and fixing a few issues with spacing of elements inside each row).

I think it might be good to merge as it is, and I would continue in another PR


semantic/src/site/collections/grid.overrides
```diff
@@ -1,3 +1,27 @@
...
+
+.stripes .ui.grid:nth-child(odd) .stripedlabel {
+  background: @surfacesLight;
+}
+
+.ui[class*='middle aligned'].grid > .column:not(.row) {
+  padding: 0px;
+  margin-left: 10px;
+  margin-right: 10px;
+}
```
CarlosNZ, line 27: Hey, I don't think it's a good idea to go over-riding the default styles for all Grids, since Grids are already in use in other places. Primarily, the main application form page from my last PR (#647) now has weirdly wide margins due to these changes, which makes the Progress bar look manky.

How it should be:

Screen Shot 2021-05-02 at 11 23 10 PM

How it is now:
Screen Shot 2021-05-02 at 11 24 13 PM


Probably the best thing to do is to use `#id`s on the specific elements you want to over-ride. Since grids are very much a "custom" thing that changes on a case-by-case basis, we should probably leave the defaults as is (unless you think there is a strong reason not to?)

src/components/Review/ReviewSectionRowAssigned.tsx
```diff
@@ -14,84 +13,61 @@ const ReviewSectionRowAssigned: React.FC<ReviewSectionComponentProps> = ({
...
-                  : strings.LABEL_REVIEWED_BY}
-                <Reviewer user={assignment.reviewer} isCurrent={isAssignedToCurrentUser} />
-              </>
-            }
-          />
-        )}
-      </div>
+    <Grid.Column className="padding-zero">
+      {isAssignedToCurrentUser && action === ReviewAction.canSelfAssign ? (
+        <Label className="simple-label" content={strings.LABEL_ASSIGNMENT_SELF} />
```
CarlosNZ, line 18: Just wondering why this needs to be a "Label" component as opposed to just straight text? Maybe the formatting (i.e. vertical alignment) would work better for straight text. Doesn't seem like anything is really being "labelled".

semantic/src/site/elements/label.overrides
```diff
@@ -22,7 +27,26 @@
...
+  background-color: transparent;
+  color: @blue;
+}
+
+.ui.active.label.link-label:hover {
+  background-color: transparent;
+  border-color: none;
+  opacity: @hoverOpacity;
+  color: @blue;
+}
```
CarlosNZ, line 52: You should be able to achieve this by just over-riding the variables, I don't think you need to override the class definitions. (since it's really just colours your re-mapping, right?)

CarlosNZ: For example, see "Form.variables" -- there's a bunch of variables where I've just over-ridden the colour definitons to point to our colour palette.

semantic/src/site/elements/icon.overrides
```diff
@@ -1,3 +1,12 @@
...
 *******************************/
+
+i.icon {
+  color: @darkGrey;
+}
+
+link.icon {
+  font-size: @fontXXsmall;
+  font-weight: @semiboldWeight;
+}
```
CarlosNZ, line 12: (Similar to my comment below): You should be able to just over-ride variables to achieve this, no need to re-write class definitions.

semantic/src/site/globals/site.overrides
```diff
@@ -374,27 +393,6 @@ a:hover {
...
-}
-
-.stripes .ui.grid:nth-child(even) .stripedlabel {
-  background: white;
-}
-
-.stripes .ui.grid:nth-child(odd) .stripedlabel {
-  background: rgb(248, 248, 248);
-}
-
```
CarlosNZ, line 397: Great, thanks for deleting this. I've been wanting to but wasn't sure it wasn't needed.

src/components/Application/ApplicationContainer.tsx
```diff
@@ -1,26 +1,29 @@
...
+const ApplicationContainer: React.FC<ApplicationContainerProps> = ({
+  template,
+  currentUser,
+  children,
+}) => {
+  const { replace } = useRouter()
   const { code, name } = template
   return (
-    <Container id="application-area">
+    <Container>
```
CarlosNZ, line 19: Why remove the id? I know there was no longer any CSS referring to it, but it's still helpful to have ids for thing, especially major structural components like this. (Makes it a lot easier seeing what's what in the Inspector and relating it back to the code).

src/components/Review/ReviewContainer.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+export interface ReviewContainerProps {
+  application: ApplicationDetails
+}
+
+const ReviewContainer: React.FC<ReviewContainerProps> = ({ application, children }) => {
+  const { replace } = useRouter()
+  const { template, name, org } = application
+  return (
+    <Container>
```
CarlosNZ, line 15: Probably want to chuck in an id for this, since it's a fairly major element.

semantic/src/site/elements/label.overrides
```diff
@@ -22,7 +27,26 @@
   background-color: transparent;
   color: @darkGrey;
   font-weight: @regularWeight;
-  font-size: @fontXXsmall;
   height: 20px;
   padding-left: 0;
 }
+
+.ui.label.text-italic {
```
CarlosNZ, line 34: I don't think this needs its own class. To put a label in italics, just wrap it the text in `<em></em>` tags.

semantic/src/site/elements/label.overrides
```diff
@@ -22,7 +27,26 @@
...
+  font-style: italic;
+}
+
+.ui.label.simple-label > .text-blue {
```
CarlosNZ, line 38: Wouldn't we just make a generic "info-text" class that can be used for ALL text elements that need to be coloured `@information` colour? No need to restrict it to one particular label type.

CarlosNZ (changes requested):
I'll have a closer look at this in the morn, I just highlighted a few things that I noticed from a quick look through.

Most important is the Grid over-rides. (see comments)

semantic/src/site/collections/grid.overrides
nmadruga: Sorry, I didn't realise it had changed the layout of the In progress page : (
I thought that the `border-radius: 6px` would be something global to all grids, but didn't realise the margin was going to be a bad decision. I'll add that to a specific `id`. All the other styles are under a **className** `.stripes`, so should be ok, right?

src/components/Review/ReviewSectionRowAssigned.tsx
nmadruga: Can you give me an example please? I was using like that to keep all the Global layout we have set for Semantic-ui components, and using text, not wrapped with something seemed difficult to set styles, since there is no tags related to it.

semantic/src/site/collections/grid.overrides
CarlosNZ: > All the other styles are under a **className** `.stripes`, so should be ok, right?

Yeah, looks like it.

semantic/src/site/elements/label.overrides
nmadruga: Sorry, end up not using this, using the `<Icon link ...>` styled seemed as a better approach. Will remove this unused definitions now.

src/components/Review/ReviewSectionRowAssigned.tsx
CarlosNZ: Yeah okay, I see what you're saying.  "simple-label" works as an extension of "Label". Cool.

Just wondering what's causing the vertical alignment problem though. Looking at the `.simple-label` definition, I think the problem is the "height" specification. It's artificially forcing a height onto something whose height really should be determined by the text itself. If I remove `height: 20px;`, it correctly centers vertically. There's still an issue with the rows being taller than what you want (I think), but it looks better and it's more of an issue of wrangling margins and padding to make it narrower, I'd say.

semantic/src/site/collections/grid.overrides
nmadruga: I tried something else, have changed the margin inside each row, and started using the Grid.Row with this purpose on `ReviewSectionRow` and `AssignmentSectionRow`, both are contained in a semantic component (Segment) that is using the `.stripes` class. Hope that is a better solution to this: 32c935a

src/components/Review/ReviewSectionRowAssigned.tsx
nmadruga: > Looking at the .simple-label definition, I think the problem is the "height" specification

Yes, I noticed that too, and have removed on commit 32c935a

src/components/Application/ApplicationContainer.tsx
nmadruga: NP. Putting this back.

src/components/Review/ReviewContainer.tsx
nmadruga: Yep, will add that too. Thanks

semantic/src/site/elements/label.overrides
nmadruga: Cool, didn't know about this one. Changing this now

nmadruga: Done: 985669f

semantic/src/site/elements/label.overrides
nmadruga: Sure. Good idea. I'll move to the `site.overrides` file

semantic/src/site/elements/icon.overrides
nmadruga: After talking in person. Done: 43e1624

CarlosNZ (changes requested):
Cool, looking good. Just a few finicky things:

- I think it's quite important to get the "white" rows to have the same height as the "grey" ones, as per the design. You'll have to play around with the row margins etc that you set in grid.overrides, so it looks like this (the exact heights are not so important, just that they match each other):

- Less important, but should be an easy fix: see if you can get the Segment border colour and border radius to match the design. I *think* it should be safe to do it globally, in which case you just need to add the following variable overrides to Segment.less:
   - `@borderRadius: 8px;`
   - `@border: 1px solid #777777;`  (not 100% sure of colour here)
  You might want to do a quick check of the other pages though to make sure those overrides don't mess something else up.

nmadruga:
> Cool, looking good. Just a few finicky things:
> 
> * I think it's quite important to get the "white" rows to have the same height as the "grey" ones, as per the design. You'll have to play around with the row margins etc that you set in grid.overrides, so it looks like this (the exact heights are not so important, just that they match each other):
>   
> * Less important, but should be an easy fix: see if you can get the Segment border colour and border radius to match the design. I _think_ it should be safe to do it globally, in which case you just need to add the following variable overrides to Segment.less:
>   
>   * `@borderRadius: 8px;`
>   * `@border: 1px solid #777777;`  (not 100% sure of colour here)
>     You might want to do a quick check of the other pages though to make sure those overrides don't mess something else up.

Thanks for this addition, I'm running out on time today. 

I have merged #600 into this now. I think this can be merged, then I create a tiny other one to make this lasts changes requested tomorrow. Sounds alright?

nmadruga:
> * `@borderRadius: 8px;`
> * `@border: 1px solid #777777;`  (not 100% sure of colour here)
>   You might want to do a quick check of the other pages though to make sure those overrides don't mess something else up.

Done! Seems safe didn't find other places that seemed to be used it.

CarlosNZ:
> I have merged #600 into this now. I think this can be merged, then I create a tiny other one to make this lasts changes requested tomorrow. Sounds alright?

Sure, can you either make a new issue to address this, or add it to #662 (mopping up)

nmadruga:
> * I think it's quite important to get the "white" rows to have the same height as the "grey" ones, as per the design. You'll have to play around with the row margins etc that you set in grid.overrides, so it looks like this (the exact heights are not so important, just that they match each other)

I'm working on it - almost good : )

CarlosNZ (approved):
Maybe hold off merging until you approve #600?

nmadruga:
OK - Is [PR#600](#647) ready for re-review?

CarlosNZ:
> OK - Is [PR#600](#647) ready for re-review?

Yup

nmadruga:
Nice! : ) I have finish my changes here. Will have another look on your PR now

Issue #571, opened 2021-04-08
Need "default" parameter for shortText (and maybe other) plugins


CarlosNZ:
We have a "placeholder" value, but that's not actually stored in the Response -- it's just a "guide" for users. It occurs to me that if we're going to have templates to allow editing (of user or company details), we'll need to have the text responses pre-filled with the existing values. So we'll need to explicitly define a "default" parameter for shortText and have it store it as the response on load (like the default selections for Dropdown and Radiochoice do currently)

Issue #332, opened 2021-02-18
Epic #32 - Emailing action


nmadruga:
- Send notifications
- Localisation

PR #652, opened 2021-05-02
#571 Default parameter for text input

CarlosNZ:
Applies a "default" parameter to shortText and longText plugins, which allows a certain value to be "pre-filled" when launching an application. 

This will be useful for an "Edit user details" form, where the fields can be pre-filled with existing values.

Have made a back-end to demonstrate this,  but it's just a rudimentary "Edit User" template, without full functionality.

Use branch from [back-end PR #332](https://github.com/openmsupply/application-manager-server/pull/332)

To see the "default" values in use, start a UserEdit template with: http://localhost:3000/application/new?type=UserEdit


src/formElementPlugins/longText/src/ApplicationView.tsx
```diff
@@ -14,12 +14,15 @@ const ApplicationView: React.FC<ApplicationViewProps> = ({
...
-  const { label, description, placeholder, lines, maxLength } = parameters
+  const { label, description, placeholder, lines, default: defaultValue, maxLength } = parameters
 
   const { isEditable } = element
 
   useEffect(() => {
-    onUpdate(value)
+    if (!value && defaultValue) {
+      onSave({ text: defaultValue })
+      setValue(defaultValue)
```
andreievg, line 24: I am getting confused now, how the ApplicationViewWrapper interacts with ApplciationView and useGetFullApplicationStructure, sound like we need another diagram : /

I am not sure what the difference here is between calling onUpdate and not calling it

andreievg (approved):
Approved, but need to compile a digram sometime (and in depth one), I'll try to do it towards the end of the week (if you haven't done it already)

CarlosNZ:
> Approved, but need to compile a digram sometime (and in depth one), I'll try to do it towards the end of the week (if you haven't done it already)

Okay, have made an issue, hopefully get to it soon (ish) :)

src/formElementPlugins/longText/src/ApplicationView.tsx
CarlosNZ: onUpdate just updates the local state "value", whereas onSave saves it to the database. However, onSave *includes* a call to onUpdate within it, so no need to do onUpdate if you're doing onSave :)

PR #666, opened 2021-05-05
#594(2) ProgressBar Area

CarlosNZ:
Implements remainder of #594 -- namely styling for ProgressBar area.

**Note**: have re-named `<ProgressBar>` component to `<ProgressArea>`, since it has separate ProgressBars within it. Will wait till this PR is approved, and then will rename the file too, but didn't want to make the diff larger for now.

ProgressArea is hidden on mobile.

Regarding the icons, I have used the designs to inform the logic, with some extrapolation. The component currently distinguishes between 5 states (although some share the same icon, this is just for extensibility), and are implemented with the following priority (highest takes precendence):

| State            	| SECTION ICON 	| PAGE ICON 	|
|--------------------	|--------------	|-----------	|
| `error`            	| Error         	| Error      	|
| `complete`         	| Complete         	| Green dot     	|
| `current`/`active` 	| Current         	| Blue dot      	|
| `incomplete`       	| Empty          	| (blank)       	|
| `notStarted`       	| Empty         	| (blank)      	|

Let me know if you recommend any different icons/precedence for these.


src/components/Application/ProgressBar.tsx
```diff
@@ -68,95 +79,205 @@ const ProgressBar: React.FC<ProgressBarProps> = ({
...
+      return (
+        <Progress
+          percent={percent}
+          size="tiny"
+          success={!error}
+          error={error}
+          style={{
+            width: length,
+            transformOrigin: '0 0',
+            transform: `rotate(90deg) translate(2px, -10px)`,
```
andreievg, line 129: can we do those in .less files please

src/components/Application/ProgressBar.tsx
```diff
@@ -68,95 +79,205 @@ const ProgressBar: React.FC<ProgressBarProps> = ({
...
+  const getIndicator = (
+    progress: ApplicationProgress,
+    isStrict: boolean,
+    isActive: boolean,
+    type = ProgressType.section
+  ) => {
+    const { completed, valid } = progress
+    const isStrictlylInvalid = !valid || (isStrict && !completed)
+    const isStarted = progress.doneNonRequired + progress.doneRequired > 0
+    const typeIndex = type === ProgressType.section ? 0 : 1
```
andreievg, line 98: nice, I like this approach

not a required change, but it's good to use descriptive object structure:
```
const progressIconMap = {
 error: {
    section: ...jsx
    page: ...jsx
  },
..
}
```

src/components/Application/ProgressBar.tsx
```diff
@@ -68,95 +79,205 @@ const ProgressBar: React.FC<ProgressBarProps> = ({
...
-      </Label>
-    ) : (
-      <Icon name="circle outline" />
+    return (
+      <div className="page-list-container">
+        <div className="section-progress-bar">
+          <SectionProgressBar
+            percent={calculateBarPercent()}
+            length={calculateBarLength()}
+            error={!sectionProgress.valid}
```
andreievg, line 156: this should also include strict and incomplete

src/components/Application/ProgressBar.tsx
```diff
@@ -68,95 +79,205 @@ const ProgressBar: React.FC<ProgressBarProps> = ({
...
+    const calculateBarPercent = () =>
+      sectionProgress.totalRequired === 0
+        ? 100
+        : (sectionProgress.doneRequired / sectionProgress.totalRequired) * 100
 
-    if (isStrictlylInvalid) return <Icon name={'exclamation circle'} color={'red'} size={size} />
-    if (completed && valid) return <Icon name={'check circle'} color={'green'} size={size} />
+    const calculateBarLength = () => {
+      // Adjust these constants to match existing style settings
+      const pixelsPerPage = 27.8
```
andreievg, line 142: I am ok with this a little hacky compromise to make it look awesome, it's well contained, only improvement would be to move those to constants at the top of file

src/components/Application/ProgressBar.tsx
```diff
@@ -68,95 +79,205 @@ const ProgressBar: React.FC<ProgressBarProps> = ({
...
       },
     }
   })
 
   return (
-    <Sticky as={Container} offset={135}>
-      <Header as="h4" style={inlineStyles.top}>
-        {strings.TITLE_INTRODUCTION}
-      </Header>
+    <Sticky as={Container} id="application-progress" offset={135} className="hide-on-mobile">
```
andreievg, line 227: it's not part of the changes you've made, but I would make this offset a constant at the top of this file (it can even be global constants, but at the top here would be enough for now)

andreievg (changes requested):
A couple of small picky things.

Code looks good, clean and solid and the feature is great

src/components/Application/ProgressBar.tsx
CarlosNZ: The value of `length` is dynamically calculated at runtime (based on Page count), so we can't put this in external CSS.

Have moved the transform properties over though 9840c8a8

src/components/Application/ProgressBar.tsx
CarlosNZ: Yup. Thanks. baf10d83

src/components/Application/ProgressBar.tsx
CarlosNZ: Like this? 4452d378

src/components/Application/ProgressBar.tsx
CarlosNZ: Cool, I've started a new `styleConstants` file and replaced the 135 in a couple of extra places too: 2f408884

andreievg (approved):
Thanks for changes !

PR #669, opened 2021-05-05
#668 Quick fix - Showing "Assigned to you" wrong label 

nmadruga:
Super quick bug fix! Fixes #668.

Also organised the file a little to be better to read.

Issue #275, opened 2021-01-19
Update ReviewPage with Consolidator showing reviewer and consolidation comments


nmadruga:
Show ReviewPage with previous reviewers comments and the **Review** button (or another label for Consolidators?) after the consolidation was created and issue #388 is done. 
If no consolidation should redirect to start page as described in #269.

Consolidator Schema: #281

First needs to outline what is the differences in UI (and query) between lvl 1 review and consolidation review page.

UI Design

Reviewer
![Screen Shot 2021-01-29 at 12 19 15 PM](https://user-images.githubusercontent.com/16461988/106210836-7969ec00-622c-11eb-8e90-9accd94382bb.png)
![Screen Shot 2021-01-29 at 12 19 30 PM](https://user-images.githubusercontent.com/16461988/106210842-7c64dc80-622c-11eb-8e09-2985e2cd7e2c.png)

Consolidator
![Screen Shot 2021-01-29 at 12 19 00 PM](https://user-images.githubusercontent.com/16461988/106210845-7ec73680-622c-11eb-83e3-0ef17f651e3b.png)
<img width="616" alt="Screen Shot 2021-05-06 at 8 37 47 PM" src="https://user-images.githubusercontent.com/16461988/117268238-2cd92100-aeab-11eb-984e-d0496a1c7815.png">

andreievg:
Thought to clarify the new shape of structure for this:

editable: status = 'Draft'.

if thisReviewLatestResponse.reviewResponseLink, then it's consolidation

latestOriginalReview response to be used in consolidation to render review response being consolidated (not editable for consolidation regardless of status)

i think `ReviewResponseElement` needs to take two reviewResponses as parameter:
for review it would be just thisReviewLatestResponse (for both, and render edit icon when thisReviewLatestResponse.decision exists)
for consolidation it would be latestOriginalReviewResponse and thisReviewLatestResponse (and if it's consolidation only render the 'Review' when thisReviewLatestResponse.decision does not exist)

I think we need another issue to render section progress (in home and in review page), logic for this would need to be decided from the new reviewProgress mentioned in https://github.com/openmsupply/application-manager-web-app/pull/631

PR #672, opened 2021-05-06
#634 show progress in consolidation

### LINKED ISSUE #634: Add consolidation progress (review home and review page)
Based on structure below:

changes from review progress to consolidation progress if thisReview.level > 1.

progress = (doneAgreeConform + doneAgreeNonConform + doneDisagree) / (totalReviewable - totalPendingReview)
color = red if at least on doneAgreeNonConform or one doneDisagree otherwise green
caption = 'Disagree' if at least on doneDisagree, 'Non Conform' if at least on doneAgreeNonConform else 'Completed' if progress is 100 

Level 2 Review Structure Changes (2)

nmadruga:
Fixes #634 

Although not testable yet, Consolidator approval/disagreements seems to not be updating current `consolidationProgress` will check if required #276 will find out, but I wanted to leave this PR already created.

### Based on PRs:
#632 
#669 - not based on previous one so changes will show here (Ignore files - `ReviewSectionRowAssigned.tsx` in this PR)

### Changes
- Types: 
  - Add new Type `SectionAssignment` and rename field in `SectionState.reviewAction` to `SectionState.assignment`
  - Add `SectionState.assignment.isConsolidation`
  - Combine `ReviewAndConsolidationProgress` to be an extended type of `reviewProgress` and `consolidationProgress`
- Component: 
  - New `ConsolidationProgressBar` to display consolidation progress per section
  - Add consolidationProgress on the `ReviewHome`
  - Add consolidationProgress on the `ReviewPage


src/utils/helpers/structure/generateReviewSectionActions.ts
```diff
@@ -83,9 +100,14 @@ const generateReviewSectionActions: GenerateSectionActions = ({
   currentUserId,
 }) => {
   const isCurrentUserReview = reviewAssignment.reviewer.id === currentUserId
+  const isConsolidation = reviewAssignment.level > 1
 
   sections.forEach((section) => {
-    const isReviewable = (section.reviewAndConsolidationProgress?.totalReviewable || 0) > 0
+    const { totalReviewable, totalPendingReview, totalActive } = isConsolidation
```
andreievg, line 106: I was wondering what was the logic behind changing to combined progress (i.e. extra added code/logic and + testing + extra time taken to review refactor). not a required changed but I am trying to figure out what was the reason behind the change, likewise for the reviewAction -> assignment change

src/components/Sections/SectionWrapper.tsx
```diff
@@ -44,7 +44,7 @@ const SectionWrapper: React.FC<SectionProps> = ({
                 <Header as="h2" content={details.title} style={sectionStyles.title} />
               </Grid.Column>
               <Grid.Column floated="right" textAlign="right">
-                {extraSectionTitleContent && extraSectionTitleContent(section)}
```
andreievg, line 47: hmm also was wondering about this change, what was wrong with React.ReactNode vs React.FC<SectionState>

src/components/Sections/SectionProgress.tsx
```diff
@@ -1,41 +1,83 @@
...
+ * Application
+ */
+
+const getApplicationProgressTitle = ({ completed, valid }: ApplicationProgress) => {
+  if (!valid) return strings.LABEL_SECTION_PROBLEM
+  else if (completed) return strings.LABEL_SECTION_COMPLETED
+  return null
+}
+
+const ApplicationProgressBar: React.FC<ApplicationProgress> = (applicationProgress) => {
```
andreievg, line 21: would be good to keep this in the same place when making changes (easier for reviewer to see diff)

src/components/Sections/SectionProgress.tsx
```diff
@@ -1,41 +1,83 @@
...
+const ConsolidationSectionProgressBar: React.FC<SectionProgressBarProps> = (props) => {
+  const {
+    doneAgreeNonConform,
+    doneAgreeConform,
+    doneDisagree,
+    totalReviewable,
+    totalPendingReview,
+  } = getConsolidationProgressDefaults(props)
+  return (
+    <Progress
```
andreievg, line 60: should there be a label (disagree, agree non-conform, agree conform) or something like that ?

andreievg (changes requested):
@nmadruga only request is label added to progress bar.

I spent most of the time reviewing refactor changes not related to the issue. It was a stylistic refactor imo, which is completely fine in new changes, but we should refrain from it in existing code (we can refactor existing code if it adds clarity or readability, but if it's a stylistic preference, we would end up going back and forwards refactoring based on individuals preference in every PR)

src/utils/helpers/structure/generateReviewSectionActions.ts
nmadruga: Sorry if this has been difficult to review. The reason was only for readability, it seemed easier to see things as separated entities - either Reviewer or Consolidator. I know this was just changed by you for break-down the logic, but it wasn't helpful to me, because then I had to get things from 2 different places to come up with a progress.

Changes on  `reviewAction -> assignment` also for consistency. I needed to create a type - because I was using a variable that would be of this same type (not defined yet), then there was already a `ReviewAction` type, which was one of the fields of this old field on `SectionState`, so it used to be:
```
SectionState {
...
reviewAction: {
   action: ReviewAction
   isAssignedToCurrentUser
   etc
}
```
Because I wanted to use a type for this instance I didn't know how to name it properly, so changed it to assignment, maybe it would be more correct to be `sectionAssignment`. I agree it adds to this change, you are right. But it would be really difficult if every change wasn't allowed during development of each new feature when it's considered refactoring. Even if it's a refactoring of a variable name.

src/components/Sections/SectionWrapper.tsx
nmadruga: I can revert that, I thought it was clearer this way. I did that because I felt more confident use React.FC, will check out what is the difference

nmadruga: Reverted now: 65790b0

src/components/Sections/SectionProgress.tsx
nmadruga: Oopsy 😯

src/components/Sections/SectionProgress.tsx
nmadruga: Yes, it would be nice. I guess I wasn't sure what to display exactly. Any suggestions? I can try something here quickly now : )

nmadruga: Done change here 4639a35, but haven't tested yet. Will do that on my current unPRd of #275 & #276 - Should be on a PR today : )

nmadruga:
Ok, last change before this can be re-reviewed please @andreievg. I have the labels running - tested using the current PR I'm working on, so won't be possible to test using this one. Just trust me : o )

I took some prints tough, so you can check and make suggestion for changes if desired...
When **Agreed** with everything, show details number of conform and non-conforms:
Screen Shot 2021-05-12 at 3 01 22 PM

When **Disagrees** show in red, and the number of disagreements vs. agreements:
- current
Screen Shot 2021-05-12 at 2 56 46 PM
- alt: Just show the number of disagreements
Screen Shot 2021-05-12 at 2 55 08 PM

andreievg:
@nmadruga I wasn't requesting any changes in my comments about refactor (and not requesting any changes now), I wanted to understand why 50 lines of code turned to 350. At this stage of the project, I am not against refactor (would be different after MVP), I wanted to hear your logic for the changes in this case though, (see if it can be avoided, or if this pattern would cause backwards and forwards refactor due to different devs having different opinions about naming/structure, or see if i am missing something that's makes the extra lines of codes necessary for the introduced feature).

I guess my conclusion is this:

Every developer has a preference for style, structure and consistency. We just have to be aware of that when making changes.

And I am ok with reviewing refactor changes, but it would be nice to minimise diff for reviewer. (reviewing can catch potential regressions, and it becomes harder when diff is large) 

I am just pointing out below to communicate the reason for me even going into the topic (and spending the time to help us improve)

> because then I had to get things from 2 different places to come up with a progress.

ok, so it was for ?

https://github.com/openmsupply/application-manager-web-app/blob/b48b249319308a1488473f5fa8f5cda5d8b4c932/src/components/Sections/SectionProgress.tsx#L101-L105

`vs`

https://github.com/openmsupply/application-manager-web-app/blob/27049a6223bd652b980fdae5059c1ef2886d91d7/src/components/Sections/SectionProgress.tsx#L32-L39

> Changes on reviewAction -> assignment also for consistency. I needed to create a type - because I was using a variable that would be of this same type (not defined yet), then there was already a ReviewAction type, which was one of the fields of this old field on SectionState, so it used to be:

not sure about `consistency`, but i look like you needed to create a type to destructure  `{ isConsolidation, isReviewable, isAssignedToCurrentUser }` from optional `section.reviewAction`.  So that it can be used `!isAssignedToCurrentUser` vs `!reviewAction?.isAssignedToCurrentUser`

So in the end we have:

```
const SectionRowStatus: React.FC<SectionState> = (section) => {
  const { assignment } = section
  const { isConsolidation, isReviewable, isAssignedToCurrentUser } = assignment as SectionAssignment

  if (!isAssignedToCurrentUser)
    return <Label className="simple-label" content={strings.LABEL_ASSIGNED_TO_OTHER} />
  if (!isReviewable)
    return (
      <Label
        icon={<Icon name="circle" size="mini" color="blue" />}
        content={strings.LABEL_ASSIGNED_TO_YOU}
      />
    )
  if (isConsolidation && section.consolidationProgress)
    return <ConsolidationSectionProgressBar consolidationProgress={section.consolidationProgress} />
  if (section.reviewProgress)
    return <ReviewSectionProgressBar reviewProgress={section.reviewProgress} />
  return null // Unexpected
}

```
vs
```
const SectionRowStatus: React.FC<SectionState> = (section) => {
  const { reviewAction } = section

  if (!reviewAction?.isAssignedToCurrentUser)
    return <Label className="simple-label" content={strings.LABEL_ASSIGNED_TO_OTHER} />
  if (!!reviewAction?.isReviewable)
    return (
      <Label
        icon={<Icon name="circle" size="mini" color="blue" />}
        content={strings.LABEL_ASSIGNED_TO_YOU}
      />
    )
  if (!reviewAction?.isConsolidation)
    return <ConsolidationSectionProgressBar consolidationProgress={section.consolidationProgress} reviewAndConsolidationProgress={section.reviewAndConsolidationProgress} />
  if (section.reviewProgress)
    return <ReviewSectionProgressBar reviewProgress={section.reviewProgress}
reviewAndConsolidationProgress={section.reviewAndConsolidationProgress} />
  return null // Unexpected
}

```

but + 300 or so lines of code changes (touching a lot of areas, and probably quite a long time spent refactoring), personally i dont' think it's worth it.

Anyways, I am sorry, I didn't want to pick on you or the changes, just want to see if we can improve to be more efficient.

And, I am also sorry in advance for pointing this out in the future.

andreievg (approved):
Looks good, thanks for the changes !

PR #664, opened 2021-05-04
#563 & #581 - Review submission improvements

nmadruga:
Fixes #563 
Fixes #581 (WIP - reason why this is a Draft PR)

### Changes
- Add styles to Review page (only for Submission area) and remove old inline styles on `ReviewSubmit` file
- Update style of Submission to be similar to what is in designs - still not quite, but getting there
Current:
Screen Shot 2021-05-05 at 9 39 46 AM

Design:
Screen Shot 2021-05-04 at 10 43 12 PM

  - `checkbox` checked colour to be using `@primaryColour` by default (the border is still on the wrong colour - not sure why)  
  - New `strong-label` type for labels in black and `@semiboldWeight`

Issue #598, opened 2021-04-12
Update layout and use global styles: Review  page + Submission modal


nmadruga:
Remember: Before changing styles take a screenshot on current page and add which style to use on each element - post to the issue so others can have a look and agree.

Just change elements with existing style. Anything not defined should be done in another (to be created) issue.

CarlosNZ:
FYI -- I will do the main "Review" containers etc, along with the SectionWrapper styles as part of #595 (since they share a lot of components), so this issue should focus on the Submission component

PR #673, opened 2021-05-06
#598 Review page styles

CarlosNZ:
Still some improvements to be done, mainly to do with later reviews, change requests, etc. but this is a solid base I think.

It uses many of the same basic container and section layouts as ApplicationSummary.

Please note, this PR has both #671 (Summary page) AND #664 (Review submission improvements), so probably best to wait till at least one of those is merged so it's clear what's actually been done here (then I'll rebase to the other one).

Screen Shot 2021-05-06 at 5 14 35 PM


nmadruga:
Merged PR #664 - Review submission improvement - to make this PR a little simpler to review

semantic/src/site/globals/site.overrides
```diff
@@ -319,6 +329,51 @@ a:hover {
...
+/*---------------------------------------
+      Review page
+----------------------------------------*/
+
+// NOTE: The review page shares a lot of styles with Application Summary
+
+.review-comment-area {
+  background-color: @reviewCommentBackground;
+  margin-top: 10px;
+  margin-left: -20px;
```
nmadruga, line 341: I don't understand why the negative margins here I think I was seeing something weird because of this when trying to update the comment to also show Consolidator comments in between the Reviewer comment and Application response

semantic/src/site/globals/site.overrides
```diff
@@ -319,6 +329,51 @@ a:hover {
...
+  margin-right: -20px;
+  margin-bottom: -14px;
+  border-top: 1px solid @surfacesMed;
+  border-bottom: 1px solid @surfacesMed;
+}
+
+.review-comment-grid.ui.grid {
+  padding: 10px 20px;
+  font-size: 0.85em;
```
nmadruga, line 350: Should probably use one of the predefined one in `site.variables`

semantic/src/site/globals/site.overrides
```diff
@@ -319,6 +329,51 @@ a:hover {
...
+  color: @headersInteractiveLow;
+}
+
+.review-approve-all-button {
+  margin-top: 20px;
+}
+
+#review-submit-area {
+  // display: flex;
```
nmadruga, line 359: Probably can be removed?

semantic/src/site/globals/site.overrides
```diff
@@ -548,6 +606,11 @@ a:hover {
   margin-top: 20px;
 }
 
+// Hide sticky bar when it jumps to bottom
+.summary-section .bound.bottom {
+  display: none;
+}
```
nmadruga, line 612: 🎉 🎉 🎉 🎉 🎉 🎉 🎉

semantic/src/site/globals/site.overrides
```diff
@@ -488,6 +543,8 @@ a:hover {
       Application Summary area
 ---------------------------------*/
 
+// Note: these layout ids/classes also shared with Review page
```
nmadruga, line 546: I think that's the case only for `summary-section`, `.summary-section-header`, `.summary-page`, `.ui.basic.segment.summary-page-element` and `.summary-section .bound.bottom`. So maybe just move these to the "General section" or even create one specific area for Section/Page elements styles?

src/components/PageElements/Elements/ReviewDecisionElement.tsx
```diff
@@ -73,34 +85,4 @@ const ReviewDecisionElement: React.FC<ReviewDecisionElementProps> = ({
 
 // Styles - TODO: Move to LESS || Global class style (semantic)
```
nmadruga, line 86: Please remove this comment too

semantic/src/site/globals/site.overrides
nmadruga: Ok, just found out that  `application-summary` and  `application-summary-header` are in use on ReviewHeader as well, so probably we can rename to a more general name i.e. `summary-header-container` and `summary-header`

nmadruga: And `application-summary-content` on **ReviewPage**, same move to the "General" section, although the name for this one seems concise, right? It is a application-summary-content : )

nmadruga (commented):
Just a few comments... Will run and see if anything remaining. Really excited to see everything working together when I get to PR #684!

src/components/Review/ReviewComment.tsx
```diff
@@ -40,13 +40,19 @@ const ReviewComment: React.FC<ReviewCommentProps> = ({ reviewDecisionId, isEdita
...
-        defaultValue={initialComment}
-        onChange={(_, { value }) => setComment(String(value))}
-        onBlur={() => updateComment({ variables: { reviewDecisionId, comment } })}
-      />
-    </Form>
+    <div>
+      <p>
+        <strong>{strings.LABEL_REVIEW_OVERALL_COMMENT}:</strong>
+      </p>
+      <Form.Field id="review-commment-content">
```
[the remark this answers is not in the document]
CarlosNZ: It's not, as it didn't need any additional styles. However, I think it's useful to id key items regarldess, it makes it much easier to understand the mark-up when developing/debugging

nmadruga:
Hoovering over the **Review** label doesn't show the mouse as hand, as it should when hovering an clickable text, it shows the mouse as you can start typing on it, which is weird. Have a try to see what I mean.

semantic/src/site/globals/site.overrides
CarlosNZ: The Element wrapper has some padding, which means that when `review-comment-area` sits inside it, there's white space outside it. The negative margin overcomes this while still allowing the parent component to have sensible padding, otherwise it looks like this:
Screen Shot 2021-05-10 at 12 25 46 PM

nmadruga:
Nice one : )
Screen Shot 2021-05-10 at 12 27 38 PM

semantic/src/site/globals/site.overrides
CarlosNZ: You mean font-size? I kind of wanted it to be a relative size here (and we probably should have done a bit more of it). The important thing is that this text is "a bit smaller" than the paragraph text, absolute size is not particularly important. So if we change the paragraph text size, this will automatically change proportionally, which is much easier to manage. We could make some classes that do basic proportioning in `ems` at some point, but might be best to go through once we're ready for a tidy-up. I've got a bunch of little things that can be consolidated, but it's best to wait till the end I think.

semantic/src/site/globals/site.overrides
CarlosNZ: Yes, thanks b3f209ba

semantic/src/site/globals/site.overrides
CarlosNZ: Thank @andreievg  for that one!

semantic/src/site/globals/site.overrides
CarlosNZ: I agree, but again I think it'll be easier to do this all at once at the end.

src/components/PageElements/Elements/ReviewDecisionElement.tsx
CarlosNZ: 25625ede

nmadruga:
There is a problem that I noticed here - but not related to this one so created a new issue to solve that: #685

And have been wondering for another issue as well - Should we add a Stick title to the Review page, it seems like the Application name, stage and start date are very relevant and could be kept in the "always showing area", on the top of the current Sticky section title:
Screen Shot 2021-05-10 at 12 28 38 PM

CarlosNZ:
> Hmmm weird - I see you did already one of the improvements desired and listed on issue #588, for the sticky not to display on the bottom. But I can still see a tiny problem with style of sticky. There is a gap, and you can see text flowing behind it:


Yes, I've fixed this somewhere else. Can't actually remember which PR, but it's definitely fixed later on.

CarlosNZ:
> Hoovering over the **Review** label doesn't show the mouse as hand, as it should when hovering an clickable text, it shows the mouse as you can start typing on it, which is weird. Have a try to see what I mean.

Fixed in #684  (or by then, not sure exactly which one). (It's just a matter of adding the "clickable" class to it.

semantic/src/site/globals/site.overrides
nmadruga: ok

CarlosNZ:
> Should we add a Stick title to the Review page, it seems like the Application name, stage and start date are very relevant and could be kept in the "always showing area", on the top of the current Sticky section title:

Yeah, maybe, let's discuss it at our sprint meeting. In a similar vein I was wondering if we should add the page number with the section heading on application pages rather than having to put them in the template as TextInfo elements.

src/components/Review/ReviewComment.tsx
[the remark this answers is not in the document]
nmadruga: Ah! Yes, good point

semantic/src/site/globals/site.overrides
nmadruga: Ahhh, alright : )

nmadruga (approved):
Great work. All good yey!

PR #680, opened 2021-05-06
#600 Plugin styles

CarlosNZ:
Not much to do here, since most of the plugins look fine with the default and already-implemented styles.

The only significant addition here is to extend the "TextInfo" plugin to have a range of different styles (based on Semantic's Message component styles). See docs for details.

Requires [back-end PR #338](https://github.com/openmsupply/application-manager-server/pull/338) to see changes to TextInfo. In particular, Section 3 of FeatureShowcase has a TextInfo element with a style selector:
Screen Shot 2021-05-07 at 11 03 00 AM


src/formElementPlugins/SummaryViewWrapper.tsx
```diff
@@ -46,7 +46,7 @@ const SummaryViewWrapper: React.FC<SummaryViewWrapperProps> = ({
   const DefaultSummaryView: React.FC = () => {
     const combinedParams = { ...parameters, ...evaluatedParameters }
     return (
-      <Form.Field required={isRequired}>
+      <Form.Field className="element-summary-view" required={isRequired}>
```
nmadruga, line 49: Maybe this should be an id?

src/formElementPlugins/textInfo/src/ApplicationView.tsx
```diff
@@ -2,13 +2,53 @@ import React from 'react'
...
+interface TextInfoProps {
+  parameters: any
+  Markdown: any
+}
+
+// Extract main content to another Component so it can be shared with SummaryView (and TemplateView)
+export const TextInfoElement: React.FC<TextInfoProps> = (props) => {
+  const { parameters, Markdown } = props
+  const { title, text, style } = parameters
+  if (['basic', 'info', 'warning', 'success', 'positive', 'negative', 'error'].includes(style)) {
```
nmadruga, line 14: Maybe nice to define a type for this:
`type StyleType = 'basic' | 'info' | 'warning' | 'success' | 'positive' | 'negative' | 'error'`

And then change to 

```suggestion
  const { title, text } = parameters
  const style: StyleType = parameters.style
  if (style) {
```

nmadruga (approved):
Looks great. Just a minor suggestion

Issue #662, opened 2021-05-03
Update layout and use global styles: All remaining elements


CarlosNZ:
Add things here for elements that still need styling, so we can "mop up" in one issue:

- [ ] Review decision area styles (part 1 in PR #682)
- [x] `isChangeRequest` in PageElements (in PR --  #684)
- [ ] Error message pointer should be label with icon as per design
- [x] "Help" sidebar implementation & style (in PR -- #683)
- [ ] Application Summary page:
   - [ ] Change requests (should match layout and styles of Review page)
   - [ ] Summary Review Response Element (should be very similar (if not a variant of) ReviewDecisionElement
- [ ] Check application/reviews in all states (i.e. changes requested, etc.)
- [ ] Consolidate/refactor common styles

nmadruga:
@CarlosNZ also can this one be closed? I think that's the only remaining issue for [Epic#10](https://github.com/openmsupply/application-manager-web-app/issues/78)

CarlosNZ:
I dunno, maybe move it to another epic if you want to close #10. I feel like there's some things we need to have a more thorough look at (the unchecked items) but they're not an immediate priority.

nmadruga:
Yes, good idea to create separated issues for those and link to Epic #25 - Others. So we can close this issue. I have already moved Epic #10 to Column "Epics Done" so we can have a better feel of what is currently in progress really

PR #682, opened 2021-05-07
#662(1) Review decision area -- part 1

CarlosNZ:
Since design discussion for this are ongoing, I've just done a basic layout and made the existing panel in line with designs (mostly)

I've made a container for a "document-viewer" alongside the details panel, but that's currently hidden

Also -- I was getting tired of having to reload the panel every time the page refreshed, so I made the panel open state part of the url query. So now the query is the primary source of truth for which Decision panel is open. E.g.
`http://localhost:3000/application/23456/review/5?active-sections=S1&open-response=Q1`

### DIFF
--- a/semantic/src/site/modules/checkbox.overrides
+++ b/semantic/src/site/modules/checkbox.overrides
@@ -5,5 +5,5 @@
 .ui.checkbox {
   // font-size: @fontXsmall;
   color: @headersInteractiveLow;
-  padding-left: 10px;
+  // padding-left: 10px;
 }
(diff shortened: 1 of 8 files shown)


semantic/src/site/modules/checkbox.overrides
```diff
@@ -5,5 +5,5 @@
 .ui.checkbox {
   // font-size: @fontXsmall;
   color: @headersInteractiveLow;
-  padding-left: 10px;
+  // padding-left: 10px;
```
nmadruga, line 8: Please remove commented lines

nmadruga (approved):
LGTM. Looking very good so far - will check Part 2.

Issue #681, opened 2021-05-06
Help text in application form sidebar


CarlosNZ:
Just putting this here as a discussion location for how we should handle this. 

Deb's design has this "help" text in the right-hand sidebar:
<img width="942" alt="Screen Shot 2021-05-07 at 11 24 05 AM" src="https://user-images.githubusercontent.com/5456533/117377279-048f0800-af27-11eb-926a-9f4b29563ffb.png">

At first I thought we could just add another field to template elements, but that won't be straightfoward since the right-hand column is outside the ApplicationViewWrapper component, so won't have access to plugin values.

Here are some suggestions of how we could handle it:

1. Make the elements column (middle) and the right-hand column just one wider column, with the elements limited by width. Then we could display help tips inline with the element. I think it'll be fine for overflow to hang down into the area of the next element, with a little CSS magic.
2. Keep help text separate from elements completely, and just store them as part of the template.  Each would contain configuration of its text, the section/page to associate with, and a vertical offset.
3. Do away with it completely and find some other use for this extra space. (We can include any relevant "helper" info with the element description field)

What are your thoughts?

Personally I'm leaning towards option 3, but maybe option 1 (don't really care for option 2 the more I think about it). We could see what Deb comes up with as an alternative perhaps? One good thing about option 1 is that we could have an option for a TextInfo plugin to "span" the whole width and not just be limited to the main elements "column".

(Note: I'm also assuming we would hide this content on mobile)

andreievg:
1 - I like this but I think overlap would be an issue ?
2 - Could be per page I guess, in designs it seems to be directed at the form element though
3a - Could do a kind of a help accordion (with help icon as the title), this would be more portable to mobile
3b -Like one and 3a, show help tip, restricted by the height of form element (add ...), when form element in focus, show full help tip (some css trickery still)

Can ask Debs about 3a and 3b or opened ended 3.
I think it's worth asking, since in her reply, she didn't give any guidance of what happens if two consecutive  form elements have the help tip (and it looks like they will overlap since the help tip in the example spans beyond the form element's height) 
Andrei
```
Info/descriptions in application form
We were wondering about the info on the right hand side of form elements, do they appear for currently ‘entered’ elements or all of them ?
```
Debs
```
My thinking on this "Info description" text is to only show on certain inputs where you need to show extra information (paragraph) to help user with extra info. So show all the time, not just on entered elements. Not to be used on every element though.
```

CarlosNZ:
[...]

Issue #645, opened 2021-04-29
Element plugins -- add "help" field


CarlosNZ:
This is the text that will display in the right-hand sidebar in application forms. It is intended to be used sparingly, and is distinct from "description".

To think about: should it be a field on its own in the "template_element" table, or just part of "parameters" JSON? My initial thought is that it should be separate, since it'll be rendered _outside_ the main ApplicationView element, so should probably be handled earlier.

![2021-04-30 11 07 37](https://user-images.githubusercontent.com/5456533/116628955-50c8ce00-a9a4-11eb-8793-890f665eba35.jpg)

CarlosNZ:
Implement as per discussion [here](https://github.com/openmsupply/application-manager-web-app/issues/681)

PR #683, opened 2021-05-09
#645 Implement help text tips

CarlosNZ:
Implements #645, using "Option 1" from [this discussion](https://github.com/openmsupply/application-manager-web-app/issues/681#issue-878183541)

Requires branch from [back-end PR #389](https://github.com/openmsupply/application-manager-server/pull/339)

Screen Shot 2021-05-09 at 10 13 29 PM

Note that the help text can quite happily hang below the level of the next question, which means that template builders will have to mindful not to overlap them, or else this happens:

Screen Shot 2021-05-09 at 10 16 25 PM

Have added more "demonstration" help text to the Feature Showcase template, but I've also added some more "realistic" help tips to some of the other templates (see back-end changes)

### DIFF
--- a/src/components/PageElements/PageElements.tsx
+++ b/src/components/PageElements/PageElements.tsx
@@ -1,5 +1,5 @@
 import React from 'react'
-import { Form, Segment } from 'semantic-ui-react'
+import { Form, Grid, Segment } from 'semantic-ui-react'
 import {
   ApplicationDetails,
   ElementState,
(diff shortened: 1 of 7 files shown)


src/components/PageElements/PageElements.tsx
```diff
@@ -67,7 +68,20 @@ const PageElements: React.FC<PageElementProps> = ({
               currentResponse: responsesByCode?.[element.code],
             }
             // Wrapper displays response & changes requested warning for LOQ re-submission
-            return <ApplicationViewWrapper key={`question_${element.code}`} {...props} />
+            return (
+              <div className="form-element-wrapper" key={`question_${element.code}`}>
```
andreievg, line 72: hmm, wondering if this should be in ApplicationViewWrapper or here ?

I can see that it would complicate `ApplicationViewWrapper` return statement

I originally thought that the changes required border would be outside of ApplicationViewWrapper too. I guess we can see if it needs a refactor in the future.

andreievg (approved):

src/components/PageElements/PageElements.tsx
CarlosNZ: Yeah, originally it was outside `ApplicationViewWrapper`, but the spacing between elements pushed the "Changes Required" comments too far away from the field they relate too. Worked better having them inside it, and I kind of think it works if everything related to one element sits within one `<Form.Field>`.

But yeah, can maybe refactor later if we want.

PR #693, opened 2021-05-12
#342B extra restriction field (sessionId) in application

CarlosNZ:
Front-end complement of [back-end PR #350](https://github.com/openmsupply/application-manager-server/pull/350) 

### Summary of changes

- fix `fetchUserInfo` so it doesn't re-write a new JWT on fetching userInfo (so original login sessionId persists)
- ApplicationCreate mutation inserts sessionId into application table when it creates application


andreievg (changes requested):
I think on front end this PR should just deal with just setting sessionID when application is created (and storing it in userID).
templatePermissions may update, and we want to make sure we update JWT token in that case.

Back end should add logic to make sure sessionID does not change if JWT is validated (but recalculate other things, templatePermission etc..)

CarlosNZ:
> I think on front end this PR should just deal with just setting sessionID when application is created (and storing it in userID).
> templatePermissions may update, and we want to make sure we update JWT token in that case.
> 
> Back end should add logic to make sure sessionID does not change if JWT is validated (but recalculate other things, templatePermission etc..)

Okay, have reverted `fetchUserInfo` to save the new JWT (and back-end changes to return the same one if provided) 085d63a2

I assume the reason why you wanted to do it like this is that a users template permissions could have changed, so it makes sense to get a fresh token each visit with the latest permissions in it. However, one thing to consider is that with a fresh token, it gets a new "iat" time, so if we wanted to force logout after a certain period of time, we'd need to add an additional "originalLoginTime" field to the JWT in order to check expiry.

andreievg (approved):

andreievg:
> I assume the reason why you wanted to do it like this is that a users template permissions could have changed, so it makes sense to get a fresh token each visit with the latest permissions in it. However, one thing to consider is that with a fresh token, it gets a new "iat" time, so if we wanted to force logout after a certain period of time, we'd need to add an additional "originalLoginTime" field to the JWT in order to check expiry.

Sure, we can do this if it's deemed necessary, also since we are acquiring the token on every refresh, we have a little bit more control

PR #695, opened 2021-05-12
#275 & #276 UI consolidation page

### LINKED ISSUE #275: Update ReviewPage with Consolidator showing reviewer and consolidation comments

### LINKED ISSUE #276: Agree/disagree Consolidation modal

nmadruga:
Fixes #275 & fixes #276

### Branched of PR
#672 

### Changes
- Removed many `<Something>Element.tsx` components and combine into one of the 3 components:
  - SummaryInformationElement
  - ApplicantResponseElement
  - ReviewDecisionElement
#### Applicant
- The `ApplicantResponseElement` is used in both cases of first application or Updated - displaying the Updated label and only editable elements in `ChangesRequested` if they have been updated. Maybe this was a little out of scope.

#### Review/Consolidation
- Change `DecisionArea` to display and update reviewResponse of an Consolidator
- The `ReviewDecisionElement` is used to display either a Reviewer or Consolidator review, the one not active (not very proud of `isReviewActive` prop) will display in white background.
- Set Edit/Review button dynamically outside of the component, so it can be defined in `PageElements`
- Ideally would have combined the loop over elements in the Page/Summary/Review in one, but haven't done that yet. But it might be included in another iteration of changes after I get some feedback from you.


src/components/Review/DecisionArea.tsx
```diff
@@ -1,137 +1,169 @@
...
+      : reviewUpdate?.decision === ReviewResponseDecision.Decline
+
+  // TODO: Improve - how to use shorthand was on my mind...
+  const getOptions = () => {
+    return isConsolidation ? (
+      <>
+        <strong>{strings.LABEL_CONSOLIDATE}</strong>
+        <Form.Field>
+          <Radio
+            label={strings.LABEL_CONSOLIDATION_AGREEMENT}
```
andreievg, line 47: could have an object describing decision and label:
```
const optionsMap = {
  consolidation: [{
    label: strings.LABEL_CONSOLIDATION_AGREEMENT,
    decision: ReviewResponseDecision.Agree
  },...],
  review:...
}
```
then 
```
const options = isConsolidation ? optionsMap.consolidation : optionsMap.review
...
   {options.map(({decision. label}) => (
   <Form.Field key={decision}>
          <Radio
            label={label}
            value={decision
            name="decisionGroup"
            checked={reviewUpdate.decision === decision}
            onChange={() =>
              setReviewUpdate({
                ...reviewUpdate,
                decision: decision,
              })
            }
          />
   )}
```

src/components/Review/DecisionArea.tsx
```diff
@@ -1,137 +1,169 @@
...
 
   const submit = async () => {
-    // TODO do we need to handle update error ?
-    await updateResponse(review)
     updateQuery({ openResponse: null })
+    // TODO do we need to handle update error ?
+    await updateResponse(reviewUpdate)
+  }
+
+  const validateDecisionComment = () =>
```
andreievg, line 35: `isValidComment` for name maybe (can be a const, doesn't need to be a method)

src/components/Review/DecisionArea.tsx
```diff
@@ -1,137 +1,169 @@
...
-      setReview(reviewResponse)
-      setPreviousToggle(toggle)
-      setIsOpen(true)
-    }
-  }, [toggle])
+  const {
+    query: { openResponse },
+    updateQuery,
+  } = useRouter()
+  const [reviewUpdate, setReviewUpdate] = useState(reviewResponse)
```
andreievg, line 26: i think it would be more clear to name this one `reviewResponse` and the prop can be `reviewResponse: initialReviewResponse`

andreievg (changes requested):
Just had a look at the modal so far, and had a quick look overall, i think the structure is sound. Will add another review for remainder

semantic/src/site/globals/site.overrides
```diff
@@ -345,6 +344,10 @@ a:hover {
   border-bottom: 1px solid @surfacesMed;
 }
 
+.changable-background {
```
CarlosNZ, line 347: Spelling: `changeable`

CarlosNZ (changes requested):
Haven't had a thorough look, but just a couple of UI regressions I noticed:

Screen Shot 2021-05-12 at 9 55 53 PM

1. Something has made this panel gain an extra scroll bar on the right. It shouldn't scroll at all -- if the History panel gets too big, then that part can have its own scroll bar, not the whole container
2. The "Submit" button has become disabled, even when there is a comment in the box, so can't mark any responses "Request resubmission"

src/components/PageElements/Elements/ApplicantResponseElement.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+type ApplicantResponseElementProps = ResponseElementProps & SummaryViewWrapperProps
+const ApplicantResponseElement: React.FC<ApplicantResponseElementProps> = ({
+  applicationResponse,
+  isResponseUpdated = false,
+  children,
+  ...summaryViewProps
+}) => {
+  const backgroudColour = isResponseUpdated ? 'changable-background' : ''
+  // TODO: Fix tiny margin showing on left and right on grid.item
```
CarlosNZ, line 17: If possible, you should change the background of the parent element, the `summary-page-element`, and then you won't have to deal with this margin discrepancy. If you add the `changeable-background` (dynamically) class to the same div as `summary-page-element`, the blue colour should over-ride the white (assuming it's later in the CSS file.)

andreievg (changes requested):
Had a more thorough review, and left some comments, overall this is looking really good.

I have a couple of overall comments

**Destructuring props**

I think I prefer to keep types specific to elements, i remember originally adding de-structured props for summary view, but it was just those summaryProps (and nothing else).
It's hard to keep track of where props are coming from, imagine coming back to this piece of code in 6th month and trying to trace back where props for summary view are coming from in application response for consolidation.
Anyways I think we should pass through props for other components as key value. Happy with destructuring props if it's a full set of props for that component (and nothing else).

**Using JSX for top down of complex view**

I think JSX is great for top down templating, quite often we want to keep as much logic in one place as possible while at the same time keeping code digestable and dry. I noticed that consolidation is almost the same as review block, and that you've created nicely reusable elements. imo, ApplicantResponseElement and ReviewDecision element was all that's needed to create readable, dry and unified JSX. Only thing that was needed is positioning review action (i think logic for it is good to keep in one place).

Anyways if i was to do it, it would be something like this: What do you think ?

```JSX
return (
      <div>
        <Form>
          {visibleElements.map(
            ({
              element,
              thisReviewLatestResponse,
              isNewApplicationResponse,
              isActiveReviewResponse,
              latestApplicationResponse,
              latestOriginalReviewResponse
              
            }) => {
              const summaryViewProps = getSummaryViewProps(element)

              if (element.category === TemplateElementCategory.Information) {
                return (
                  <RenderElementWrapper key={element.id}>
                    <SummaryInformationElement {...summaryViewProps} />
                  </RenderElementWrapper>
                )
              }
              // Renders an Icon or Button
              const ReviewResponseAction = (
                <ReviewElementTrigger
                  isIcon={!!thisReviewLatestResponse?.decision}
                  title={
                    isNewApplicationResponse /* can add check for isNewReviewResponseAlso */
                      ? strings.BUTTON_RE_REVIEW_RESPONSE
                      : strings.BUTTON_REVIEW_RESPONSE
                  }
                />
              )
              const decisionExists = !!thisReviewLatestResponse?.decision

              const renderReviewResponseAction = (location: ReviewResponseActionLocation) => {
                if (!isActiveReviewResponse) return null
        
                if(location === 'consolidationResponse') {
                  if(!isConsolidation) return null
                  // render action on consolidation response if decision exists
                  if(decisionExists) return ReviewResponseAction
                }
                if(location === 'applicationResponse') {
                  if(!isReview) return null;
                  // render action on application response response if decision does not exists
                  if (!decisionExists) return ReviewResponseAction
                }

                // location must be reviewResponse
                // render action on review response if consolidation decision exists
                if(isConsolidation && decisionExists) return ReviewResponseAction
                // render action on review response if review decision exists
                if(isReview && decisionExists) return ReviewResponseAction
                // no other option
                return null
              }

              const toggleDecision = openResponse === element.code
              const applicationResponseProps = {applicationResponse: latestApplicationResponse, ...summaryViewProps}
            
              return (
                <RenderElementWrapper key={element.id}>
                  {/* Application Response */}
                  <ApplicantResponseElement {...applicationResponseProps}>
                    {renderReviewResponseAction('applicationResponse')}
                  </ApplicantResponseElement>
                  {/* Consolidation Response */}
                  {isConsolidation && decisionExists && (
                    <ReviewDecisionElement reviewResponse={thisReviewLatestResponse} isCurrentReview={true}>
                      {renderReviewResponseAction('consolidationResponse')}
                    </ReviewDecisionElement>
                  )}
                  {/* Review Response */}
                  <ReviewDecisionElement reviewResponse={latestOriginalReviewResponse} isCurrentReview={isConsolidation}>
                    {renderReviewResponseAction('reviewResponse')}
                  </ReviewDecisionElement>
                  {toggleDecision && <DecisionArea reviewResponse={thisReviewLatestResponse}/>}
                </RenderElementWrapper>
              )
            }
          )}
        </Form>
      </div>
    )
```

src/components/Review/DecisionArea.tsx
nmadruga: Yes, that's a very good solution : ) I'll add that in, thanks

nmadruga: Done: 7519f83

src/components/Review/DecisionArea.tsx
nmadruga: Done: b06f9dc

nmadruga:
> 2\. The "Submit" button has become disabled, even when there is a comment in the box, so can't mark any responses "Request resubmission"

Done: b06f9dc

nmadruga:
> 1. Something has made this panel gain an extra scroll bar on the right. It shouldn't scroll at all -- if the History panel gets too big, then that part can have its own scroll bar, not the whole container

I was seeing this scrollbar before. Not sure what have caused it.... Maybe if we just reduce the size of the container, so the "History panel" area gets a little less height it might solve that?

src/components/Review/DecisionArea.tsx
nmadruga: Yes, good idea

nmadruga: Done: 04741bb

semantic/src/site/globals/site.overrides
nmadruga: Done: 02b6d85

src/components/PageElements/PageElements.tsx
```diff
@@ -100,90 +118,244 @@ const PageElements: React.FC<PageElementProps> = ({
...
+  updateQuery: Function
+}
+
+const ReviewButton: React.FC<ReviewButtonProps> = ({
+  isActive,
+  isNewReview,
+  elementCode,
+  updateQuery,
+}) =>
+  isActive ? (
```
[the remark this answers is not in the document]
nmadruga: When displaying a reviewResponse (aka `originalLatestReviewResponse`)

src/components/PageElements/PageElements.tsx
```diff
@@ -100,90 +118,244 @@ const PageElements: React.FC<PageElementProps> = ({
...
+
+              const consolidationProps: ReviewElementProps = {
+                applicationResponse: latestApplicationResponse,
+                reviewResponse: thisReviewLatestResponse as ReviewResponse,
+                originalReviewResponse: latestOriginalReviewResponse as ReviewResponse,
+                isConsolidation,
+                ExtraReviewAction: () =>
+                  canConsolidate ? <ReviewButton {...reviewButtonProps} /> : null,
+                ExtraEditReviewAction: () =>
+                  thisReviewLatestResponse?.status === ReviewResponseStatus.Draft ? (
```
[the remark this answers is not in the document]
nmadruga: Will change to use your suggestion for `PageElements` combining consolidation and reviewer - so rewriting this.

src/components/PageElements/PageElements.tsx
```diff
@@ -100,90 +118,244 @@ const PageElements: React.FC<PageElementProps> = ({
...
+  isNewReview: boolean
+  elementCode: string
+  updateQuery: Function
+}
+
+const ReviewButton: React.FC<ReviewButtonProps> = ({
+  isActive,
+  isNewReview,
+  elementCode,
+  updateQuery,
```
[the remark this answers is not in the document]
nmadruga: Not sure. Have to think about - this PR involved too many decisions (and finding out why bugs were happening) so I tried to avoid adding more to it. But once all other changes are made we can see if this can be addressed as part of it.

src/components/PageElements/PageElements.tsx
```diff
@@ -100,90 +118,244 @@ const PageElements: React.FC<PageElementProps> = ({
...
+          {visibleElements.map(
+            ({
+              element,
+              thisReviewLatestResponse,
+              latestOriginalReviewResponse,
+              latestApplicationResponse,
+              isNewApplicationResponse,
+              isActiveReviewResponse,
+            }) => {
+              const canConsolidate =
```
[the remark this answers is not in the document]
nmadruga: Will change to use your suggestion for `PageElements` combining consolidation and reviewer - so rewriting this.

src/components/PageElements/PageElements.tsx
```diff
@@ -100,90 +118,244 @@ const PageElements: React.FC<PageElementProps> = ({
...
+                (isActiveReviewResponse as boolean) && !thisReviewLatestResponse?.decision
```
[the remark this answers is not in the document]
nmadruga: Will change to use your suggestion for `PageElements` combining consolidation and reviewer - so rewriting this.

src/components/PageElements/Elements/ApplicantResponseElement.tsx
nmadruga: Hi @CarlosNZ I tried to do that but it didn't work as simple as this. I'll probably need some help with that, if we can have a look together in about 15 minutes?

nmadruga:
> **Destructuring props**
> 
> I think I prefer to keep types specific to elements, i remember originally adding de-structured props for summary view, but it was just those summaryProps (and nothing else).
> It's hard to keep track of where props are coming from, imagine coming back to this piece of code in 6th month and trying to trace back where props for summary view are coming from in application response for consolidation.
> Anyways I think we should pass through props for other components as key value. Happy with destructuring props if it's a full set of props for that component (and nothing else).

No problem. That can be changed back to be similar to what it was before.

src/components/PageElements/Elements/ApplicantResponseElement.tsx
CarlosNZ: I had a quick play last night, and I thought it was working ok -- let's take a quick look when you're ready

CarlosNZ:
> > 1. Something has made this panel gain an extra scroll bar on the right. It shouldn't scroll at all -- if the History panel gets too big, then that part can have its own scroll bar, not the whole container
> 
> I was seeing this scrollbar before. Not sure what have caused it.... Maybe if we just reduce the size of the container, so the "History panel" area gets a little less height it might solve that?

This is a strange one. Looking at the inspector, I can see that the "dimmer" element has an additional `overflow: auto` property that it doesn't on master, which is what is causing the problem. Yet I can't see any changes you've made that might have caused this. 😕

I'll take a look and suggest a fix.

CarlosNZ:
> > > 1. Something has made this panel gain an extra scroll bar on the right. It shouldn't scroll at all -- if the History panel gets too big, then that part can have its own scroll bar, not the whole container
> > 
> > 
> > I was seeing this scrollbar before. Not sure what have caused it.... Maybe if we just reduce the size of the container, so the "History panel" area gets a little less height it might solve that?
> 
> This is a strange one. Looking at the inspector, I can see that the "dimmer" element has an additional `overflow: auto` property that it doesn't on master, which is what is causing the problem. Yet I can't see any changes you've made that might have caused this. 😕
> 
> I'll take a look and suggest a fix.

Okay, found it. In master the content inside the Modal is only rendered if "isOpen":
```
 {!isOpen ? null : (
        <div id="review-decision-container">
          <div id="review-decision-content">
            <div id="document-viewer" className="hidden-element">
             ...
```

This branch doesn't have that condition. I don't really understand why it matters (after all, the whole Modal is hidden if it's not 'open'), but it seems to cause a change in the class that the fullscreen Modal uses, so it's probably simplest to put this qualifier back in (or an equivalent using your changes).

nmadruga:
Hey @andreievg I have added now the attempt to combine blocks of Review and Consolidation in `PageElements` and removing the difficult union types being passed to components. Hopefully this is simpler. Although I ended up putting the code block of consolidation (`ConsolidateReviewDecision`) and review (`ReviewApplicantResponse`), and wasn't able to use the `renderReviewResponseAction` here.

nmadruga:
> o it's probably simplest to put this qualifier back in (or an equivalent using your changes).

Nice finding. Thanks, I removed because it wasn't clear why in use. But I'll revert it back now...

nmadruga:
Phew! There is still one missing bit that was noticed now - have checked and it was there. So, to keep this one focused, I have created a new issue just to deal with showing what is missing on the Applicant summary view after changes requested are made

How is on the master:
Screen Shot 2021-05-13 at 4 28 43 PM

 And the designs:
Screen Shot 2021-05-10 at 11 52 26 PM

src/components/PageElements/Elements/ConsolidateReviewDecision.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+  return (
+    <div>
+      {/* Application Response */}
+      <ApplicantResponseElement
+        applicationResponse={applicationResponse}
+        summaryViewProps={summaryViewProps}
+      />
+      {/* Consolidation Response */}
+      {!isNewConsolidation && reviewResponse && (
```
andreievg, line 43: (not required change): although it's pretty easy to see now, but would be great to have this one as variable:
`canShowConsolidationResponse`

src/components/PageElements/Elements/ConsolidateReviewDecision.tsx
```diff
@@ -0,0 +1,84 @@
...
+        <ReviewResponseElement
+          isCurrentReview={true}
+          isConsolidation={isConsolidation}
+          applicationResponse={applicationResponse}
+          reviewResponse={reviewResponse}
+          originalReviewResponse={originalReviewResponse}
+        >
+          {isActiveReviewResponse && decisionExists && <UpdateIcon onClick={showModal} />}
```
andreievg, line 51: it seems this will always have an edit button ? (unless it's not a draft review)

src/components/PageElements/Elements/ReviewApplicantResponse.tsx
```diff
@@ -0,0 +1,76 @@
...
+  summaryViewProps,
+  reviewResponse,
+  isNewApplicationResponse,
+  isActiveReviewResponse,
+  showModal,
+}) => {
+  const decisionExists = !!reviewResponse?.decision
+  // After review is submitted, reviewResponses are trimmed if they are not changed duplicates
+  // or if they are null, we only want to show reviewResponses that are linked to latestApplicationResponse
+  const isNewReview = applicationResponse.id !== reviewResponse?.applicationResponse?.id
```
andreievg, line 28: isActiveReviewResponse ?

src/components/PageElements/Elements/ReviewApplicantResponse.tsx
```diff
@@ -0,0 +1,76 @@
...
+              isNewApplicationResponse /* can add check for isNewReviewResponseAlso */
+                ? strings.BUTTON_RE_REVIEW_RESPONSE
+                : strings.BUTTON_REVIEW_RESPONSE
+            }
+            onClick={showModal}
+          />
+        )}
+      </ApplicantResponseElement>
+      {/* Review Response */}
+      {!isNewReview && reviewResponse && (
```
andreievg, line 49: i think the logic is now different from what it was for review (not consolidation)
should show review response if decision does exists right ?)
would be good to have a variable for this: `const canShowReviewResponse = decisionExists`

src/components/PageElements/Elements/ReviewApplicantResponse.tsx
```diff
@@ -0,0 +1,76 @@
...
+        <ReviewResponseElement
+          isCurrentReview={true}
+          isConsolidation={false}
+          applicationResponse={applicationResponse}
+          reviewResponse={reviewResponse}
+        >
+          {isActiveReviewResponse && decisionExists && <UpdateIcon onClick={showModal} />}
```
andreievg, line 56: this would always show unless review not draft ?

src/components/PageElements/PageElements.tsx
```diff
@@ -100,90 +109,125 @@ const PageElements: React.FC<PageElementProps> = ({
...
-        )}
-      </Form>
+              // Information - no review
+              if (element.category === TemplateElementCategory.Information) {
+                return (
+                  <RenderElementWrapper key={element.code}>
+                    <SummaryInformationElement {...summaryViewProps} />
+                  </RenderElementWrapper>
+                )
+              } else {
```
andreievg, line 169: return would exit the method, would be nice to reduce indentation in next block by remove the else

src/components/Review/DecisionArea.tsx
```diff
@@ -9,129 +9,137 @@ import { SummaryViewWrapperProps } from '../../formElementPlugins/types'
...
   const submit = async () => {
-    // TODO do we need to handle update error ?
-    await updateResponse(review)
     updateQuery({ openResponse: null })
+    // TODO do we need to handle update error ?
+    await updateResponse(reviewResponse)
   }
 
+  const isValidComment =
+    (!!reviewResponse.comment && reviewResponse.comment?.trim() !== '') ||
```
andreievg, line 59: (not required change): just realised we doing a lot of `!`, when generating this one and when this one is used. Maybe `isInvalidComment` is easier ? (if no comment && Disagree or Decline)

src/components/PageElements/PageElements.tsx
```diff
@@ -100,90 +109,125 @@ const PageElements: React.FC<PageElementProps> = ({
...
-            linkToPage: `/application/${serial}/${sectionCode}/Page${pageNumber}`,
-            summaryProps: getSummaryViewProps(element),
-            latestApplicationResponse: latestApplicationResponse,
-            previousApplicationResponse: previousApplicationResponse,
-          }
+          const { element, isChanged, isChangeRequest, latestApplicationResponse } = state
 
-          const changedQuestionResponse = !isChangeRequest && !isChanged
+          const isResponseUpdated = !!isChangeRequest || !!isChanged
+          const canApplicantEdit = isUpdating ? isResponseUpdated && canEdit : canEdit
```
andreievg, line 115: is updating is set to isChangeRequest and passed through via props twice, can we just use isChangeRequest ?

andreievg: Also this logic here seems easier as `canEdit && (!!isChangeRequest || !!isChanged)`

andreievg (changes requested):
Nice, looking very good.

Just a couple of comments, can ignore (fix in new issue) styles ones. But some minor logic comments that would be good to fix before approval:

src/components/PageElements/Elements/ConsolidateReviewDecision.tsx
nmadruga: I think this is touching the "Changes requested" area so I'm a little confused since I can't really test it right now.
I had removed this check because of the previous [comment](https://github.com/openmsupply/application-manager-web-app/pull/695#discussion_r632237408) and I think this is already covered. The `ReviewResponseElement` would return `null` in case this review haven't got a decision done, which should be the case when a **New Review** is done (after changes requested by lvl 2 reviewer) and the consolidator does the re-review. Right?

src/components/PageElements/Elements/ConsolidateReviewDecision.tsx
nmadruga: This would show if (all is true):
- Review is in progress
- Reviewable by consolidator (isActiveReviewResponse)
- Consolidator has reviewed (decisionExists)

Do you see any case which is should/shouldn't display?

src/components/PageElements/Elements/ReviewApplicantResponse.tsx
nmadruga: I have changed in this and `ConsolidationReviewDecision` so the Review element shows if there is some `reviewResponse` received (which is the `thisLatestReviewResponse` only created when the previous check `applicationResponse.id !== reviewResponse?.applicationResponse?.id` was true) and the **Review Button** is the one that display only if the `isActiveReviewResponse` is present, so only the Review button is removed on submitted reviews.

src/components/PageElements/Elements/ReviewApplicantResponse.tsx
nmadruga: I can rename it to be clearer - will do that on Consolidator as well.

nmadruga: I hope that with latest changes the logic is more consistent.

src/components/PageElements/Elements/ReviewApplicantResponse.tsx
nmadruga: Similar to other one:

src/components/Review/DecisionArea.tsx
nmadruga: 97e507a

src/components/PageElements/PageElements.tsx
nmadruga: Sorry, didn't understand the first comment. Also, when not a change (`!isUpdating`) we still want to check `canEdit`, which is the case of `canEdit=true`, `isChangeRequest=false` and `isChanged=false` would return false...

src/components/PageElements/PageElements.tsx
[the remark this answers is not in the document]
nmadruga: I'm copying this comment to the new issue to move the decision to inline on the Review page : )

src/components/PageElements/Elements/ApplicantResponseElement.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+const ApplicantResponseElement: React.FC<ApplicantResponseElementProps> = ({
+  applicationResponse,
+  summaryViewProps,
+  isResponseUpdated = false,
+  children,
+}) => {
+  return (
+    <Grid columns="equal" className={'element-grid'}>
+      <Grid.Column width={isResponseUpdated ? 4 : 14} textAlign="left">
```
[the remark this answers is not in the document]
nmadruga: Removed

nmadruga:
Alright finished 3rd round of review changes. Let me know if anything else is needed, and hopefully this one can be merged today.

src/components/PageElements/Elements/ConsolidateReviewDecision.tsx
andreievg: thisReviewLatestResponse => will exists even if the decision is not made. I think this check needs to be isActiveReviewResponse && decisionExists ?

src/components/PageElements/Elements/ConsolidateReviewDecision.tsx
andreievg: > Consolidator has made decision (decisionExists)
I was thinking, the parent element should only show if decision exists

> Review is in progress
> Reviewable by consolidator (isActiveReviewResponse)
True, need to check isActiveReviewResponse

no need for change

src/components/PageElements/PageElements.tsx
andreievg: >Also, when not a change (!isUpdating) we still want to check canEdit, which is the case of canEdit=true, isChangeRequest=false and isChanged=false would return false...

Hmm, sorry for confusion, I started this comment thinking `isUpdating` is not needed for this check, and forgot that we show edit icon in summary view differently based on draft review of application as change request differently to when we show the edit button on first time draft. your change looks good., but a comment would be helpful here

andreievg (approved):
Thank for all of the changes ! 

Good to merge, although i still think that we shouldn't be showing reviewResponse and consolidationResponse when decisionDoesNotExist, and then the UpdateIcon within reviewResponse and consolidation response would only show if review response isActiveReviewResponse (draft)

src/components/PageElements/Elements/ConsolidateReviewDecision.tsx
nmadruga: Note: `ReviewResponseElement` would return null in case this review haven't got a decision done

src/components/PageElements/PageElements.tsx
nmadruga: Added a comment

Issue #539, opened 2021-03-26
Epic #81 - Templates by category


nmadruga:
## Description

Use category to display application from different templates in the ~~same list (maybe a new issue if we decide it's important)~~ on the Dashboard

Probably 1 day of this is design time

## Documentation
- [ ] TODO: Add as part of [Epic #38](338) Public docs & tutorials

PR #697, opened 2021-05-13
698 - Dashboard

andreievg:
Fixes: #698 

Needs back end branch `dashboard`

### Changes

* Added generic logic for some new application list filter (for number and boolean type)
* Added relative date selection for date application filter and moved date filter to generics)
* Removed dev selection of templates
* useGetListApplication now return filters and categories
* Relocated layouts home page to another menu ('Layout helpers')
* Introduced 'Dashboard', at root navigation
* Added styles

### Testing

Play around with dashboards : ), the following filters are configured if you want to test them all:

Applicant
 * Approved Applications
 * Submitted Applications
 * Draft Applications
 * Changes Required
Reviewer
 * Available for Self Assignment
 * Ready to Review
 * Ready to Re-Review
 * Draft Reviews
Assigner
 * Awaiting Assignment
 * Re Assign

### Questions

How did i do with styles ?

Is it ok to have all of the dashboard logic in one file ? (i didn't see the need to move components around, only 160 lines of code, and none of the components seem reusable


semantic/src/site/globals/site.overrides
```diff
@@ -737,6 +737,136 @@ a:hover {
...
+}
+
+#dashboard .template-category .title {
+  display: flex;
+  flex-direction: row;
+  align-items: center;
+}
+
+
+#dashboard .templates {
```
CarlosNZ, line 767: There's a lot of repetition here with existing flex layouts. It's fine for now, but at some point it'd be good to go through and consolidate flex layouts into a small set of re-useable classes.

src/components/Dashboard.tsx
```diff
@@ -0,0 +1,160 @@
+import React, { useEffect } from 'react'
```
CarlosNZ, line 1: `useEffect` not in use

src/components/Dashboard.tsx
```diff
@@ -0,0 +1,160 @@
...
+  usePageTitle(strings.PAGE_TITLE_HOME)
+
+  return (
+    <div id="dashboard">
+      {templatesByCategory.map(
+        ({ templates, templateCategory: { categoryIcon, categoryTitle } }) => (
+          <div key={categoryTitle} className="template-category">
+            <div className="title">
+              {categoryIcon && (
+                <Icon size="large" color="grey" name={categoryIcon as SemanticICONS} />
```
CarlosNZ, line 36: Instead of `color="grey"` you can use `className="dark-grey"` to get our custom colour grey.

Either that or map semantic "grey" to `@headersInteractiveLow` in site.variables (line 18)

src/components/Dashboard.tsx
```diff
@@ -0,0 +1,160 @@
...
+      .join('&')}`
+
+  if (loading) return null
+  if (applicationCount === 0) return null
+
+  return (
+    <div className="filter">
+      <Link to={constructLink()}>
+        {filter.icon && (
+          <Icon color={filter.iconColor as SemanticCOLORS} name={filter.icon as SemanticICONS} />
```
CarlosNZ, line 105: Wonder why we need the `as` here, haven't had to use it before. 🤷‍♂️

src/components/LayoutHelpers.tsx
```diff
@@ -5,7 +5,7 @@ import { useUserState } from '../contexts/UserState'
 import strings from '../utils/constants'
 import usePageTitle from '../utils/hooks/usePageTitle'
 
-const Home: React.FC = () => {
+const LayoutHelpers: React.FC = () => {
```
CarlosNZ, line 8: Can you update the page name for this one (line 18): `usePageTitle("Layout Helpers")`

src/containers/Dev/DevOptions.tsx
```diff
@@ -1,24 +1,13 @@
...
-
-  const { filteredTemplates } = useListTemplates(templatePermissions, isLoading)
   return (
     <div id="dev-options" style={menuStyle}>
       <Dropdown item icon="user">
         <UserSelection />
       </Dropdown>
-      <Dropdown item icon="map">
-        <AppMenu templatePermissions={filteredTemplates} />
-      </Dropdown>
```
CarlosNZ, line 21: Why did you remove this?

src/utils/constants.ts
```diff
@@ -128,4 +130,6 @@ export default {
   PAGE_TITLE_CREATE: 'New Application | Application Manager',
   PAGE_TITLE_APPLICATION: 'Application %1 | Application Manager',
   PAGE_TITLE_REVIEW: 'Review %1 | Application Manager',
+  DEFAULT_TEMPLATE_CATEGORY_TITLE: 'General',
+  DEFAULT_TEMPLATE_CATEGORY_ICON: 'globe',
```
CarlosNZ, line 134: I think we should keep this file for just "translatable" strings, maybe make a new one for other constants? (Cos we don't want some hapless translator accidentally translating the word "globe" for example)

src/utils/helpers/list/buildQueryFilters.ts
```diff
@@ -20,7 +23,65 @@ export default function buildQueryFilters(filters: BasicStringObject) {
...
+      'reviewAssignedNotStartedCount',
+      'reviewAssignedCount',
+      'reviewAvailableForSelfAssignmentCount',
+      'reviewDraftCount',
+      'reviewPendingCount',
+      'reviewChangeRequestCount',
+      'reviewSubmittedCount',
+      'assignReviewerAssignedCount',
+      'assignReviewersCount',
+      'assignCount',
```
CarlosNZ, line 46: These will need to change when we make the change to return enums from application_list right?

CarlosNZ (commented):
Awesome, looking really good and working well. Just a few comments really, not so much change requests.

src/containers/Dev/DevOptions.tsx
andreievg: i thought we would just click to dashboard now, should i reinstate it?

CarlosNZ: Oh okay, so the Dashboard should show all the templates that are available in that list? If so, then that's cool, leave it out now.

semantic/src/site/globals/site.overrides
andreievg: hmm, sure, i still not sure what the best practice is but defo can reduce repetion. just had a quick look, in less we can do this

how does it look after this commit 063886fe061b5e804983f9debde2ff7effa945e8 (hard to see the change in commit, best to look at overall diff
(added nesting, added .class(); within a class

src/components/Dashboard.tsx
andreievg: 275040f18ccef37f150fb988502d35fb95aeff78

src/components/Dashboard.tsx
andreievg: it looks too dark with our custom one, so idk, ?

src/components/Dashboard.tsx
andreievg: .iconColor and .icon are both strings, need to cast to semanticICON (make it that type and cast in the hook too, not sure where is the best place)

src/components/LayoutHelpers.tsx
andreievg: b6f2e2b5ff34f616824c9adfc6cf9113eebc30e1

src/utils/constants.ts
andreievg: I was thinking this can be changed by the admin user, but you are right, this potentially could be translated by accident. not sure what to do

src/containers/Dev/DevOptions.tsx
andreievg: yep, it would match the list

src/utils/helpers/list/buildQueryFilters.ts
andreievg: yeah, and it sounds like we dont' need a number type, although it could be useful somewhere ? (number of reviewers ?) or number of change request iterations ?

src/components/Dashboard.tsx
CarlosNZ: Screen Shot 2021-05-14 at 1 51 49 PM

I reckon it looks pretty good. Though we probably want to let Deb give some overall design suggestions for this page, so it's fine for now.

src/components/Dashboard.tsx
```diff
@@ -0,0 +1,160 @@
...
+  usePageTitle(strings.PAGE_TITLE_HOME)
+
+  return (
+    <div id="dashboard">
+      {templatesByCategory.map(
+        ({ templates, templateCategory: { categoryIcon, categoryTitle } }) => (
+          <div key={categoryTitle} className="template-category">
+            <div className="title">
+              {categoryIcon && (
+                <Icon size="large" color="grey" name={categoryIcon as SemanticICONS} />
```
CarlosNZ, line 36: Also here -- rather than use `as`, it might be better to define the type where categoryIcon is first declared (i.e. useListTemplates line 88):
`let categoryIcon: SemanticICONS`

semantic/src/site/globals/site.overrides
```diff
@@ -737,6 +762,119 @@ a:hover {
...
+  }
+
+  .filters {
+    margin: 5px;
+    padding: 3px
+  }
+
+  .view-all {
+    padding: 5px;
+    font-weight: 600;
```
nmadruga, line 863:
```suggestion
    font-weight: @semiboldWeight;
```

semantic/src/site/globals/site.variables
```diff
@@ -164,3 +164,6 @@
 
 // Box shadow -- used in Login form
 @basicBoxShadow: rgb(204, 201, 201) 0px -3px 10px 0px;
+
+// Box shadow -- used in Dashboard 
+@templateBoxShadow: 0 1px 3px 0 #d4d4d5, 0 0 0 1px #d4d4d5;
```
nmadruga, line 169: Good idea to defined this colour "#d4d4d5" in some variable

src/components/LayoutHelpers.tsx
```diff
@@ -15,7 +15,7 @@ const Home: React.FC = () => {
     return null
   }
 
-  usePageTitle(strings.PAGE_TITLE_HOME)
+  usePageTitle('Layout Helpers')
```
nmadruga, line 18: Localisation?

src/utils/graphql/fragments/template.fragment.ts
```diff
@@ -8,5 +8,21 @@ export default gql`
     isLinear
     startMessage
     submissionMessage
+    templateCategory {
+      title
+      icon
+    }
```
nmadruga, line 14: Oh nice! We had one [Epic#81](https://github.com/openmsupply/application-manager-web-app/issues/539) just for categories work - I guess we can mark it as done (move to awaiting public docs actually) after this PR is merged - possibly nice to have some internal docs created regarding that too

src/utils/types.ts
```diff
@@ -430,6 +433,23 @@ interface StageDetails {
   description?: string
 }
 
+interface TemplateCategoryDetails {
+  categoryTitle: string
+  categoryIcon: string
+}
```
nmadruga: Maybe doesn't need to be `categoryTitle` and `categoryIcon` just title and icon?

src/components/Dashboard.tsx
andreievg: 60b783cb8c841ed6744add9dc7c03868b26e540b

Remembered to relocate icon color to style and still need to cast filter.icon, since it's a string type in DB

andreievg:
A couple of things left outstanding, I am not sure how to deal with them: `DEFAULT_TEMPLATE_CATEGORY_ICON` -> 

Color of category icon (see what Debs says)
https://github.com/openmsupply/application-manager-web-app/pull/697#discussion_r632157547

src/utils/constants.ts
CarlosNZ: Let's change this file to "language strings" and make a new constants.ts?

src/utils/hooks/useListTemplates.tsx
```diff
@@ -17,30 +37,96 @@ const useListTemplates = (templatePermissions: TemplatePermissions, isLoading: b
...
+      if (filteredTemplates.length === 0) return setLoading(false)
+
+      const templates = filteredTemplates.map((template) =>
+        convertFromTemplateToTemplateDetails(template, templatePermissions)
+      )
+
+      setTemplatesData({
+        templates,
+        templatesByCategory: getTemplatesByCategory(templates),
       })
```
nmadruga, line 54: Kind of a big change - but it seems a little more straight-forward to me. You could do in the same loop a separation of templates per category in a object like this:
```
[
   {
      category: TemplateCategoryDetails
      templateList: TemplateInList[]
   },
   ...
]
```

And perhaps the category.title and category.icon doesn't need to go inside the `TemplateInList` also, just removing it from the type. I'll continue looking with that in mind see if there is any flaws.

src/components/Dashboard.tsx
```diff
@@ -0,0 +1,158 @@
...
+  return (
+    <div className="template">
+      <div className="title">
+        <Header as="h5">{name}</Header>
+
+        {hasApplyPermission && (
+          <Button inverted size="small" as={Link} to={`/application/new?type=${code}`} primary>
+            {constants.BUTTON_DASHBOARD_NEW}
+          </Button>
+        )}
```
nmadruga, line 61: Would there be another one for `hasNonApplyPermission`?

src/components/Dashboard.tsx
```diff
@@ -0,0 +1,158 @@
...
+
+  const applicationListUserRole =
+    filter.userRole === PermissionPolicyType.Apply ? USER_ROLES.APPLICANT : USER_ROLES.REVIEWER
+
+  const constructLink = () =>
+    `/applications?type=${templateType}&user-role=${applicationListUserRole}&${Object.entries(
+      filter.query
+    )
+      .map(([key, value]) => `${key}=${value}`)
+      .join('&')}`
```
nmadruga, line 94: Maybe nice to keep this in a separated helper file -- don't. know which one, but there is a folder for `helpers`, it might be nice to have in there helpers related to "updating the URL"

src/components/Dashboard.tsx
```diff
@@ -0,0 +1,158 @@
...
+
+  const renderLink = (permissionTypes: string[], applicationListRole: USER_ROLES) => {
+    if (permissionTypes.length === 0) return null
+    return (
+      <Link
+        key={applicationListRole}
+        className="view-all clickable"
+        to={`/applications?type=${code}&user-role=${applicationListRole}`}
+      >
+        <div style={{ display: 'flex', alignItems: 'center' }}>
```
nmadruga, line 139: A missing inline type here

src/components/Dashboard.tsx
```diff
@@ -0,0 +1,158 @@
...
+          {constants.BUTTON_DASHBOARD_VIEW_ALL}
+        </div>
+      </Link>
+    )
+  }
+
+  return (
+    <div className="view-all-area">
+      {renderLink(applicantRoles, USER_ROLES.APPLICANT)}
+      {renderLink(reviewerRoles, USER_ROLES.REVIEWER)}
```
nmadruga, line 153: That's cool - and if we have different user-roles (for now it's really just the difference around these too) we can add more. I would like to have this user-role selection in a separated place, but I guess that what was already there was hard to use. It could be a nice refactoring issue to replace the existing generation of columns on the list to also use this simplified one: Applicant or Reviewer basically for all different combinations of review + assign permissions

src/utils/helpers/list/buildQueryFilters.ts
```diff
@@ -72,13 +126,20 @@ const inList = (values: string) => ({ inInsensitive: splitCommaList(values) })
...
 })
 
+// Can represent dates as relative numbers (number of days), i.e. -1, +4, 4, -3
+const convertRelativeDates = (dateStrings: string[]) =>
+  dateStrings.map((dateString) => {
+    if (!dateString.trim().match(/^[-+\d]+$/g)) return dateString
+    return datePlusDays(Number(dateString))
+  })
+
 const parseDateString = (dateString: string) => {
```
nmadruga, line 140: I think the logic for this should be working well - but it's a little difficult to understand what this is doing on the first look. Not sure what this line 83 is doing:
`...getGenericTypes(),`
I'm log it now and see.

But my question here is, how does the Dates logic (after converting to enum like `today`, `yesterday` works? Do we set the URL with these values and then convert into dates to pass to GraphQL?

nmadruga (changes requested):
First run on reviewing this one. The UI looks real good, and the logic is nice to follow. 

I only found it very tricky on the `buildQueryFilters` it's hard to understand what goes to where. Maybe just a comment at the top can easy the comprehension. And I have one suggestion on `useListTemplates`. Other than that just tiny easy fix/improvements.

CarlosNZ (approved):
Cool, thanks for tidying up the style file, makes it cleaner.

I think this is pretty good until we get some further design feedback.

Cheers.

nmadruga:
I'm not familiar with one filter here, I can see it is getting set to the `type` filter, but I wonder what does it mean:
Screen Shot 2021-05-14 at 3 08 39 PM

src/components/LayoutHelpers.tsx
CarlosNZ: It's just a dev page -- probably won't be kept.

src/components/Dashboard.tsx
CarlosNZ: There's actually a way you can pass state data directly along with a `.push` if you want to look into that (I used it a little bit in `AuthenticatedWrapper.tsx`)

So you could just keep the parameters bundled in the object, pass them through, then use "updateQuery" in the receiving page (ie. applications) to set the URL etc.

CarlosNZ:
> I'm not familiar with one filter here, I can see it is getting set to the `type` filter, but I wonder what does it mean:
> Screen Shot 2021-05-14 at 3 08 39 PM

It's translating the URL query `type=...` to a GraphQL filter, in this case `equalToInsensitive`, which mean a case-insensitive match. Is that what you mean?

nmadruga:
> It's translating the URL query `type=...` to a GraphQL filter, in this case `equalToInsensitive`, which mean a case-insensitive match. Is that what you mean?

Yep. Thank you!

semantic/src/site/globals/site.variables
andreievg: b71df009f78493415b2a33e96c17bb96a0be4652

src/components/LayoutHelpers.tsx
andreievg: yeah it's a dev page, so i thought to leave it out of localisations

src/utils/types.ts
andreievg: 7e81af9d3528a9842db91f456956ab1bfb6ece27

Although when it's destructured in dashboard it's i revert to categoryIcon and categoryTitle (to make it easier to see what the title and icon refers to in JSX

src/utils/hooks/useListTemplates.tsx
andreievg: Hmm, prefer separating calculations based on their result, similar to how the progress calculations are done for review vs application. Similar in the review structure calculation, could potentially have a big loops calculating isPendingReview, isNewApplicationResponse, isActiveReviewResponse, thisReviewLatest response.. etc..)  -> but I think it's easier to digest logic when it's separated this way (and easier to find and change pieces of logic). It does create redundancy with space (having templateCategory in the TemplateInList) and it's not optimised (having to loop over the same element multiple times). I prefer readability over premature optimisation though. But maybe it's not as readable as i think ?

src/components/Dashboard.tsx
andreievg: not sure what you mean, this is to control the new application button.

src/components/Dashboard.tsx
andreievg: > Maybe nice to keep this in a separated helper file -- don't. know which one, but there is a folder for helpers, it might be nice to have in there helpers related to "updating the URL"

I was wondering if we intend to use it anywhere else ? If so it would be good to move to a common place. 

> There's actually a way you can pass state data directly along with a .push if you want to look into that (I used it a little bit in AuthenticatedWrapper.tsx)
> So you could just keep the parameters bundled in the object, pass them through, then use "updateQuery" in the receiving page (ie. applications) to set the URL etc.

We already have quite a few useEffect in `applications` pages, I wonder if this would complicate it further. I really like the idea of just navigating to applications page with just the url and for it to do the job of applying filters to the list

CarlosNZ: > We already have quite a few useEffect in `applications` pages, I wonder if this would complicate it further. I really like the idea of just navigating to applications page with just the url and for it to do the job of applying filters to the list

Yeah that seems fine in this case, since it's just creating new URL from scratch. It's only really important when you want to update existing queries.

src/components/Dashboard.tsx
andreievg: 7a37bbb6043feb7042b58eac3f496d21d4849e69

src/components/Dashboard.tsx
andreievg: Yeah, it sounds like we only really need to two roles. Good point, the selection should be shared, I did look into trying to use existing code, but could't, without pretty big change

src/utils/helpers/list/buildQueryFilters.ts
andreievg: I am not that happy with implementation of genericTypes tbh, I've added some comments to explain it: 27329d470e6e5a7ef217a0f6a942c05f6a9074b8

Also added more descriptive type (so it's more clear what the key and value definitions in `mapQueryToFilterField` are):

0bccde0bb28d19a0775dc8f2a5c730c6a876d246

andreievg:
@nmadruga @CarlosNZ 

I think I've dealt with all of the comments.

There is this one still remaining: https://github.com/openmsupply/application-manager-web-app/pull/697#issuecomment-840961666

Mainly about the constants 'globe', I wanted it customisable, maybe the default category should be built into the system (kind of like the nonRegistered user), and should be linked to every template by default ?

semantic/src/site/globals/site.variables
```diff
@@ -164,3 +164,7 @@
 
 // Box shadow -- used in Login form
 @basicBoxShadow: rgb(204, 201, 201) 0px -3px 10px 0px;
+
+// Box shadow -- used in Dashboard 
+@templateBoxShadowColor: #d4d4d5;
+@templateBoxShadow: 0 1px 3px 0 @templateBoxShadowColor, 0 0 0 1px @templateBoxShadowColor;
```
nmadruga, line 170: Nice! Can you move this color `@templateBoxShadowColor` up to be among other colous definitions? I think it can also be used on the `@basicBoxShadow` definition. it's pretty much the same.

Also - not related, but since we are discussing colors here. Can you add the latest colour from Debbie's design `#F0F0F0` to replace the color used for Application background?

src/utils/hooks/useListTemplates.tsx
nmadruga: Yes, to me the double loop is compromising to quickly read. It would be nice to have all in a big loop, but I don't quite see the connection between this one and the deduction of fields in the progress... So I didn't get the correlation, if you don't want to change, it's fine. I was just pointing out something that could be improved from my perspective.

nmadruga (approved):

src/components/Dashboard.tsx
nmadruga: Nice. Ii was just asking if you have created also the variable `hasNonApplyPermission` to be used in a similar way to go straight to a review page for example. I haven't seen that, maybe it's there.

src/components/Dashboard.tsx
nmadruga: Well, we can leave to do when needed to use in another place. That's fine. I'm just keen to keep the code cleaner by having functions that does only one thing. We have been doing that in many places which is pretty cool. I though it was a nice one to point out, it is doing a separated logic inside here.

src/utils/helpers/list/buildQueryFilters.ts
nmadruga: Cooments look great. Thanks

semantic/src/site/globals/site.variables
andreievg: 0370ff680b3a11ec8b080b32272bf61ea2c8d68e

andreievg:
> Yes, to me the double loop is compromising to quickly read. It would be nice to have all in a big loop, but I don't quite see the connection between this one and the deduction of fields in the progress... So I didn't get the correlation, if you don't want to change, it's fine. I was just pointing out something that could be improved from my perspective.

@nmadruga, with my comment I meant to point out two styles of computation, and I can understand that there is difference in how individuals digest/read code. Although I assumed (perhaps wrongly) that it's easier to break complexity into smaller chunks and compute one chunk at a time. I've written an example of this, but thought to leave it for now (let me know if you want to see it). I think the main difference between the two progress calculations is running totals (vs filtering and getting length) and having pre sorted pages and sections (which wasn't available yet when application progress was implemented)

PR #699, opened 2021-05-13
#654 Show your assignment rows and review progress rows

### LINKED ISSUE #654: Bug: Self-assignment rows showing for all reviews that have permission to self-assign
The assignment rows are showing as one per possible user that can be selfAssigned to review a submitted application. 

✅  When selection "Everyone" on the dropdown, the result is correct, you should see all possible self assigners, and the option to Self-assign or do assignment to others (if you have the permission to) in just one assignment row:
Screen Shot 2021-05-04 at 3 09 17 PM

🚫 After self assigning or after the review has been assigned to another reviewer all other reviews should change to `Self-assigned by another`, at that stage we want don't want to display them anymore on rows as currently:
Screen Shot 2021-05-04 at 3 09 40 PM

Should only show the possible assignments at this stage, so basically only the rows for:
- Next level reviewers that can self-assign
- Assignment row if you can re-assign
- Current reviewer assigned / Assigned to yourself row

nmadruga:
Fixes #654 

Very basic changes, perhaps can be improved. I wasn't sure why `isReviewable` is set to true when the review isn't in **Draft**. Anyway, just a quick previous of the changes in this PR. Please test more cases - although to test Consolidation one it would better to first merge in PR #695.

I have test using this application:

Before the user self-assign it show other possible self-assigners:
Screen Shot 2021-05-13 at 5 06 46 PM

After the review is assigned to one and locked to the other is doesn't show to others:
Screen Shot 2021-05-13 at 5 06 56 PM

But the review is visible from your review home page (logged as Consolidator 2):
Screen Shot 2021-05-13 at 5 07 08 PM

And other reviews which are already submitted would show as well. I think that was the idea.

### DIFF
--- a/src/containers/Review/ReviewSectionRow.tsx
+++ b/src/containers/Review/ReviewSectionRow.tsx
@@ -58,7 +58,9 @@ const ReviewSectionRow: React.FC<ReviewSectionRowProps> = ({
   }
 
   const canRenderRow =
-    section?.assignment?.isReviewable || section?.assignment?.action === ReviewAction.canSelfAssign
+    section?.assignment?.isAssignedToCurrentUser ||
+    section?.assignment?.action === ReviewAction.canSelfAssign ||
+    (section?.assignment?.action === ReviewAction.canView && section?.assignment?.isReviewable)
 
   return (
     <>


CarlosNZ (approved):
LGTM! Thanks

PR #700, opened 2021-05-14
#299F Update templatePermission allowedSections outside restriction

nmadruga:
Front-end branch to update types and query for [Back-end#299](https://github.com/openmsupply/application-manager-server/issue/299)

### Back-end branch PR
[#299](https://github.com/openmsupply/application-manager-server/pull/358)

### Changes
- Update generated types
- Fix queries to use the new field allowedSections on reviewAssignment

### DIFF
--- a/src/utils/graphql/queries/getReviewInfo.query.ts
+++ b/src/utils/graphql/queries/getReviewInfo.query.ts
@@ -12,7 +12,7 @@ export default gql`
         levelNumber
         reviewerId
         isLastLevel
-        templateSectionRestrictions
+        allowedSections
         trigger
         reviewer {
           id
--- a/src/components/Dashboard.tsx
+++ b/src/components/Dashboard.tsx
@@ -1,0 +1,158 @@
+import React from 'react'
+import { Link } from 'react-router-dom'
+import { Header, Button, Icon, SemanticCOLORS, SemanticICONS } from 'semantic-ui-react'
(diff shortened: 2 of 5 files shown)


src/utils/graphql/queries/getReviewInfo.query.ts
```diff
@@ -12,7 +12,7 @@ export default gql`
         levelNumber
         reviewerId
         isLastLevel
-        templateSectionRestrictions
+        allowedSections
```
CarlosNZ, line 15: Do we also want to get `canSelfAssign` and `restrictions` here as well?

CarlosNZ (commented):
Looks good, thanks.

Just wondering -- where is the `canSelfAssign` value being retrieved from currently? Do we need update that too?

src/utils/graphql/queries/getReviewInfo.query.ts
nmadruga: The only field present in `reviewAssignment` is `allowedSections`. The other mentioned fields are part of `templatePermission` and used for generating the next assignments based on `restrictions` and if the reviewer to be assigned `canSelfAssign`.

CarlosNZ (approved):
LGTM :)

nmadruga:
Bah -- merged `master` in and somehow that wasn't good. Will have to created a new branch without this merge and work from there.

PR #701, opened 2021-05-16
#Bump evaluator and make Password element work with UserEdit template

CarlosNZ:
Was supposed to be a simple version bump for the evaluator package, but ended up needing to make some changes to the Password element to get it to work well as a current password verification.

Was initially just going to use a shortText plugin, but then realised that it would store the entered password, which we don't want. Thought about adding a "dontSaveResponse" parameter to shortText, but that was going to involve too many changes to the structure of how ApplicationViewWrapper works. In the end, it was simplest to update the Password element with an additional `requireConfirmation` parameter, which can be set to `false` to prevent the Password element from displaying the additional "Confirmation" entry field. This way we can use the Password element for checking existing passwords without saving them in plain text.

Requires branch from [back-end PR #361](https://github.com/openmsupply/application-manager-server/pull/361)


src/formElementPlugins/password/src/ApplicationView.tsx
```diff
@@ -45,7 +52,9 @@ const ApplicationView: React.FC<ApplicationViewProps> = ({
       setPassword(e.target.value)
       const responses = { thisResponse: password || '' }
       const customValidation = await validate(validationInternal, validationMessageInternal, {
-        objects: { responses },
+        objects: { responses, currentUser, applicationData },
+        APIfetch: fetch,
```
nmadruga, line 56: Does this need also binding? Please check with Andrei

src/formElementPlugins/password/src/ApplicationView.tsx
```diff
@@ -91,25 +103,27 @@ const ApplicationView: React.FC<ApplicationViewProps> = ({
...
+          onChange={handleChange}
+          onBlur={handleLoseFocus}
+          onFocus={setIsActive}
+          value={passwordConfirm ? passwordConfirm : ''}
+          disabled={!isEditable}
+          type={masked ? 'password' : undefined}
+          error={
+            passwordConfirm !== password && passwordConfirm !== ''
+              ? {
+                  content: 'Passwords do not match',
```
nmadruga, line 120: Hmmm I wonder how to localise things that are internal to plugins...

nmadruga (changes requested):
An overall comment - the need to create a new permission to add the `editUser` template to the list of permissions granted to a new registered user - and allow exisiting user to have it listed on the Dashboard as well.

src/formElementPlugins/password/src/ApplicationView.tsx
CarlosNZ: We have constants.ts files for each plugin, although most aren't using it (Checkbox has one). I can move this one into a localisation file for this plugin too.

src/formElementPlugins/password/src/ApplicationView.tsx
CarlosNZ: Don't think so, it's exactly the same evaluator parameters used throughout, and there's no problem with the API calls from the front-end.

src/formElementPlugins/password/src/ApplicationView.tsx
CarlosNZ: Done 9c80005b

CarlosNZ:
> An overall comment - the need to create a new permission to add the `editUser` template to the list of permissions granted to a new registered user - and allow exisiting user to have it listed on the Dashboard as well.

Cool -- have made those changes over on [back-end branch](https://github.com/openmsupply/application-manager-server/pull/361).

src/formElementPlugins/password/src/ApplicationView.tsx
andreievg: I think the only time we want to bind is when we are accessing fetch through an object.
i.e.
```JS
const obj = { fetch }
obj.fetch(...) // should be a problem without bind
const { fetch: fetchMe } = object
fetchMe(...) // should work
```

andreievg (changes requested):
I've checked one form (google create gmail account), it looks like it's perfectly fine to show validity on loose focus (vs on every change), this would be good to use here (especially since we are reusing the component for password confirmation, we are probably going to add a limit, server side, of how many times a certain user can query for password, and doing  a request on loose focus vs key stroke would be better).

CarlosNZ:
> I've checked one form (google create gmail account), it looks like it's perfectly fine to show validity on loose focus (vs on every change), this would be good to use here (especially since we are reusing the component for password confirmation, we are probably going to add a limit, server side, of how many times a certain user can query for password, and doing a request on loose focus vs key stroke would be better).

Yup, just working on that now.

CarlosNZ:
> I've checked one form (google create gmail account), it looks like it's perfectly fine to show validity on loose focus (vs on every change), this would be good to use here (especially since we are reusing the component for password confirmation, we are probably going to add a limit, server side, of how many times a certain user can query for password, and doing a request on loose focus vs key stroke would be better).

Okay, have updated so that (when using single field password checking) we only show error on lose focus. Have also made the error display behaviour more consistent with other plugins in normal conditions, too, in that it won't show all errors onChange until you've left the input field at least once. 9f29fb28

nmadruga (approved):

andreievg (approved):
I don't think we should validate on every key stroke. Can you suggest a reason for it ? In fact it could be dangerous for password confirmation
* I type my well remembered password, but miss last character (by accident)
* When confirming I get an indicator  

Actually, in general I dont' see a reason for validating on key stroke. (of course not asking to change, but I think we can get rid of a chunk of logic by not allowing this)

Let's discuss this.

There is one change request, but it shouldn't stop the approval. Looks like we are putting post body parameters in the url string

CarlosNZ:
> I don't think we should validate on every key stroke. Can you suggest a reason for it ? In fact it could be dangerous for password confirmation

For password confirmation, it's NOT validating on every keystroke, only on lose focus.

For other situations, it only validates on keystroke AFTER you've left the input field for the first time. So if you just go through and put proper values in each field, it'll only ever be validating as you leave each field. This is the way it works for all the text input plugins, I thought we were happy with this?

CarlosNZ:
> There is one change request, but it shouldn't stop the approval. Looks like we are putting post body parameters in the url string

Yeah, that's a good point. I had thought that it wouldn't matter, since irrelevant query params don't make any difference, but I see how this could be somewhat dodgy for passwords.

Have made an issue to fix (should be fairly trivial): https://github.com/openmsupply/application-manager-server/issues/366

CarlosNZ:
@andreievg I've made the password confirmation field only show mismatch error on lose focus, as discussed.

Had to create an additional local state variable `confirmationIsActive` though to keep track of whether the field is active or not. (Didn't have to do this for the main input field, since there were other ways to check) 

a0f3ed42

PR #710, opened 2021-05-17
#638 Fix Application Home start message

### LINKED ISSUE #638: Bug: Start message (sometimes) not showing up on Application Home page
See these comments for description of the problem and a probable cause:

https://github.com/openmsupply/application-manager-web-app/pull/622#issuecomment-824429620

https://github.com/openmsupply/application-manager-web-app/pull/622#pullrequestreview-642863267

CarlosNZ:
Fixes #638.

An unevaluated "startMessage" was being passed to ApplicationHome. Have fixed so that both startMessage and submissionMessage are evaluated in `useLoadApplication` and stored in structure.info (previously it was just submissionMessage), so evaluated startMessage string can be available in ApplicationHome.

To test: start a new application. You should see a correctly rendered "Start Message" box on both the "ApplicationCreate" page `/new?...` and the "ApplicationHome" page (`/application/<serial>`).

### DIFF
--- a/src/containers/Application/ApplicationHome.tsx
+++ b/src/containers/Application/ApplicationHome.tsx
@@ -65,7 +65,7 @@ const ApplicationHome: React.FC<ApplicationProps> = ({ structure, template }) =>
     <>
       <ChangesRequestedTitle status={current?.status} isChangeRequest={isChangeRequest} />
       <ApplicationHomeWrapper
-        startMessage={template.startMessage}
+        startMessage={structure.info.startMessage}
         name={template.name}
         ButtonSegment={SummaryButtonSegment}
       >
--- a/src/containers/Application/ApplicationWrapper.tsx
+++ b/src/containers/Application/ApplicationWrapper.tsx
@@ -20,7 +20,7 @@ const ApplicationWrapper: React.FC = () => {
     userState: { currentUser, isNonRegistered },
   } = useUserState()
 
-  const { error, isLoading, structure, template } = useLoadApplication({
+  const { error, isLoading, structure } = useLoadApplication({
     serialNumber,
     currentUser: currentUser as User,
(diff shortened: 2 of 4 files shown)


andreievg (commented):
Looks good,

Although I am not sure why we have `template` as output of `useLoadApplication`, do you think it's worth removing it in this PR, structure.info.template is available for `ApplicationContainer`, `ApplicationHome` in ApplicationWrapper (that uses useLoadApplication), do you think it' worth removing in this PR ?

CarlosNZ:
> Although I am not sure why we have `template` as output of `useLoadApplication`

Yes, I was wondering that too, but I was wary of touching it as I assumed it would be used elsewhere. But it does seem like everything in "template" is also being saved in "structure.info", so perhaps we should remove it?

I'll spend a few minutes trying to make sense of it, and see what the consequences of removing it are. But if it causes other problems, then it's probably best to leave it to another issue, right?

CarlosNZ:
> I'll spend a few minutes trying to make sense of it, and see what the consequences of removing it are. But if it causes other problems, then it's probably best to leave it to another issue, right?

@andreievg Okay, I've removed `template` from the output of useLoadApplication, and replaced usage of it with `structure.info.template`, which contains the same info.

Seems to working okay with no noticeable breakages 🤞 c972cb8f

andreievg (commented):
thanks, one more little change please

useLoadApplication.ts

- 38: const [template, setTemplate] = useState<TemplateDetails>()

- 91 through to 96

CarlosNZ:
> thanks, one more little change please
> 
> useLoadApplication.ts
> 
>     * 38: const [template, setTemplate] = useState()
> 
>     * 91 through to 96

Sweet 05a66d3e

andreievg (approved):

PR #714, opened 2021-05-18
#703 List application name links to Review home for reviewers

### LINKED ISSUE #703: Change links in the list to go to the Review home page for reviewers and hide the expansion icon

CarlosNZ:
Fixes #703.

Passed the useRouter "query" object into ColumnComponents, then ApplicationNameCell can render the link dependent on whether user is acting as "applicant" or "reviewer"/"consolidator"

Edit: updated to remove "Assigner" and "Consolidator" user roles throughout, and used the USER_ROLE enum for ApplicationNameCell


andreievg (changes requested):
I fee like this needs to be a new cell. I thought that infrastructure that was put in place was meant to abstract this logic away from cells

src/components/List/Cells/ApplicationNameCell.tsx
```diff
@@ -1,17 +1,16 @@
...
-    {application.name as string}
-  </Link>
-)
+const ApplicationNameCell: React.FC<CellProps> = ({ application, query }) => {
+  const { userRole } = query
+  const linkUrl =
+    userRole === USER_ROLES.REVIEWER
+      ? `/application/${application.serial || 0}/review`
+      : `/application/${application.serial || 0}`
+
```
nmadruga, line 12: I agree with Andrei, this would be better to be passed on and received here only as a `linkUrl` to be called

src/utils/helpers/list/findUserRole.ts
```diff
@@ -16,9 +16,7 @@ type UserRoles = {
 
 const userRoles: UserRoles = {
   applicant: [PermissionPolicyType.Apply],
-  consolidator: [PermissionPolicyType.Assign, PermissionPolicyType.Review],
-  reviewer: [PermissionPolicyType.Review],
-  assigner: [PermissionPolicyType.Assign],
+  reviewer: [PermissionPolicyType.Review, PermissionPolicyType.Assign],
```
nmadruga, line 19: I think this will be checking if both match! It might be needed to update the code that verifies this to use an `or` statement

src/components/List/Cells/ApplicationNameCell.tsx
CarlosNZ: Okay, I think I'm getting two slightly different suggestions. Andrei suggests having a completely different Cell component, Nicole is suggesting same component, but with "linkUrl" passed in as prop.

I think passing in the link URL sounds better, since it seems overkill to create an additional cell component type for what is essentially a trivial difference between the two cases.

So are we cool if I:
- calculate "linkURL" string in ApplicationsList (while iterating over applications)
- pass it as additional prop (instead of `query`) to cells
- ApplicantNameCell just renders the link it's passed

andreievg:
> I think passing in the link URL sounds better, since it seems overkill to create an additional cell component type for what is essentially a trivial difference between the two cases.

I think a new cell, it's cleaner and probably less code

andreievg:
> I think a new cell, it's cleaner and probably less code

Maybe not less code, but it's cleaner, and we've added all the mechanism for exactly that reason I thought

CarlosNZ:
> Maybe not less code, but it's cleaner, and we've added all the mechanism for exactly that reason I thought

Sure, not sure what the "boundary" is though, as quite a few of the existing cell components have quite a bit of logic in them.

CarlosNZ:
> Maybe not less code, but it's cleaner, and we've added all the mechanism for exactly that reason I thought

And a new cell will have to have the same column header "Name" -- isn't that confusing?

andreievg:
> Sure, not sure what the "boundary" is though, as quite a few of the existing cell components have quite a bit of logic in them.

We currently differentiate between applicant and non applicant, and so far we have no logic within cells that do that sort of differentiation (as far as I know)

Btw the status column would also need to be changed eventually (it is currently showing the same think for reviewer and applicant). i.e. shows edit 'draft' for reviewer.

> And a new cell will have to have the same column header "Name" -- isn't that confusing?

Reviewer and Applicant action have the same column header, which is '', status as per above, should also be different but will have the same header name

The reason I am strong about this is there is logic for this already, if there wasn't then it's fine, but it's made specifically for differentiating between the two types of roles

CarlosNZ:
> > Sure, not sure what the "boundary" is though, as quite a few of the existing cell components have quite a bit of logic in them.
> 
> We currently differentiate between applicant and non applicant, and so far we have no logic within cells that do that sort of differentiation (as far as I know)
> 
> Btw the status column would also need to be changed eventually (it is currently showing the same think for reviewer and applicant). i.e. shows edit 'draft' for reviewer.
> 
> > And a new cell will have to have the same column header "Name" -- isn't that confusing?
> 
> Reviewer and Applicant action have the same column header, which is `''`, status as per above, should also be different but will have the same header name
> 
> The reason I am strong about this is there is logic for this already, if there wasn't then it's fine, but it's made specifically for differentiating between the two types of roles

Okay, makes sense, have separated them into two different cells (both in the same file though since they're practically identical): d55b5655

(Not sure if my naming is great, feel free to suggest alternatives)

Slightly off-track, I also just fixed up the React key error that was happening in ReviewerActionCell: 95b433e3

src/utils/types.ts
```diff
@@ -139,6 +139,7 @@ interface BasicStringObject {
 
 interface CellProps {
   application: ApplicationListShape
+  query: BasicStringObject
```
andreievg, line 142: i think this one is not needed anymore

src/components/List/ApplicationsList.tsx
```diff
@@ -20,6 +20,7 @@ const ApplicationsList: React.FC<ApplicationsListProps> = ({
   handleSort,
   loading,
 }) => {
+  const { query } = useRouter()
```
andreievg, line 23: i think this one is not needed anymore

src/components/List/ApplicationsList.tsx
```diff
@@ -50,6 +51,7 @@ const ApplicationsList: React.FC<ApplicationsListProps> = ({
                 index,
                 columns,
                 application,
+                query,
```
andreievg, line 54: also here

andreievg (changes requested):
thanks or changes, a couple of things to remove and good to go

CarlosNZ:
> thanks or changes, a couple of things to remove and good to go

Sweet, thanks for spotting those. e93ac907

andreievg (approved):

PR #715, opened 2021-05-18
#310B New filters by reviewer and assigner

nmadruga:
### Changes
- Add new filters for `assigner` and `reviewer` filtering. 
- Problem is that we don't set the `assigner_id` yet, so can't test that. 
- To test with `reviewer` I was using:
Go to see the list of `ReviewTest` applications
- Test 1: Filter by "testReviewer1" with URL `http://localhost:3000/applications?type=ReviewTest&user-role=consolidator&reviewer=testReviewer1` who is assigned to all applications, so nothing changes.
- Test 2: Filter by "testReviewer2" with URL`http://localhost:3000/applications?type=ReviewTest&user-role=consolidator&reviewer=testReviewer2` now only 3 applications will show because this user isn't assigned to one of the applications

### DIFF
--- a/src/utils/helpers/list/buildQueryFilters.ts
+++ b/src/utils/helpers/list/buildQueryFilters.ts
@@ -145,7 +145,7 @@ const mapQueryToFilterField: FilterMap = {
 const splitCommaList = (values: string) => values.split(',')
 
 // Use this to find string in an array of strings
-const inArray = (value: string) => ({ anyEqualTo: value })
+const inArray = (values: string) => ({ overlaps: splitCommaList(values) })
 
 // Use this if the values can be free text strings (e.g. stage name)
 const inList = (values: string) => ({ inInsensitive: splitCommaList(values) })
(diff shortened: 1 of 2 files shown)


src/utils/helpers/list/buildQueryFilters.ts
```diff
@@ -139,6 +144,9 @@ const mapQueryToFilterField: FilterMap = {
 
 const splitCommaList = (values: string) => values.split(',')
 
+// Use this to find string in an array of strings
+const inArray = (value: string) => ({ anyEqualTo: value })
```
andreievg, line 148: I've checked another filter type, which can accept an array:

```GraphQL
query MyQuery2 {
  applicationList(
    filter: {
      reviewerUsernames: {
        overlaps: ["testReviewer2", "testReviewer1"]
      }
    }
  ) {
    nodes {
      id
      reviewerUsernames
    }
  }
}

```

this seems to work correctly, would be great to use this one (split values into array, similar to enum list filtering)

andreievg (changes requested):
looks good, but one small change request, so that we can filter by multiple reviewers/assigners at ta time

nmadruga:
About this issue, I have added something to work on adding this #718 

> Problem is that we don't set the assigner_id yet, so can't test that.

src/utils/helpers/list/buildQueryFilters.ts
nmadruga: I was looking for it! Thanks, I'll add this implementation to this PR now : )

nmadruga: Done: 8fe1bd8

andreievg (approved):
Looks good, thanks

PR #720, opened 2021-05-18
#651 Element loading improvements

### LINKED ISSUE #651: Element plugins -- better "loading" information
Currently, if an element plugin takes a long time to load (e.g. a slow API call), there is no information for the user that something is even expected to show up. I've added a crude "Loading" spinner for the time being, but we really should have other information show up where possible.

The problem is that, with the current way it works, *all* element parameter need to finish being evaluated *before* the "ApplicationView" element starts rendering anything, so even basic things like the Label and Description don't appear until all evaluations are finished.

The reason it's like this is that there is a single "parameters" object passed to each plugin, and if you pass in an unevaluated property (i.e. it's still an object), React throws an error. Currently it waits until the `parametersReady` state variable is `true` before trying to render to avoid this problem.

We'll need to restructure it somewhat so that unevaluated parameters are never passed to the plugins, but there is a seperate "evaluatedParameters" field that gets incrementally updated as values are evaluated. For incomplete parameters, there should be a "default" placeholder, such as "Loading" that is used. That way the plugin can start rendering and displaying useful info to the user before slow API calls are resolved.

CarlosNZ:
Fixes #651 

I thought it would be good to sort this out before doing the Ingredients list.

I'm much happier with how the dynamic parameter are being handled now, we no longer need the annoying `parametersReady` value to wait for parameters to be evaluated, and form elements render immediately while they wait for additional parameters to load (like from a slow API call).

### Summary of changes

- Removed "dynamic parameters" from pluginConfig files, as it was silly to have to specify this. Now the ApplicationViewWrapper just checks each parameter at load and only evaluates the ones that have dynamic expressions _for this instance_
- Parameters are separated into "simpleParameters" and "parameterExpressions" -- simple (ie. not evaluator expressions) ones are passed to the plugin element immediately. A "default" value (usually "Loading...") is passed for parameters that are awaiting dynamic evaluation.
- "parameterExpressions" are evaluated in parallel, and the "evaluatedParameters" object is updated immediately each time one of them gets a result
- "Default" loading values are specified in each element's "pluginConfig.json". If not provided, the string "Loading..." is used, but it's important to specify valid Loading value for arrays (for example), as the plugin will give an error if it gets a string value when it's expecting an array of options.
- Out of scope: fixed a React "key is not a prop" error in PageElements

One issue: I had to specify a "key" for the "ApplicationView" element (line 153 of ApplicationViewWrapper) in order to force the child components to re-render after a parameter update. This seems clunky, and there's probably a better way. I noticed that it was _only_ the dynamic checkbox list in the checkbox plugin that was affected by this -- every other plugin, including Dropdowns with dynamic options, seemed to be okay and would update when the parameters are evaluated. No idea why only checkbox is affected, but if someone can explain it to me (and how to do it better), that'd be rad.


src/formElementPlugins/ApplicationViewWrapper.tsx
```diff
@@ -151,12 +150,13 @@ const ApplicationViewWrapper: React.FC<ApplicationViewWrapperProps> = (props) =>
 
   const PluginComponent = (
     <ApplicationView
+      key={JSON.stringify(evaluatedParameters)} // Forces re-render after dynamic parameter update
```
andreievg, line 153: is this really required ?
I thought parameter change would cause re-render

src/formElementPlugins/ApplicationViewWrapper.tsx
```diff
@@ -51,24 +50,24 @@ const ApplicationViewWrapper: React.FC<ApplicationViewWrapperProps> = (props) =>
...
-  // This value prevents the plugin component from rendering until parameters have been evaluated, otherwise React throws an error when trying to pass an Object in as a prop value
-  const [parametersReady, setParametersReady] = useState(false)
-
   const { ApplicationView, config }: PluginComponents = pluginProvider.getPluginElement(pluginCode)
 
-  const dynamicParameters = config?.dynamicParameters
-  const dynamicExpressions =
-    dynamicParameters && extractDynamicExpressions(dynamicParameters, parameters)
+  const parameterLoadingValues = config?.parameterLoadingValues
+  const [simpleParameters, parameterExpressions] = buildParameters(
```
andreievg, line 56: any reason for using array vs object ?

src/formElementPlugins/ApplicationViewWrapper.tsx
```diff
@@ -250,32 +245,19 @@ const getDefaultIndex = (defaultOption: string | number, options: string[]) => {
...
-  const evaluatedExpressions: any = await Promise.all(expressions)
-  const evaluatedParameters: ElementPluginParameters = {}
-  for (let i = 0; i < fields.length; i++) {
-    evaluatedParameters[fields[i]] = evaluatedExpressions[i]
+  const simpleParameters: any = {}
+  const parameterExpressions: any = {}
+  for (const [key, value] of Object.entries(parameters)) {
+    if (value instanceof Object && !Array.isArray(value)) {
+      parameterExpressions[key] = value
+      simpleParameters[key] = parameterLoadingValues?.[key] ?? 'Loading...'
```
andreievg, line 257: wonder if we need to localise loading

andreievg (commented):
Minor comments, also a question, is the reason for JSON stringify for key, so that previous response that relies on api call to render is shown ?

I think the problem we need to solve is evaluating a big list of options (via API), which needs to be done within the plug itself .. Can't really image any other use cases for evaluation of parameters that are required for plugin to render.

src/formElementPlugins/ApplicationViewWrapper.tsx
CarlosNZ: It should, but for some reason it doesn't for the checkbox with dynamic options, as per my comment in the original post. Would really like to know why. Don't suppose you have any ideas? I'll have a closer look at it.

src/formElementPlugins/ApplicationViewWrapper.tsx
CarlosNZ: No particular reason, I just like this style for functions that return multiple values. I think it's more elegant for simple variables, but that's just personal taste.

(Although it would seem React has similar thoughts: `[ value, setValue ] = useState()` ;) )

src/formElementPlugins/ApplicationViewWrapper.tsx
CarlosNZ: Yeah, we probably do, although I wasn't sure how to handle localisations that are in the config.json file. (And I didn't think there was much point moving this to constants if the config values weren't also.)

src/formElementPlugins/ApplicationViewWrapper.tsx
CarlosNZ: Okay, I understand why now. But we can't use the key, that's a clunky and inefficient way to force a new instance of every element component for every change.

The problem is that the Checkbox plugin is using "derived state", in that it has a "useState" value (the checkboxes array) that is built from incoming parameters. But when the parameters change, the state doesn't get updated. I should be able to come up with a solution for this.

In the meantime, have removed the key 1e838555

CarlosNZ: Okay, got it sorted for Checkboxes now. Made a small `useEffect` to watch for changes in `parameters.checkboxes` and update local state, but dont' actually save the state if it's still "Loading..." 1e838555

src/formElementPlugins/ApplicationViewWrapper.tsx
andreievg: sure, all good with personal prefs.

react uses it to allow you to name it to whatever you want with minimum syntax

src/formElementPlugins/ApplicationViewWrapper.tsx
andreievg: But this is not in config.json ?

CarlosNZ: `parameterLoadingValues` are in pluginConfig.json, this is just a fallback. What I was meaning is that there isn't a lot of point localising this string unless we can also localise the ones in the .json file.

andreievg (approved):
looks good, thanks

PR #722, opened 2021-05-19
#674 replace actions filter from number to enum type

### LINKED ISSUE #674: Change how we return actions in application list to be enums per column
For now, let's just ADD a new column, then we can migrate front-end calls over gradually.

nmadruga:
Fixes #674 

### Needs back-end issue in PR
[#674F](https://github.com/openmsupply/application-manager-server/pull/367)

### Changes
- Replace counts (to set review_actions) for a new `reviewer_action` ENUM
- Fix query to use new filter as enum - not testes
- Fixed filters on Dashboard
- Replace counts for assigner_action with new ENUM - UPDATED 20/05


nmadruga:
This is ready for a re-review now. I have finished all the work-in-progress

andreievg (changes requested):
This is all working very well.

Can you add assignerAction to `buildQueryFilters` and change the last two filters in back end to use assignerActions please (i've checked it seems to work with those changes)

nmadruga:
> This is all working very well.
> 
> Can you add assignerAction to `buildQueryFilters` and change the last two filters in back end to use assignerActions please (i've checked it seems to work with those changes)

Done

andreievg (approved):

PR #724, opened 2021-05-19
#718 Update assignerId on review_assignment when assign or self-assign

### LINKED ISSUE #718: Update review_assignment field of assigner_id when assigning another user to a review

nmadruga:
Fixes #718 

### Branched off PR
#722 

### Changes
- Add assignerId to updateReviewAssignment mutation
- Also added the user id of someone that self-assign to show in the `assignerId`, not sure if that's a change on what we planning. My idea was that with that we can filter by this self-assigner and perhaps have a new action for assigner (self-assigner only) to un-assign. That's for the future, but just to be consider to adding the user id as `assignerId` or not.


src/containers/Review/AssignmentSectionRow.tsx
```diff
@@ -27,7 +28,7 @@ const AssignmentSectionRow: React.FC<AssignmentSectionRowProps> = (props) => {
     if (!assignment) return
 
     try {
-      await assignSectionToUser({ assignment, sectionCode })
+      await assignSectionToUser({ assignment, assignerId: userId, sectionCode })
```
andreievg, line 31: what do you think about getting userId in the hook vs passing it through as props ?

nmadruga: Hmmm for this one it should be fine. That was my first trial, but I was having a hook error thrown because the other place calling wasn't a React component. But I can change for this one

nmadruga: Done: 21d664a

andreievg: i was meaning something like this: `https://github.com/openmsupply/application-manager-web-app/compare/assignerId`

andreievg (approved):
Approved but FYI it think could be a little simpler: https://github.com/openmsupply/application-manager-web-app/pull/724#discussion_r637520877

nmadruga:
> Approved but FYI it think could be a little simpler: [#724 (comment)](https://github.com/openmsupply/application-manager-web-app/pull/724#discussion_r637520877)

Yes, sorry I haven't gone back to this review before. I was very focused on other PR. I can make this change now

Issue #713, opened 2021-05-17
Show changes that were made by reviewer to consolidator


andreievg:
Can have something similar to reviewer re-review ? 'Re-Review' with number on review home page, and Re-Review instead of review on consolidation/review page.

PR #730, opened 2021-05-21
675 part 1 update consolidation in progress status after lower level review is submitted

### LINKED ISSUE #675: When consolidation is in progress and new review is submitted needs to either show in the UI that the consolidation should be re-started or reload the ReviewHome

nmadruga:
Fixes #675

### Back-end branch PR to use:
[675F](https://github.com/openmsupply/application-manager-server/pull/369)

### Things to discuss
- This is label `part 1` since we might want to have something extra - this is a pretty simple solution. I wasn't sure if the warning is better, but though it could be strange to show one snackbar while **Review buttons** disappears.
- We could use something cleaver on subscriptions to reload the page on a Review.status update. But I think that's bit too much for now. Might be a good idea to do that as part of other changes that also uses something similar to do live updates UI.
- When the review has status PENDING we show the action to **Start** on the `ReviewHome` page was wondering if this would be improved, but now I remembered Andrei already gave some good ideas on issue #713: 
 > Can have something similar to reviewer re-review ? 'Re-Review' with number on review home page, and Re-Review instead of review on consolidation/review page.

### Changes
- Simple change to show **View** instead of **>** on the `ReviewHome` for the current user:
Screen Shot 2021-05-21 at 2 17 08 PM
Screen Shot 2021-05-21 at 2 17 15 PM

- Add new message and modal on `ReviewPage` for when loading a review with PENDING status:
Screen Shot 2021-05-21 at 2 07 54 PM

- After the **OK** button is clicked loads the `ReviewHome`:
Screen Shot 2021-05-21 at 2 24 33 PM


src/containers/Review/ReviewPage.tsx
```diff
@@ -67,10 +74,28 @@ const ReviewPage: React.FC<{
...
+    setShowWarningModal({
+      open: true,
+      title,
+      message,
+      option,
+      onClick: () => {
+        setShowWarningModal({ open: false })
+        push(`/application/${fullReviewStructure.info.serial}/review`)
+      },
+      onClose: () => setShowWarningModal({ open: false }),
```
andreievg, line 95: do we want to make sure we navigate back to home either way (even is modal is somehow closed ?)

src/containers/Review/ReviewPage.tsx
```diff
@@ -67,10 +74,28 @@ const ReviewPage: React.FC<{
       current: { stage },
       name,
     },
+    thisReview,
     attemptSubmission,
     firstIncompleteReviewPage,
   } = fullReviewStructure
 
+  const status = thisReview?.status || null
```
andreievg, line 82: technically this is not needed, but don't need to change

andreievg:
This looks pretty good.

We also need to do a check on submission (I think it's fine to do in both places ?). 

The reason we need to do it just before submission, is that there is a use case that status 'pending' will be lost otherwise:
i.e.

Reviewer1, Reviewer2, Consolidator;

Review1 submits review.
Consolidator starts review, with disagreements
Reviewer2 submits review -> Consolidator's status changes to pending
Consolidator hasn't made any decision and submits change request to Reviewer1
Now the pending status will be 'lost' and will have to wait for Reviewer1 to submit review before looking at Reviewer2 consolidation.

For the pre-submission check we need to do something extra in in ReviewSubmit:

`https://github.com/openmsupply/application-manager-web-app/compare/675-part-1-update-consolidation-in-progress-status-after-lower-level-review-is-submitted...675-check-status-before-review-submission`

andreievg:
As per above, it would also be a good place to check that review is locked

src/containers/Review/ReviewPage.tsx
nmadruga: Yea. Good point

nmadruga: Done: da72742

nmadruga:
@andreievg I tried adding the check for Pending status as you suggested, with the modal to display when review is pending. Have also updated the test data to have Application serial: 12345 with one review submitted and other in progress, to check the workflow with **Consolidator1** start the consolidation and before trying to submit I submit the other Review. But have been finding problems with the `updateReview` mutation now. It's time for me to go, so if you like to have a look on what is the problem that would be great. Otherwise I can have another look on this tomorrow.

andreievg:
@nmadruga 

> @andreievg I tried adding the check for Pending status as you suggested, with the modal to display when review is pending. Have also updated the test data to have Application serial: 12345 with one review submitted and other in progress, to check the workflow with Consolidator1 start the consolidation and before trying to submit I submit the other Review. But have been finding problems with the updateReview mutation now. It's time for me to go, so if you like to have a look on what is the problem that would be great. Otherwise I can have another look on this tomorrow.

It worked as expected for me (sorry did't see 12345 test data that you mention, i used back end branch: #675F-part-1-update-upper-level-status-on-new-lower-level-submission-with-consolidation-in-progress), I did the following:

* create new review from 12345, approve all and submitted
* assign stage 2 to reviewer 1 and 2
* reviewed with reviewer 1 and submitted
* opened a new incognito window and reviewed as consolidator reviewer 1 changes, disagreed with one, and did't press submit
* reviewed with reviewer 2 and submitted
* changed to incognito window and tried submitted, go the warning
* was taken to home page, and was able to start reviewer 2 consolidation

nmadruga:
Yep, that sounds good. I have revert the change on application serial: 12345 then and fixed something that was wrong on the test (back-end). Will merge both PRs now if you agreed @andreievg

andreievg (approved):
looks good, I would be tempted to dry things up a little bit more if i touch the code though.

andreievg:
> Yep, that sounds good. I have revert the change on application serial: 12345 then and fixed something that was wrong on the test (back-end). Will merge both PRs now if you agreed @andreievg

👍

Issue #489, opened 2021-03-22
Ingredients list plugin (Table in original IRIMS)


nmadruga:
From "IRIMS description truncated user guide.pdf":

![Screen Shot 2021-05-17 at 12 53 36 PM](https://user-images.githubusercontent.com/5456533/118419057-0c149500-b70f-11eb-8127-b4464561d0ea.png)

![Screen Shot 2021-05-17 at 12 53 42 PM](https://user-images.githubusercontent.com/5456533/118419063-0f0f8580-b70f-11eb-83a1-8612d690586e.png)

nmadruga:
Take a screenshot from original IRIMS so we can have an idea of what to use as input to decide - use table or something else

CarlosNZ:
I have a thought about how this might work utilising existing plugins (partly).

Because it seems each "ingredient" requires the filling in of its own "mini-form" that uses the same kind of fields we already have.

How would it be if we created the "mini-form" for inputting ingredients using our existing elements (shortText, radioChoice etc.), then we make a new plugin called "buildList". It's parameters would specify a range of _existing_ responses to take in as input and create one "ingredient"?

The plugin UI would have:
- a button to "Add ingredient" which takes the current responses and adds them to an ingredient in a list (while resetting those responses)
- each "ingredient" can be displayed using Semantic's "Card" component (something like [this](https://react.semantic-ui.com/views/card/)
- each ingredient can be deleted.
- selecting an ingredient in the list re-populates the previous form entries with their values for that ingredient.
- a "Ingredients completed" button, which sets the visibility of the other responses to `false` (so they don't show up in the Summary View). It can be toggled back to "incomplete", but the "validation" of this plugin would include making sure this button is "ON".

I dunno, there may be technical limitations and other problems with this approach that I haven't forseen, but it'd be good to be able to build this fairly complicated ingredients input interface without duplicating existing plugin functionality.

If you think this is a workable idea, let me know and I'll come up with a more detailed description.

CarlosNZ:
Here's a quick mockup of how it might look:

![Ingredients List](https://user-images.githubusercontent.com/5456533/118428773-4ccad900-b724-11eb-87f2-e33c955f7c99.png)

andreievg:
I see a few challenges in implementing this (a lot of them are dealt with by above, but not all)

**Validation**

I think we would need a 'custom' validation, it would go through every row and do the validation

**Visibility of the form**

[...]

PR #732, opened 2021-05-21
#489 List builder (Ingredients list) plugin

CarlosNZ:
Just putting this here in Draft so you can have a look and give feedback. Styles are not really addressed at all, and there's still a couple of quirks with the validation that I need to explore, but the basic idea is working well, I think.

If we like the general idea, I can do a screencast to send to Deb for design/UI ideas, like we did with the Filters.

Please use back-end branch `#489F-ingredients-list-demo` (now [in PR](https://github.com/openmsupply/application-manager-server/pull/372)) and go to Section 3, Page 2 of **Feature Showcase** template. You can switch between "table" or "card" layout using the Dropdown in the next question, and you should be able to Add, Edit, and Delete list items with validation from the Modal (it won't put items into the list if the input fields are invalid)

Have a look at the template mutation to see the input/parameter structure.

### Still To-Do

- The styling of the Application List table is currently broken. That's because the table style over-rides done for that make the "generic" table look stupid. For now, I've just commented out all the table overrides, so please ignore that -- I will make them specific to a particular class or id before finishing this PR.
- ~~No proper types defined or localisation strings handled yet, just making it all work~~ b377d8b9
- It doesn't yet do additional validation of its own response object (the list items), it's relying on the check before it puts items into the list. We'll want to do another validation check, but that'll require an internal validation method. We can discuss how much of a priority this is.
- Haven't tried working with dynamic parameters or element fields really, they'll probably do bad things. If we want to have elements within the Input Modal change visibility dynamically based on other choices within the modal, that'll add another layer of complexity I haven't delved into yet. Perhaps we leave this for now?
- ~~Haven't documented input parameters yet. You'll see I've created a custom substitution string (look at the template), replacing element codes with values, and it's also split into "header", "meta", and "description" as per [Semantic's "Card" component](https://react.semantic-ui.com/views/card/).~~ Documentation now done

### Changes to ApplicationViewWrapper to get it working nicely:

(without it affecting existing functionality)
- Moved `[value, setValue]` out of ApplicationViewWrapper and into plugins (shortText and longText were the only ones even using it), as it was causing child elements to affect each other through a shared method
- made the "responseMutation" method a passed-in method, which is replaced with a custom method when the listBuilder spawns its child elements. (Still defaults to "responseMutation" in normal functioning)
- ~~Had to create a new parameter `isFocusChange` that is passed into `calculateValidationState`, so the listBuilder could discriminate between an onUpdate validation and an onSave validation. Normal plugins use the changing value of "currentResponse.text" to cause a state update, but the list builder doesn't get any such info from its _own_ "currentResponse", so it wasn't getting a proper validationstate update on changing Focus. Open to better suggestions for how to handle this.~~ Seems to have been fixed, have removed


src/formElementPlugins/ApplicationViewWrapper.tsx
```diff
@@ -108,10 +115,10 @@ const ApplicationViewWrapper: React.FC<ApplicationViewWrapperProps> = (props) =>
         })
       if (jsonValue === null)
         // Reset response if cleared
-        await responseMutation({
+        await onSaveUpdateMethod({
           variables: {
             id: currentResponse?.id as number,
-            value: null,
+            value: { text: undefined },
```
andreievg, line 121: is this just 'cosmetic' or was there something up with null ?

src/formElementPlugins/listBuilder/src/ApplicationView.tsx
```diff
@@ -0,0 +1,332 @@
...
+                  key={`list-${element.code}`}
+                  element={element}
+                  isStrictPage={inputError}
+                  allResponses={allResponses}
+                  currentResponse={
+                    selectedListItem === null
+                      ? {
+                          id: index,
+                          code: element.code,
+                          text: currentInputResponses?.[index]?.value.text,
```
andreievg, line 183: would the be a case where this has value ?

src/formElementPlugins/ApplicationViewWrapper.tsx
```diff
@@ -75,7 +81,7 @@ const ApplicationViewWrapper: React.FC<ApplicationViewWrapperProps> = (props) =>
     onUpdate(currentResponse?.text)
   }, [currentResponse, isStrictPage])
 
-  const onUpdate = async (value: LooseString) => {
+  const onUpdate = async (value: LooseString, isFocusChange = false) => {
```
andreievg, line 84: I still don't quite understand about isFocusChange, I've read your explanation, but still can't quite figure it out.
To me it seems that ApplicationViewWrapper shouldn't care who is calling it.

andreievg:
Had a quick look through code, and did a little test, will do a more thorough look. 

One comments was about state, I would have thought we store it as:

```JS
type CurrentListItem =  { [code: string]: InputResponseField }
type ListItems = CurrentListItem[]
```

I think it would be easier to use down the line. And when doing custom validation (and if we want to either validate inner form elements against each other or introduce evaluation of validity etc, which btw seems quite easy now that i've looked at `addEvaluatedResponsesToStructure`, it just needs to export `evaluateAndValidateElements`, but this is for v2 i think of this plugin)

src/formElementPlugins/listBuilder/src/ApplicationView.tsx
```diff
@@ -0,0 +1,332 @@
...
+            )}
+            {meta && (
+              <Card.Meta onClick={() => editItem(index)}>
+                <Markdown text={substituteValues(meta, item)} semanticComponent="noParagraph" />
+              </Card.Meta>
+            )}
+            {description && (
+              <Card.Description onClick={() => editItem(index)}>
+                <Markdown
+                  text={substituteValues(description, item)}
```
andreievg, line 292: I wonder if we really need this, or if we can just have description with `Title` and `Value`. I can see where you were going with dynamic card formatting, but not sure if it's really needed

src/formElementPlugins/listBuilder/src/ApplicationView.tsx
```diff
@@ -0,0 +1,332 @@
...
+        }}
+        onOpen={() => setOpen(true)}
+        open={open}
+        trigger={<Button primary content={createModalButtonText} />}
+      >
+        <Segment>
+          <Form>
+            <Markdown text={modalText} />
+            {inputFields.map((field: TemplateElement, index: number) => {
+              const element = {
```
andreievg, line 154: probably a good idea to extract this out of jsx, maybe even map inputFields into something that can be iterated over here

CarlosNZ:
> One comments was about state, I would have thought we store it as:
> 
> ```js
> type CurrentListItem =  { [code: string]: InputResponseField }
> ```

Yeah, that makes sense. The top-most (ListItems) still needs to be a list, since order is important, but the input responses for each list item can be an object indexed by code as you suggest. I think the reason I went the way I did is that there is not really a clear link between the response that comes back (via onSave) and the original element, except through ID. So I was saving the code lookup until the "updateList" method, and doing it all there at once. But yeah, happy to do it the other way, I can't see any problems arising.

src/formElementPlugins/ApplicationViewWrapper.tsx
CarlosNZ: There was a reason -- I think it's because the listBuilder needs the .text and it was causing a crash somewhere when it was getting no text field. This might change though once I refactor to the data shape you described, so will revert this if I can.

src/formElementPlugins/listBuilder/src/ApplicationView.tsx
CarlosNZ: No, but if there's no "text" field present, the inner item goes immediately to "strict" error mode, so shows validation errors immediately. I can't totally remember why this happens, but can confirm that if I set this to "" or `null`, you get immediate validation fail on loading the modal.

src/formElementPlugins/ApplicationViewWrapper.tsx
CarlosNZ: Yeah, I'll talk it through with you in person, it'll be easier I think.

src/formElementPlugins/listBuilder/src/ApplicationView.tsx
CarlosNZ: You mean the "element" fields definitions? Yeah sure, I'd planned to organise this a bit better once I had a clearer idea of which fields were going to be coming from where (and which would just have "dummy" values, etc.)

src/formElementPlugins/listBuilder/src/ApplicationView.tsx
CarlosNZ: I think we need some kind of substitution system, and this seemed the simplest to me. But we can discuss options. (But if a card is only ever going to show "Title: Value", then there's not really much point displaying cards, you may as well always show Table).

CarlosNZ:
> > One comments was about state, I would have thought we store it as:
> > ```js
> > type CurrentListItem =  { [code: string]: InputResponseField }
> > ```
> 
> Yeah, that makes sense. The top-most (ListItems) still needs to be a list, since order is important, but the input responses for each list item can be an object indexed by code as you suggest. I think the reason I went the way I did is that there is not really a clear link between the response that comes back (via onSave) and the original element, except through ID. So I was saving the code lookup until the "updateList" method, and doing it all there at once. But yeah, happy to do it the other way, I can't see any problems arising.

Refactored to store ListItems as you described. 3290001f

The conversion from the array of input responses to code-indexed object only happens at the last step though -- updateList(). It's a lot easier to keep the actual collection of InputResponses as an array, since virtually all the processing on them is using array methods (.map, .reduce, etc.)

src/formElementPlugins/listBuilder/src/ApplicationView.tsx
CarlosNZ: a80f5def

src/formElementPlugins/listBuilder/src/ApplicationView.tsx
CarlosNZ: Have added a default/fallback -- if "displayFormat" is not provided,  the card will just show "Title: Value" as you describe.

As you'll see, it looks _uuuugly_, so it's nice to have the fancy option too, if configurer can be bothered. :) 

86212533

src/formElementPlugins/ApplicationViewWrapper.tsx
CarlosNZ: Have reverted 55d5b51e -- seems to be okay

src/formElementPlugins/listBuilder/src/ApplicationView.tsx
CarlosNZ: Ah yes, that's right. This does have a value, as it keeps getting updated whenever a fields loses focus.

src/formElementPlugins/ApplicationViewWrapper.tsx
CarlosNZ: Okay, it looks like I can get rid of this now. Must have avoided the problem in the process of fixing/improving something else.

db57beec

andreievg:
>  think the reason I went the way I did is that there is not really a clear link between the response that comes back (via onSave) and the original element, except through ID. 

Is that because ID is an number, so can't use 'code' ?

CarlosNZ:
> Is that because ID is an number, so can't use 'code' ?

No, it's because the Response is limited to the shape of the existing `onSave` method, which is:
```
await onSaveUpdateMethod({
          variables: {
            id: currentResponse?.id as number,
            value: jsonValue,
          },
        })
```,

So no "code" included. Have to re-link to template element by looking up id.

File: src/formElementPlugins/listBuilder/src/ApplicationView.tsx [hunk 6]
```diff
@@ -0,0 +1,347 @@
...
+  )
+
+  const [listItems, setListItems] = useState<ListItem[]>(initialValue?.list ?? [])
+  const [selectedListItem, setSelectedListItem] = useState<number | null>(null)
+  const [open, setOpen] = useState(false)
+  const [inputError, setInputError] = useState(false)
+
+  useEffect(() => {
+    onSave({
+      text: 'Not required', // Text value never displayed to user
```
andreievg, line 71: hmm this is what we use to deduce if required is empty. I guess custom validation within this plugin would handle absence of value ? Although i don't think there is an easy way to display isRequired but empty (but only do it in strict mode) if validation is done within the plugin

CarlosNZ: This plugin doesn't actually have a way to enforce "isRequired" yet. Can come up with one (i.e. using presence or absence of `.text`), or could just specify a validation criteria such that `listItems.length > 0` ?

andreievg:
> > Is that because ID is an number, so can't use 'code' ?
> 
> No, it's because the Response is limited to the shape of the existing `onSave` method, which is:
> 
> ```
> await onSaveUpdateMethod({
>           variables: {
>             id: currentResponse?.id as number,
>             value: jsonValue,
>             isValid: validationResult.isValid,
>           },
>         })
> ```
> 
> So no "code" included. Have to re-link to template element by looking up id.

Hmm to me it seem this is easier (instead of buildListItem and getInputResponses):

https://github.com/openmsupply/application-manager-web-app/blob/db57beecdbb1b2d845acc18ad0064ad73180fcf0/src/formElementPlugins/listBuilder/src/ApplicationView.tsx#L171
to 

```JSX
onSaveUpdateMethod={innerElementUpdate(element.code}
```

```JS
  const innerElementUpdate = (code: string) => async ({ variables: response }: { variables: InputResponseField }) => {
    const newResponses ={ ...currentInputResponses, [code]: response } 
    setCurrentInputResponses(newResponses)
    if (!anyErrorItems(newResponses, inputFields)) setInputError(false)
  }
```

src/formElementPlugins/listBuilder/src/ApplicationView.tsx
andreievg: btw suggestion is to have at least something there or undefined if itemList.length === 0

src/formElementPlugins/listBuilder/src/ApplicationView.tsx
```diff
@@ -0,0 +1,347 @@
...
+      list: listItems,
+    })
+  }, [listItems])
+
+  // This is sent to ApplicationViewWrapper and runs instead of the default responseMutation
+  const innerElementUpdate = async ({ variables: response }: { variables: InputResponseField }) => {
+    const newResponses = [...currentInputResponses]
+    newResponses[response.id as number] = response
+    setCurrentInputResponses(newResponses)
+    if (!anyErrorItems(newResponses, inputFields)) setInputError(false)
```
andreievg, line 81: should this only happen on submit add/update button ? (an setInputError(false) after adding or updating ?)

src/formElementPlugins/listBuilder/src/ApplicationView.tsx
```diff
@@ -0,0 +1,347 @@
...
+  currentInput.some(
+    (response, index) => inputFields[index]?.isRequired !== false && !response.value.text
+  )
+
+const anyErrorItems = (currentInput: InputResponseField[], inputFields: TemplateElement[]) =>
+  anyInvalidItems(currentInput) || anyIncompleteItems(currentInput, inputFields)
+
+const substituteValues = (parameterisedString: string, item: any) => {
+  const getValueFromCode = (_: string, $: string, code: string) => item[code].value.text || ''
+  return parameterisedString.replace(/(\${)(.*?)(})/gm, getValueFromCode)
```
andreievg, line 261: I know I've added a comment about custom card formatting being probably out of scope, but this is too cool.

CarlosNZ:
> Hmm to me it seem this is easier (instead of buildListItem and getInputResponses):
> 
> https://github.com/openmsupply/application-manager-web-app/blob/db57beecdbb1b2d845acc18ad0064ad73180fcf0/src/formElementPlugins/listBuilder/src/ApplicationView.tsx#L171
> to
> 
> ```js
> onSaveUpdateMethod={innerElementUpdate(element.code}
> ```

Ah right, yeah, didn't realise you can pass down the function with a parameter from the outer scope. Looks good, thanks, I'll work it in.

src/formElementPlugins/listBuilder/src/ApplicationView.tsx
CarlosNZ: The main `setInputError` happens on Submit/update. But this is an extra one so that if you have the error message showing on the modal, it will disappear as soon as you leave the last invalid field. (i.e. you don't want to the message to persist once you've fixed all the invalid fields)

src/formElementPlugins/ApplicationViewWrapper.tsx
```diff
@@ -251,7 +255,7 @@ export const buildParameters = (
   const simpleParameters: any = {}
   const parameterExpressions: any = {}
   for (const [key, value] of Object.entries(parameters)) {
-    if (value instanceof Object && !Array.isArray(value)) {
+    if (value instanceof Object && !Array.isArray(value) && 'operator' in value) {
```
andreievg, line 258: as per our convo, would prefer for a method from evaluator to deduce this (not required for now)

src/formElementPlugins/listBuilder/src/ApplicationView.tsx
andreievg: ahh yep, sounds good

src/formElementPlugins/listBuilder/src/ApplicationView.tsx
CarlosNZ: Cool, done as part of 1aeae727

CarlosNZ:
> > Hmm to me it seem this is easier (instead of buildListItem and getInputResponses):
> > https://github.com/openmsupply/application-manager-web-app/blob/db57beecdbb1b2d845acc18ad0064ad73180fcf0/src/formElementPlugins/listBuilder/src/ApplicationView.tsx#L171
> > 
> > to
> > ```js
> > onSaveUpdateMethod={innerElementUpdate(element.code}
> > ```
> 
> Ah right, yeah, didn't realise you can pass down the function with a parameter from the outer scope. Looks good, thanks, I'll work it in.

Done in 1aeae727, thanks for the suggestion (managed to avoid using the `id` completely)

src/formElementPlugins/ApplicationViewWrapper.tsx
CarlosNZ: Cool, will bear it in mind next time we update the evaluator.

src/formElementPlugins/ApplicationViewWrapper.tsx
```diff
@@ -138,11 +144,11 @@ const ApplicationViewWrapper: React.FC<ApplicationViewWrapperProps> = (props) =>
...
   }
 
-  const setIsActive = () => {
+  const setIsActive = (text: string | undefined | null = '') => {
     // Tells application state that a plugin field is in focus
     setUpdateTrackerState({
       type: 'setElementEntered',
-      textValue: value || '',
+      textValue: text || '',
     })
```
andreievg, line 152: Hmm, i just realised this breaks validity check (if values is not changed). i.e start test review application, fill the first page, refresh the page, go into a field (in and out), try to press review and summary, nothing happens.

I wont' go into details but data is not changed but validity checks thinks it should have changed.

`const setIsActive = (text: string | undefined | null = '') => { ` -> `const setIsActive = () => { ` 
`textValue: text || '',` => `textValue: currentResponse?.text || '',`

src/formElementPlugins/ApplicationViewWrapper.tsx
```diff
@@ -279,9 +283,7 @@ const calculateValidationState = async ({
   const validationResult = validationExpression
     ? await validate(validationExpression, validationMessage as string, evaluationParameters)
     : { isValid: true }
-
   if (!validationResult.isValid && currentResponse?.text !== undefined) return validationResult
-  // !responses.thisResponse, check for null, undefined, empty string
```
andreievg, line 284: keep comment ?

src/formElementPlugins/listBuilder/src/ApplicationView.tsx
```diff
@@ -0,0 +1,339 @@
...
+    displayFormat,
+    editItem,
+    deleteItem,
+    fieldTitles: inputFields.map((e: TemplateElement) => e.title),
+    codes: inputFields.map((e: TemplateElement) => e.code),
+    Markdown,
+    isEditable,
+  }
+
+  const DisplayComponent =
```
andreievg, line 125: can this be component to export and use in SummaryView, it's a small tidy up request (since we are checking for const value rather then enum in SummaryView)

src/formElementPlugins/listBuilder/src/ApplicationView.tsx
```diff
@@ -0,0 +1,339 @@
...
+    addButtonText = strings.BUTTON_ADD,
+    updateButtonText = strings.BUTTON_UPDATE,
+    deleteItemText = strings.BUTTON_DELETE,
+    inputFields,
+    displayFormat = getDefaultDisplayFormat(inputFields),
+    displayType = DisplayType.CARDS,
+  } = parameters
+
+  const [currentInputResponses, setCurrentInputResponses] = useState<ListItem>(
+    resetCurrentResponses(inputFields)
```
andreievg, line 58: anther smallTidy up (being a-little bit picky), resetCurrentResponse, setOpen(false) and selectedListItem(null) seem to be coupled ? up to you if you want either combine their state or make a helper that resets them all at once, or neither (it should be ok as it is)

src/formElementPlugins/listBuilder/src/ApplicationView.tsx
```diff
@@ -0,0 +1,339 @@
...
+  )
+
+// Separate components, so can be shared with SummaryView
+export interface ListLayoutProps {
+  listItems: ListItem[]
+  displayFormat: { header?: string; meta?: string; description: string }
+  Markdown: any
+  fieldTitles?: string[]
+  codes?: string[]
+  editItem?: Function
```
andreievg, line 262: a little bit stronger typed please

src/formElementPlugins/listBuilder/src/ApplicationView.tsx
```diff
@@ -0,0 +1,339 @@
...
+      newList[selectedListItem] = currentInputResponses
+      setListItems(newList)
+    }
+    setCurrentInputResponses(resetCurrentResponses(inputFields))
+    setOpen(false)
+    setSelectedListItem(null)
+  }
+
+  const editItem = async (index: number) => {
+    if (!isEditable) return
```
andreievg, line 102: hmm, just for consistency sake, would be good to either remove or also add to deleteItem

src/formElementPlugins/ApplicationViewWrapper.tsx
CarlosNZ: Cool, thanks 35cf5a02

src/formElementPlugins/ApplicationViewWrapper.tsx
CarlosNZ: Restored 👍

src/formElementPlugins/listBuilder/src/ApplicationView.tsx
CarlosNZ: Do you just mean make "DisplayComponent" a component and export that so that SummaryView can just render it without any checking logic?

src/formElementPlugins/listBuilder/src/ApplicationView.tsx
```diff
@@ -0,0 +1,339 @@
...
+                  applicationData={applicationData}
+                />
+              )
+            })}
+            <Button
+              primary
+              content={selectedListItem !== null ? updateButtonText : addButtonText}
+              onClick={updateList}
+            />
+            {displayType === 'table' && selectedListItem !== null && (
```
andreievg, line 177: recently I've been liking the idea of keeping conditions in JSX free of 'logic'. As in extracting into constants and using constants (not suggesting a change, but just communicating my recent preference).
i.e. 

shouldRenderDeleteButton &&

andreievg (commented):
just a few comments. look pretty good.

Something else I've noticed

* we seems to be getting initialValue and currentResponse and using one or the other (within plugins)

The above comment also leads to a comment I've mentioned to Nicole in one of the PR's, this is not related to current changes, but existing way ApplicationViewWrapper is calling ApplicationView. Destructuring props vs passing them one by one, i think it's ok to destructure props only if it's a full set of props for the component, otherwise would prefer one by one. (I think I am to blame for that, as I've started that practice, although originally it was the full set of props in one destructure statement)

src/formElementPlugins/listBuilder/src/ApplicationView.tsx
CarlosNZ: Sure, I made a `resetModalState` method:
f31b871a
382b15af
(Sorry, two commits)

src/formElementPlugins/listBuilder/src/ApplicationView.tsx
CarlosNZ: It's actually only needed on `editItem`, as for the delete method, the UI elements that trigger it are either hidden or not reachable.

You still think it should have it just for consistency?

CarlosNZ: How about like this? (override the methods rather than logic in the method)

src/formElementPlugins/listBuilder/src/ApplicationView.tsx
CarlosNZ: Sure, will bear it in mind.

src/formElementPlugins/listBuilder/src/ApplicationView.tsx
CarlosNZ: Sure 3b3898fc

src/formElementPlugins/listBuilder/src/ApplicationView.tsx
```diff
@@ -0,0 +1,332 @@
...
+  }, [listItems])
+
+  // This is sent to ApplicationViewWrapper and runs instead of the default responseMutation
+  const innerElementUpdate =
+    (code: string) =>
+    async ({ variables: response }: { variables: InputResponseField }) => {
+      const newResponses = { ...currentInputResponses, [code]: response }
+      setCurrentInputResponses(newResponses)
+      if (!anyErrorItems(newResponses, inputFields)) setInputError(false)
+    }
```
andreievg, line 80: to fix the default value (it was race condition between LB3 and 6.

```suggestion
  const innerElementUpdate =
    (code: string) =>
    async ({ variables: response }: { variables: InputResponseField }) => {
      // need to get most recent state of currentInputResponse, thus using callback
      setCurrentInputResponses((currentInputResponses) => {
        const newResponses = { ...currentInputResponses, [code]: response }
        if (!anyErrorItems(newResponses, inputFields)) setInputError(false)
        return newResponses
      })
    }
```

CarlosNZ:
>     * we seems to be getting initialValue and currentResponse and using one or the other (within plugins)

Yeah, I've noticed that for a while. I've been leaning towards `currentResponse` and trying to (slowly) deprecate `initialValue`. (Though I was never 100% sure if they were exactly the same) 
> 
> The above comment also leads to a comment I've mentioned to Nicole in one of the PR's, this is not related to current changes, but existing way ApplicationViewWrapper is calling ApplicationView. Destructuring props vs passing them one by one, i think it's ok to destructure props only if it's a full set of props for the component, otherwise would prefer one by one. (I think I am to blame for that, as I've started that practice, although originally it was the full set of props in one destructure statement)

Yeah, there are a *lot* of props being passed into ApplicationView, once all existing props and element is destructured:
Screen Shot 2021-05-26 at 5 11 39 PM

src/formElementPlugins/listBuilder/src/ApplicationView.tsx
andreievg: or use enum in the summaryView

src/formElementPlugins/listBuilder/src/ApplicationView.tsx
andreievg: that commit looks good, thanks !

andreievg (approved):
Thanks for all the changes, looks good

PR #734, opened 2021-05-24
Publish to wiki fixes

CarlosNZ:
As mentioned in Telegram chat:

> Noticed that the images for the documentation weren’t showing up in front end wiki. Turns out images must be in the format `[[relative/path/to/images | alt-text]]` format and doens’t work (needs path from root) in the `!(alt-text)[image/path]]`, which is what we’d been using. Good news is we can do without the python conversion script, but will need to go through and replace images links with the other Markdown syntax. (Better than having to specify full paths)

> But I still think we need a small script to copy to the “_wiki” folder for publishing. That way we can keep a copy in the main repo and review documentation changes along with PRs, and then “publish” them once approved. If we edit them directly int he submodule folder (/_wiki) then we won’t see the changes to docs in our PRs.

> Although I’m not super happy with that solution, since the `[[ ]]`  syntax seems to be very much non-standard markdown (_update: it's MediaWiki format_), so it won’t be compatible in many other renderers. I think we’ll still need the conversion script, but it’ll just change “standard” image links into Github wiki-friendly image link formats

To summarise: the only way to make relative links work properly when published is to use MediaWiki link syntax, or put all content at top-level of folder structure. I've opted for the former for Images (easier to convert) and the latter for other pages (they're all rendered at top-level, so don't need to actually move the files, just update the links).

BTW -- if anyone is interested in trying to make sense of what the Regexes are doing, this will help:

https://regex101.com/r/b0A6C8/1/ (for the Markdown page links)
https://regex101.com/r/nfaUlY/1 (for the image links)

### DIFF
--- a/utils/convert_docs_to_wiki.py
+++ b/utils/convert_docs_to_wiki.py
@@ -6,7 +6,7 @@ import re
 docs_path = "./documentation"
 output_path = "./documentation/_wiki"
 markup_formats = [".md", ".markdown"]
-media_types = ["png", "jpg", "gif", "svg"]
+media_types = [".png", ".jpg", ".gif", ".svg"]
 ignore = ["external"]  # File or folder names to ignore
 ext_ignore = [".DS_Store"]  # File extensions to ignore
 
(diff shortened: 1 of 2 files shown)


andreievg (approved):
Would just trust that it works, at some point regex becomes quite complex to reason about, for me anyways (especially when thinking of use cases that can cause bugs). 
I am not arguing against using them btw, an in dev tools like this readability is not top priority.

CarlosNZ:
> Would just trust that it works, at some point regex becomes quite complex to reason about, for me anyways (especially when thinking of use cases that can cause bugs).

Yes for sure. The syntax is horrible, but they're very powerful and very fast. Good case for having robust unit tests for any regex in production code.

PR #736, opened 2021-05-25
#711 & #707 Review changes requested progress

### LINKED ISSUE #707: Create new review_responses when change request is started

### LINKED ISSUE #711: Add change request status to review progress
Similar to what's done for applicant (see `Add change request status to review progress`):

(we would need to get linked review responses, in reverse, `reviewResponsesByReviewResponseLinkId`)

'isChangeRequest'
'isChanged'

Validation:

https://app.diagrams.net/#G1prBh132KxxAze5UnjS39lJMvStzqCdbS (the first `on validate`)

Basically we look to see if previous review response is disagreement, if new review response must differ (in comment or decision)

Progress:

Want to consider isChangeRequest and !isChanged as incomplete

`pending` should also include isChangeRequest before review is started (thisReviewLatestResponse has change request)

nmadruga:
Fixes #711 & fixes #707

## Back-end PR to be used:
[#711F](https://github.com/openmsupply/application-manager-server/pull/371)

### Changes 
- #707 had the action to create duplicated reviews from previous review already implemented. The additions done here was only on `ReviewSectionRowAction` to correctly display the `action` **Update** and allow the Review to be started
- Added new fragment `consolidatorResponsesFragment` and added to `getReviewResponsesQuery` to get the connected nodes using `reviewResponseLinkId` on each returned `thisReviewResponse`
- Renamed some files to get a similar way to construct progress for each users...
- The new structure changes has now basically added:
```
{
  sortedPages: [
    {
      state: [
        {
           thisReviewLastesReponse: {
            reviewResponsesByReviewResponseLinkId: {
              nodes: [...]: ReviewResponse[]
            }
         },
         thisReviewPreviousReponse: {
            reviewResponsesByReviewResponseLinkId: {
              nodes: [...]: ReviewResponse[]
            }
          },
          changeRequestsProgress: {
           totalChangeRequests: number,
           doneChangeRequests: number
          },
          ...
       ]: PageElement
      },
    },
  ]: Page,
  sortedSections: [
    changeRequestsProgress: {
      totalChangeRequests: number,
      doneChangeRequests: number
    },
  ],
  isChangeRequest: boolean
}
``` 

### Test

- Use application serial: 34567 to test. It has CHANGES_REQUESTED from **Consolidator2** to **Reviewer1** on Section 1. 
- Click on the review HomePage on the **Update** action and check out the progress and validation considering those. 
- The progress at the start is considering both: responses reviewed + updated changes requested
- It's not visible on the Review page yet, but only the 2 first questions have changes requests, so changing those 2 should update the progress to **Completed** and allow the review to be submitted.


src/utils/helpers/structure/addChangeRequestForReviewer.ts
```diff
@@ -0,0 +1,51 @@
...
+
+const hasConsolidatorReviewResponse = (reviewerReviewResponse?: ReviewResponse) =>
+  (reviewerReviewResponse?.reviewResponsesByReviewResponseLinkId?.nodes || []).length > 0
+
+const addChangeRequestForReviewer = (structure: FullStructure) => {
+  const questionElements = Object.values(structure?.elementsById || {}).filter(
+    ({ element }) => element?.category === TemplateElementCategory.Question
+  )
+
+  const hasConsolidatorReviewResponses = questionElements.some(
```
andreievg, line 17: this methods looks great, one suggestion (not required change at all), would be to 'filter' for consolidatorReviewResponses here and us the in the forEach below

src/utils/helpers/structure/generateReviewerChangesRequestedProgress.ts
```diff
@@ -0,0 +1,52 @@
...
+  const doneChangeRequests = totalChangeRequests.filter((element) => element.isChanged)
+
+  page.changeRequestsProgress = {
+    totalChangeRequests: totalChangeRequests.length,
+    doneChangeRequests: doneChangeRequests.length,
+  }
+}
+
+const addIsChangeRequest = (newStructure: FullStructure) => {
+  const isPageRequest = ({ changeRequestsProgress }: Page) =>
```
andreievg, line 27: was wondering about the naming, did you want to make it specific to the page ? or should it be isChangeRequest ?

andreievg:
Looks great, https://github.com/openmsupply/application-manager-web-app/issues/711 also has validation.

I think the easiest way to do validation would be to exclude isChangedRequest and !isChanged from (doneConform or doneNonConform)

https://github.com/openmsupply/application-manager-web-app/blob/9c8d84ef88cd8f6fd4b286d721aef591525b09bc/src/utils/helpers/structure/generateReviewerResponsesProgress.ts#L23-L36

src/utils/helpers/structure/addChangeRequestForReviewer.ts
nmadruga: Ah, I think the last lines of the `forEach` block are meant to update **every** reviewResponse with `isChangeRequest` and `isChanged` fields, so we don't have undefined ones. This is to simplify other areas of the code.

src/utils/helpers/structure/generateReviewerChangesRequestedProgress.ts
nmadruga: Hmmm I would prefer to change this so this doesn't set the isChangeRequest here. I'll make this change (was done as part of previous commit, but now removed when I reverted)

src/utils/hooks/useGetReviewStructureForSection/helpers.tsx
```diff
@@ -87,13 +88,20 @@ const generateReviewStructure: GenerateReviewStructure = ({
   // review info comes from reviewAssignment that's passed to this hook
   newStructure.thisReview = review
 
+  // when changes requested by consolidator (included in thisReviewPreviousResponse OR thisReviewLatestResponse)
+  const isChangeResquest = addChangeRequestForReviewer(newStructure)
+
   newStructure = addIsPendingReview(newStructure, level)
   newStructure = addIsActiveReviewResponse(newStructure)
 
+  if (isChangeResquest) {
```
andreievg, line 97: tbh, I preferred previous way tbh, it keeps everything in one place and it keeps it consistent with how progress is calculated in application

nmadruga:
> Looks great, #711 also has validation.
> 
> I think the easiest way to do validation would be to exclude isChangedRequest and !isChanged from (doneConform or doneNonConform)

Thanks for the idea. I tried to implement this way, but it was hard to add the checking for changeRequested (done vs. total) on the validation and progress, so I tried something different - Now I really need to update documentation for that so we know exactly what each one means! - the `doneConform` and `doneNonConform` consider every review, similar to `totalReviewable`, so we also keep track of other changes (not requested). See how I have managed to show on the progress and on the validation to see what do you think.

src/utils/hooks/useGetReviewStructureForSection/helpers.tsx
andreievg: dont' need to change though

andreievg:
hmm, i wonder why the name was changed to generateReviewResponsesProgress ?  I am not requiring a change but was wondering why the name change, the method does more then just reviewResponses it also ads validation.

nmadruga:
> generateReviewResponsesProgress

Just to make a clear difference to `generateReviewChangeRequestsProgress`

andreievg:
> Just to make a clear difference to generateReviewChangeRequestsProgress

Ok but both of them deal with reviewResponses 🤷‍♂️

nmadruga:
> > Just to make a clear difference to generateReviewChangeRequestsProgress
> 
> Ok but both of them deal with reviewResponses 🤷‍♂️

One is to generate the progress for `changesRequested`, other to generate progress for `reviewProgress` and another to generate progress for `consolidationProgress`, I though it would be nice to name the files as:
- generateReviewChangeRequestsProgress
- generateReviewerResponsesProgress
- generateConsolidatorResponsesProgress

If you don't really like now, I can change back to how it was, is that what you would like?
- generateReviewChangeRequestsProgress - Just not to be confusing with other generateApplicantChangesRequestProgress
- generateReviewProgress
- generateConsolidationProgress

src/utils/hooks/useGetReviewStructureForSection/helpers.tsx
nmadruga: Ok. I can change this back. No problem

nmadruga:
Please have a look on latest changes - I have separated `generateReviewValidity` and `generateConsolidationValidity` in new files, since they will need to be also checking `changesRequest` progress, so I thought it would be tidier. Even though we haven't done the update to consolidationProgress by checking if changesRequest haven't been changed yet.

src/utils/helpers/structure/generateReviewValidity.ts
```diff
@@ -0,0 +1,41 @@
...
+
+  const incompletedUpdatesToChangesRequest = totalChangeRequests > doneChangeRequests
+  const incompletedReviews = doneNonConform === 0 && totalReviewable > doneConform
+
+  if (incompletedUpdatesToChangesRequest || incompletedReviews) {
+    const firstIncomplete = sortedPages.find(
+      ({ reviewProgress }) =>
+        reviewProgress?.totalReviewable !==
+        (reviewProgress?.doneConform || 0) + (reviewProgress?.doneNonConform || 0)
+    )
```
andreievg, line 25: I think i had it slightly wrong previously, it should have been just checking for doneConform === totalReviewable.

Now I think it needs to check (for a valid page):

* all doneConform ?
* at least one doneNonConform AND totalChangeRequest === doneChangeRequests ?

I think it needs to be it's own methods with comments, as per below

```suggestion
  const pageIsNotValid = ({ reviewProgress, changeRequestsProgress }: Page) => {
    const { doneChangeRequests = 0, totalChangeRequests = 0 } = changeRequestsProgress || {}
    // Not valid if remaining change requests
    if (doneChangeRequests !== totalChangeRequests) return true
    const { totalReviewable, doneConform } = reviewProgress || {}
    // Valid if all conform
    if (totalReviewable === doneConform) return false
    // Valid if at least one non conform
    if (doneNonConform > 0) return false
    // Otherwise invalid
    return true
  }

  if (incompletedUpdatesToChangesRequest || incompletedReviews) {
    const firstIncomplete = sortedPages.find(pageIsNotValid)

    if (!firstIncomplete) return
    else
      firstIncompleteReviewPage = {
        sectionCode: firstIncomplete.sectionCode,
        pageNumber: firstIncomplete.number,
      }
  }

```

andreievg:
> Please have a look on latest changes - I have separated generateReviewValidity and generateConsolidationValidity in new files, since they will need to be also checking changesRequest progress, so I thought it would be tidier. Even though we haven't done the update to consolidationProgress by checking if changesRequest haven't been changed yet.

All the name changes make sense now, and i like the validity checks in new methods, thanks

andreievg (approved):

PR #743, opened 2021-05-27
fix default values for password and shortText

andreievg:
need a better solution i think, and an issue, quick replication of what is fixed with this issue -> try editing user details

### DIFF
--- a/src/formElementPlugins/password/pluginConfig.json
+++ b/src/formElementPlugins/password/pluginConfig.json
@@ -4,6 +4,7 @@
   "displayName": "Secure Password Input",
   "parameterLoadingValues": {
     "label": "Loading...",
+    "default": "",
     "description": ""
   },
   "//": "categories types: 'Input'|'Informative'",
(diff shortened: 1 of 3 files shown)


CarlosNZ:
Hmm, yeah I had this problem with the Checkbox element too, and it took me a while to figure out why it was happening. I'll take a closer look this morn and see if I can apply the same or similar solution to this.

CarlosNZ:
Haven't tested it yet, but the immediate problem I can see with this solution is that it will fill all text field with empty string values on load. And we use the "text" value being undefined as a check for knowing when it's the first time focusing on the input field. 

Which will probably mean that you get immediate validation on every keystroke rather than after first leaving the field.

CarlosNZ:
> need a better solution i think, and an issue, quick replication of what is fixed with this issue -> try editing user details

Made an issue for both these, with suggested fixes: https://github.com/openmsupply/application-manager-web-app/issues/745

CarlosNZ:
> Haven't tested it yet, but the immediate problem I can see with this solution is that it will fill all text field with empty string values on load. And we use the "text" value being undefined as a check for knowing when it's the first time focusing on the input field.
> 
> Which will probably mean that you get immediate validation on every keystroke rather than after first leaving the field.

Never mind, this seems good.

PR #744, opened 2021-05-27
ignore validationInternal parameter evaluation

andreievg:
password validation stoped working because we evaluating all parameters. this pr not to be merged, but needed for demo server. @CarlosNZ please reinstate plugin config option that specifies which parameters need to be "pre" evaluated


CarlosNZ:
Yeah, I actually came across this problem with the search plugin too, as I needed the source expression to be passed down rather than evaluated in the Wrapper.  For that I just passed down a separate prop containing all parameters unevaluated. 

However, I think the best solution is, rather than specify all the possible "dynamic" parameters in config like we used to, to specify an array of parameters to NOT evaluate in the wrapper. Call it "internalParameters" or something like that, and those ones get put straight into the "simpleParamaters" object. 

So it's "opt out" rather than "opt in". (We'll also be able to get rid of that "operator" check that I added that you weren't happy with). 

Sound good?

Issue #120, opened 2020-11-13
New question plugin free text Search to match entity in a table


nmadruga:
## Requirements (suggested)

- lookup table is a evaluator expression (i.e. the work of getting the options is done by evaluator, so can be API or database query)
- User types a free-text code, is presented with matching entities from lookup table
- lookup table can be mapped values (i.e type a code, get a name), or matching name directly (it's a setting)
- Allow multiple selections?
- parameter for "minimum characters" that need to be typed before results appear (default 3?)

**NOTE**: the only entities we have in the database currently that have unique-value -> name mapping is application serial, so probably want to use this for testing. Should we also add a "Code" field to organisation for this purpose?

### Suggested implementation: 

Probably use the "Search" component: https://react.semantic-ui.com/modules/search/

CarlosNZ:
Just to clarify -- this should use the Semantic "Search" component to find and display results:
![Screen Shot 2021-05-17 at 12 46 50 PM](https://user-images.githubusercontent.com/5456533/118418821-20a45d80-b70e-11eb-9ab0-542741e7a19a.png)

https://react.semantic-ui.com/modules/search/

CarlosNZ:
Done. Looks like it wasn't linked to a PR.

Issue #745, opened 2021-05-27
Fixes for parameter evaluation: default in text fields, and pass down "internal" expressions


CarlosNZ:
As discussed:

- Here: https://github.com/openmsupply/application-manager-web-app/pull/744

and

- https://github.com/openmsupply/application-manager-web-app/pull/743

### Suggested fixes

For internal validation: 

> However, I think the best solution is, rather than specify all the possible "dynamic" parameters in config like we used to, to specify an array of parameters to NOT evaluate in the wrapper. Call it "internalParameters" or something like that, and those ones get put straight into the "simpleParamaters" object. 
> So it's "opt out" rather than "opt in". (We'll also be able to get rid of that "operator" check that I added that you weren't happy with). 

For "default" value: TBC

CarlosNZ:
Have added my solution for the first problem, and @andreievg 's existing solution into my latest [PR for Search plugin](https://github.com/openmsupply/application-manager-web-app/pull/747) (since I needed to fix the first one any way to make that work)

CarlosNZ:
This is done

PR #747, opened 2021-05-28
#120 Search (lookup) plugin

CarlosNZ:
First iteration of Search plugin -- PR in Draft for early feedback.

Requires back-end branch `#120F-search-field-plugin` from [PR #380](https://github.com/openmsupply/application-manager-server/pull/380). Go to Section3, Page3 of **FeatureShowcase** application to test.

See docs for explanation of the parameters.

### Comments

- Had the same problem mentioned by @andreievg in #744, since the search field needs to get the "source" query unevaluated. I implemented the solution I suggested in [this comment](https://github.com/openmsupply/application-manager-web-app/pull/744#issuecomment-849923924), and I think it's working well.
- Also includes @andreievg 's fix from #743 (until we come up with a better solution)
- Because API results can be any shape ranging from simple strings to complex, nested objects and arrays, it's hard to create a "display" solution to elegantly handle everything that might come along and map it nicely to the expected display fields. The solution I've implemented here will dealt with objects only 1 level deep, and expects the query result to be either an object or an array of objects. This seems to take care of most of the use cases we will deal with (that I know of), but can extend this to be more comprehensive if we decide it's important. Also, please suggest APIs to query that would make good test cases.
- I created a useDebounce hook based on [this tutorial](https://typeofnan.dev/writing-a-custom-react-usedebounce-hook-with-typescript/). I tried other debounce methods, but it's hard getting setTimeout and clearTimeout to work nicely in React (there's ways using higher-order functions and useCallback, but they were getting a bit complicated and this solution seemed quite straightforward and elegant). Let me know if there's a better approach, or if we should put this hook in our main project and just import it here.
- Renamed `jsonValue` in ApplicationViewWrapper -> onSave to just `response` so it's consistent with everything else -- been meaning to do this for a long time 😄
- The "DisplaySelection" component is basically the same as the Card Display component from ListBuilder.


nmadruga:
@CarlosNZ is this one ready for review?

CarlosNZ:
> @CarlosNZ is this one ready for review?

Yeah.

nmadruga:
The first look is nice! I really like how it display with the helpers at the right side.

Screen Shot 2021-06-01 at 9 48 08 AM

nmadruga:
2 comments about this:
Screen Shot 2021-06-01 at 9 49 28 AM

1. [For back-end] Should we change this to only search for `startsWith`? I think it makes more sense for usernames...
2. Can the popup box to select one of the results have a smaller size and show a scrollbar? Imagine if this list had 1000 items

documentation/internal/Element-Type-Specs.md
```diff
@@ -457,6 +458,56 @@ _Allows user to build a list of items, such as an **Ingredients List**_
...
+- **multiSelect**: `boolean` -- whether or not the user can select multiple items for their response (default: `false`)
+- **minCharacters**: `number` -- the minimum number of characters the user must type before the search query executes (default: 1). This is useful in situations where need the user to look up a specific item without being able to freely browse through the entire results list. For example, to look up organisation in our system using "registration" code, we set `minCharacters = 6`, so the user will need to know an exact code rather than being able to try characters one at a time.
+- **displayFormat**: `object` -- defines how to display the search results and the user's selection cards. See `displayFormat` for the [List Builder](#list-builder) (above) for detailed explanation. In this case, however, instead of a `code` substitution, the display string should contain property names from the result object. For example:
+  ```
+  displayFormat:  {
+                    title: "\${firstName} \${lastName}"
+                    description: "\${username}"
+                  }
```
nmadruga, line 492:
```suggestion
  displayFormat: {
    title: "\${firstName} \${lastName}"
    description: "\${username}"
  }
```

documentation/internal/Element-Type-Specs.md
```diff
@@ -457,6 +458,56 @@ _Allows user to build a list of items, such as an **Ingredients List**_
...
+  ```
+  If not specified, a generic "default" display will be shown, using the first 1-2 properties on the result object.
+- **resultFormat**: `object` -- same as `displayFormat`, but used when specifying a format for the "result" display that is different to the selection card display. If not specified, `resultFormat` will just be the same as `displayFormat`.
```
nmadruga, line 495: Maybe needs to say "where" does this `resultFormat` will be displayed. That is in the result list after the user adds a valid input to search, right?

semantic/src/site/modules/search.overrides
```diff
@@ -1,3 +1,7 @@
 /*******************************
          Site Overrides
 *******************************/
+
+// .ui.search .prompt {
+//   border-radius: @promptBorderRadius !important;
+// }
```
nmadruga, line 7: Maybe this can be removed?

src/formElementPlugins/ApplicationViewWrapper.tsx
```diff
@@ -100,19 +102,19 @@ const ApplicationViewWrapper: React.FC<ApplicationViewWrapperProps> = (props) =>
...
         await onSaveUpdateMethod({
           variables: {
             id: currentResponse?.id as number,
-            value: jsonValue,
+            value: response,
             isValid: validationResult.isValid,
           },
         })
-      if (jsonValue === null)
+      if (response === null || response?.text == undefined)
```
nmadruga, line 117:
```suggestion
      if (response === null || !response?.text)
```

I think it's more descriptive... Or `!!response?.text` in case `text` can also be null

src/formElementPlugins/ApplicationViewWrapper.tsx
```diff
@@ -250,12 +252,14 @@ const getDefaultIndex = (defaultOption: string | number, options: string[]) => {
...
-  parameterLoadingValues: any
+  parameterLoadingValues: any,
+  internalParameters: string[]
 ) => {
   const simpleParameters: any = {}
   const parameterExpressions: any = {}
   for (const [key, value] of Object.entries(parameters)) {
-    if (value instanceof Object && !Array.isArray(value) && 'operator' in value) {
+    if (internalParameters.includes(key)) simpleParameters[key] = value
+    else if (value instanceof Object && !Array.isArray(value)) {
```
nmadruga, line 262: I wonder what is this new `internalParameters` doing here. Is it worth adding a quick comment to explain this?

nmadruga:
Another style suggestion... sorry! Would be nicer in my opinion to show the result inside the input box. What do you think?
Like this:
Screen Shot 2021-06-01 at 10 23 23 AM
Screen Shot 2021-06-01 at 10 23 37 AM

Instead of this:
Screen Shot 2021-06-01 at 9 53 35 AM

Although this example from [semantic-ui](https://react.semantic-ui.com/modules/dropdown/#types-multiple-selection) only shows the dropdown without a input entered, so not sure how it would look for also allowing the user to enter new text for another search.

src/formElementPlugins/search/src/SummaryView.tsx
```diff
@@ -0,0 +1,18 @@
...
+    selection: response?.selection,
+    displayFormat,
+    Markdown,
+    isEditable: false,
+  }
+
+  return <DisplaySelection {...displayProps} />
+}
+
+export default SummaryView
```
nmadruga, line 18: Something strange happening in the Summary view. 

After entering all required fields in a "Features showcase" application, and having added a organisation.registration in the form - Page 3. I got this error:
Screen Shot 2021-06-01 at 10 29 06 AM

Showing it's in the Search plugin:
Screen Shot 2021-06-01 at 10 29 17 AM

And the summary view is showing this error for Search plugin (organisation):
Screen Shot 2021-06-01 at 10 29 45 AM

src/formElementPlugins/search/src/useDebounce.ts
```diff
@@ -0,0 +1,20 @@
...
+  useEffect(() => {
+    const debounce = setTimeout(() => {
+      setDebounceOutput(debounceInput)
+    }, delay)
+    return () => {
+      clearTimeout(debounce)
+    }
+  }, [debounceInput])
+
+  return [debounceOutput, setDebounceInput]
```
nmadruga, line 19: There seems to be a problem with that - when using UC API and searching for the 2nd drug, it doesn't stop trying to fetch data and show is "loading" for ever: as pointed out [here](https://github.com/openmsupply/application-manager-server/pull/380#discussion_r642689412)

nmadruga (changes requested):
My first review so far - hoping to have another go today so we can finish up this one and use in other templates. So much fun to use. It's really good overall!

CarlosNZ:
> * [For back-end] Should we change this to only search for `startsWith`? I think it makes more sense for usernames...

In a real situation, for sure. But I was just trying to create a demo that would get several results, and "startsWith" was too limited. I don't think it matters just for the showcase.
 
> * [For back-end] I think `firstName` and `lastName` should have a space between

There is -- you're just not seeing any users with lastNames in your results there  😉

(I think only me and "John Smith" have surnames in our database

> * Can the popup box to select one of the results have a smaller size and show a scrollbar? Imagine if this list had 1000 items

I was kind of thinking that it would be up to the query constructor to manage result sizes, like putting a "first: 25" parameter in or something. Currently a long result list just flows off the page out of sight. I kind of don't mind this, as the user would know to just type more characters to narrow it. 

But, yes a scrollable list would be a bit prettier. I'll have a look and see if it's do-able without major changes to the Semantic styles (it's currently just out-of-the-box layout). Will get back to you.

CarlosNZ:
> I'll have a look and see if it's do-able without major changes to the Semantic styles (it's currently just out-of-the-box layout). Will get back to you

Yay, fortunately it was pretty easy and looks good 👍🏼 ddafc17a

Screen Shot 2021-06-01 at 11 38 50 AM

Thanks for the suggestion

CarlosNZ:
> Another style suggestion... sorry! Would be nicer in my opinion to show the result inside the input box. What do you think?

That would be fine for results that are a single line and can fit in the input box, but there's no reason why they should be (none of my examples are, nor any of the [Semantic examples](https://react.semantic-ui.com/modules/search/).

It makes sense for a Dropdown, as the input box is the same size as the results elements, but that's not the case for Search.

I agree that the format of the selected elements could be improved, but I was planning to probably just copy whatever we end up doing for the list builder. The card display here is pretty much identical to the list builder card display, so when we update that (based on Deb's design suggestions) we can update this one too (if it works for this).

documentation/internal/Element-Type-Specs.md
CarlosNZ: Okay, I added a couple of screenshots to clarify: d643f1e2

semantic/src/site/modules/search.overrides
CarlosNZ: Done  (as part of ddafc17)

src/formElementPlugins/ApplicationViewWrapper.tsx
CarlosNZ: We have to be careful with this, it's caught us out before. It's checking _specifically_ for undefined in text, not anything that could be falsy (e.g. `""`).  I think your suggestion wouldn't distinguish between undefined or "".
Screen Shot 2021-06-01 at 12 07 26 PM

src/formElementPlugins/ApplicationViewWrapper.tsx
CarlosNZ: Yeah, this is the problem Andrei and I were discussing [here](https://github.com/openmsupply/application-manager-web-app/pull/744), and I made an issue for it: #745.

What I've done here implements my suggested solution, but I'm not 100% sure that'll be what we go with -- I'll wait till @andreievg weighs in, or we can discuss it IRL tomorrow.

src/formElementPlugins/search/src/SummaryView.tsx
CarlosNZ: Yup, I know what this it -- the Summary view doesn't have any fallback for "default" display when `displayFormat`is not defined. I'll need to add this.

CarlosNZ: It's also the same problem for the List builder. Will get onto it, thanks for spotting.

CarlosNZ: > It's also the same problem for the List builder. Will get onto it, thanks for spotting.

Actually, it wasn't quite the same problem, for list builder, but similar. Basically, SummaryView failed if `listItems`, `selection` or `displayFormat` wasn't defined. ApplicationView had defaults specified for these, so I've just added the same defaults to SummaryView.

323bd501

src/formElementPlugins/search/src/useDebounce.ts
CarlosNZ: Great catch, this one -- it only happens if you enter the same search term immediately after selecting a result.

Turns out if the input is the same, the useEffect for `debounceOutput` never runs as the value of `debounceOutput` doesn't change.

Just needed to reset `debounceInput` after the search finishes.   43d00cfd

src/formElementPlugins/ApplicationViewWrapper.tsx
nmadruga: ok. I was considering that "" wouldn't also be a valid response.

CarlosNZ: Yeah, if you delete your response in a Text input, the value goes back to `""`, not undefined -- it never goes undefined again.

nmadruga (approved):

nmadruga:
Very good one! Ready to go : )

src/formElementPlugins/search/src/useDebounce.ts
nmadruga: Nice : ) 👍

PR #748, opened 2021-05-28
#731 Checkbox negative text

### LINKED ISSUE #731: Improvement: Add a "negative" text value for Checkbox plugin
Currently when a checkbox is unselected, the resulting text value in Summary view (and elsewhere) is just the text _`<Nothing Selected>`_. This is fine in Summary view, but it looks pretty bad in the Ingredients list, say.

We currently specify a text value for when the checkbox is selected (e.g. "Yes"), but we should also be able to specify a "No" value

CarlosNZ:
Fixes #731 

Tiny bit of sugar to allow checkbox values to appear more elegantly when not selected -- specify a negative text value as well as a "selected" value.

Makes this ugly thing:
Screen Shot 2021-05-28 at 4 42 53 PM

Into this:
Screen Shot 2021-05-28 at 4 42 00 PM

New parameter is `textNegative`, and it defaults to "Nothing Selected" as above.

Use back-end branch `#731-checkbox-negative-text-value` (PR ??) to test (only applied to the checkbox on the ListBuilder, all others should be unchanged)


nmadruga (approved):
LGTM

PR #746, opened 2021-05-28
#741 create locked review when assignment is locked

nmadruga:
Fixes #741 

### Back-end branch PR
[#741F](https://github.com/openmsupply/application-manager-server/pull/378)

### Changes
- Show modal when review has status LOCKED when trying to submit:

- When review was locked before started (`reviewAssignment.isLocked`) a new action (done on back-end) setup for ON_REVIEW_CREATED will change the status of this review to LOCKED, and in the front end we show like this:

Review page doesn't have "Submission area":
Screen Shot 2021-05-28 at 12 22 50 PM

On the Review home, there isn't a link to view the Review (probably we should show a **View** option now):
Screen Shot 2021-05-28 at 12 22 32 PM

### Test
~~- First you need to run `yarn database_init` on the back-end branch
- Login as **Consolidator2** and go to the review page for new Application serial 45678: http://localhost:3000/application/45678/review
- Open in an incognito tab, login with user **Reviewer2** and also open the review page: http://localhost:3000/application/45678/review
- Click `start` and then approve all or decline one to get a decision read to submit. Leave like this
- Go to the other instance (as Consolidator2) and submit the LOQ. This will lock Reviewer2 review.
- Go to Reviewer2 now and try to submit... Should show the modal~~
~~- PENDING: Noticed a problem with the current implementation to get the latest ReviewStatus, so need to work on a fix for that.~~ Won't fix now

Another test - you will need to restart the database....

- Login as Consolidator 2, and submit review of http://localhost:3000/application/45678/review as LOQ
- Login as Reviewer 2, start the review and check that it is already locked: No submission area available

Issue #721, opened 2021-05-19
Padding change in summary view


andreievg:
Seems like there were some minor style changes done in current master, and styles look different now:

I used "0370ff680b3a11ec8b080b32272bf61ea2c8d68e" on web app and "7dfb3ea28ca36d95152f8e8aa518d6aaa17c035b" on server (although compilation errors in web app, but still works)

Previous:

![Screen Shot 2021-05-19 at 12 21 21 PM](https://user-images.githubusercontent.com/26194949/118739508-c2a68000-b89d-11eb-9099-5d2552dd8237.png)

Now:

![Screen Shot 2021-05-19 at 12 12 11 PM](https://user-images.githubusercontent.com/26194949/118739550-d356f600-b89d-11eb-9b34-406a1030059f.png)

Issue #646, opened 2021-04-29
Update layout and use global styles: Navigation Area


CarlosNZ:
Basic part of this done, but we think we'd like to move the "Prev/Next" buttons up to the Application form area (as per designs).

Also, waiting on Designer suggestions for what to do with the bar when it's wide. Full-width of the window looks a bit much, and truncating the sides also doesn't look great. The current solution (drop-shadow around all) is also not ideal.

Issue #308, opened 2021-02-03
Improve message after review is submitted and update review page on reload to not display the submission button again


nmadruga:

nmadruga:
I was thinking to work on this at some point - just display a snackbar on the top of the page after the Review is submitted with a message like "Review submitted with success!". Although I'm not sure when to display this, it should be on every time the Review view the `ReviewPage`, so not sure how to keep this state of **just** submitted. Any ideas @andreievg @CarlosNZ ?

andreievg:
> I was thinking to work on this at some point - just display a snackbar on the top of the page after the Review is submitted with a message like "Review submitted with success!". Although I'm not sure when to display this, it should be on every time the Review view the ReviewPage, so not sure how to keep this state of just submitted. Any ideas @andreievg @CarlosNZ ?

If this is successful: https://github.com/openmsupply/application-manager-web-app/blob/c10b9b91eac1574419d92c16f001553a5aa12c32/src/containers/Review/ReviewSubmit.tsx#L118

CarlosNZ:
> I was thinking to work on this at some point - just display a snackbar on the top of the page after the Review is submitted with a message like "Review submitted with success!". Although I'm not sure when to display this, it should be on every time the Review view the `ReviewPage`, so not sure how to keep this state of **just** submitted. Any ideas @andreievg @CarlosNZ ?

Should we just mimic the Application submission process and have a seperate `/submission` page when submitting review? In fact, you _might_ even be able to re-use the ApplicationSubmission component (rename it just "Submission") and pass in an `isReview` prop to render a different message (but the main layout would be the same).

nmadruga:
> If this is successful: https://github.com/openmsupply/application-manager-web-app/blob/c10b9b91eac1574419d92c16f001553a5aa12c32/src/containers/Review/ReviewSubmit.tsx#L118

That's a good solution. But we still want to display at the top of the `ReviewPage` container, so I think something as a state to be retrieved from ReviewSubmit?

>  Should we just mimic the Application submission process and have a seperate /submission page when submitting review? In fact, you might even be able to re-use the ApplicationSubmission component (rename it just "Submission") and pass in an isReview prop to render a different message (but the main layout would be the same).

[...]

PR #753, opened 2021-05-31
#708 consolidator changes request and updates by reviewer in review page

### LINKED ISSUE #708: Show on Review page: consolidator comments, agreements and disagreements and Review options to update change requests

nmadruga:
Fixes #708 

### Based on PR
[#646-#721-and-other-style-changes](#747 
With changes merged from PR [#741](#746) 

### Back-end branch
Using linked back-end of [#741F](#741F create locked review when assignment is locked)

### Changes
- Please ignore files `src/containers/Review/ReviewSubmit.tsx` and `src/utils/messages.ts` which are changes from #741 
- Added  element of **consolidation response** of change requested to lower level on `ReviewApplicantResponse`
- After the change is done, so now `isNewReviewResponse` is defined use the `shouldDim` to grey out the Consolidator changes request to previous Reviewer review response.

### Screenshots
- Also display on the Review Home the number of changes to be updated by Reviewer:
Screen Shot 2021-05-31 at 1 00 47 PM
- Although on the list is still showing only **Update** - let me know what do you think:

Screen Shot 2021-05-31 at 1 00 39 PM

- On the Review Page show the label **Update** on each Review with changes requested not updated:
Screen Shot 2021-05-31 at 1 01 04 PM

- After is updated will show the Consolidation greyed out and the Edit icon (as others Review responses with decision):
Screen Shot 2021-05-31 at 1 40 07 PM

- Something that I don't like is that after submitting the updated review, the Consolidator is showing (related to previous reveiwResponse) and that's misleading:
Screen Shot 2021-05-31 at 1 40 20 PM
Some solutions to that:
- Show the **Updated** icon on the reviews with new ReviewResponse
- Don't show the Consolidator below the ReviewResponse after it's updated (just the Review history icon)
- Show the previous ReviewResponse on the Consolidation response box 

### Tests 
- Use Application serial: 45678 and logged in as **Consolidator2**
- Change some of the decision to "Disagree" and submit a Changes Request to **Reviewer1**
- Login as Reviewer1 and see what is to be updated, do the update and resubmit.
- You can also see the **Consolidator2** received updated changes on their Review page - no other indication other that a button to **Review** showing on the interface though.


andreievg:
> Something that I don't like is that after submitting the updated review, the Consolidator is showing (related to previous reveiwResponse) and that's misleading:

Maybe we should hide consolidator decision and comment when not draft status ?

src/components/PageElements/Elements/ReviewResponseElement.tsx
```diff
@@ -1,39 +1,41 @@
...
-        {!shouldHideDecision && (
-          <ElementDecisionLabel isConsolidation={isConsolidation} reviewResponse={reviewResponse} />
-        )}
-        {!reviewResponse.comment ? null : (
-          <div className="comment-container">
-            <Icon name="comment alternate outline" color="grey" />
-            <p className="secondary">{reviewResponse.comment}</p>
-          </div>
-        )}
+        <ElementDecisionLabel
```
andreievg, line 30: hmm was wondering why we would add an extra prop and extra changes to change from shouldHideDecision to isDecisionVisible ?

src/components/PageElements/Elements/ReviewApplicantResponse.tsx
```diff
@@ -10,24 +10,40 @@ interface ReviewApplicantResponseProps {
...
 }
 
 const ReviewApplicantResponse: React.FC<ReviewApplicantResponseProps> = ({
   applicationResponse,
   summaryViewProps,
   reviewResponse,
+  previousReviewResponse,
   isNewApplicationResponse,
   isActiveReviewResponse,
+  isNewReviewResponse = false,
```
andreievg, line 28: I think we should show 'updated' whenever response is updated (when isChanged === true), currently isChanged would only be set during change request phase (so this should work correctly and shouldn't interfere with re-reviews).

It looks like isNewReviewResponse =  isChangeRequest && isChanged, i think it should just be = isChanged.

src/components/PageElements/Elements/ReviewApplicantResponse.tsx
```diff
@@ -49,16 +65,26 @@ const ReviewApplicantResponse: React.FC<ReviewApplicantResponseProps> = ({
...
           >
-            {isActiveReviewResponse && <UpdateIcon onClick={showModal} />}
+            {getReviewDecisionOption()}
           </ReviewResponseElement>
           {/* div below forced border on review response to be square */}
           <div />
+          {/* Consolidation Response */}
+          {consolidationReviewResponse && (
+            <ReviewResponseElement
+              isCurrentReview={false}
```
andreievg, line 75: would be good to indicate in review response that it's been updated `<UpdatedLabel />`

src/components/PageElements/Elements/ReviewApplicantResponse.tsx
```diff
@@ -49,16 +65,26 @@ const ReviewApplicantResponse: React.FC<ReviewApplicantResponseProps> = ({
...
+              isConsolidation={true}
+              reviewResponse={consolidationReviewResponse}
+              shouldDim={isNewReviewResponse}
+            ></ReviewResponseElement>
```
andreievg, line 79:
```suggestion
              shouldDim={isNewReviewResponse}/>
```

src/components/PageElements/Elements/ReviewResponseElement.tsx
```diff
@@ -1,39 +1,41 @@
...
+          isConsolidation={isConsolidation}
+          reviewResponse={reviewResponse}
+          isDecisionVisible={isDecisionVisible}
+        />
+        <div className="comment-container">
+          {canReviewerEdit ? <Icon name="comment alternate outline" color="grey" /> : null}
```
andreievg, line 36: Hmm, the only place I could find in Miro where the comment icon is hidden is for application change request that have not been updated. So i don't quite understand this, bit of logic

src/components/PageElements/Elements/ReviewResponseElement.tsx
andreievg: The only difference I can see is showing date of review. But it's not actually shown in the summary view of change requests for applicant (based on Miro)

src/containers/Review/ReviewSubmit.tsx
```diff
@@ -92,23 +92,19 @@ const ReviewSubmitButton: React.FC<ReviewSubmitProps & ReviewSubmitButtonProps>
   }
 
   const showPendingReviewWarning = () => {
-    const { title, message, option } = messages.REVIEW_STATUS_PENDING
```
andreievg, line 95: I am a bit confuse about this changes in this file

src/components/PageElements/PageElements.tsx
```diff
@@ -195,26 +198,28 @@ const PageElements: React.FC<PageElementProps> = ({
...
+                isNewReviewResponse: isChangeRequest && isChanged,
+                showModal: () => updateQuery({ openResponse: element.code }),
+                summaryViewProps: summaryViewProps,
+              }
+
               return (
                 <RenderElementWrapper key={element.code}>
                   {isConsolidation ? (
                     <ConsolidateReviewDecision
-                      applicationResponse={latestApplicationResponse}
```
andreievg, line 202: not required changed

hmm ConsolidationReviewResponse doesn't have isNewReviewResponse as part of it's props, I guess it doesn't hurt passing it through, it does show why destructuring props is considered a be a bad practice. (i.e. if we remove from ReviewApplicationResponse, there wouldn't be any indication that it's unused)

src/components/Review/ReviewSectionRowAction.tsx
```diff
@@ -92,11 +95,19 @@ const StartContinueOrRestartButton: React.FC<ReviewSectionComponentProps> = ({
...
   const getButtonName = () => {
-    const reReviewCount = reReviewableCount(reviewProgress)
-    if (reReviewCount > 0) return `${strings.BUTTON_REVIEW_RE_REVIEW} (${reReviewCount})`
     switch (action) {
-      case ReviewAction.canUpdate:
-        return strings.ACTION_UPDATE
+      case ReviewAction.canUpdate: {
+        const changeRequestsCount = changesRequestedCount(changeRequestsProgress)
+        return strings.ACTION_UPDATE.concat(
+          changeRequestsCount > 0 ? ` (${changeRequestsCount})` : ''
```
andreievg, line 102: this looks neater then full on string template literals 👍

andreievg (changes requested):
Overall looks pretty good. A few comments, main change request and clarifications are:

* show `updated` on top of edit button (if updated)
* confusing about hiding comment icon
* pending review modal changes

src/components/PageElements/Elements/ReviewApplicantResponse.tsx
nmadruga: Hmmm I'll add this to the list of things to check with Debbie. It seems this makes more sense from the Applicant's perspective, since we show only in the Summary page at last so they can see all changes. In the Review page, you they can see all their current reviews with the "edit" icon or Update button - when changes are required. Not sure if would be better to display something extra about changes done in this review... Let's see.

src/components/PageElements/Elements/ReviewResponseElement.tsx
nmadruga: I think the date is a nice one to have - like is used in some other places that aren't part of designs either.
Also changed because thought the code looks simples to read, by moving this logic to be inside.

src/components/PageElements/PageElements.tsx
nmadruga: Yes, I get your point. Not sure how it would be the best practice for this either. I wonder if the `interface` defined in each component isn't enough indication of what is used in there. I was annoyed about how many same props both of them had...

src/components/PageElements/Elements/ReviewApplicantResponse.tsx
andreievg: Ok, i thought it would be the same as change request to applicant (show updated indicator when it's been updated, with edit button).

src/containers/Review/ReviewSubmit.tsx
nmadruga: Yes, sorry that was a mistake. But didn't remember to revert. Reverted now: f67f7e3

An remaining change to cover the case - when trying to submit and review status is changed to LOCKED is added as part of showing different messages of successful or failed submission on issue #308.

src/components/PageElements/Elements/ReviewApplicantResponse.tsx
nmadruga: Good catch : ) 6c35a54

src/components/PageElements/Elements/ReviewResponseElement.tsx
nmadruga: Yes, that was not correct. I updated now, was just trying to keep the same logic as before (after removing the statement around all) but now revert to only check for reviewResponse.comment is there and show icon: 9d70064

src/components/PageElements/PageElements.tsx
andreievg: fair point, i am not sure how to deal with this either (apart from having shared props as one object, but that could be cumbersome). I guess it should be ok as long as we can see clearly what is being destructured (as it is in this case)

andreievg (approved):
thanks for the changes, looks good, but I still think we should show 'updated' and the edit icon if isChanged

nmadruga:
Create a second PR to merge this one, after having problems with a conflict resolution when merging master in. Will merge second PR straight away.

PR #754, opened 2021-05-31
#706 Inline review

### LINKED ISSUE #706: Inline decision - only shows when click Review and has a Submit button to collapse
Check out this [comment](https://github.com/openmsupply/application-manager-web-app/pull/695#discussion_r630914944) to move  the updateQuery to be a hook

CarlosNZ:
Fixes #706 

Implemented as per latest designs, except added a "Cancel" button to exit without having to enter a review. 

Working for initial review and consolidation -- please test both.

### Summary of Changes

- Moved reviewing functionality out of DecisionArea and into new component ReviewInlineInput
- Renamed DecisionArea to HistoryPanel (it's just a placeholder for now until we implement History functionality)
- a new state variable `isActiveEdit` in ConsolidateReviewDecision and ReviewApplicantResponse (we should probably rename those to something more consistent with each other) that tracks whether to display the inline review component, or the current review/consolidation response.

I'm wondering about the layout. I don't really like the large vertical space it takes up, esp when there's so much excess space on the right. We've widened the form area from Deb's designs, which is why the TextArea doesn't fill the space (you can't really have a text input more than about 50-70 characters wide). I agree that it looks a bit cluttered if I push the text area to the right, but it might be good to think about other options, or ask Deb?


nmadruga (approved):
Looks good. Thanks!

I agree with getting back to Debbie about what you have explained in the description there. Can you keep this in a note on the board of things to get her feedback by the end of this week, please?

Issue #751, opened 2021-05-30
Option on ShortText plugin to have maxWidth


nmadruga:
Default should be kept to show in full width, but if the new field `maxWidth` is defined show as defined in the question plugin of shortText.
e.g.:  Quantity (number) field in the design.

As per in latest designs:

![Screen Shot 2021-05-31 at 11 20 36 AM](https://user-images.githubusercontent.com/16461988/120123717-96c3bc80-c204-11eb-933d-6b623c6a2938.png)

CarlosNZ:
Fixed with #755

PR #755, opened 2021-06-01
#750 #751 Radio button and shortText updates

### LINKED ISSUE #750: Option on RadioChoice plugin to display inline
When there are less options to select by the user, the UI looks cleaner with radio vertically aligned  in one line as per latest designs - for List builder:
Screen Shot 2021-05-31 at 11 16 09 AM

CarlosNZ:
Fixes #750 #751

Requires back-end branch `#750F-radio-buttons-inline` ([PR #382](https://github.com/openmsupply/application-manager-server/pull/382)) -- have only implemented for elements that are part of the Ingredients list (Section 3, page 2, FeatureShowcase) -- to bring UI closer to the designs for the Ingredients list

### Summary of changes

- Add "layout: inline" parameter to radioChoice and implement custom style based on this selection
- Add "maxWidth" parameter to shortText and implement custom style based on this value
- Since this solution is more elegant than what was used in "Checkbox" element, I've updated checkbox to match this pattern, too.

### DIFF
--- a/src/formElementPlugins/checkbox/src/ApplicationView.tsx
+++ b/src/formElementPlugins/checkbox/src/ApplicationView.tsx
@@ -57,16 +57,23 @@ const ApplicationView: React.FC<ApplicationViewProps> = ({
     setCheckboxElements(checkboxElements.map((cb, i) => (i === index ? changedCheckbox : cb)))
   }
 
+  const styles =
+    layout === 'inline'
+      ? {
+          display: 'inline',
+          marginRight: 10,
+        }
+      : {}
+
   return (
     <>
       <label>
         <Markdown text={label} semanticComponent="noParagraph" />
       </label>
       <Markdown text={description} />
-      {checkboxElements.map((cb: Checkbox, index: number) => {
-        return layout === 'inline' ? (
+      {checkboxElements.map((cb: Checkbox, index: number) => (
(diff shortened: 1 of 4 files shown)


src/formElementPlugins/checkbox/src/ApplicationView.tsx
```diff
@@ -57,36 +57,32 @@ const ApplicationView: React.FC<ApplicationViewProps> = ({
...
   return (
     <>
       <label>
         <Markdown text={label} semanticComponent="noParagraph" />
       </label>
       <Markdown text={description} />
-      {checkboxElements.map((cb: Checkbox, index: number) => {
-        return layout === 'inline' ? (
+      {checkboxElements.map((cb: Checkbox, index: number) => (
+        <Form.Field key={`${index}_${cb.label}`} disabled={!isEditable} style={styles}>
```
nmadruga, line 75: We can't be using the LESS classes here because it's a plugin, right?

nmadruga (approved):
Just one concern. But the changes overall looks good. Simple and nice. I like Semantic.ui for sure!

src/formElementPlugins/checkbox/src/ApplicationView.tsx
CarlosNZ: We could, but it makes sense to plugin styles contained within their bundle.

Issue #534, opened 2021-03-26
Epic #76 - Expose tables to public


andreievg:
## Description

An outcome of application can be exposes to public eye (noRegistered user).
This issue is for configuration (should be as simple as adding a new policy and giving access to public user) and also make sure UI aspect works (can just add a policy type called view)

## Documentation
- [ ] TODO: Add as part of [Epic #38](338) Public docs & tutorials

### Sprint 35:
- [x] #376 outcome displays

PR #758, opened 2021-06-02
#376 outcome displays

andreievg:
Fixes back end: https://github.com/openmsupply/application-manager-server/issues/376

Required for back end PR: https://github.com/openmsupply/application-manager-server/pull/392
Requires back end branch: #376-custom-fields-and-tables-outcomes

### Changes
* New routes `/outcomes`, `/outcomes/:code`,  `/outcomes/:code/:id` 
* New components to display outcomes list, outcome table and outcome details
* New hook that handles most of the functional requirements of the components
   * Puts outcome display query results into consumable form
   * Generate dynamic queries and getters for them

### Testing
* Create new user
* Apply for organisation
* Review organisation with admin user (approve all)
* Navigate to `/outcomes` (explore)

TODO
* link to epic: https://github.com/openmsupply/application-manager-web-app/issues/534
* create documentation issue


CarlosNZ (commented):
Looks good. I think we need some design improvements obviously, but that can wait.

My main comment is the same one I added to the back-end - I think the additional columns added to the tables that are purely for displaying the responses in the Outcomes table should be bundled together in one JSON object. It's very weird having "username" column AND a "username_value" column, and same for logo on organisation. The "response" values could all just go in one "display_info" or "response_data" column -- makes it much tidier and easier to understand the structure of the tables.

andreievg:
> My main comment is the same one I added to the back-end - I think the additional columns added to the tables that are purely for displaying the responses in the Outcomes table should be bundled together in one JSON object. It's very weird having "username" column AND a "username_value" column, and same for logo on organisation. The "response" values could all just go in one "display_info" or "response_data" column -- makes it much tidier and easier to understand the structure of the tables.

I think only one column should be used for a particular value

> "username_value"

that's an artefact of testing, not needed

> and same for logo on organisation

Can probably use image display on logo_url in the display outcomes config, or we could have used response value of image plugin in organisation table (I think I was pushing for that at the start)

> makes it much tidier and easier to understand the structure of the tables.

It's tidier when looking at the table (without data) but messier when looking at the jsonb field, dont' see how it's easier to understand (apart from indicating what are the 'core' schema fields and which ones have been dynamically added).

CarlosNZ:
> It's tidier when looking at the table (without data) but messier when looking at the jsonb field, dont' see how it's easier to understand (apart from indicating what are the 'core' schema fields and which ones have been dynamically added).

I'm not saying _all_ the additional fields should be in a single field, just the ones that are only there for storing responses for showing in the Outcomes tables. To me, it's very confusing having two fields that are basically the same, but with different versions of the same info. I think it makes sense to keep all the fields that show actual data relating to the entity as separate fields, but storing the responses is essentially a kind of "metadata", and I find it weird that they're mixed in freely with the "real" entity data without any clear indication of what they're for.

Am I correct in assuming that these "response" fields are only there for providing structured data for the SummaryView in the Outcomes tables? If so, they're only ever going to queried as one "thing" -- "response data", and then iterated over for display. We're not querying them as seperate fields at all, are they? You could probably then do away with one of the outcome display tables, as the "response data" object would already contain all the info about the column names in their field keys (maybe?)

andreievg:
@CarlosNZ 
> I'm not saying all the additional fields should be in a single field, just the ones that are only there for storing responses for showing in the Outcomes tables. To me, it's very confusing having two fields that are basically the same, but with different versions of the same info. I think it makes sense to keep all the fields that show actual data relating to the entity as separate fields, but storing the responses is essentially a kind of "metadata", and I find it weird that they're mixed in freely with the "real" entity data without any clear indication of what they're for.

> Am I correct in assuming that these "response" fields are only there for providing structured data for the SummaryView in the Outcomes tables? If so, they're only ever going to queried as one "thing" -- "response data", and then iterated over for display. We're not querying them as seperate fields at all, are they? You could probably then do away with one of the outcome display tables, as the "response data" object would already contain all the info about the column names in their field keys (maybe?)

I think they will be used to pre-populate default values, and possibly queried by external API.

For example user job history (list builder), might need to be changeable.

### On another note: 

Also that's another reason why default values should really be handled in wrapper. (I am still leaning toward simplifying plugin responsibility, I can see use cases where 'things' need to be done within plugin when all parameters are evaluated, which is very hard to do with existing way of parameter evaluation. Still pushing toward previous way of handling parameters, this would also allow default to be evaluated and 'saved' by the wrapper, and response parameter passed to applicationView to be changed, without any extra logic in plugin to handle this)

It's also possible to do evaluation of parameters inside structure building (similar to visibility). This would mean one loading indicator for the whole page, which might mitigate your concerns of UI with previous way of evaluating parameters. It would be reasonable easy to expose that evaluation to be used inside list builder and outcomes.

Lastly there is another issue i just saw that's related to 'Loading..', i think (check out a summary page of submitted Organisation registration, in network tab you'll see an api call to 'loading').

CarlosNZ:
> Also that's another reason why default values should really be handled in wrapper. (I am still leaning toward simplifying plugin responsibility, I can see use cases where 'things' need to be done within plugin when all parameters are evaluated, which is very hard to do with existing way of parameter evaluation. Still pushing toward previous way of handling parameters, this would also allow default to be evaluated and 'saved' by the wrapper, and response parameter passed to applicationView to be changed, without any extra logic in plugin to handle this)

Yeah, I take your point.

> It's also possible to do evaluation of parameters inside structure building (similar to visibility). This would mean one loading indicator for the whole page, which might mitigate your concerns of UI with previous way of evaluating parameters. It would be reasonable easy to expose that evaluation to be used inside list builder and outcomes.

Yeah, not sure if its easier (for list builder) to have the parameters evaluated in Wrapper, or with Structure. Probably need to try doing it with Structure and seeing if it causes any problems.

> Lastly there is another issue i just saw that's related to 'Loading..', i think (check out a summary page of submitted Organisation registration, in network tab you'll see an api call to 'loading').

👍

CarlosNZ:
Looking good, except I'm getting this error on some fields in the front-end:
Screen Shot 2021-06-04 at 9 41 17 AM
Is it possible you need to update the Outcome details/columns table, as I think it might be defining columns that no longer exist (with your latest changes)?

andreievg:
@CarlosNZ 

> Looking good, except I'm getting this error on some fields in the front-end:

Hmm, it's because the join table wasn't created yet, should work after creating a user. Should I remove the error message ? Similar thing would happen if user field is not created yet. Ideally we would have a seperate configuration for 'outcome tables' where fields are added manually and join table would be created, but modifyRecord seems the fastest way to do this (give our time constraints)

CarlosNZ:
> Hmm, it's because the join table wasn't created yet, should work after creating a user. Should I remove the error message ? Similar thing would happen if user field is not created yet. Ideally we would have a seperate configuration for 'outcome tables' where fields are added manually and join table would be created, but modifyRecord seems the fastest way to do this (give our time constraints)

Ah okay, I see. Yeah, if there's some straightforward way to suppress the error message in this case, that'd be great. But not a big deal if it's too complicated.

andreievg:
> Ah okay, I see. Yeah, if there's some straightforward way to suppress the error message in this case, that'd be great. But not a big deal if it's too complicated.

Should it just be for the join or also for details and table ? (as all three are susceptible to graphQL error)

CarlosNZ:
> Should it just be for the join or also for details and table ? (as all three are susceptible to graphQL error)

If it's due to user/admin config error, then sure it should probably show error. But if it's just due to the circumstance you described (i.e. haven't created a user yet so join table doesn't exist), then it seems find to suppress it.

But not too fussed, I think we can update it once we've used it for a while and get a feel for what's important.

PR #765, opened 2021-06-07
Parameter evaluations in structure

andreievg:
https://github.com/openmsupply/application-manager-web-app/pull/758#issuecomment-854186075

@CarlosNZ this is what it would be like (draft PR for now), it seems slower because for every change we re-evaluate all parameters (not just current page parameters), and since we do lots of api calls in feature showcase template, there is a delay (for example when changing from card to table view). I think that's ok though (I dont' see us using lookups to api to change evaluated parameters frequently, most api calls will use 'search' plugin.

Also we can add an optimisation to specify which questions codes evaluations for form element instance relies on.

### Changes

* Evaluate dynamic parameters in structure building
* Remove loading states
* Two parameters passed to plug in `evaluatedParameters` and `parametersExpressions`, `parmetersExpressions` is used in search and password plugin for unevaluated parameters
* Refactor list builder to account for changes (now it should work pretty much the same as general structure, with isVisible, isRequired, isEditable -> evaluatable, check out with back end branch `evaluation-in-structure-test`)

`note:` both front end and back end test branch rely on new version of evaluator, can add it from file for now as per testing in 761 pr

### DIFF
--- a/src/formElementPlugins/ApplicationViewWrapper.tsx
+++ b/src/formElementPlugins/ApplicationViewWrapper.tsx
@@ -2,16 +2,9 @@ import React, { useEffect, useState } from 'react'
 import { ErrorBoundary, pluginProvider } from '.'
 import { ApplicationViewWrapperProps, PluginComponents, ValidationState } from './types'
 import { useUpdateResponseMutation } from '../utils/generated/graphql'
-import {
-  EvaluatorNode,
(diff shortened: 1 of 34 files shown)


andreievg:
closing PR but keeping branch for reference ( @CarlosNZ can still be used for evaluation visibility etc for listBuilder )

Issue #544, opened 2021-03-28
Epic # - Demo server


andreievg:
* Info page for the project (.io styles) and demo server

nmadruga:
@andreievg  I think you have nailed this one already. Maybe just need one issue to deal with the cretion of .io info page and the rest one is pretty much done, right? Can we move it to the "Epic (waiting public docs)" column?

PR #771, opened 2021-06-09
Build updates

andreievg:
Build updates from previous demo server and optimisations.

Managed to bring overall build down to just under 2mb (would be a just under 500kb gzipped), quite happy with that.

### Changes

* Simplified webpack slightly by moving common things to common
* Added minimisation for both css and js
* Removed chunking for vendor
* Add alias in webpack and paths in tsconfig to redirect semantic-ui-react to src/semanticUIReactExport, which import each component individually (to avoid all importing all components)
* reduce semantic ui less build by removing brand icons with variable override and by only importing 'ttf' icon files
* config file is now .ts to allow changing server and graphql to relative routes
* use file loader for image loading


build-utils/webpack.common.js
```diff
@@ -1,66 +1,60 @@
...
-    }
+        test: /\.(png|svg|jpg|jpeg|gif)$/i,
+        loader: 'file-loader',
+        options: {
+          name: '[path][name].[ext]',
+        },
+      },
+
+      {
+        test: /\.less$/,
```
CarlosNZ, line 40:
```suggestion
        test: /\.(less|css)$/,
```
I was playing around with another plugin that had its own .css file, and it was giving a compile error until I explicitly added a css handler in webpack config. So would probably be handy to include this here now.

nmadruga:
Just so we use the same jobs in IRIMS - I will be adding my time to review this as part of [Epic#86](https://github.com/openmsupply/application-manager-web-app/issues/544)

build-utils/webpack.prod.js
```diff
@@ -1,31 +1,30 @@
-const commonPaths = require('./common-paths')
-const webpack = require('webpack')
-const MiniCssExtractPlugin = require('mini-css-extract-plugin')
+const CssMinimizerPlugin = require('css-minimizer-webpack-plugin')
+const BundleAnalyzerPlugin = require('webpack-bundle-analyzer').BundleAnalyzerPlugin
```
nmadruga, line 2: Is this one used?

src/semanticUiReactReExport.ts
```diff
@@ -0,0 +1,195 @@
...
+  StrictTableHeaderCellProps,
+} from 'semantic-ui-react/dist/commonjs/collections/Table/TableHeaderCell'
+
+export {
+  default as Checkbox,
+  CheckboxProps,
+  StrictCheckboxProps,
+} from 'semantic-ui-react/dist/commonjs/modules/Checkbox'
+
+export { SemanticCOLORS, SemanticICONS } from 'semantic-ui-react/src/generic'
```
nmadruga, line 195: Trying to understand what is this file for and where it's used. In `ApplicationViewWrapper` it seems to be importing straight from `semantic-ui-react`: `import { SemanticICONS } from 'semantic-ui-react'`

tsconfig.json
```diff
@@ -12,13 +12,17 @@
     "preserveConstEnums": true /* Do not erase const enum declarations in generated code. */,
     "removeComments": true /* Do not emit comments to output. */,
 
+    "baseUrl": "src" /* paths are relative to baseUrl */, 
+    "paths": {
+      "semantic-ui-react": ["semanticUiReactReExport"]
+    },
```
nmadruga, line 18: Ah! I see, that's nice. So basically just a simplification of this path? Or is this something else?

nmadruga (approved):

nmadruga:
> Add alias in webpack and paths in tsconfig to redirect semantic-ui-react to src/semanticUIReactExport, which import each component individually (to avoid all importing all components)

Ah! Didn't read that first. That's great. Probably most of the sizing was due to this then?

andreievg:
> Ah! Didn't read that first. That's great. Probably most of the sizing was due to this then?

It didn't really make that much difference. I'll keep it for now as it is also useful for listing what components we are using.

build-utils/webpack.common.js
andreievg: bf834f3b4f4250dc30d44a26d0b6635398ebe7e6

build-utils/webpack.prod.js
andreievg: yep, it's quite useful when analysing build size, but it's commented out below

Issue #717, opened 2021-05-18
Bug: Trying to click "review and submit" button after going to summary and then back with Edit the button stop not working


nmadruga:
Seems to occur on some templates, not sure exactly which, but UserEdit and OrgRego have been shown to be affected

steps:
1. create new EditUser template, without changing the password. And go to the summary page
2. Click on edit the option to use the password, going back to page 1. Don't change anything
3. Try clicking on "Review and submit" doesn't work
4. Try just focusing on any field and try to clicking the button again. It works... 
5. Mistery 🤔

CarlosNZ:
Is this with the latest changes in #701 (and back-end complement) ?  Cos I can't seem to reproduce it.

CarlosNZ:
Ah I see it now. However, it's not just UserEdit, I can create the same situation in RegisterOrg (and probably others). I'll update the issue title to reflect this

CarlosNZ:
It's also affecting the "Next" button, but only on "linear" restricted templates.

I notice that I can click "Next", nothing happens, then as soon as I change one of the responses, the page jumps to the next page, without the button needing to be clicked again.

I can see that it's getting stuck waiting for the "requestRevalidation" method to complete. I think the logic around `shouldProcessValidation` and `lastRevalidationRequest` must be getting messed up somehow.

CarlosNZ:
@andreievg -- maybe you want to have a look into this one, as I think it might be to do with the "FormElementUpdateTrackerState" module. Not sure, but it's becoming quite a noticeable problem with the new Laos templates we're working on.

PR #772, opened 2021-06-09
make sure validation doesn't break when plugin saves the same value

andreievg:
Fixes: #717 

will make an issue tomorrow for this:

replication
* create org rego
* fill the fist page
* go to second page
* go back to first page
* try to go to next page

made validation a little bit more resilient (didn't realise plugins would be changing values, even if it's not change, and on subsequent loads)

### DIFF
--- a/src/contexts/FormElementUpdateTrackerState.tsx
+++ b/src/contexts/FormElementUpdateTrackerState.tsx
@@ -2,7 +2,7 @@ import React, { createContext, useContext, useReducer } from 'react'
 
 interface ContextFormElementUpdateTrackerState {
   elementEnteredTimestamp: number
-  elementEnteredTextValue: string
+  elementEnteredTextValue?: string
   elementUpdatedTimestamp: number
   elementUpdatedTextValue: string
   isLastElementUpdateProcessed: boolean
@@ -17,6 +17,7 @@ export type UpdateAction =
   | {
       type: 'setElementUpdated'
       textValue: string
+      previousValue: string
     }
 
 type FormElementUpdateTrackerProps = { children: React.ReactNode }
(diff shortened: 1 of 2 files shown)


CarlosNZ:
Is this the same issue described in #717

src/contexts/FormElementUpdateTrackerState.tsx
```diff
@@ -35,11 +36,17 @@ const reducer = (
         elementUpdatedTimestamp: Date.now(),
         elementUpdatedTextValue: action.textValue,
       }
+
+      const notEntryState = typeof newState.elementEnteredTextValue === 'undefined'
+      let wasElementChanged = action.textValue !== newState.elementEnteredTextValue
+      if (notEntryState) wasElementChanged = action.textValue !== action.previousValue
```
CarlosNZ, line 42: Can this not just be:
```
const wasElementChanged =
        typeof newState.elementEnteredTextValue === 'undefined'
          ? action.textValue !== action.previousValue
          : action.textValue !== newState.elementEnteredTextValue
```

Personally I find it easier to read, because the name "notEntryState" is kind of meaningless to me. So I have to parse its contents anyway, whereas the ternary structure at least formats it in a readable shape. Just a personal preference, no biggie.

CarlosNZ (approved):
Cool, looks like this solves #717, so good to get that cleared up. Thanks.

Just one tiny suggestion, nothing major.

src/contexts/FormElementUpdateTrackerState.tsx
andreievg: I agree: 3a241b6e76a272941409fa5faab75e5e1f0bd0da

andreievg:
Thanks for review

PR #778, opened 2021-06-10
#550 Date picker plugin

### LINKED ISSUE #550: Template element plugin -- Date Picker
Since Semantic-UI doesn't have a DatePicker component, suggest using this package: https://www.npmjs.com/package/react-semantic-ui-datepickers

CarlosNZ:
Fixes #550 

Just wanted to put this here for initial feedback.
Screen Shot 2021-06-10 at 3 16 30 PM

Requires back-end branch `date-picker-plugin` to see it in action (**Feature Showcase**, one on the first page (date of birth), and more examples in Section 3, page 4) 

Uses [react-semantic-ui-datepickers](https://www.npmjs.com/package/react-semantic-ui-datepickers) package -- will need a fresh `yarn install` to get the package (as well as an updated version of Luxon). Most of the parameters and functionality available here is dictated by the functionality of this component.

Implementation is fairly straightforward, the main complication being that we have to convert between three different types of Date:

- The database requires dates as ISO strings (in JSON response object)
- the [DatePicker component ](https://www.npmjs.com/package/react-semantic-ui-datepickers) uses Javascript `Date` objects (and switches between either `Date` or `Date[]` depending on whether single date or range)
- date calculations and saving ISO dates without timestamp (date only) is most easily done with Luxon

### Parameters:
- `allowRange` (boolean) -- specifies if the selected Date is a single Date or a start/end date (it's built into the component) (default: false)
- `locale` (e.g. "en-US") -- easy, built-in Date localisation (default:  "en-US")
- `displayFormat` -- a range of options for displaying the date in Text format (and stored in `text` field) (default: "short")
- `entryFormat` -- format to enter when typing in manually (as opposed to using the widget selector) e.g. "YYYY-MM-DD", and what shows the "textInput" field after entry/selection (default: "YYYY-MM-DD")
- `minDate`/`maxDate` -- to specify how far into the past/future the selector should go
- `minAge` / `maxAge` -- same as above, but a number relative to the _current_ date (for example, if an applicant was required to be over 18 years old, the date-of-birth picker wouldn't allow dates newer than 18 years before now). Note: This number currently specifies "years" -- do you think it needs to be more fine grained than that?  (default -1000/1000 i.e. effectively no restriction)
- `firstDayOfWeek` (default "Sunday")
- `default` -- specified as ISO string

Obviously most of the above have defaults, so won't need to be specified most of the time.

### Response shape:
```
{
  text: <text format of date (range) as specified in "displayFormat">
  date: {
     start: <ISO YYYY-MM-DD date>
     end?:  <ISO YYYY-MM-DD date>
  }
```
Note, if it's a single date, only `start` will be specified. If `end` is present, we know it's a range


andreievg (approved):
@CarlosNZ at a quick glance, this looks good to me, approving (I think a quick glance is sufficient since code changes are directed at one new plugin), but pending some info in the docs (or an issue to do the docs), cheers.

andreievg:
https://github.com/openmsupply/application-manager-server/issues/463

and

https://github.com/openmsupply/application-manager-web-app/issues/849

PR #779, opened 2021-06-10
#773 Option to hide "Add item" button in listBuilder

### LINKED ISSUE #773: Should hide the "Add item" on listBuilder when isEditable is false - so used for visualisation only

CarlosNZ:
Fixes #773.

Adds an additional (optional) parameter `displayOnly` which (if `true`) completely hides the "Add item" button for adding new ingredients (as opposed to just disabling it when `isEditable = false`). This is so a previous listBuilder response can be displayed in a form (as defaultValue) without there being any indication of editability by the user.

### DIFF
--- a/documentation/internal/Element-Type-Specs.md
+++ b/documentation/internal/Element-Type-Specs.md
@@ -401,6 +401,8 @@ _Allows user to build a list of items, such as an **Ingredients List**_
   - `visibility_condition` / `is_editable` -- not required, always `true` (for now, may be implemented later)
   - `parameters`, `title`, and `code` are essential
 
+- **displayOnly** `boolean` (default `false`) -- if `true` the "Add item" button will be hidden completely (as opposed to just disabled when `isEditable = true`)
+
 - **displayType** `string` (must be either `table` or `cards` (default)) -- how to present the list of items, as shown here:
 
   - **table** view:
--- a/src/formElementPlugins/listBuilder/src/ApplicationView.tsx
+++ b/src/formElementPlugins/listBuilder/src/ApplicationView.tsx
@@ -53,6 +53,7 @@ const ApplicationView: React.FC<ApplicationViewProps> = ({
     inputFields,
     displayFormat = getDefaultDisplayFormat(inputFields),
     displayType = DisplayType.CARDS,
+    displayOnly = false,
   } = parameters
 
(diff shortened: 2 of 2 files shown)


andreievg:
I think we should just have an option on element to always display a summary view (together with making it an information category type)

CarlosNZ:
> I think we should just have an option on element to always display a summary view (together with making it an information category type)

You mean for any kind of element? Yeah I guess that makes sense. 

So, a new field on template_element? Or do we just set any element to Information type which forces a Summary display?

andreievg:
> So, a new field on template_element? Or do we just set any element to Information type which forces a Summary display?

I think that's a great idea

CarlosNZ:
> > So, a new field on template_element? Or do we just set any element to Information type which forces a Summary display?
> 
> I think that's a great idea

Hmmm, this is not super straightforward, since "INFORMATION" type elements don't get responses created, and so the "default" value we want to load won't be stored as the response for this element. I'm reluctant to change the logic around this at this point cos I think it could cause more problems than it solves. What do you think?

I agree that it might be handy to have an option to have elements displayed in SummaryView in application forms, but I think changing the type from QUESTION to INFORMATION is not the best way to do it at this point. 

How about we infer it from "isEditable" and "defaultValue"? -- if `isEditable === false` AND `defaultValue` is NOT NULL, then show Summary View?

CarlosNZ:
> How about we infer it from "isEditable" and "defaultValue"? -- if `isEditable === false` AND `defaultValue` is NOT NULL, then show Summary View?

I made a separate branch where I did this -- see what you think: `#773-2-listBuilder-hide-button-option`

CarlosNZ:
Okay, since @andreievg  has implemented a version of this (allowing "SummaryView" when elements set to "INFORMATION -- now in master), we can probably close this. Just leaving it open for now as a reminder to update the templates that need to do this.

andreievg:
> Okay, since @andreievg has implemented a version of this (allowing "SummaryView" when elements set to "INFORMATION -- now in master), we can probably close this. Just leaving it open for now as a reminder to update the templates that need to do this.

I think the only template that's doing this is the licence one and it should be used there now, ok to close ?

CarlosNZ:
> I think the only template that's doing this is the licence one and it should be used there now, ok to close ?

Sweet. Closing now.

PR #791, opened 2021-06-15
#777 Fix for empty file upload responses

### LINKED ISSUE #777: File upload not required generating reviews for empty responses - on initial application submission and on re-submission  updated from changes-required
Using PR for new company registration. After assigned to Admin and started review, some file upload responses had reviews created!

Screen Shot 2021-06-10 at 12 58 35 PM

Have done a LOQ back to the applicant (Declined the only uploaded document) and after changes were done by applicant - on Re-review had even more questions to be Reviewed - that didn't have any file uploaded.

Expected: When no file is uploaded no review should be requested to be done.

CarlosNZ:
Fixes #777 

In the `useEffect` watching the "uploadedFiles" array, we now check if the length is 0, in which case we (re)set the saved response to `null`.

### DIFF
--- a/src/formElementPlugins/fileUpload/src/ApplicationView.tsx
+++ b/src/formElementPlugins/fileUpload/src/ApplicationView.tsx
@@ -59,6 +59,11 @@ const ApplicationView: React.FC<ApplicationViewProps> = ({
   const fileInputRef = useRef<any>(null)
 
   useEffect(() => {
+    // Set response to null if no files
+    if (uploadedFiles.length === 0) {
+      onSave(null)
+      return
+    }
     // Only store files that aren't error or loading
     const fileDataToSave = uploadedFiles
       .filter(({ loading, error, fileData }) => !(loading || error) && fileData)


nmadruga:
The code seems alright, but while testing this I noticed the problem still happens after a changes-request with LOQ to the applicant. Steps to reproduce this problem:
- Login as **Valerio**, and create a new "Join Organisation" application. Leave the last option File upload untouched.
- After application is submitted, login as **Assigner1** and assign it to **Nicole** or **Carl**. 
- Login as the reviewer selected:
   - Check that the last question doesn't have a **Review** since there is no file uploaded
   - Now set Non-conformity the other file uploaded required. Submit the review with LOQ to the Applicant.
- Login as **Valerio** again and update the file upload as requested - keeping the last one without a file still. Re-submit
- Now there is something wrong with the Dashboard for reviewer - for me, it isn't showing the application re-submitted, but if you simply go to the application review page you will see the option to **Re-review**.
- When you start the re-review you can see the problem now: the **Review** button is showing for the question without a file uploaded, so the problem seems to be fixed but only partially : (
Screen Shot 2021-06-22 at 11 41 50 AM

CarlosNZ:
> The code seems alright, but while testing this I noticed the problem still happens after a changes-request with LOQ to the applicant.

Good catch, that's a strange one. Inspecting the database, I can see that the responses are NULL as expected, and they are being pruned correctly after each submission. So there must be a problem with the logic that checks whether to display "Review".

I'll have a quick look at it now and see if I can see what it is. I suspect that it isn't just File Uploads with this problem (since the responses are behaving correctly now), in which case it would probably be a new issue.

Anyway, I'll have a quick look and see if it something straighforward and test it with other element types. I'll make a new issue if need be.

CarlosNZ:
> Anyway, I'll have a quick look and see if it something straighforward and test it with other element types.

Yeah, as I suspected, this problem is not just for FileUpload plugins -- I added another question to the template and the same problem occurs:
Screen Shot 2021-06-22 at 3 19 18 PM

CarlosNZ:
Have made a new issue for that outstanding problem: #804 

Hopefully not too complicated for someone who understands the display logic here.

nmadruga (approved):
: ) Thanks for having a look in the other bug, I think it's clear enough. I'll just add as part of Epic 85 and tag as to be done in this Sprint.

nmadruga:
Oh sorry, haven't seen you already took care of doing that! : ) Cheers

PR #792, opened 2021-06-16
#705 show history button

### LINKED ISSUE #705: Display icon for Review history
### Design
From Miro "Consolidator Reviewer" (13/06):
Screen Shot 2021-06-16 at 3 05 14 PM

## Where to show it?
Ideally it should only appear when there are previous reviews/applicant responses/consolidations done. From changes done on PR #790 there are only a maximum of 3 elements displaying for each question part of either: 
1. Applicant summary page
2. Review page
3. Consolidator page

Below all existing elements belonging to the same questions, would show the "Show History" at last child.

#### Applicant Summary
- has previous applicant response

#### Review Page
- have received changes-request + has done new review for **current** applicant response, or
- has previous review done by you for a **previous** applicant response

#### Consolidation Page
- has previous consolidation done by you to lower level review, or
- has previous consolidation done by you to applicant, or
- have received changes-request + has done new review for **current** applicant response

### Questions
- Should "Show history" appear on submitted reviews?

nmadruga:
Fixes #705 

### Changes
- Show "View history" in Review page if there are previous reviews existing
- Added new `ApplicantElementWrapper` to keep all related component that display in each question on Application summary page.
- Show "View history" in Application summary page when there is previous application responses


semantic/src/site/elements/label.overrides
```diff
@@ -30,6 +30,11 @@
   padding-left: 0;
 }
 
+.ui.label.link-label {
+  .ui.label.simple-label();
+  color: @interactiveHigh;
+}
```
CarlosNZ, line 36: Is `.link-label` one of our classes, or part of Semantic? If the latter, then are you sure this won't effect any other labels we're using?

semantic/src/site/globals/site.overrides
```diff
@@ -862,6 +862,17 @@ a:hover {
   }
 }
 
+.history-container {
```
CarlosNZ, line 865: Just wondering why this needs its own class -- the container looks exactly the same as the main response-container.

src/components/PageElements/ViewHistoryButton.tsx
```diff
@@ -0,0 +1,17 @@
+import React from 'react'
+import { Label } from 'semantic-ui-react'
+import strings from '../../utils/constants'
+
+const ViewHistoryButton: React.FC = () => {
+  return (
+    <div className="history-container">
+      <Label
+        className="link-label clickable"
+        icon="history"
```
CarlosNZ, line 10: Just my preference, but I think the icon could be a bit bigger.

CarlosNZ (commented):
Looks good, and seems to be working well. I just had a few little comments, nothing significant.

semantic/src/site/elements/label.overrides
nmadruga: It's our classes, used to display the Show History as a link to open the HistoryPanel modal

semantic/src/site/globals/site.overrides
nmadruga: The main difference is that it doesn't have the `flex-row` because elements display in a horizontalBox.

src/components/PageElements/ViewHistoryButton.tsx
nmadruga: No problem I'll use the same size as other icons

nmadruga: Done

CarlosNZ (approved):

Issue #793, opened 2021-06-16
formatjs localisation demo


sworup:
A demo would be created to show how format js could be used in our web app.

PR #794, opened 2021-06-16
#793 formatjs localisation demo

sworup:


sworup:
@andreievg the demo as you have requested.

- The plugin would not play nice with `awesome-typescript-loader` so I replaced it with `ts-loader`.
- `translations/en.json` is generated by the `yarn extract` script I have added in `package.json`. The idea is that this file will be descributed to translators and they will deliver `fr`, `de`, `nl` etc. language translation json files. They are all collected and added in this translations folder. `yarn compile` will be run on it to generate the final json files (with proper ids) in `src/intl/compiled-lang/` directory. 
- The `translations` directory is supposed to be excluded from git. We only commit `src/intl/compiled-lang/` in git.
- We might have to adjust the directory and write some webpack script to get the compiled translations to production build.

nmadruga:
Hi @sworup sorry it took so long to review this, we had a few rushing weeks around here. Could you work on fixing these conflicts please so I can have a look on this soon?

sworup:
Hi @nmadruga, this is a demo @andreievg asked me to build to see how easily we can formulate localization into our system with formatjs. This has a single demo page (with no logic) page that has been localized. This should not affect any code that had been written earlier, except for `awesome-typescript-loader` being replaced by `ts-loader` that is proposed with this PR.

andreievg:
@sworup, can we please organise a chat with you and IRIMS team, tomorrow or Thursday, at 4pm our time ?

sworup:
> @sworup, can we please organise a chat with you and IRIMS team, tomorrow or Thursday, at 4pm our time ?

@andreievg sure... tomorrow is fine. Can you please set a calendar event invite for me so that I could be sure of the time. Seems it would be 9:45AM my time.

andreievg:
@sworup 9:45am sounds good, I'll send you a zoom link closer to the time.

nmadruga:
Hi @sworup just wanted to have an idea on how is the work going for changes discussed about this PR please.
We would like to have the localisation module in place very soon so we can demonstrate to countries interested in using the App in another language besides English. 
After the demo have some of the proposed ideas on our meeting last week we would still need to update the whole App to be using the new Localisation module before some translations can be done for a presentation. So, let me know if you have another PR coming soon please. Also if you found any existing open source tools that can be used to help out when making translations to a new Language based on formatJs it would be great for us to have a look so we start making a plan to use that.

sworup:
> Hi @sworup just wanted to have an idea on how is the work going for changes discussed about this PR please.
> We would like to have the localisation module in place very soon so we can demonstrate to countries interested in using the App in another language besides English.
> After the demo have some of the proposed ideas on our meeting last week we would still need to update the whole App to be using the new Localisation module before some translations can be done for a presentation. So, let me know if you have another PR coming soon please. Also if you found any existing open source tools that can be used to help out when making translations to a new Language based on formatJs it would be great for us to have a look so we start making a plan to use that.

@nmadruga I had been busy with another issue so have not been able to do a through research so I was waiting till then.
Currently there is no PR and if made any changes I would make them to this PR, which would mostly be the demo changes @andreievg asked me to research on. e.g. demo the purpose of description field.

I did find a resource that las the list of localisation resources, some of them include the translation services which do seem promising but I have not been able to use the services and test:
https://github.com/mbiesiad/awesome-translations

nmadruga:
@sworup I don't know the priority of other work you are doing currently but if not as urgent, could you prioritise this PR please? It's very important to have this on the first iteration with the clients that are interested so we can show how the system is highly configurable to be used in any language and how to setup with the App - that's probably going to be our next next. So maybe, first you can have a look on the changes required to this and after do the research of the tool. Sounds good? 

Can you also update the issue description with a summary of changes please? 

If you think that using to transform all existing strings to be using the new Localisation module should be in a separated issue, that's fine. We would just like to have the functionality closed so we can move to the next steps on this module.

nmadruga:
Just a summary of changes required to this PR - based on what we talked a couple of weeks ago:
1) We would like to have a wrapper added to this structure. 
2) Call to the wrapper on components should be simplified so its cleaner and possible to reuse strings.
3) Some mechanism to insert new localised strings. Ideally all the localised strings are constructed in the same place. 
4) I'm still not sure I see how it would generate different files for each language. Would it just generate one and we would rename and add all translations to other languages? How about we specify (in a config file) a key for each language and have a script to just generate one file for each language and it can be easy to just add to the folder of translated files? 

... Something to add @andreievg @CarlosNZ ?

sworup:
@andreievg @nmadruga 
Answer to @andreievg's question, whether or not we can have multiple instances of text having the same default messages but separate descriptions. (If allowed this would enable cases where the same English word might have different translations based on context).

Answer: Yes, you can

https://user-images.githubusercontent.com/683452/128148158-9343b7cd-69a6-42ca-ae5e-df812bf8700b.mp4

sworup:
- Google translate API is not free, it is paid service.

- This tool seems promising... support format.js and does the translation (from google, microsoft etc. translation service) automatically. But is not free:
  - https://www.codeandweb.com/babeledit

sworup:
@andreievg @nmadruga 
Example of auto-translation happening with BabelEdit. And an example of the whole extraction and compilation workflow. Hope this might help with @nmadruga's confusion about how the translation files are handled. Even without BabelEdit, manual steps should be pretty much the same.

https://user-images.githubusercontent.com/683452/128476953-ab3f41f7-e091-4400-99d7-235a8bb946f8.mp4

sworup:
> Just a summary of changes required to this PR - based on what we talked a couple of weeks ago:
> 
> 1. We would like to have a wrapper added to this structure.

- I am thinking of just creating an alias. Kinda like this

  ```js
  
  // Create alias
  const intl = useIntl()
  const translate = intl.formatMessage
  
  // use case inside render
  <Header as="h1" content={translate(messages.registered_products)} />
  
  ```

> 2. Call to the wrapper on components should be simplified so its cleaner and possible to reuse strings.

- Create a wrapper as above, add it to helper file or add it to the context or something... just using as above itself is quite simple.

> 3. Some mechanism to insert new localised strings. Ideally, all the localised strings are constructed in the same place.

- We can do that with `defineMessages` method.
  Could be like this

  ```js
  / / messages.ts
  import { defineMessages } from 'react-intl'

  const ProductListDemoTranslations = defineMessages({
    registered_products: {
      description: 'Header 1', 
      defaultMessage: 'Registered Products', 
    },
    demo_manual_id_00222: {
      id: 'manual-id-00222', // Example of custom ID
      description: 'Manual id test',
      defaultMessage: 'Manual id test',
    },
  })
  
  export default ProductListDemoTranslations
  ```
- We can create separate files for different modules or have them together.
- We can import this file into our component file and use it as `messages.demo_manual_id_00222`. Please see example.

> 4. I'm still not sure I see how it would generate different files for each language. Would it just generate one and we would rename and add all translations to other languages? 

- The `extract` command will create a `translations/en.json` file. This file will have entries for each translation in code, added with `intl.formatMessage` method. It will give IDs to the entries that had no id in code. Every time `extract` is called from terminal, the command will look into the code for `intl.formatMessage`, defineMessages, defineMessage etc. instances to list out all of the messages requiring translation.
- Only the `en.json` file is extracted as of my code. 

> How about we specify (in a config file) a key for each language and have a script to just generate one file for each language and it can be easy to just add to the folder of translated files?

  - `yarn extract` is actually a small script's command written in package.json to extract English, so you can use the script that is in package.json and modify it to extract other language files too but all of them will have English text. 
    - Translation would have to be done on those files.
    - Also running the extraction script for let's say french file, while fr.json already exists in the directory with some translations, would override all of the translations to English so **doing this is not desirable**

#### The most desirable flow is 

- Add translation method calls in code with `intl.formatMessage`.
- Use `yarn extract` to extract all of these translation call's default messages to an `en.json` file.
- Send that `en.json` file to be translated.... could be a person, we could manually translate them one by one or use some online service... google translate API is also an option (it is also not free though) to get them done altogether.
- Take the translated file, say `fr.json` and add it to the same directory as the `en.json` file.
- Run `yarn compile`.  This will take all of the files in translation folder, strip out unnecessary data (used as context for translator) and just leaves a `ID : message string` pair, and put the files in translated folder.
- This folder is read by the development app... and production app... so we have to write some webpack script to transfer these files to build folder (not done in this example PR yet).

sworup:
About @andreievg's question about whether we can add translations in server API, the answer is we can. 

The way I see us doing this is:
- We can extract and compile the JSON files here with the normal workflow. 
- And upload that JSON file to the server.
- A API endpoint is created to call for these translations.
- In our code, 
    - For development, we use the local json files
    - For production, we do the API call then add the returned JSON into the messages object.

Another way would be to extract the en.json file. And do the rest of the steps sever side. Format.js is not limited to react. 

The more I think about it I do not see a good use-case for this unless we write a microservice to do these translations on the fly whenever the en.json file changes during development.

nmadruga:
Closing out-dated

PR #410, opened 2021-03-05
380 review page

andreievg:
Branched from #367 
Also has #386 merged into it

Fixes: #380 
Fixes: #349 

Due to the two above it's probably best to wait for some merging before reviewing (quite a large diff otherwise)

### Change

Will sum it up when diff reduces

PR #795, opened 2021-06-16
#409B Link company registration application changes (plus misc tweaks)

CarlosNZ:
Front-end tweaks to support [back-end PR #410](https://github.com/openmsupply/application-manager-server/pull/410)

### Summary of changes

- added a cheeky hack so we can quickly submit applications without filling in the whole thing -- when filling an application go to `.../<serial>/summary?dev=true` and you'll be able to view summary and submit application even with incomplete elements
  - TO-DO: add condition to not allow this in production (I couldn't figure out how to do this)
  - updated `useQuerySectionActivation` to accomodate this, so it uses "updateQuery" instead of manually re-writing the query strings (it should be doing this anyway)
  - need to be a bit careful with this -- viewing the invalid applications will go into a weird loop after they've been approved -- this is purely for convenience when needing to quickly submit an application
- hyperlinks in "TextInfo" default to open in new tab (but there is an optional parameter to suppress this if ever needed)
  - updated docs with this additional parameter, as well as removing deprecated references to "dynamic parameters" (now that any parameter can be dynamic)
- when SummaryView elements are displayed on Application pages (i.e. when set to "INFORMATION" type), I thought they were looking a bit tightly packed together. Added and extra prop `extraSpacing` to `RenderElementWrapper` that adds additional space to bottom of each -- applied only in this one situation


andreievg:
>TO-DO: add condition to not allow this in production (I couldn't figure out how to do this) 

https://github.com/openmsupply/application-manager-web-app/blob/e403051f238626e81a9605464757140d812ac9ea/src/config.ts#L4

semantic/src/site/globals/site.overrides
```diff
@@ -857,6 +857,11 @@ a:hover {
   }
 }
 
+// when SummaryView elements used on Application pages
+.summary-extra-spacing {
```
andreievg, line 861: can we use .form-element with child selector for this?

.form-area {
...
  .summary-page-element-container {
    margin-bottom: 20px;
 }
}

maybe at the same time add variable for both element-application-view and this one for margin

CarlosNZ:
> > TO-DO: add condition to not allow this in production (I couldn't figure out how to do this)
> 
> https://github.com/openmsupply/application-manager-web-app/blob/e403051f238626e81a9605464757140d812ac9ea/src/config.ts#L4

Okay, cool thanks. I just added this directly to ApplicationSummary, since you haven't merged your "config.ts" file into master yet. Will update later. 0bab4840

semantic/src/site/globals/site.overrides
CarlosNZ: Yeah good point. This what you mean? 121b923b

andreievg:
> Okay, cool thanks. I just added this directly to ApplicationSummary, since you haven't merged your "config.ts" file into master yet. Will update later. 0bab484

ok, it doesn't' really need the config.ts btw, if webpack mode is set to production, it would replace rocess.env.NODE_ENV with 'production'

andreievg (approved):

PR #797, opened 2021-06-17
#585 Organisation selector

### LINKED ISSUE #585: Org display in Header needs to have a selector to switch orgs
Should also have a "No org" option in the list.

CarlosNZ:
Fixes #585 

Adds a drop-down to Organisation in header to easily switch logged-in org.

Had to make a small change to back-end `/login-org` endpoint to allow it to accept "no organisation" without giving error message (it just returns userInfo with no org defined, like when you initially log in with no org selected). [Back-end PR #411](https://github.com/openmsupply/application-manager-server/pull/411)

UserState now keeps the "orgList" as part of its state, so it can be accessed by the selector.

One problem: The Dashboard and List don't reliably refresh with the new template permissions when changing (and when changing user in the Dev menu too). It seems like those components are not always re-rendering in response to a change in `templatePermissions`, but not sure why, and I don't understand why the small change I did here would have changed the behaviour of the dev menu. Keeping this PR in Draft until we figure this out -- suggestions welcome.


CarlosNZ:
@andreievg -- can you have a quick look at the problem I've described above? Do you think it's a separate issue, or is there a quick fix for it I can do here?

src/containers/User/UserArea.tsx
```diff
@@ -54,16 +56,42 @@ const MainMenuBar: React.FC = () => {
...
 }
 
-const OrgSelector: React.FC<{ user: User }> = ({ user }) => {
-  // TO-DO: Make into Dropdown so Org can be selected
+const OrgSelector: React.FC<{ user: User; orgs: OrganisationSimple[]; onLogin: Function }> = ({
+  user,
+  orgs,
+  onLogin,
+}) => {
+  const LOGIN_AS_NO_ORG = 0 // Ensures server returns no organisation
```
andreievg, line 64: not required change, and not part of this PR, but I like the idea of using context state vs passing through params (`useUserState` here i mean)

src/containers/User/UserArea.tsx
```diff
@@ -54,16 +56,42 @@ const MainMenuBar: React.FC = () => {
...
+
+  const JWT = localStorage.getItem('persistJWT') as string
+
+  const handleChange = async (_: SyntheticEvent, { value: orgId }: any) => {
+    await attemptLoginOrg({ orgId, JWT, onLoginOrgSuccess })
+  }
+  const onLoginOrgSuccess = async ({ user, orgList, templatePermissions, JWT }: LoginPayload) => {
+    await onLogin(JWT, user, templatePermissions, orgList)
```
andreievg, line 72: maybe too much abstraction ?

andreievg (approved):
One suggestion is to change from generic `Function` type to:https://github.com/openmsupply/application-manager-web-app/blob/d140f22a8e2bbdbd6c64ac045290d1b1942610e8/src/contexts/UserState.tsx#L79

Not required change but would possibly help with finding where onLogin is used with correct params.

>@andreievg -- can you have a quick look at the problem I've described above? Do you think it's a separate issue, or is there a quick fix for it I can do here?

It looks like it's not refreshing only when there are 0 templates available for user/org, which i fixed in template builder branch 


https://github.com/openmsupply/application-manager-web-app/commit/a8cbdcb832b5c3632934e7592bc0dd8bb6465850#diff-d9985f23958c1bc55c434545cafda68d6b8e8746882baa403eaf214e9d70882f

CarlosNZ:
> It looks like it's not refreshing only when there are 0 templates available for user/org, which i fixed in template builder branch

Okay, great. Yeah, it switches okay when there are templates available, so it should be sweet once we merge your branch 👍

src/containers/User/UserArea.tsx
CarlosNZ: I think it was just to keep parallel structure with what we already did with `onLoginSuccess`.

Issue #477, opened 2021-03-21
Show all the history on for all the previous responses and review (from consolidator and reviewer)


nmadruga:
Just to show the list as in:
<img width="498" alt="Screen Shot 2021-05-06 at 8 38 29 PM" src="https://user-images.githubusercontent.com/16461988/117269505-6eb69700-aeac-11eb-88a2-7fafd97b32de.png">

Can be done on a separated issues, or as part of this one: selecting one review from the list:
<img width="592" alt="Screen Shot 2021-05-06 at 8 39 23 PM" src="https://user-images.githubusercontent.com/16461988/117269468-665e5c00-aeac-11eb-8909-81a107ef64dc.png">

This is how it would be displaying the current - in preview mode (to be done in [Epic #33 - Application to PDF templates](#333)):
<img width="578" alt="Screen Shot 2021-05-06 at 8 39 04 PM" src="https://user-images.githubusercontent.com/16461988/117269453-63636b80-aeac-11eb-9c63-6fcc95edd25f.png">

nmadruga:
This one has been updated to display the reviewHistory on it's own modal instead of previous decision modal (which is dead). 
Get all the reviewer responses and applicant responses displaying with existing components from review/summary view in order bottom to top including the current ones. No decision can be made in here and has a close button.

nmadruga:
Trying to use the following designs (from "Consolidator and review" on 13/06) to display in the History panel:
<img width="408" alt="Screen Shot 2021-06-21 at 1 42 54 PM" src="https://user-images.githubusercontent.com/16461988/122696298-e68f2400-d296-11eb-96d6-7d3c7c2cbf3a.png">

I think the latest version for history that we have it's a little more complex than we would like to do now:
<img width="710" alt="Screen Shot 2021-06-21 at 1 43 35 PM" src="https://user-images.githubusercontent.com/16461988/122696295-e2fb9d00-d296-11eb-8e81-2656ea8af00c.png">

PR #798, opened 2021-06-18
#477 - part 1 - Refactoring structure stage date on Application, review and assignment

nmadruga:
Partial fix for #477 

The idea is to have the `stageDate` available for Review and Application when showing the history, but just prefer to break-down this PR which is doing a small refactor so it doesn't mix with specific changes to show the History modal next (part 2) 

### Linked PRs
[#477F-part1](https://github.com/openmsupply/application-manager-server/pull/412)

### Changes
- A lot of files with only variable changes....
- Changes on types was to unify a little how we keep stages information:
   - Removed type `StageAndStatus`
   - Replaced  `StageAndStatus` with `ApplicationStage` in FullStructure of application
   - Replace `Review.stage` from `ApplicationStage` to new type `ReviewStage` and move status to inside of that
   - Extends `ReviewStage` and `ApplicationStage` from same type `StageDetails`
- Renamed in `FullStructure` (for Application) field `current` to `currentStage` it keeps both dates `currentDate` for this stage and `statusCreatedDate` for the current `status` property.
- Fix on `ReviewSectionRowLastActionDate` to show correct date related to the status (instead of stage as it was before)


src/components/Review/ReviewSectionRowLastActionDate.tsx
```diff
@@ -16,15 +16,15 @@ const ReviewSectionRowLastActionDate: React.FC<ReviewSectionComponentProps> = ({
         return (
           <LastDate
             title={strings.ACTION_DATE_REVIEW_STARTED}
-            indicator={getSimplifiedTimeDifference(thisReview?.timeStatusCreated)}
+            indicator={getSimplifiedTimeDifference(thisReview?.stage.statusCreatedDate)}
```
andreievg, line 19: i think this is different now, as previously it was date review was started, and I think stage.statusCreatedDate is when application jumped to new stage

andreievg: Ahh i see you've changed to putting statusCreated together with stage, which is a bit confusing for me, I think the stage no ReviewDetails should be explicitly 'applicationStage' -> to avoid the confusion, and maybe keep review status related field within the review, rather then stage ?

src/utils/graphql/queries/getReviewInfo.query.ts
```diff
@@ -6,13 +6,19 @@ export default gql`
     reviewAssignments(condition: { applicationId: $applicationId }, orderBy: TIME_UPDATED_DESC) {
       nodes {
         id
-        levelNumber
```
andreievg, line 9: I think this mike cause a problem ?

src/utils/hooks/useLoadApplication.ts
```diff
@@ -99,15 +107,14 @@ const useLoadApplication = ({ serialNumber, networkFetch }: UseGetApplicationPro
...
-        },
+      currentStage: {
+        id: stageId as number,
+        name: stage as string,
+        number: stageNumber as number,
+        colour: stageColour as string,
+        createdDate: statusHistoryTimeCreated,
         status: status as ApplicationStatus,
-        date: statusHistoryTimeCreated,
+        statusCreatedDate: stageHistoryTimeCreated,
```
andreievg, line 117: is this correct ? there is no graphQL query for this and stage history time created relates to creation time of stage not latest status. similar to comment about review, maybe we should seperate currentStage state from currentStatus state

src/utils/graphql/queries/getReviewInfo.query.ts
```diff
@@ -6,13 +6,19 @@ export default gql`
...
         reviewerId
         isLastLevel
         allowedSections
+        stage {
+          id
+          number
+          title
+          colour
+        }
+        stageDate
```
andreievg, line 21: was wondering is this the only 'data' addition for this PR ?

src/utils/graphql/queries/getReviewInfo.query.ts
nmadruga: It's a duplication : )

src/utils/graphql/queries/getReviewInfo.query.ts
nmadruga: Yes. Pretty much, I have had to refactor things to also avoid using different types to keep the same data:
* Removed type StageAndStatus
* Replaced StageAndStatus with ApplicationStage in FullStructure of application
* Replace Review.stage from ApplicationStage to new type ReviewStage and move status to inside of that

src/utils/hooks/useLoadApplication.ts
nmadruga: Good catch, this is wrong. It should be the opposite way. Changing this now

src/components/Review/ReviewSectionRowLastActionDate.tsx
nmadruga: Yes... I can see it might be a bit confusing indeed. Before we had a composed property:
```

**Comment by nmadruga** [comment]
@andreievg any further comments on this PR? I would like to continue dealing with problem/suggestion on #477 - part 2.

File: src/components/Review/ReviewSectionRowLastActionDate.tsx (hunk 1 above)
Comment (nmadruga) [reply src/components/Review/ReviewSectionRowLastActionDate.tsx:19]: Will be changing this to the above suggestion, using fields for times more aligned with how is stored in `review` table:
interface StageAndStatus {
  stage: StageDetails
  status: ApplicationStatus
  timeStageCreated: Date
  timeStatusCreated: Date 
}
```

nmadruga:
Hi @andreievg I have changed this to be back on using the `current` field (instead of `status` inside `stage` field) and updated the fields related to timestamps on `StageAndStatus`, `AssignmentStageAndStatus` and `ReviewStageAndStatus`, hope this is clearer now. 

I'm currently working on a required change for #477-part2 to show reviewResponses in order, will commit my changes to the following PR which will have this PR already up-to-date.

If you have time today to review this one that will be great and probably practical to get more things included in either `development` or `master` branch. Depending on how we choose to proceed.

src/utils/hooks/useLoadApplication.ts
andreievg: What about the graphQL part of the comment? I can't see where it's queried

nmadruga: Oh, I didn't see this one. Thanks! Fixing now : )

andreievg (approved):

Issue #694, opened 2021-05-12
Document consolidation progress


andreievg:
based on diagrams:

https://app.diagrams.net/#G1ttYeAHmdxWCnxyTY5UX1VSRDFPe-5Nvd
https://app.diagrams.net/#G1345qSOpVZ3PIw-_YXsZg_qgvm8CmIMDy
https://app.diagrams.net/#G1XyiT_SSpjy521KJA-hJ8WjGgdqj8m5Ru
https://app.diagrams.net/#G1JmZc2daHZswUyjOFWZicJy8sttpS_-eZ

Issue #272, opened 2021-01-18
Display and reflect in URL selected filters by user in UI


nmadruga:
List application: Clicking on filter should update interface to list only applications with that filter. 
The selected filter should be visible too. Still missing UI design for that, should be consider in the next meeting with Debbs.

Issue #604, opened 2021-04-13
Create endpoint to fetch possible assigners/reviewers per template


nmadruga:

CarlosNZ:
Why does this need to be an endpoint as opposed to just a regular GraphQL query from front-end? (I know we talked about it, I just can't remember)

CarlosNZ:
It's to do with permissions, isn't it?

PR #799, opened 2021-06-20
#272 #604 #702 filter UI

andreievg:
fixes: #272 
fixes: #694 
fixes: #702

requires back end branch in this pr: https://github.com/openmsupply/application-manager-server/pull/415

a bit bigger diff then I hoped, should be pretty straight forward though (i hope).

### Changes

* Wanted to combine filter definitions for both building and displaying/applying UI filters (so we can use this same logic in outcomes and lookup tables), thus:
   * buildQueryFilters was refactored to define filterTypeDefinitions (filterType -> method for generating filter based on string value)
   * One change in logic from existing filtering is that stage is using includesInsensitive for all (even for stage)
   * Added applicationFilters definitions which maps application filters to types 
   * Filter definitions also has extra options (like enum list, or query to use where option list comes from a query, or remapping from filter name to columnName)
* Added new queries for filter option queries (added them to .query files rather then inline in applicationFilters, so that they can be checked at compile time (captured by yarn generate)
* Filter UI
   * ListFilter is the main component for adding filters (on page reload filters from query are used)
   * OptionsFilters and BooleanFilter for individual sets of filters
   * common filter definitions
   * URL query is the main source of filter criteria state (apart from new filters that were added without selected criteria)

`note about getListQuery` -> since some option sets need to be selected from an api query which may require some extra information about the context of what is being filtered, we need to pass through some parameters when instantiating ListFilters, which in turn are passed to getListQuery (like templateCode and userId for applicationList, but would be different for lookup table and outcomes)

### Testing

* Using prerequisite back end branch and pg without permission restrictions, try filtering review test applications in dev locale. (see how things look when navigating to application list from dashboard filter link)


src/containers/List/ListFilters/ListFilters.tsx
```diff
@@ -0,0 +1,145 @@
...
+      })
+    }
+
+    const availableFilterNames = filterNames.filter(
+      (filterName) => !activeFilters.includes(filterName)
+    )
+
+    return (
+      <div id="list-filter">
+        <Dropdown text="+ Filter" icon={null}>
```
CarlosNZ, line 66: Localise "+ Filter" string?

CarlosNZ: Also -- should the "+" be an icon? (You can maybe re-use the styling for the "+ New Application" button on the right)

src/containers/List/ListFilters/ListFilters.tsx
```diff
@@ -0,0 +1,145 @@
+import React, { useEffect, useState } from 'react'
+import { Dropdown } from 'semantic-ui-react'
+import { useRouter } from '../../../utils/hooks/useRouter'
+import { FilterDefinitions } from '../../../utils/types'
+import BooleanFilter from './BooleanFilter'
+import { EnumFilter, SearchableListFilter, StaticListFilter } from './OptionFilters'
+import { GetMethodsForOptionFilter } from './types'
+
```
CarlosNZ, line 8: Just putting this here, not sure where it belongs:

Seems like a bug on the filter list drop-down: when first loading, the drop-down has an additional "check" item at the top of the list, which does nothing. It disappears after selecting it.
Screen Shot 2021-06-22 at 11 05 42 AM

src/containers/List/ListFilters/OptionFilters.tsx
```diff
@@ -0,0 +1,110 @@
+import React, { useState } from 'react'
```
CarlosNZ, line 1: Any particular reason why these 3 types are together in one file, but Boolean Filters has its own file?

CarlosNZ: FWIW -- I found the term "optionsFilters" quite confusing until I looked through it properly, so might be nicer just to have one "FilterTypes.tsx" file, or put them all seperate?

CarlosNZ (changes requested):
Cool, looking really good.

Just a few additional comments/questions/suggestions
- How does the UI determine what filters should be available? (e.g if I'm an applicant, I shouldn't see "Reviewers" as a possible filter), and is this being enforced yet? I could see all filters all the time, regardless of which user I switched to.
- Suggestion: Might be nice to have a "Clear all filters" button, as it can be quite tedious to "Remove Filter" many times once you've added a few filters. Perhaps it only would show up once there's more than 2 filters added?
- Some basic (high-level) documentation would  be nice, though understand if not main priority this week. I just find it a bit tricky to follow all the logic here, so maybe just quickly explain it in person tomorrow?

src/containers/List/ListFilters/BooleanFilter.tsx
```diff
@@ -0,0 +1,34 @@
+import React from 'react'
```
nmadruga, line 1: Suggestion to add these to a `components/List` folder since it seems to be more visualisation done here - not much logic involved

src/containers/List/ListFilters/ListFilters.tsx
```diff
@@ -0,0 +1,145 @@
...
+      setInactiveOption: (option: string) =>
+        updateQuery({
+          [filterName]: getArrayFromString(query[filterName])
+            .filter((_option) => _option != option)
+            .join(','),
+        }),
+    })
+
+    const addFilter = (filterName: string) => setActiveFilters([...activeFilters, filterName])
+    const getOnRemove = (filterName: string) => () => {
```
nmadruga, line 53: not sure what this function meant - is this used when removing all options for a given filterName?

src/containers/List/ListFilters/ListFilters.tsx
```diff
@@ -0,0 +1,145 @@
...
+              </Dropdown.Item>
+            ))}
+          </Dropdown.Menu>
+        </Dropdown>
+
+        {activeFilters.map((filterName) => {
+          const filter = displayableFilters[filterName]
+          if (!filter.title) return null
+
+          if (filter.type === 'enumList') {
```
nmadruga, line 80: I recommend using switch case here - it's just easier to follow although you need to have a default case, I would use the enumList which seems to be the most used one.

The code is looking great though, congrats!

src/containers/List/ListFilters/OptionFilters.tsx
```diff
@@ -0,0 +1,110 @@
...
+  const activeOptions = getActiveOptions()
+  let matchedOptions: string[] = []
+
+  if (!error && data) matchedOptions = resultExtractor(data).list
+
+  return (
+    <FilterContainer selectedCount={activeOptions.length} title={title} onRemove={onRemove}>
+      <Input
+        icon="search"
+        placeholder="Start typing..."
```
nmadruga, line 91: Localisation string missing

src/containers/List/ListFilters/common.tsx
```diff
@@ -0,0 +1,65 @@
...
+        {selectedCount}
+      </Label>
+    )}
+    <Dropdown multiple text={title}>
+      <Dropdown.Menu>
+        {children}
+        <Dropdown.Divider />
+        <Dropdown.Item className="remove-filter" key="removeFilter" onClick={() => onRemove()}>
+          <Icon name="remove circle" />
+          Remove filter
```
nmadruga, line 26: Localisation string missing

src/utils/data/applicationFilters.ts
```diff
@@ -0,0 +1,128 @@
...
+    return {
+      query,
+      variables: {
+        searchValue,
+        templateCode,
+      },
+      resultExtractor: (resultQuery) => ({
+        list: resultQuery?.[queryMethod]?.nodes || [],
+        totalCount: resultQuery?.[queryMethod]?.totalCount || 0,
+      }),
```
nmadruga, line 27: Oh, that's interesting. Can this be used in any GraphQL query?

src/utils/data/applicationFilters.ts
```diff
@@ -0,0 +1,128 @@
...
+    }
+  }
+  return getListQuery
+}
+
+export const APPLICATION_FILTERS: FilterDefinitions = {
+  lastActiveDate: {
+    type: 'date',
+    title: 'check',
```
nmadruga, line 36: Found the `check` here @CarlosNZ  : )

src/utils/data/applicationFilters.ts
```diff
@@ -0,0 +1,128 @@
...
+    type: 'boolean',
+    title: 'Assignment Status',
+    options: {
+      booleanMapping: {
+        true: 'Fully Assigned',
+        false: 'Not Fully Assigned',
+      },
+    },
+  },
+  assignReviewerAssignedCount: {
```
nmadruga, line 48: Why not `reviewersAssignedCount` instead? I haven't really understood what is this filter though. Not a visible one so only available on the URL correct?

nmadruga (commented):
Hi, just wanted to have a look on this. Made a few comments, and I'm excited to see all in there. Will have a go testing this later, but it seems very wrapped around. The types in the main `types` file and in the `ListFilters/type` aren't still all clear in my head, but probably after a while dealing with the code it might get clearer.

Great work Andrei, looking forward to get this in today

semantic/src/site/globals/site.overrides
```diff
@@ -1060,6 +1060,76 @@ a:hover {
...
+  padding: 5px;
+  padding-left: 10px;
+  padding-right: 10px;
+  background-color: @surfacesWhite;
+  margin-right:10px;
+  color: @darkGrey;
+
+  &.visible {
+    border: solid 1px @blue;
+  }
```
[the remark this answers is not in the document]
andreievg: e35082241a32ff22b5a64f373f8fe8c47c6c84fa

semantic/src/site/globals/site.overrides
```diff
@@ -1060,6 +1060,76 @@ a:hover {
...
+
+  .dropdown.icon {
+    margin-left: 6px;
+  }
+
+}
+
```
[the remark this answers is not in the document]
andreievg: eb33c85cabf0f8b00db8aa5f5b1bfa578ab60c64

src/containers/List/ListFilters/ListFilters.tsx
```diff
@@ -0,0 +1,145 @@
...
+
+const getDisplayableFilters = (filterDefinitions: FilterDefinitions) => {
+  const result: FilterDefinitions = {}
+  Object.entries(filterDefinitions).forEach(([filterName, filterValue]) => {
+    // Missing title determines if filters should be shown in UI or not
+    if (!filterValue.title) return
+    result[filterName] = filterValue
+  })
+
+  return result
```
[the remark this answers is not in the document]
andreievg: I prefer to do this in forEach tbh, happy to have these stylistic differences in code btw, if other prefer reduce for these sorts of things (something about defining the starting value at the 'end' bugs me)

src/containers/List/ListFilters/ListFilters.tsx
```diff
@@ -0,0 +1,145 @@
...
+
+    const displayableFilters = getDisplayableFilters(filterDefinitions)
+
+    const filterNames = Object.keys(displayableFilters)
+    const [activeFilters, setActiveFilters] = useState<string[]>([])
+
+    // Add fitlers from URL to activeFilters, filters may not have criteria/options yet thus we have to use activeFilter state variable
+    useEffect(() => {
+      const filersInQuery = filterNames.filter((filterName) => !!query[filterName])
+      const newFilters = filersInQuery.filter((filterName) => !activeFilters.includes(filterName))
```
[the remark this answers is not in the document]
andreievg: 0f4eaba74aee8eda996dfd5864af29e3ebd4d923

src/containers/List/ListFilters/ListFilters.tsx
andreievg: 76c191b7e88206665d7a00966598a805b761acaf

src/containers/List/ListFilters/ListFilters.tsx
andreievg: e572da065eab52ada930e0b46201740ee4af7115

src/containers/List/ListFilters/OptionFilters.tsx
andreievg: they kind of all have the same UI, and set of options, and boolean behaves different. I am not happy with the name 'options', happy to change if there is anything better.  Can't think of a better term atm

src/containers/List/ListFilters/BooleanFilter.tsx
andreievg: Hmm, I've been pretty placid on the folder structure front, but I really think it's more important to group components/container by functional areas rather then by differentiation of how much logic they contain. I am wondering if I am missing something, what is the benefit of differentiating between components and containers ?

src/containers/List/ListFilters/ListFilters.tsx
andreievg: Yeah, it's for removing the whole filter option, and the `get` part of it, so we don't have to do the following in each filter instantiation below
```
  setActiveFilters(activeFilters.filter((_filterName) => _filterName !== filterName))
      updateQuery({
        [filterName]: undefined,
      })
```

src/containers/List/ListFilters/ListFilters.tsx
andreievg: fa1860d3aca2cd4f8b3bc2473e1220ef920021c8
so much better, thanks

src/containers/List/ListFilters/ListFilters.tsx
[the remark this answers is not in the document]
CarlosNZ: Weirdly long commit hash. I think it only shows up with auto-linking if you add the first 8 chars (this is what shows in Github Desktop) 0f4eaba7

src/containers/List/ListFilters/OptionFilters.tsx
andreievg: 29ec70fe38a02cab62de9aa53692bba17afed202

src/containers/List/ListFilters/common.tsx
andreievg: 037a9ebc909b8b7d509350aa26518178b121fbb4

src/utils/data/applicationFilters.ts
andreievg: all of the filter list/options queries have a common interface, this was just done to avoid the need of creating custom hook to extract the data from them, can be used for others but most other queries differentiate from one another and require extra processing

src/utils/data/applicationFilters.ts
andreievg: Thanks, this is fixed in upstream branch, but I've fixed it here too

src/utils/data/applicationFilters.ts
andreievg: yeah, not sure what it's for either.. it was there so decided to expose it (since there are no other number filters, so used this one as a template for future number filters)
Actually I can't see it in the list at all now, so commented it out: d49e730902ff11f538f39fa8e6de66cf4f06769f

src/containers/List/ListFilters/ListFilters.tsx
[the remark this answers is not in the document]
andreievg: fixed, i typically just copy sha (straight from commit), i think this wasn't pushed yet when i made the comment

src/containers/List/ListFilters/BooleanFilter.tsx
nmadruga: I think the components should be something that can be more easily reused if they aren't allocated inside one logic-area folder. But since our components some times aren't following the strict of only doing visualisation that might not be the case. I'm just pointing out that this is in the wrong folder with the current folders structure we have.

nmadruga: To understand more about the container and component pattern design se here: 
That's not a rule, and if we disagree to use it, I would rather move things to the correct folder, so it makes sense to everyone than keeping as it without using the proper sorting of files.

nmadruga (approved):
Thanks for the changes, Andrei

src/containers/List/ListFilters/BooleanFilter.tsx
andreievg: Good point, it's definitely not consistent with existing projects structure, so far we have deviations in lookup tables, filter and template builder. I feel like re-structuring the whole project would be a big task, was thinking, maybe it's ok to have mix of different structure styles ?

andreievg:
Approved by @nmadruga and I've made all requested changes by @CarlosNZ, merging

Issue #805, opened 2021-06-22
Add date filter


andreievg:
Add date filter as per designs

PR #806, opened 2021-06-22
#805 date filters

andreievg:
Fixes: #805 

Sorry another massive PR, I got a little bit carried away, but happy with result.

Didn't end up adding main date filter to application page, but made it possible (should be able to easily use generic filter logic).

Functionality in this PR should work for named date (mapped date), and relative dates:

`last-active-date=this-week`

`last-active-date=this-year` 

`last-active-date=this-week`

`2021-06-01:2021-06-24`

`:2021-06-24` before that date

`2021-06-01:` after that date

`2021-06-01:30_day` relative date (should remain as relative unless from date is manually selected and should work the same as normal date when other date is missing). `day`, `year`, `month`, `quarter`, `week`

`last-active-date=-1_year:1_quarter` (notice the minus), no UI for relative days yet, we may decide to add it in the future

### Changes

* moved date methods from buildQueryFilters to parseDateRange
* namedDates (previously called mapped i think), is now an option in filter definitions
* added namedDates to filter definitions (extend to a few more)
* added icons to filter drop down (not 100% happy with it)
* added date filter to UI
   * DateFilter/helpers, logic methods for date picker
   * DateFilter/DateFilters, UI elements
* made common active filter drop down trigger extendable, and extending it for date filter (to show selected date range or namedDate) <- this can be easily used to display stand alone date filter

Screen Shot 2021-06-22 at 11 32 40 PM

**note** couldn't see how we can use available npm packages to easily make UI and functionality work, thus created our own, maybe we can use it for date plugin element


CarlosNZ:
Haven't had a proper look yet, just a couple of initial comments:

- The general button styling has been broken -- the "New" buttons are all washed out/faded:
Screen Shot 2021-06-23 at 10 07 22 AM
Screen Shot 2021-06-23 at 10 07 16 AM
- Do we need to create our own Date UI -- why not just use the same package I used for the date-picker plugin? [react-semantic-ui-datepickers](https://www.npmjs.com/package/react-semantic-ui-datepickers) Either way, we should probably co-ordinate them so all date interfaces are using the same UI.

andreievg:
> Do we need to create our own Date UI -- why not just use the same package I used for the date-picker plugin? react-semantic-ui-datepickers Either way, we should probably co-ordinate them so all date interfaces are using the same UI.

I've made a note at the end of the PR comments (hard to see):

**note** couldn't see how we can use available npm packages to easily make UI and functionality work, thus created our own, maybe we can use it for date plugin element

andreievg:
> The general button styling has been broken -- the "New" buttons are all washed out/faded:

Oops I overwritten ligthBlue, changed:

89c95400fdc8bdcaa2e6a625e40b9ed55c15403a

CarlosNZ:
> **note** couldn't see how we can use available npm packages to easily make UI and functionality work, thus created our own, maybe we can use it for date plugin element

Okay, but the package has a lot of functionality that wouldn't want to have to implement ourselves, for example:
- enforce either single date or date range selection
- limit date range (e.g. can only select 3 months into the future) (which also affects where the calendar opens up to by default)
- automatic localisation (mostly)

(It has loads more, but those are just the things I've taken advantage of in the date picker plugin)

andreievg:
In our conversation with @CarlosNZ, it was deemed ok to keep both the date picker package for plugin and have this functionality.

PR #809, opened 2021-06-23
#796 bug fix after re review loq not showing to applicant

### LINKED ISSUE #796: Bug: Applicant can't start updating application after Changes-required after a second-review (when consolidator had disagreed previously with decision)
The list show action for applicant to **Update** their responses but in Application home page there is no Update button, therefor the application cannot be started.

nmadruga:
Fixes #796

- Retrieve reviewResponseLinked to compare correctly before setting to be visible to applicant. This should fix the problem with submitting a list of questions after review with consolidation (consider more than one iteration between reviewer <-> consolidator).


CarlosNZ (approved):
LGTM!

PR #810, opened 2021-06-23
#800 Verification UI

### LINKED ISSUE #800: UI to show link to user for verification

CarlosNZ:
Fixes #800

Requires [back-end PR #424](https://github.com/openmsupply/application-manager-server/pull/424)

Creates route (`/verify`) and basic UI for verification page. Note that this route is outside the "Authenticated Content" wrapper, as it needs to be accessible to non-logged in users.

To test:
- follow instructions from the back-end PR
- notice that if you try and visit a verification page more than once you will get an error UI instead.

FYI -- the layout is basically just a copy of the "Submission" page. Keeping it simple.


nmadruga:
Oh, some conflicts to solve there. I'll check with you on Telegram if ok to do this conflict resolution to speed this review.

Also have tested the verification now and it's looking great. Very good work : )

Have a minor request though, I think linking the Dashboard bellow the message you receive once you are verified is odd, it would be better to say **Login** instead : )
Screen Shot 2021-07-07 at 12 22 22 PM

CarlosNZ:
> Oh, some conflicts to solve there. I'll check with you on Telegram if ok to do this conflict resolution to speed this review.

It was just the graphql.tsx file, so I've updated and re-generated it. All good now :)

CarlosNZ:
> Have a minor request though, I think linking the Dashboard bellow the message you receive once you are verified is odd, it would be better to say **Login** instead : )

The thing is, this Verification page can be used for other kinds of verifications too, and the user might already be logged in. I've added an additional `isLoggedIn()` check to determine what text to display. (The link doesn't need to change, as it'll redirect to `/login` if not logged in.)

941d1ff8

nmadruga (approved):
LGTM 🎉 

Thanks!

PR #817, opened 2021-06-28
#815 ListBuilder dynamic display

### LINKED ISSUE #815: ListBuilder: evaluate isVisible, isEditable and isRequired dynamically

CarlosNZ:
Fixes #815 

Adds an evaluation to listBuilder to allow `isVisibile`, `isEditable` and `isRequired` to be evaluated dynamically for each inner element, just like "normal" elements.

To see it in action, use back-end branch: `#815F-list-builder-evaluation` (This branch not intended for merging, just for demo-ing). Use the "Drug Registration" template in "dev" folder, and go to Ingredients List. Will add to "proper" Laos drug reg template once we're happy with how it works.

(Note: there's a couple of files here in the diff that are not relevant to this PR, to do with ReviewResponses, not sure which PR they're from)


src/utils/hooks/useUpdateReviewResponse.ts
```diff
@@ -46,7 +46,7 @@ const computeVisibility = (reviewResponse: ReviewResponse | undefined) => {
   // consolidator
   if (
     reviewResponse.decision === ReviewResponseDecision.Agree &&
-    reviewResponse.reviewResponseLink?.decision === ReviewResponseDecision.Decline
+    reviewResponse.originalReviewResponse?.decision === ReviewResponseDecision.Decline
```
andreievg, line 53: this looks like a correct change (however would only affect lvl 3 consolidation by the looks of it), ok to have it in i think

src/formElementPlugins/listBuilder/src/ApplicationView.tsx
```diff
@@ -64,6 +75,18 @@ const ApplicationView: React.FC<ApplicationViewProps> = ({
   const [open, setOpen] = useState(false)
   const [inputError, setInputError] = useState(false)
 
+  const [currentResponseElementsState, setCurrentResponseElementsState] = useState<any>()
```
andreievg, line 78: would prefer some type, even if it's `{[key: string]: any} | undefined`

andreievg (approved):
looks good, one comment about type.

src/utils/hooks/useUpdateReviewResponse.ts
CarlosNZ: Hmm, I don't think this was an intended change, it doesn't really have anything to do with the issue. The commit message I wrote was "Revert unrelated changes", and now I'm not sure what it's referring to. I think I should get @nmadruga to check was this is supposed to be before I merge it.

src/formElementPlugins/listBuilder/src/ApplicationView.tsx
CarlosNZ: Cool: 29ab720d
```
const [currentResponseElementsState, setCurrentResponseElementsState] =
    useState<{ [key: string]: ElementState }>()
```

PR #820, opened 2021-06-30
#816 Nested Listbuilder -- WIP

CarlosNZ:
Just putting this here so @andreievg can have a look if you want. 

Back-end branch is `#816F-nested-list-builders` (Inline listbuilder is in "Drug Registration - General Medicines Procedure", Ingredients Page 1)

I've significantly tidied up the plugin, separating into different files for easier structure.

I haven't managed to get the Edit working for the Inline version of the plugin, as we're still getting the re-renders of the child components, not sure how to fix.

For some reason, the "outer" component (ApplicationView) is running multiple times when it loads, each time which causes its children (the DisplayComponent and its children) to be re-instantiated with initial values.

Same when the response changes -- it causes a "save" which causes "allResponses" all the way up in the structure to update, so all children get re-run -- so the list ApplicationVIew re-loads and its children get reset. You can see this most clearly when you open one of the list Accordions, and click "edit" -- we see the inline form appear, but the accordion itself is closed even though there was never any `setOpen(false)` called on it -- it's been re-initialised. 

We need to figure out how to deal with this problem before I can embed an editable form inside the Accordions (it should be straightforward after that)

### DIFF
--- a/src/formElementPlugins/listBuilder/src/displayComponents/ListInlineLayout.tsx
+++ b/src/formElementPlugins/listBuilder/src/displayComponents/ListInlineLayout.tsx
@@ -2,10 +2,10 @@ import React, { useEffect, useState } from 'react'
 import { Button, Icon, Accordion } from 'semantic-ui-react'
 import { ApplicationDetails, User } from '../../../../utils/types'
 import { ListLayoutProps } from '../types'
-
 import ApplicationViewWrapper from '../../../ApplicationViewWrapper'
 import SummaryViewWrapper from '../../../SummaryViewWrapper'
 import { buildElements, substituteValues } from '../helpers'
+import '../styles.css'
 
 const ListInlineLayout: React.FC<ListLayoutProps> = ({
   listItems,
@@ -36,7 +36,11 @@ const ListInlineLayout: React.FC<ListLayoutProps> = ({
     }, [])
     if (!currentItemElementsState) return null
     return (
-      <Accordion styled style={{ marginBottom: 5, marginTop: 10 }}>
+      <Accordion
+        styled
+        className="accordion-container"
+        style={{ maxWidth: open ? 'none' : '100%' }}
+      >
         <Accordion.Title active={open} onClick={() => setOpen(!open)}>
           <Icon name="dropdown" />
           <Markdown text={substituteValues(header, item)} semanticComponent="noParagraph" />
--- a/src/formElementPlugins/listBuilder/src/displayComponents/ListInlineLayout.tsx
+++ b/src/formElementPlugins/listBuilder/src/displayComponents/ListInlineLayout.tsx
@@ -1,0 +1,69 @@
+import React, { useEffect, useState } from 'react'
+import { Button, Icon, Accordion } from 'semantic-ui-react'
+import { ApplicationDetails, User } from '../../../../utils/types'
+import { ListLayoutProps } from '../types'
+
+import ApplicationViewWrapper from '../../../ApplicationViewWrapper'
+import SummaryViewWrapper from '../../../SummaryViewWrapper'
+import { buildElements, substituteValues } from '../helpers'
+
+const ListInlineLayout: React.FC<ListLayoutProps> = ({
+  listItems,
+  displayFormat,
+  fieldTitles = [],
+  codes = [],
+  Markdown,
+  editItem = () => {},
+  deleteItem = () => {},
+  isEditable = true,
+  inputFields,
+  responses,
+  currentUser,
+  applicationData,
+}) => {
+  // Inner component -- one for each Item in list
+  const ItemAccordion: React.FC<any> = ({ item, header, index }) => {
+    const [open, setOpen] = useState(false)
+    const [currentItemElementsState, setItemResponseElementsState] = useState<any>()
+    useEffect(() => {
+      buildElements(
+        inputFields,
+        responses,
+        item,
+        currentUser as User,
+        applicationData as ApplicationDetails
+      ).then((elements) => setItemResponseElementsState(elements))
+    }, [])
+    if (!currentItemElementsState) return null
+    return (
+      <Accordion styled style={{ marginBottom: 5, marginTop: 10 }}>
+        <Accordion.Title active={open} onClick={() => setOpen(!open)}>
+          <Icon name="dropdown" />
+          <Markdown text={substituteValues(header, item)} semanticComponent="noParagraph" />
+        </Accordion.Title>
+        <Accordion.Content active={open}>
+          {codes.map((code, cellIndex) => (
+            <SummaryViewWrapper
+              key={`item_${cellIndex}`}
+              element={currentItemElementsState[code]}
+              response={item[code].value}
+              allResponses={responses}
+            />
+          ))}
+          <Button primary content={'EDIT'} onClick={() => editItem(index)} />
+          <Button secondary content={'DELETE'} onClick={() => deleteItem(index)} />
+        </Accordion.Content>
+      </Accordion>
+    )
+  }
+
+  return (
+    <>
+      {listItems.map((item, index) => (
+        <ItemAccordion item={item} header={displayFormat.title} key={index} index={index} />
+      ))}
+    </>
+  )
+}
+
+export default ListInlineLayout
--- a/src/formElementPlugins/listBuilder/src/displayComponents/ListInlineLayout.tsx
+++ b/src/formElementPlugins/listBuilder/src/displayComponents/ListInlineLayout.tsx
@@ -48,7 +48,7 @@ const ListInlineLayout: React.FC<ListLayoutProps> = ({
         <Accordion.Content active={open}>
           {codes.map((code, cellIndex) => (
             <SummaryViewWrapper
-              key={`item_${cellIndex}`}
+              key={`item_accordion_${cellIndex}`}
               element={currentItemElementsState[code]}
               response={item[code].value}
               allResponses={responses}
--- a/src/formElementPlugins/listBuilder/src/helpers.ts
+++ b/src/formElementPlugins/listBuilder/src/helpers.ts
@@ -1,0 +1,112 @@
+import { ApplicationViewProps } from '../../types'
+import {
+  ApplicationDetails,
+  ElementState,
+  EvaluationOptions,
+  ResponseFull,
+  ResponsesByCode,
+  User,
+} from '../../../utils/types'
+import { TemplateElement, TemplateElementCategory } from '../../../utils/generated/graphql'
+import ApplicationViewWrapper from '../../ApplicationViewWrapper'
+import SummaryViewWrapper from '../../SummaryViewWrapper'
+import strings from '../constants'
+import { evaluateElements } from '../../../utils/helpers/evaluateElements'
+import { defaultEvaluatedElement } from '../../../utils/hooks/useLoadApplication'
+import { useUserState } from '../../../contexts/UserState'
+import { DisplayType, InputResponseField, ListItem, ListLayoutProps } from './types'
+
+// Formatting and Text manipulation
+export const getDefaultDisplayFormat = (inputFields: TemplateElement[]) => {
+  const displayString = inputFields.reduce(
+    (acc: string, { code, title }) => acc + `**${title}**: \${${code}}  \n`,
+    ''
+  )
+  return { title: '', subtitle: '', description: displayString }
+}
+
+export const substituteValues = (parameterisedString: string, item: ListItem) => {
+  const getValueFromCode = (_: string, $: string, code: string) => item[code]?.value?.text || ''
+  return parameterisedString.replace(/(\${)(.*?)(})/gm, getValueFromCode)
+}
+
+export const createTextString = (listItems: ListItem[], inputFields: TemplateElement[]) =>
+  listItems.reduce(
+    (outputAcc, item) =>
+      outputAcc +
+      inputFields.reduce(
+        (innerAcc, field) => innerAcc + `${field.title}: ${item[field.code]?.value?.text}, `,
+        ''
+      ) +
+      '\n',
+    ''
+  )
+
+// Data validation and state reset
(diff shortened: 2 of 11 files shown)


CarlosNZ:
Okay @andreievg , I'm pretty happy with how this is functioning now. Inline editing works propertly.

There is still a bit of an issue with UI -- I'm just using basic "Edit"/"Delete" buttons inside the inline editor, but I think we should talk about how exactly it should work/look while looking at it together before I change too much.

semantic/src/definitions/modules/accordion.less
```diff
@@ -51,14 +51,14 @@
...
-  padding-bottom: 0em;
-}
+// .ui.accordion:not(.styled) .title ~ .content:not(.ui),
+// .ui.accordion:not(.styled) .accordion .title ~ .content:not(.ui) {
+//   margin: @contentMargin;
+//   padding: @contentPadding;
+// }
+// .ui.accordion:not(.styled) .title ~ .content:not(.ui):last-child {
+//   padding-bottom: 0em;
+// }
```
CarlosNZ, line 61: Note to reviewer: the reason this is commented out is quite inexplicable to me. Basically, when this is left as it was, the padding/margin values for the Accordion take on these values, _but only on Summary page!_ I've inspected the HTML and I can't see any difference in the component/class structure between Application and Summary pages that would affect this, but somehow it's taking on these ".not" properties on the Summary page. Any insights would be welcome.

src/formElementPlugins/listBuilder/src/displayComponents/ListInlineLayout.tsx
```diff
@@ -0,0 +1,117 @@
...
+import { buildElements, substituteValues } from '../helpers'
+import '../styles.less'
+
+const ListInlineLayout: React.FC<ListLayoutProps> = (props) => {
+  const { listItems, displayFormat } = props
+  return (
+    <>
+      {listItems.map((item, index) => (
+        <ItemAccordion
+          key={index} // Not ideal, but there isn't any other unique identifier
```
nmadruga, line 16: How about `${displayFormat.title}-${index}`?

src/formElementPlugins/listBuilder/src/displayComponents/ListInlineLayout.tsx
```diff
@@ -0,0 +1,117 @@
...
+              element={currentItemElementsState[code]}
+              isStrictPage={false}
+              allResponses={responses}
+              currentResponse={item[code].value}
+              onSaveUpdateMethod={innerElementUpdate(currentItemElementsState[code].code)}
+              applicationData={applicationData}
+            />
+          ) : (
+            <SummaryViewWrapper
+              key={`item_accordion_${cellIndex}`}
```
nmadruga, line 104: Maybe something like `summary_item_${cellIndex}`? I'm still understanding this for now : )

src/formElementPlugins/listBuilder/src/displayComponents/ListInlineLayout.tsx
```diff
@@ -0,0 +1,117 @@
...
+  editItem,
+  deleteItem,
+  editItemText,
+  updateButtonText,
+  deleteItemText,
+  innerElementUpdate,
+  updateList,
+}) => {
+  const [open, setOpen] = useState(false)
+  const [edit, setEdit] = useState(false)
```
nmadruga, line 49: From my understanding so far I think it would make more sense if using:
```suggestion
  const [isEditing, setIsEditing] = useState(false)
```

nmadruga: And changes in everywhere using...

src/formElementPlugins/listBuilder/src/helpers.ts
```diff
@@ -0,0 +1,112 @@
...
+  User,
+} from '../../../utils/types'
+import { TemplateElement, TemplateElementCategory } from '../../../utils/generated/graphql'
+import ApplicationViewWrapper from '../../ApplicationViewWrapper'
+import SummaryViewWrapper from '../../SummaryViewWrapper'
+import strings from '../constants'
+import { evaluateElements } from '../../../utils/helpers/evaluateElements'
+import { defaultEvaluatedElement } from '../../../utils/hooks/useLoadApplication'
+import { useUserState } from '../../../contexts/UserState'
+import { DisplayType, InputResponseField, ListItem, ListLayoutProps } from './types'
```
nmadruga, line 17:
```suggestion
import {
  ApplicationDetails,
  ElementState,
  EvaluationOptions,
  ResponsesByCode,
  User,
} from '../../../utils/types'
import { TemplateElement, TemplateElementCategory } from '../../../utils/generated/graphql'
import { evaluateElements } from '../../../utils/helpers/evaluateElements'
import { defaultEvaluatedElement } from '../../../utils/hooks/useLoadApplication'
import { ListItem } from './types'
```

Many were not used

src/formElementPlugins/listBuilder/src/helpers.ts
```diff
@@ -0,0 +1,112 @@
...
+  const displayString = inputFields.reduce(
+    (acc: string, { code, title }) => acc + `**${title}**: \${${code}}  \n`,
+    ''
+  )
+  return { title: '', subtitle: '', description: displayString }
+}
+
+export const substituteValues = (parameterisedString: string, item: ListItem) => {
+  const getValueFromCode = (_: string, $: string, code: string) => item[code]?.value?.text || ''
+  return parameterisedString.replace(/(\${)(.*?)(})/gm, getValueFromCode)
```
nmadruga, line 30: Would be good to comment what is the string substitution/regex doing for future reference

src/formElementPlugins/listBuilder/src/helpers.ts
```diff
@@ -0,0 +1,112 @@
...
+    code: field.code,
+    pluginCode: field.elementTypePluginCode as string,
+    category: field.category as TemplateElementCategory,
+    title: field.title as string,
+    parameters: field.parameters,
+    validationExpression: field?.validation || true,
+    validationMessage: field?.validationMessage || '',
+    isVisibleExpression: field?.visibilityCondition || true,
+    isEditableExpression: field?.isEditable || true,
+    isRequiredExpression: field?.isRequired || true,
```
nmadruga, line 91: Nice! This was broken on the current ListBuilder. Will check how it is using this version :)

src/formElementPlugins/listBuilder/src/displayComponents/ListInlineLayout.tsx
```diff
@@ -0,0 +1,117 @@
...
+          {...props}
+        />
+      ))}
+    </>
+  )
+}
+export default ListInlineLayout
+
+// Inner component -- one for each Item in list
+const ItemAccordion: React.FC<any> = ({
```
nmadruga, line 29:
```suggestion
const ItemAccordion: React.FC<ListLayoutProps > = ({
```
?

nmadruga (commented):
Great! I'll be testing this one and make more comments or approve after finished.

The plan is to use this PR as base for the next one I'll be working to add error-handling for List-builder and radioChoice (and others) if these are required and no response given.

nmadruga:
Also, I have merged latest `develop` into this branch and back-end branch `#816F-nested-list-builders`. And there is a new file with changes from back-end new types. Hope that's fine, if not let me know and we can revert that.

nmadruga:
Ah, couldn't test this much further because I don't have file to upload to lookup-tables of `ingredients.csv`. Will check with you tomorrow.

nmadruga:
Tested now after import of Lookup table ingredients.csv - suggestion add the file in the lookup table for snapshot dev so it's already there for future use.

nmadruga:
Other fixes required: when showing on Summary/Review pages the edit icon and Review icon are being pushed to outside the container:
Screen Shot 2021-08-25 at 10 26 42 AM
Screen Shot 2021-08-25 at 10 25 16 AM

nmadruga:
Also required fields are showing for all elements in the Add ingredient modal, should only be showing for the ones with `isRequired` = true

Screen Shot 2021-08-25 at 10 24 09 AM

src/formElementPlugins/listBuilder/src/helpers.ts
CarlosNZ: > Also required fields are showing for all elements in the Add ingredient modal, should only be showing for the ones with `isRequired` = true

Yeah -- once again I forget that that "false" will fallback to the second parameter of an `||` operator 🙄 -- have updated to `??`: 9835533f

src/formElementPlugins/listBuilder/src/displayComponents/ListInlineLayout.tsx
CarlosNZ: Sure. thanks. a4ed71fe

src/formElementPlugins/listBuilder/src/displayComponents/ListInlineLayout.tsx
CarlosNZ: I just wanted it to be distinct as an "accordion" item when inspecting. But it doesn't really matter, it's a key identifier.

src/formElementPlugins/listBuilder/src/displayComponents/ListInlineLayout.tsx
CarlosNZ: Sure, good call: 532c123e

src/formElementPlugins/listBuilder/src/helpers.ts
CarlosNZ: Oops -- thanks 254d8b3e

src/formElementPlugins/listBuilder/src/helpers.ts
CarlosNZ: Sure. a3324c71

src/formElementPlugins/listBuilder/src/displayComponents/ListInlineLayout.tsx
CarlosNZ: Not quite so straightforward, as it has some additional props above ListLayoutProps, but i've made it work: 21ecb3f7

CarlosNZ: > Tested now after import of Lookup table ingredients.csv - suggestion add the file in the lookup table for snapshot dev so it's already there for future use.

Yeah, I'll do this when I update the dev snapshot overall with updated Feature Showcase and this listbuilder element, etc.

CarlosNZ:
> Tested now after import of Lookup table ingredients.csv - suggestion add the file in the lookup table for snapshot dev so it's already there for future use.

Yeah, I'll do this when I update the dev snapshot overall with updated Feature Showcase and this listbuilder element, etc.

CarlosNZ:
> Other fixes required: when showing on Summary/Review pages the edit icon and Review icon are being pushed to outside the container:

Okay, this was quite tricky, as I need to make sure the accordion can go wider than the container when expanded, but without pushing the review/edit buttons out.

I think I've got it sorted -- e0444a8b

CarlosNZ:
@nmadruga -- Okay, I think I've addressed everything now. The "isRequired" is fixed, and most of your notes.

CarlosNZ:
@nmadruga  -- I've also added some UI for the error state. It'll need your logic to work (probably), but you can just make it "true" to see what it looks like. 341cc6d5

nmadruga:
When selecting `inactive` can't add any other fields (isEditable = false) but still getting error when trying to submit
Screen Shot 2021-08-25 at 4 55 11 PM

nmadruga:
> > Other fixes required: when showing on Summary/Review pages the edit icon and Review icon are being pushed to outside the container:
> 
> Okay, this was quite tricky, as I need to make sure the accordion can go wider than the container when expanded, but without pushing the review/edit buttons out.
> 
> I think I've got it sorted -- [e0444a8](https://github.com/openmsupply/application-manager-web-app/commit/e0444a8ba8e5e574edf34dce0a85ceddd701e414)

Nice! This looks great 👍

CarlosNZ:
> Still something odd about `isRequired`:
> 
> When selecting `inactive` can't add any other fields (isEditable = false) but still getting error when trying to submit

Okay, this is a config issue, the plugin is following intended behaviour. There's some logic on there to only allow editing of other elements when "Active" as been selected, but they're still required. This was a hangover from something Valerio said about "active/inactive" needing to be selected first. I'll just get rid of it on the Product Reg template, but shouldn't need to change anything here.

Thanks for spotting it.

nmadruga:
Another one - sorry 🦖 

Forever on the Validating....
Screen Shot 2021-08-25 at 5 04 23 PM

Steps to reproduce:
- Ingredients list needs to have template config set to `isRequired` = true
- Complete the previous page and then click "Review and submit" to get the validation showing
- Now add an ingredient ( I usally click Add before all isRequired inner elements are added to get an extra error)
- Finish to add all required elements and then **Add**

Problem one: the validation doesn't change below **Add Formulation**
Problem two: the Review & Submit stays in a loop while disabled and showing **Validating...**

CarlosNZ:
> Another one - sorry 🦖
> 
> Forever on the Validating....

> Steps to reproduce:
> 
>     * Ingredients list needs to have template config set to `isRequired` = true
> 
>     * Complete the previous page and then click "Review and submit" to get the validation showing
> 
>     * Now add an ingredient ( I usally click Add before all isRequired inner elements are added to get an extra error)
> 
>     * Finish to add all required elements and then **Add**
> 
> 
> Problem one: the validation doesn't change below **Add Formulation**
> Problem two: the Review & Submit stays in a loop while disabled and showing **Validating...**

Ooh, this is nasty. I think the inner elements are affecting the "FormElementUpdateTrackerState", which they ideally shouldn't. Will have a look at how to fix.

However -- the state shown in your screenshot is actually correct, cos you haven't pressed "Add" on the outer list. For me at least, as soon as I did that, the error message goes away. So it seems only Problem two exists, maybe not Problem one?

PR #803, opened 2021-06-22
#477 part 2 show history panel

nmadruga:
Fixes #477 

### Linked PRs
- Backend PR [#477F](https://github.com/openmsupply/application-manager-server/pull/431) 
- Base on PR #798 
- ~~Based on PR #792~~

### Changes
- 2 queries to get all reviewResponses and applicationResponses related to a question in one template (for an application), thought it was easier to set conditions to filter reviews for applicant making specific ones for Applicant and Reviewer.
- New **hook** `userGetQuestionHistory` to call either one of the queries and built the `historyList` using the type `HistoryElement`. I tried to simplify (for now) so both reviewResponse and applicantResponse can be stored in the same.
- TODO: Show SummaryView for an ApplicantResponse - for now is just showing the text
- When the button "View history" is available, click on the button will popup the modal with the history for that question.
- Some work to do around the layout and designs (not defined yet - so I described the one I'm basing on in the issue #477 
- TODO: ReviewDecision is also just showing the decision stored in db - not localisation. Was thinking if it would be better to just use `DecisionLabel` component to render a reviewResponse. And in that case, might need to change that one to receive more generic props. What do you think?

PR #832, opened 2021-07-07
#811 show stages on history panel

### LINKED ISSUE #811: Fix problem with timeUpdated on reviewResponses and show stages for a group of applicationResponse & reviewResponses in the History Panel

nmadruga:
Fixes #811 

### Linked PRs
- [Back-end#811F](https://github.com/openmsupply/application-manager-server/pull/435)

### Changes
- Show Stage label on top of each block of elements in History Panel
- Update each `application_response` with respective application stage_number at the time it is updated - now each application_response stored the `stage_number` so this can be used when sorting responses per stage
- Change `useGetQuestionReviewHistory` to keep object with elements per stage ordered from latest to oldest to return to HistoryPanel that will only to the rendering of stage labels and respective elements (application_responses, review_responses). 

#### Reviewer
Screen Shot 2021-07-08 at 10 43 53 AM

#### Applicant
Screen Shot 2021-07-08 at 11 29 08 AM

### Tests
Refer to how to test described in PR #803

### DIFF
--- a/src/utils/hooks/useGetQuestionReviewHistory.tsx
+++ b/src/utils/hooks/useGetQuestionReviewHistory.tsx
@@ -1,6 +1,5 @@
 import { useEffect, useState } from 'react'
 import {
-  Review,
   ReviewResponseDecision,
   TemplateElement,
   useGetHistoryForApplicantQuery,
@@ -48,12 +47,13 @@ const useGetQuestionReviewHistory = ({ isApplicant, ...variables }: UseGetQuesti
 
     applicationResponses.nodes.forEach((applicantResponse) => {
       if (!applicantResponse) return
-      const { timeUpdated, stageNumber, application, id, value } = applicantResponse
+      let { stageNumber } = applicantResponse
+      const { timeUpdated, application, id, value } = applicantResponse
       const { firstName, lastName } = application?.user as User
 
       if (!stageNumber) {
-        console.log(`application_reponse ${id} without stage_number: not added to History`)
-        return
+        console.log(`Warning: application_reponse ${id} without any stage_number`)
+        stageNumber = 0
       }
 
       if (!allResponsesByStage[stageNumber]) allResponsesByStage[stageNumber] = {}
@@ -72,8 +72,6 @@ const useGetQuestionReviewHistory = ({ isApplicant, ...variables }: UseGetQuesti
       const { timeUpdated, decision, comment, review } = reviewResponse
       // Avoid breaking app when review is restricted so not returned in query (for Applicant)
       const { levelNumber, reviewer } = review ? review : { levelNumber: 1, reviewer: null }
-      console.log(levelNumber, reviewer)
-
       const stageNumber = review?.stageNumber || 0
 
       if (!allResponsesByStage[stageNumber]) allResponsesByStage[stageNumber] = {}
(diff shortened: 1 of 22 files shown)


src/utils/hooks/useGetQuestionReviewHistory.tsx
```diff
@@ -35,15 +44,22 @@ const useGetQuestionReviewHistory = ({ isApplicant, ...variables }: UseGetQuesti
...
 
     applicationResponses.nodes.forEach((applicantResponse) => {
       if (!applicantResponse) return
-      const { timeUpdated, application, value } = applicantResponse
+      const { timeUpdated, stageNumber, application, id, value } = applicantResponse
       const { firstName, lastName } = application?.user as User
 
+      if (!stageNumber) {
+        console.log(`application_reponse ${id} without stage_number: not added to History`)
+        return
```
CarlosNZ, line 56: If there's a response with a missing stage number (obviously it shouldn't happen), then maybe we should still display it in the UI (with maybe a "NO STAGE" label?), otherwise the user will never know anything is wrong.

CarlosNZ (approved):
Great, we should probably talk about the UI at some point (and maybe get Deb's input), but all works nicely.

One tiny thing for you to consider, not essential.

src/utils/hooks/useGetQuestionReviewHistory.tsx
nmadruga: Hmmm ok. That's a good point. Will work on that after break. Thanks

nmadruga: Fixed: 1c999d0

nmadruga:
I have been able to replicate the demo problem to show Review history now. I made a quick fix for it, by also showing the review in the "NOT FOUND" stage section, since there is no current way to know what is the review_response stage when the applicant user doesn't have permissions to query the `review` table. The review_response would return the review.stage to be used in determining in which stage this review would show on the `HistoryPanel`: 7d6a070

@andreievg @CarlosNZ I'm not sure how we should address that, to allow the stage to be existing in the review_response that has `is_visible_to_applicant`, the only way I can think of is doing the same thing we have done for application_response, adding the `stage_number` to the entity when it gets updated by the user.

CarlosNZ:
> I made a quick fix for it, by also showing the review in the "NOT FOUND" stage section, since there is no current way to know what is the review_response stage when the applicant user doesn't have permissions to query the `review` table

Maybe I'm missing something, but isn't each review_response linked to an application_response, and the application_response has a stage? So wouldn't the linked review_response just have the same stage?

nmadruga:
> > I made a quick fix for it, by also showing the review in the "NOT FOUND" stage section, since there is no current way to know what is the review_response stage when the applicant user doesn't have permissions to query the `review` table
> 
> Maybe I'm missing something, but isn't each review_response linked to an application_response, and the application_response has a stage? So wouldn't the linked review_response just have the same stage?

No, application_response might have been submitted in **Screening** then approved and now application is on **Assessment** (but application_response is still submitted in the **Screening**) and the reviewer ask for changes - the review should show in **Assessment** and application_response in **Screening**.

nmadruga:
Pushed changes to update `review_response` to also have `stage_number` as done for `application_response`, in order to have review stages showing correctly on Applicant's Review history. Required change since `review` table is restricted to not be assessable for applicants, and `stage` was only being set in there.

nmadruga:
Screen Shot 2021-07-11 at 4 18 03 PM

PR #833, opened 2021-07-08
#824 Dashboard layout, pluralisation and remove icons in filters

### LINKED ISSUE #824: Dashboard styles to be aligned with latest designs
Remove "view alls"
Make the header navigate to the list
Remove icons form config
Pluralisation of strings (by Carl)

### LINKED ISSUE #830: Hide unwanted items from the Dashboard
Currently a user can  access the 'Edit User Details' procedure in 2 different places. 

It would be tidier if this was just in 1 place. 

My opinion is that a classic option, such as the top menu bar or the log-in button menu would be the best place to access this type of functionality. 

So would be great if we had a way to hide unwanted items from the user in the Dashboard:

nmadruga:
Fixes #824 
Fixes #830 

### Linked PRs
- Back-end https://github.com/openmsupply/application-manager-server/pull/437

### Changes
- Dashboard layout
Screen Shot 2021-07-08 at 3 56 44 PM
- One category per row, although templates on the same category (now with fixed width) show in the same row or wrap depending on the size of page
- Update on filters to remove icons, show in grey and number after title (now pluralised on back-end)
- Title for each template also has a new `plural` field that is used to display in the Dashboard and links to "View all" applications in the list

### DIFF
--- a/semantic/src/site/globals/site.overrides
+++ b/semantic/src/site/globals/site.overrides
@@ -43,10 +43,6 @@ a:hover {
   color: @attention;
 }
 
-//.information {
-//  color: @information;
-//}
-
 .clickable {
   cursor: pointer;
 }
@@ -980,6 +976,10 @@ a:hover {
         margin: 10px;
       }
     }
+
+    .templates {
+      padding-bottom: 10px;
+    }
   }
   .template {
     margin: 0px 10px 5px 0px;
@@ -989,7 +989,10 @@ a:hover {
     border: 1px solid @surfacesDark;
     border-radius: @borderRadius;
     background-color: @surfacesWhite;
-    box-shadow: @templateBoxShadow;
+
+    .content {
+      min-height: 65px;
+    }
 
     .filters {
       margin: 2px;
--- a/semantic/src/site/globals/site.overrides
+++ b/semantic/src/site/globals/site.overrides
@@ -934,7 +934,7 @@ a:hover {
   width: 100%;
 
   .template-category {
-    .flex-column-center();
+    .flex-column();
     flex-wrap: wrap;
 
     .title {
@@ -982,7 +982,7 @@ a:hover {
     }
   }
   .template {
-    margin: 10px 10px 5px 0px;
+    margin: 0px 10px 5px 0px;
     width: 330px;
     padding: 10px;
     justify-content: space-between;
--- a/semantic/src/site/globals/site.overrides
+++ b/semantic/src/site/globals/site.overrides
@@ -949,86 +949,56 @@ a:hover {
   }
 
   .template {
-    .flex-column();
+    .flex-row();
 
-    .title {
-      .flex-row-center();
-      align-items: start;
-    }
-  }
+    .content {
+      justify-content: space-between;
+      .header {
+        .flex-row-center();
+        align-items: start;
+      }
 
-  .filters {
-    .flex-column();
-    flex-grow: 1;
-  }
+      .filters {
+        .flex-column();
+        flex-grow: 1;
+      }
 
-  .filter {
-    display: flex;
-    align-items: center;
-  }
-
-  .view-all-area {
-    display: flex;
-    max-width: 100%;
-  }
-
-  .view-all {
-    .flex-row-center();
-    align-items: center;
-    text-align: center;
-    flex-grow: 1;
-
-    div {
-      display: flex;
-      align-items: center;
+      .filter {
+        display: flex;
+        align-items: center;
+      }
     }
   }
 }
 
 // Margin and padding
 #dashboard {
-  .template {
-    margin: 15px 15px 0px 0px;
-    width: 330px;
-    border: 1px solid @surfacesLight;
-    border-radius: @borderRadius;
-    background-color: @surfacesWhite;
-    box-shadow: @templateBoxShadow;
-
-    .title {
-      padding: 10px;
-      padding-bottom: 3px;
-      justify-content: space-between;
-
-      .header {
-        margin: 2px;
-      }
-    }
-  }
-
   .template-category {
-    margin: 10px;
-
     .title {
+      margin-left: 10px;
       .header {
         margin: 10px;
       }
     }
   }
+  .template {
+    margin: 10px 10px 5px 0px;
+    width: 330px;
+    padding: 10px;
+    justify-content: space-between;
+    border: 1px solid @surfacesDark;
+    border-radius: @borderRadius;
+    background-color: @surfacesWhite;
+    box-shadow: @templateBoxShadow;
 
-  .filters {
-    margin: 5px;
-    padding: 3px;
-  }
-
-  .view-all {
-    padding: 5px;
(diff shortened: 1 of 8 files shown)


CarlosNZ:
Couple of initial UI comments:

- the category names are closer to the objects above them than they are to the object that actually belong in the category (below). I'd put the space between the category names and the templates below them quite a bit tighter.
- The fonts are weirdly tiny. Can they be the same size as the standard fonts throughout?
- Can the "actions" below the template name (e.g. Submitted Applications) be standard "links" rather than labels? (i.e. they'll appear in blue) I see there's actually 2 different versions in the Miro board, one with grey links, one with blue -- I think the standard blue links are more consistent with the rest of the app
- The "NEW" button should have the little "+" icon. I've already made a standard class for this -- see the "New Application" button on the List page
- I don't love the grey-background-on-hover effect for the links. We've never used this before, can we stick to standard link behaviour (it's a subtle opacity change if I remember rightly)
- The "box-shadow" is weird, can we just have the plain light-grey border (and no box-shadow)?

src/components/Dashboard.tsx
```diff
@@ -48,26 +48,37 @@ const Dashboard: React.FC = () => {
...
-        <Header as="h5">{name}</Header>
-
+      <div className="content">
+        <div className="filters">
+          <Label
+            className="strong-label"
+            as={Link}
+            to={`/applications?type=${code}&user-role=${userRole}`}
+          >
+            {plural}
```
CarlosNZ, line 67: I really don't like the "Plural" thing (see my back-end comment), but if we are going to have it, can it at least have a fallback to the "name" if not present?

But again, I really think we should let the localisation library handle this, not manually specify every plural.

nmadruga:
Almost done with your changes suggestions, will work on the fallback and possibly adding this thing (which is not very trivial so I might leave it out):
Screen Shot 2021-07-09 at 2 50 34 PM

nmadruga:
Current looking like this:
Screen Shot 2021-07-09 at 2 51 29 PM

nmadruga:
Current implementation (with message for applicant when no applications found):
Screen Shot 2021-07-09 at 9 21 26 PM

Overall dashboard looks good on mobile as well:
Screen Shot 2021-07-09 at 9 22 14 PM

CarlosNZ:
I think it looks weird having it all stuck over to the left like that. The "container" should probably be centered so it looks like:
Screen Shot 2021-07-12 at 12 18 53 PM

Are you imagining that three boxes would be on one row? If two is the limit, then probably just set the max-width on the containing div to just fit two elements, and then center that container.

Do you think?

Also -- shouldn't the action lines be left justified in the boxes? It seems a bit odd that some lines are left and some are centered within the boxes.

nmadruga:
> I think it looks weird having it all stuck over to the left like that. The "container" should probably be centered

Idk Carl, personally I prefer the container occupying the full width and only allowing the items to show vertically (when too narrow) or left justified as it is. I mean, this is a **new problem** we only have to consider because we wanted to allow more than one template belonging to the same category to show on the same row. I don't think we need to change this.

> Are you imagining that three boxes would be on one row? If two is the limit, then probably just set the max-width on the containing div to just fit two elements, and then center that container.

Yes, I'm thinking it can have 3 boxes in one row. I don't like the looks of centralising the container. So, no problem if that is changed after, but for now I think this PR can be merged and this sort of decision can be done later - if needed.

nmadruga:
> Also -- shouldn't the action lines be left justified in the boxes? It seems a bit odd that some lines are left and some are centered within the boxes.

That's just as is in the designs. Maybe this can be improved, but what is your suggestion, centralise both or left-align both?

CarlosNZ:
> I don't like the looks of centralising the container

How come? Every other container (Application, Summary, etc.)  is centralised within the larger "content" container, so why make this one the exception?

nmadruga:
> is centralised within the larger "content" container, so why make this one the exception?

Other ones are application-like. This one is more like a Dashboard really, so it should take most of the space.

CarlosNZ:
> Other ones are application-like. This one is more like a Dashboard really, so it should take most of the space

Don't you think it should look like the list (which is not "application-like")? (as in, make it the same width as the list, and center the container within the main content block)

It's fine if it needs to be wide enough for three items per row. But then the width should be set to the width of exactly three items, and then centered, there's even space either side of the three items. (Which'll mean it's more to the left when only two, but that's fine, at least it won't be crazy-wide).

nmadruga:
Pending @andreievg decision. I still like better the way it is, so either merge as it is, or I can add this change if more than one person agrees. Fair?

nmadruga:
Just a quick think about how we display the non-application-like page, applications list. Since it's a table the natural form of this is to be quite wide. So if I'm looking in 75% of my screen it looks as it's full-width:

Screen Shot 2021-07-16 at 12 37 26 PM

When I use 100% of my screen then it was a padding on left and right, but since it is wide it doesn't get centralised as you are suggesting for Dashboards:
Screen Shot 2021-07-16 at 12 37 40 PM

CarlosNZ:
> Pending @andreievg decision. I still like better the way it is, so either merge as it is, or I can add this change if more than one person agrees. Fair?

Just FYI -- we have global variables stored for content widths in `site.variables` for the express purpose of standardising container widths throughout the site:
```
// Widths for various key containers
@applicationFormWidth: 900px;
@applicationListWidth: 900px; // Could be wider if table is getting cluttered?
@applicationSummaryWidth: 900px;
@applicationSummaryContentWidth: 700px;
@applicationFormContentWidth: clamp(350px, 70%, 480px);
```
So no container should be wider than 900px, and those 900px containers should be centred within the whole "application" container. I just don't understand why we wouldn't stick to our own established style standards.

semantic/src/site/globals/site.overrides
```diff
@@ -930,12 +926,12 @@ a:hover {
 
 // Layouts
 #dashboard {
-  .flex-row-center();
-  flex-wrap: wrap;
+  .flex-column();
+  width: 100%;
```
CarlosNZ, line 930:
```suggestion
  .flex-column();

```

CarlosNZ:
@nmadruga -- had a wee play round and I think the only thing we need to do is remove the `width: 100%` from `#dashboard` (line 930 in site.overrides), as suggested in [this comment](https://github.com/openmsupply/application-manager-web-app/pull/833/files#r670966779).

It just means that the container is only as wide as the content, so it'll look good for two or three column layouts:

When there's three columns:
Screen Shot 2021-07-16 at 4 29 56 PM

When there's two columns:
Screen Shot 2021-07-16 at 4 29 44 PM

It's not ideal, -- it should really have a more fixed width, with the boxes and their spacing changing in a more dynamic way, but I think until we can know exactly what we want the content boxes to look like, we can't really lay it out in an ideal way. I think we need to do more to the content boxes themselves, but this will do for now, do you think?

nmadruga:
Seems good for me now @CarlosNZ I'll be merging this one

PR #843, opened 2021-07-15
#834 & #835 Final decision review page

### LINKED ISSUE #834: Display in Final decision page what was the previous stage decision

### LINKED ISSUE #835: Remove need to do create reviews per questions on final decision
And add logic to allow page to show previous review_decision (of stage.level before yours) and set your decision regardless of previous.
Make sure final user can start review even if not having any `reviewQuestionAssignments`

nmadruga:
Fixes #834 
Fixes #835

Sorry about huge PR! Just think it would be easier to check all changes in one-big PR instead of small ones that need to keep up-to-date. Bear with me, each change is explained bellow and also what is missing!

### Linked PRs
- Back-end [#834F-&-#835F](https://github.com/openmsupply/application-manager-server/pull/456)

### Briefing
As discussed in person (Andrei and I) the Final Decision stage doesn't have any reviewQuestionAssignment created, since the user doesn't need to do any review decision per question, only overall decision - that can be different from previous consolidation (in previous stage). Although we still want to display the `ReviewPage` as if this was the consolidation done in previous stage, so the idea was to duplicate reviewResponses done by lastLevel reviewer in previous stage, so now reviewer of Final stage has the same structure generation working and progress. That was almost everything needed, besides that I also had to add a new field to `GetReviewResponsesQuery` to get level 1 review, which is also used in the generation of progress (and displayed in Review page) this is the final result of a Draft Review on Final Decision stage:
Screen Shot 2021-07-20 at 1 15 55 PM
Screen Shot 2021-07-20 at 1 16 06 PM

### Changes:
- Add `isFinalDecision` to `getReviewInfo` query and set as part of FullStructure in newly created `ReviewAssignment` type to store fields that are related to the same object in GraphQL (simplified version - can change to same type if required)
- Added new action `CAN_MAKE_DECISION` and in `generateReviewSectionActions.ts` also check for this option when reviewer has `isFinalDecision` set in `ReviewAssignment` info.
- IMPORTANT change on `generateConsolidatorResponsesProgress.ts ` and `generateReviewerResponsesProgress.ts` instead of having `TotalReviewable` set if `isAssigned` now using `thisReviewResponse` is defined, to cover case of finalDecision when the question isn't assigned but should be counted as part of review (to show in progress).
- Changed query `getReviewResponses.query.ts` to also get `previousOriginalReviewResponses` in previousStage, this will only be fetch when flag **shouldIncludePreviousStage** is set to true (when `isFinalDecision`), needed to get the level 1 reviewResponses which are in a different stage.
- New mutation `remakeReview` that will receive another reviewAssignment to use for duplicating reviewResponses from lastLevel of `previousStage` review. In order to have that, added a props passed in this components:
  - `ReviewHomeWrapper.tsx`
  - `ReviewHome.tsx`
  - `ReviewSectionRow.tsx`
- Change how `ReviewHomeWrapper.tsx` stored the selectedStage to be the `number` instead of `id` and the getAssignmentByStage that can also receive one specific stage and return assignment for that stage (used to get previous)
- Use default cases in `ReviewSectionRowLastActionDate.tsx` and `ReviewSectionRowAssigned.tsx` to always show something in field of ReviewSectionRow (when a new type of action is added).
- Show new action type in each row on `ReviewSectionRowAction.tsx` displayed in ReviewHome
- Quick solution to not show option to edit reviewResponse when finalDecision, not show Edit icon when `isAssigned` is **false** in `ConsolidateReviewDecision.tsx`)

### Special attention on files:
- `src/utils/hooks/useGetReviewStructureForSection/helpers.tsx` have some special cases for when **isFinalDecision**
- New hook `useRemakePreviousReview.ts `
- `src/utils/hooks/useGetDecisionOptions.ts` now returns special case for when **isFinalDecision** to allow NON_CONFORM or CONFORM decisions - always.

### Unrelated quick fixes:
- Updated variable name in `useCreateReview.tsx` to createReview
- Fix reviews not showing for level1 after submitted in `ReviewApplicantResponse.tsx`

### To be done in following PRs:
- ~~Change outcome when final decision happens~~ In back-end PR https://github.com/openmsupply/application-manager-server/pull/464 
- Show action `Make decision` in the list (now is showing `start`)
- Check with team how/where do we want to show previous stage decision

### Bug fixes created - found during implementation - have added as **Sprint 42**:
- #845 in PR #848
- #846 WIP
- #847


CarlosNZ:
Minor thing -- after making final decision, and approving, the application has changed to "APPROVED" in database. However, it's still showing up with "Make decision" action in list.

CarlosNZ (changes requested):
Great, working pretty well.

Just that one bug with the "Make decision" action still showing in the list, even when its been finished, if you could take a look at that? (I thought it might be to do with there being no "COMPLETED" status set, but it still happens even when I manually change the status to COMPLETED) (Also, I've made a fix for that particular thing in Scheduled Actions -- status will automatically go to COMPLETED when Outcome changes from PENDING, so no need to do anything about that)

nmadruga:
> Minor thing -- after making final decision, and approving, the application has changed to "APPROVED" in database. However, it's still showing up with "Make decision" action in list.
> Screen Shot 2021-08-03 at 10 08 40 AM

Wow - thanks for finding this one. I'll fix this now : )

nmadruga:
> > Minor thing -- after making final decision, and approving, the application has changed to "APPROVED" in database. However, it's still showing up with "Make decision" action in list.
> > Screen Shot 2021-08-03 at 10 08 40 AM
> 
> Wow - thanks for finding this one. I'll fix this now : )

Fixed! In [back-end](3372d9a)

Now when Final decision has started the Review (so `review` associated to reviewAssignment != NULL) it would return one of the other actions: `continue` if the review has started or `view` after it has been submitted.
Screen Shot 2021-08-03 at 2 22 36 PM

CarlosNZ (approved):

CarlosNZ:
Looks good -- thanks!

PR #844, opened 2021-07-15
#840 Wait for back-end serial generation

### LINKED ISSUE #840: Application Create -- wait for Serial to be generated in back-end before loading application page
After discussion with @andreivg, we thought the serial should be generated by the back end on Application Create, so:
- front-end mutation doesn't enter a serial
- trigger runs on back-end, generates serial (and provisional name)
- front-end waits for trigger to finish before loading the application with the new serial

CarlosNZ:
Fixes #840.

Requires back-end branch from [PR #454 (Generate Serials)](https://github.com/openmsupply/application-manager-server/pull/454)

### Summary of changes

- the "create" method (in useCreateApplication) returns the id of the newly created application to the "handleCreate" method in "ApplicationCreate", which sets the "id" in "ApplicationState"
- "onCompleted" removed from call to "useCreateApplication"
- a new query hook "useGetApplicationSerialQuery" added to ApplicationCreate, but it's set to `skip` when there is no id set in ApplicationState
- the new hook's "onCompleted" method is set to check the trigger of the new application. The query re-runs (with pollInterval) every half second until the trigger is reset (null), or it times out (10 attempts, just like useLoadApplication).
- when trigger === null, we set the serialNumber in ApplicationState, then do the same "push" to application page 1 that we used to do in the previous "onCompleted()"

**Edit**: ~~I've just realised that this isn't working properly yet for applications that auto-start (like UserRegistration) --- will fix tonight or in the morning.~~ _Sorted -- turns out that userRego (and a couple of others) didn't have core actions, so have added action to those ones manually in back-end branch._

### DIFF
--- a/src/containers/Application/ApplicationCreate.tsx
+++ b/src/containers/Application/ApplicationCreate.tsx
@@ -43,7 +43,11 @@ const ApplicationCreate: React.FC = () => {
 
   // Need to re-query the created application to get the new serial
   const [refetchAttempts, setRefetchAttempts] = useState(0)
-  const { data, error: newApplicationError } = useGetApplicationSerialQuery({
+  const {
+    data,
+    loading: newApplicationLoading,
+    error: newApplicationError,
+  } = useGetApplicationSerialQuery({
     variables: {
       id: id as number,
     },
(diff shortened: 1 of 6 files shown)


nmadruga (approved):
Very good! Thanks : )

PR #848, opened 2021-07-21
#845 Fix wrong label in review home

### LINKED ISSUE #845: Bug: Fix showing wrong action/assigned to you on ReviewHome

nmadruga:
Fix #845 

### Linked PRs
- Base front-end #843 
- Back-end https://github.com/openmsupply/application-manager-server/pull/456

### Changes
- Noticed that `isAssignedToCurrentUser` isn't very accurate, since it can only store if the current review belongs to the current user or not. So, in ReviewHome or ReviewPage, when a section doesn't have ReviewResponses the reason can be:
  - Is Assigned to another
  - Reviewer lower level hasn't submitted
  - Section doesn't have anything to review
#### ReviewPage
- Changed what is stored per `section.assignment` if there is nothing to review (yet) set `isAssignedToCurrentUser` to **false** even though the current user is the same as the reviewer
- In Section show either `Assigned to you` or the progress (if you have started to review)

Example - not showing label "Assigned to another" since that wasn't accurate:
Screen Shot 2021-07-21 at 2 44 42 PM 

#### ReviewHome
-  In have changed the labels a little, so they show ore accurate. See the example below...
- Moved things to more generic `ReviewLabel` components to be reused more easily.
- Add case for `ReviewAction.unknown` after having add a default to be the most common case, it shouldn't show for when action is **unknown**
- Same change as done for ReviewSection to show either nothing or some progress (if you have started to review)

Looking at everyones reviews  (when selected) you can see there is already a Consolidation in progress **In review by Consolidator 2** and that the other Reviewer level 1 has submitted their review **Review submitted by Reviewer 1**. When you haven't started.
Screen Shot 2021-07-21 at 3 03 15 PM

Here is how the consolidator sees everyones review (when selected). Note that since **Reviewer 2** doesn't have a review, it doesn't show for others (since it's a ReviewAssignment that only show to assigners or to  yourself...):
Screen Shot 2021-07-21 at 2 54 12 PM

In another example, **Reviewer 1** can Update their review and see that **Reviewer 2** has submitted.
Screen Shot 2021-07-21 at 2 52 53 PM
And **Reviewer 2** showing only their review (using the selection by default: only yourself) - Old style here, see the slight difference!
Screen Shot 2021-07-21 at 2 47 16 PM


nmadruga:
After spending some time on this I was having a look on this other related issue here: #776 
image

The solution proposed is to align the ApplicationHome and ReviewHome a little more with the designs, having a separated column for status and a simpler one for actions, so reducing the actions that we have now to probably only these 5:
- Start
- Continue
- Assign/Re-assign
- View
- Update

And replacing the ones that were listed in the [diagram](https://drive.google.com/file/d/1Rnaoollk5FFuGzFOOZ3eenkgQQ8efAS9/view?usp=sharing) made for this PR, so all these specific actions would be showing on a status column instead with the action Start, Update or View without any numbers. 

I'm not sure if we should change to do this... But I see why it would be good to have the status showing on each row...

src/utils/helpers/structure/generateReviewSectionActions.ts
```diff
@@ -123,12 +136,17 @@ const generateReviewSectionActions: GenerateSectionActions = ({
...
       isPendingReview: (totalPendingReview || 0) > 0,
       isReviewActive: (totalActive || 0) > 0,
     }
 
+    console.log(section.details, checkMethodProps)
+
     const foundAction = actionDefinitions.find(({ checkMethod }) => checkMethod(checkMethodProps))
 
+    console.log(foundAction)
+
```
CarlosNZ, line 149: Couple of console.logs left here :)

CarlosNZ (approved):
LGTM -- just remove those console.logs

👍

src/utils/helpers/structure/generateReviewSectionActions.ts
nmadruga: Thanks. Removed

PR #851, opened 2021-07-22
#849 Add docs for Date Picker

### LINKED ISSUE #849: Document date picker plugin
As per: https://github.com/openmsupply/application-manager-web-app/pull/778#pullrequestreview-712324879

@CarlosNZ

CarlosNZ:
Fixes #849


documentation/internal/Element-Type-Specs.md
```diff
@@ -518,6 +519,47 @@ Once selected, items are displayed in a "card" view:
...
+<a name="date-picker"/>
+
+### Date Picker
+
+- **type/code**: `datePicker`
+- **category**: `Question`
+
+_A date picker for entering standardised dates or date ranges_
+
+![Search Results](images/Element-Type-Specs-date-picker.png)
```
nmadruga, line 531:
```suggestion
![Date picker](images/Element-Type-Specs-date-picker.png)
```

nmadruga (approved):
Great. I wonder if I should go and change places not using Date-picker to use it now.
It might be a good idea to tide up our existing TestDev and Drug Registration example (and others applications that uses dates and search to use the new ones developed recently.

CarlosNZ:
> Great. I wonder if I should go and change places not using Date-picker to use it now.
> It might be a good idea to tide up our existing TestDev and Drug Registration example (and others applications that uses dates and search to use the new ones developed recently.

Yes, we probably need to go and update the templates in a more systematic way overall. We did quite a bit for the last demo that hasn't made it back into master yet, so we should try and re-capture that. Although it might be easier to do significant updates using the template builder 🤷‍♂️

PR #854, opened 2021-07-23
Fix compilation error in master

andreievg:
We were having this compilation error:

Screen Shot 2021-07-23 at 3 15 09 PM

After merging date picker PR with `react-semantic-ui-datepickers` package added to the project (it was to do with the work I've done when trying to minimise the size of bundle, which happened at the same time as pr adding the package), anyways the optimisation done with `semanticUiReactReExport` was negligible, especially after minimisation of js bundle with TerserPlugin, thus can be removed (now it all compiles and builds)


src/components/Review/HistoryPanel.tsx
```diff
@@ -6,7 +6,7 @@ import HistoryResponseElement from '../PageElements/Elements/HistoryResponseElem
 import { useUserState } from '../../contexts/UserState'
 import { Stage } from '../Review'
 import strings from '../../utils/constants'
-import { StageDetails } from 'utils/types'
+import { StageDetails } from '../../utils/types'
```
CarlosNZ, line 9: Wonder why this needed changing? Was it also causing an error that we didn't notice?

CarlosNZ (approved):
Sweet, thanks for sorting it :)

src/components/Review/HistoryPanel.tsx
andreievg: yeah was wondering the same, but didn't have time to go over it, will investigate if the problem persists

PR #856, opened 2021-07-23
#855 Fix ApplicationCreate bug

### LINKED ISSUE #855: Bug: Something broke with application creation in the latest batch of merges
This has only started happening since the latest merges, suspect something got messed up in the change to the Application Create process.

Basically, if you try and start an application with no Start page (ie. User Reg or User Edit), it'll get into an infinite loop of creating multiple different applications, each causing their own "PROCESSING" trigger and firing off actions, etc. Very ugly.

Still works fine for templates with a Start screen.

CarlosNZ:
Fixes #855 

Turns out to be a very small fix, and seems to be behaving itself now. 🤞

I'm still mystified as to what caused this -- I'm sure it was working fine when I did the original PR last week. *Something* is causing the component to reload and re-run the useEffect, even when `template` is unchanging between renders. 🤷‍♂️

### DIFF
--- a/src/containers/Application/ApplicationCreate.tsx
+++ b/src/containers/Application/ApplicationCreate.tsx
@@ -17,7 +17,6 @@ import { useGetApplicationSerialQuery } from '../../utils/generated/graphql'
 const MAX_REFETCH = 10
 
 const ApplicationCreate: React.FC = () => {
-  console.log('Loading component...')
   const {
     applicationState: { serialNumber, id },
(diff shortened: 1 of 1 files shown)


nmadruga (approved):
Working well. Thanks for noticing this problem and fixing right away.

While testing this one I saw this one happening again (it happens every time on User registrations), not assigned to this Sprint though - just worth mentioning here #866

PR #875, opened 2021-07-28
Fix to stop edit option showing for submitted application - when not updating

nmadruga:
No linked issues.

### Briefing
Quick fix - for another quick fix (not well done - by me)
Before the application submitted was showing the **Edit** icon on all elements - allowing you to change responses after the application was submitted. So, now that's fixed.


CarlosNZ (approved):
LGTM

Issue #866, opened 2021-07-28
Bug: Need to click twice in "Review and Submit" on User registration application


nmadruga:

CarlosNZ:
I don't think this is unique to "User Registration" -- its the old problem of the Button not knowing that the responses are all valid -- it's supposed to wait until the FormElementUpdateTrackerState says it's good to go, but we've never fully fixed this problem.

(If you click outside the text input field and wait a second, you'll only need to push "Review and Submit" once (as long as all responses are valid))

CarlosNZ:
Assigning @andreievg to this, since he knows his way round FormElementUpdateTrackerState

nmadruga:
Thank you! It would be nice if we could just try and keep this very visible bugs under control. I'm aiming for a more stable version so we can release a new demo probably next week?

andreievg:
Hmm, diagnosed the issue and noticed that 'entered' event was proceeded by 'exited' event of previous form element. Best way to replicate is to set throttling to fast 3g, and then it's always replica-table.

There is a note about this in a diagram about validation (I don't think it made it to an issue). Basically need to track `FormElementUpdateTrackerState` by element, i'll spend 15 mins working on this now, if it goes beyond that i'll leave it for the weekend

PR #882, opened 2021-07-29
#866 improve revalidation element state tracking

andreievg:
Fixes: #866 

Ended up changing element tracker to track changes for individual elements, and only supplying `latestChangedElementUpdateTimestamp` to ApplicationPageWrapper and in turn to useGetApplicationStructure as minRefetchTimestampForRevalidation.

Also introduced `sequence` to tracked elements, to make sure overlapping exit event in password plugin doesn't undermine enter event it overlaps.

### Changes

Basic flow is as follows:

Each element can indicate when it's 'entered' via `setIsActive` function prop, ApplicationViewWrapper (AVW) will call setUpdateTracker `setElementEntered` with that element code. Then when element is saved AVW will call setUpdateTracker with `setElementUpdated` with that code. Entered and exited (including start and end value) events for all element are tracked, and if element is active (in focus), latestChangedElementUpdateTimestamp will be Infinity, otherwise, if element looses focus and was changed, then latestChangedElementUpdateTimestamp will be set to Date.now() (more precisely to the latest  changedTimestamp of all updated elements since page was opened). 

When revalidation is requested, ApplicationPageWrapper will pass along `latestChangedElementUpdateTimestamp` to `useGetApplicationStructure` which will re evaluate isValid field when responses are updated (refetched) after `latestChangedElementUpdateTimestamp`

Refer to this behemoth of a diagram: https://github.com/openmsupply/application-manager-web-app/wiki/Application-and-Review-Flow#revalidation-diagram

Also took the liberty of showing a loading indicator (replaces review and summary button content when isValidating)

`NOTE`: I tried to be careful not to break existing functionality, and did some minor test, seems to be all good 🤞 

### Testing

This backend branch  `#866F-testing` has user registration form with just first name and password field, can check out the following with and without this front end branch:

* Create new user registration
* Change to slow or fast 3g
* Enter things in the application and tab through inputs
* When doing confirmation of password, don't tab out but press the review and submit button

After summary page is reached go back to application and try changing password so it's not matching confirmed password anymore, without loosing focus from password input try review and submit button.

### Thoughts

Not happy about the complexity of the thing, but should be safe now for use cases like:
* Long validation on text input (say some slow endpoint)
* Making changes in that long validation input then quickly tabbing out into another field and tabbing out of that one and quickly pressing review and summary

### TODO

Need to update docs and diagram

### DIFF
--- a/semantic/src/site/globals/site.overrides
+++ b/semantic/src/site/globals/site.overrides
@@ -666,6 +666,9 @@ a:hover {
     .flex-row();
     flex-grow: 1;
     justify-content: flex-end;
+    .loader {
+      margin-left: 10px;
+    }
   }
 }
 
--- a/src/components/Application/Navigation.tsx
+++ b/src/components/Application/Navigation.tsx
@@ -1,5 +1,5 @@
 import React from 'react'
-import { Button, Container, Icon } from 'semantic-ui-react'
+import { Button, Container, Icon, Loader } from 'semantic-ui-react'
 import {
   MethodRevalidate,
   MethodToCallProps,
@@ -16,6 +16,7 @@ interface NavigationProps {
   sections: SectionsStructure
   serialNumber: string
   requestRevalidation: MethodRevalidate
+  isValidating: boolean
 }
 
 const Navigation: React.FC<NavigationProps> = ({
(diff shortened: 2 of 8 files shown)


CarlosNZ (approved):
Great, works well, and I really like the "Validating" loader on the button, makes it very clear to user.

Issue #877, opened 2021-07-28
Question about the list UI -- should we show "Outcome" column?


CarlosNZ:
An "expired" application looks exactly the same as an "approved" application in the list. Should we show the Outcome column, or maybe some other way to highlight "dead" applications (like a different colour?)?

Issue #829, opened 2021-07-06
Main menu bar -- design and implement


CarlosNZ:
The current menu links are pretty ad hoc, we should probably come up with something a bit more "intentional".

Issue #677, opened 2021-05-06
Add Menu of applications list at the top and show the current selected as the current list 


nmadruga:
(Maybe also: the current template type of application/review opened)

Issue #636, opened 2021-04-27
Main header bar -- collapse to "narrow" version on scroll


CarlosNZ:
Current thinking - show it narrowed always

As per Deb's email:

> From our research, in most modern sites, the header shows in ‘full’ when the page content is scrolled fully to the top, and then diminishes as page is scrolled down. Should we consider this ?
> Yes you can make the header behave like modern sites

PR #885, opened 2021-08-02
#829 #677 #877 #785 #636 UI improvements

CarlosNZ:
Implements #829 #677 #877 #785 (and partially #636, which should do for now)

Many small UI tweaks to match the design specs and take care of some outstanding layout issues

Requires [back-end branch from PR#473](https://github.com/openmsupply/application-manager-server/pull/473)

UI improvements

### Summary of Changes

- Full-width Header and Footer (and slightly narrower to save space)
- Added "brand area" to Header with logo and application name -- strings and logo can be replaced by each country to suit (using "mFlow Application Manager" for now, will change when we have an official name)
- General tweaking and dynamic sizing of all main container widths (using `clamp()`)
- Update menu bar: Dashboard, Application List (dropdown), Outcomes (dropdown), and Admin Config (dropdown)
   - the currently active link shows in **bold**
   - items only show if allowed (e.g. isAdmin, or restricted Outcomes/Templates list)
- Update user menu (button with name) to have dropdown for "Edit User details" and "Log out" (will add more as required)
- Moved lots of UI strings that were hard-coded  into constants 

Screen Shot 2021-08-02 at 11 35 47 PM

- Tweaks to Application List table:
   - Applicant and Org now show (in same cell as application name)
   - Outcome column visible (with coloured badges reflecting outcome) (Note: only shows on applications no longer "in-progress")


semantic/src/site/globals/site.overrides
```diff
@@ -285,7 +304,7 @@ a:hover {
   align-items: center;
   height: 100%;
   max-height: 55px;
-  margin-top: 10px;
+  // margin-top: 10px;
```
nmadruga, line 307: Remove please

semantic/src/site/globals/site.overrides
```diff
@@ -531,7 +551,7 @@ a:hover {
 
 #list-container {
   padding: 25px 0;
-  max-width: @applicationListWidth;
+  width: @applicationListWidth;
```
nmadruga, line 554: Maybe this is making the table looks bad in some resolutions... is cutting off: from 770px lenght - not a bug deal though

src/components/List/Cells/OutcomeCell.tsx
```diff
@@ -1,7 +1,27 @@
...
 import { CellProps } from '../../../utils/types'
+import { Label, SemanticCOLORS, SemanticICONS } from 'semantic-ui-react'
+import strings from '../../../utils/constants'
+import { ApplicationOutcome } from '../../../utils/generated/graphql'
 
-const OutcomeCell: React.FC<CellProps> = ({ application }) => <p>{application.outcome}</p>
+const outcomeDisplayMap: {
+  [key in ApplicationOutcome]: { icon: SemanticICONS; color: SemanticCOLORS; text: string }
+} = {
+  PENDING: { icon: 'hourglass half', color: 'grey', text: '' }, // Not used
```
nmadruga, line 10: Shouldn't we show something for pending?

src/components/List/Cells/OutcomeCell.tsx
```diff
@@ -1,7 +1,27 @@
...
+  APPROVED: { icon: 'check circle', color: 'green', text: strings.OUTCOME_APPROVED },
+  REJECTED: { icon: 'cancel', color: 'pink', text: strings.OUTCOME_REJECTED },
+  EXPIRED: { icon: 'time', color: 'orange', text: strings.OUTCOME_EXPIRED },
+  WITHDRAWN: { icon: 'user cancel', color: 'yellow', text: strings.OUTCOME_WITHDRAWN },
+}
+
+const OutcomeCell: React.FC<CellProps> = ({ application }) => {
```
nmadruga, line 17: This looks very nice. : D

src/components/Main/Footer.tsx
```diff
@@ -4,7 +4,7 @@ import strings from '../../utils/constants'
 const logo = require('../../../images/logos/logo_512.png').default
 
 const Footer: React.FC = () => (
-  <Container id="footer">
+  <Container id="footer" fluid>
```
nmadruga, line 7: Not sure why, Footer on mobile look strange when looking at the Applications list:
Screen Shot 2021-08-03 at 3 48 47 PM

Screen Shot 2021-08-03 at 3 48 53 PM


In Dashboard look good:
Screen Shot 2021-08-03 at 3 48 08 PM

src/containers/User/UserArea.tsx
```diff
@@ -108,14 +198,21 @@ const OrgSelector: React.FC<{ user: User; orgs: OrganisationSimple[]; onLogin: F
...
-          {user?.firstName || ''} {user?.lastName || ''}
-        </Button.Content>
-        <Button.Content hidden>
-          <Icon name="log out" />
+          <Dropdown text={`${user?.firstName || ''} ${user?.lastName || ''}`}>
+            <Dropdown.Menu>
+              <Dropdown.Item
+                icon="edit"
+                text={strings.MENU_EDIT_USER}
+                onClick={() => push('/application/new?type=UserEdit')}
```
nmadruga, line 211: Should this be included in some config?

nmadruga (changes requested):
Overall comments:
- should we rename file `image_resources/msupply logo/msupply_logo_irims_variant.psd ` to mFlow?
- Mobile is looking good, but probably need a hamburger for menus so titles don't collapse this way (Add a new issue?) see screenshot with mobile resolution used to test:
Screen Shot 2021-08-03 at 3 22 01 PM
- One thing that is very annoying and could possibly be included in this batch of improvements is where we should the buttons **preview** and **next**, if the page is not long enough to reach the bottom, the position changes slightly from one page to to the next. It is annoying to navigate. Maybe we can set a minimum size for page height
- On the top menus, you mentioned:
> the currently active link shows in bold
It is a little difficult to see that, the item is not only bold but the background has a different colour, which is the same as the `default` item selected - when not seeing that page. So it's not clear what is a default and what is the current page selected. See example - while looking at Company rego list page, both menus show something:
Screen Shot 2021-08-03 at 3 31 55 PM
Screen Shot 2021-08-03 at 3 32 04 PM
- Not part of this PR but just keep noticing that tables look very ugly on mobile currently:
Screen Shot 2021-08-03 at 3 37 10 PM
Screen Shot 2021-08-03 at 3 36 59 PM
- The repetition of the Applicant action on `StatusCell` is not ideal... I think we need to work on that - in a different issue though: #776 
- When trying to load the **Edit user** template this error shows:
Screen Shot 2021-08-03 at 3 51 44 PM

semantic/src/site/globals/site.overrides
CarlosNZ: 122b59cf

semantic/src/site/globals/site.overrides
CarlosNZ: The table always went really ugly at ~750px, so I set a minimum to that until we can improve mobile table display properly. So it'll spill off the screen, but at least it won't go completely unreadable like it does if you let it go smaller than 750.

src/components/List/Cells/OutcomeCell.tsx
CarlosNZ: I think from a user's point of view, having no outcome makes sense -- outcome implies "finished". We're kind of just using "Pending" as an internal state for our purposes, but don't think we need to show it to the user.

Also, I figured that the majority of the time, people are going to be looking at active applications, so no need to have a million badges everywhere. I just wanted the completed applications to stand out from the rest.

I'm open to better ways to style this, but I think it requires a bit more design thought. (one idea was to shade the row background to show different outcome states, but I didn't want to go too far down this road without a designer having a crack)

CarlosNZ:
>     * should we rename file `image_resources/msupply logo/msupply_logo_irims_variant.psd ` to mFlow?

Probably not yet. But soon hopefully :) 

> Mobile is looking good, but probably need a hamburger for menus so titles don't collapse this way (Add a new issue?) see screenshot with mobile resolution used to test:

Yeah, we've kind of neglected mobile a bit. For now I'll just add "hide-on-mobile" to the "brand-area" which at least makes it useable. Maybe we should start a new Epic for "Mobile display" and deal with all of this stuff (cos there's quite a bit!) at once when we have more time?

>     * One thing that is very annoying and could possibly be included in this batch of improvements is where we should the buttons **preview** and **next**, if the page is not long enough to reach the bottom, the position changes slightly from one page to to the next. It is annoying to navigate.

I thought Andrei's last improvements to this were okay? Much better than it was at least. Maybe add a new issue with your suggested solution?

> Maybe we can set a minimum size for page height?

I quickly tried that, but it just makes the container bigger and doesn't move the button row. It's a bit more involved and would require tweaking the surrounding containers flex settings. 

> Not part of this PR but just keep noticing that tables look very ugly on mobile currently:

Yup!

> the currently active link shows in bold
> It is a little difficult to see that, the item is not only bold but the background has a different colour, which is the same as the `default` item selected - when not seeing that page. So it's not clear what is a default and what is the current page selected. See example - while looking at Company rego list page, both menus show something:

Oh, I think you misunderstood. I'm not talking about the Dropdown selection -- that's all just default behaviour. I just meant the link headings ("Dashboard", "Outcome", etc.) show in bold for the current page (it's fairly subtle, but I think it's useful)

> The repetition of the Applicant action on StatusCell is not ideal... I think we need to work on that - in a different issue though

I agree. I didn't touch that cos I wasn't sure what a good solution would be -- we can discuss it at some point.

> When trying to load the Edit user template this error shows:

Hmmm, permissions problem, but not related to this issue (happens on master too). We'll need to make a new issue I guess

src/components/Main/Footer.tsx
CarlosNZ: I can't reproduce this. Looks okay for me. (in Firefox)

- I just tried it in Chrome, and it looks weird as, scaling is all wrong. Seems correct in Safari too

src/containers/User/UserArea.tsx
CarlosNZ: What do you mean?

CarlosNZ:
> > When trying to load the Edit user template this error shows:
> 
> Hmmm, permissions problem, but not related to this issue (happens on master too). We'll need to make a new issue I guess

Okay, I've made an issue for this: https://github.com/openmsupply/application-manager-server/issues/489

CarlosNZ:
Also, I've gone through and localised all the other values I could find, including column names, filter names and filter enums:

afaf6a2 
261f47d
1464e44

nmadruga (approved):

src/containers/User/UserArea.tsx
nmadruga: UserEdit - if code changes we need to find where is used and change.
Would be better to have that as a config that users can easily change in case they would like to have another one instead.

PR #886, opened 2021-08-03
#869 Checkboxes bug fix

### LINKED ISSUE #869: Bug: something weird with checkboxes
Specifically: Start a Feature Showcase application and go to the checkboxes page (S1, P4). Start toggling checkboxes in the "Lotsa boxes" question. You'll see there is quite a delay in the response, and also, sometimes they automatically switch themselves off after being selected. 

Must be something to do with how the local state "value" and the stored response are interacting with each other.

CarlosNZ:
Fixes #869.

I'm not entirely sure why this is necessary. Basically, the "checkboxes" array in Checkbox ApplicationView was triggering the useEffect every time the parameters were re-evaluated (i.e. on response changes), even when it hadn't changed its value(s). Which was in turn triggering a re-calculation of the checkboxElements objects and re-saving the response, causing an infinite loop.

I've added a value equality check to the Wrapper before updating the evaluatedParameters, which seems to prevent this loop (I did it in the Wrapper so it would also work for any other affected plugins), but it seems a bit hacky to have to manually check for changed values when React normally takes care of that for us.

### DIFF
--- a/src/formElementPlugins/ApplicationViewWrapper.tsx
+++ b/src/formElementPlugins/ApplicationViewWrapper.tsx
@@ -12,6 +12,7 @@ import {
 import { useUserState } from '../contexts/UserState'
 import validate from './defaultValidate'
 import evaluateExpression, { isEvaluationExpression } from '@openmsupply/expression-evaluator'
+import { isEqual } from 'lodash'
 import { Form, Icon } from 'semantic-ui-react'
 import Markdown from '../utils/helpers/semanticReactMarkdown'
 import strings from '../utils/constants'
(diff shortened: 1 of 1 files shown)


nmadruga (approved):
Looks great. I don't think it's hacky we need to find out the reason why it's not doing a proper check, but all this code does is do a proper check. So all good to me.

CarlosNZ:
> we need to find out the reason why it's not doing a proper check

I think I know why -- it's a completely new copy of the `checkboxes` object, and I guess React only compares by reference. Was just wondering if we should be doing it a different way, but I can't actually think how since it gets re-evaluted and returns a new instance of the object.

Anyway, all good -- thanks for taking a look.

PR #887, opened 2021-08-03
#774 #775 Fix sticky glitches

### LINKED ISSUE #774: Progress bar not smooth transition - should be fixed to go along with page on new/edit application
After scrolling to see the next question, the progress bar just skips to the bottom of the container like this:
Screen Shot 2021-06-10 at 12 48 07 PM

### LINKED ISSUE #775: Flickering sections on application summary
Screen Shot 2021-06-10 at 12 54 11 PM

Ah, the picture is not showing it, but happens when crossing 2 sections. It just keeps both at a flickering state. Weird.

CarlosNZ:
Fixes #774 -- prevents ProgressArea from jumping to bottom of container.

**Edit**: Also fixes #775.

Specified a Ref as per [Semantic docs](https://react.semantic-ui.com/modules/sticky/#types-pushing) so the Sticky knows what container to be connected to.


nmadruga:
Nice one!! Will start reviewing this now.

nmadruga (changes requested):
The sections stop flickering! Which is great. But I found a bug... if you scroll down is good, but after you reach the end and start to scroll up, the first section doesn't show on the top:
Screen Shot 2021-08-03 at 4 04 41 PM

CarlosNZ:
> The sections stop flickering! Which is great. But I found a bug... if you scroll down is good, but after you reach the end and start to scroll up, the first section doesn't show on the top:

Good catch, thanks. I think I've fixed it by removing the "bottomOffset" e58ce5df

nmadruga:
The bug is fixed! Yey and the original one as well!!

Now, just trying to use the time to make this perfect... Please bare with me, hopefully is not a big change required:
1. When scrolling down the title passes a tiny bit out of the section box, it only disappears when touches the next title
Screen Shot 2021-08-04 at 10 16 26 AM

2. After the 2 title sticks if you scroll back up there is no title displaying in the top section - you need to scroll all the way til the top - that's very minor... But ideally the title should show again once the page start showing the section above. We can discuss this IRL if that's better. I will post a screenshot to illustrate anyway : )
Screen Shot 2021-08-04 at 10 20 31 AM

CarlosNZ:
>     1. When scrolling down the title passes a tiny bit out of the section box, it only disappears when touches the next title

Okay, sorted by moving the ref inside the div dbe7706d

>     2. After the 2 title sticks if you scroll back up there is no title displaying in the top section - you need to scroll all the way til the top - that's very minor... But ideally the title should show again once the page start showing the section above. We can discuss this IRL if that's better. I will post a screenshot to illustrate anyway : )

Yeah, I think that's beyond what we can do here. That would involve modifying the semantic-ui code I think. The current behaviour matches all the demos on the Semantic website.

CarlosNZ:
Bad news -- there's still a bit of a flicker when the Accordion is closed. :(

I actually think the problems are due to the Sticky being inside an Accordion. There's this note on the website:
> Sticky behavior is erratic inside Sidebar.Pushable. To avoid issues, place Sticky outside of any Sidebar.Pushable element. See Semantic-Org/Semantic-UI-React#2897 for more details.

Makes me think a similar problem exists with a sticky inside Accordion -- don't think it's really meant to be used that way. 🤷‍♂️

nmadruga:
> Bad news -- there's still a bit of a flicker when the Accordion is closed. :(
> 
> I actually think the problems are due to the Sticky being inside an Accordion. There's this note on the website:
> 
> > Sticky behavior is erratic inside Sidebar.Pushable. To avoid issues, place Sticky outside of any Sidebar.Pushable element. See [Semantic-Org/Semantic-UI-React#2897](https://github.com/Semantic-Org/Semantic-UI-React/issues/2897) for more details.
> 
> Makes me think a similar problem exists with a sticky inside Accordion -- don't think it's really meant to be used that way. 🤷‍♂️

Can we disable it when the Accordion is closed? I think it shouldn't do that while closed right. So we could add - sticky component when open or just regular Segment/div when closed.

CarlosNZ:
> Can we disable it when the Accordion is closed? I think it shouldn't do that while closed right. So we could add - sticky component when open or just regular Segment/div when closed.

Good idea -- I'll give it a go...

CarlosNZ:
> > Can we disable it when the Accordion is closed? I think it shouldn't do that while closed right. So we could add - sticky component when open or just regular Segment/div when closed.
> 
> Good idea -- I'll give it a go...

Sweeet! 🥳 Great suggestion -- works a treat, and actually makes more sense from a user POV. 7f4f1536

nmadruga (approved):

PR #894, opened 2021-08-05
#893 Radio button default fix

### LINKED ISSUE #893: Small bug: In Feature showcase Section1 page2 the radioChoice which has default: 1 is starting unselected

CarlosNZ:
Fixes #893

I think I must have removed this line for the "onSave" condition when I was last updating the default-saving on first load issue.

Good spotting, @nmadruga -- thanks!


nmadruga (approved):
LGTM Thanks : )

PR #895, opened 2021-08-09
#768 Number plugin

### LINKED ISSUE #768: Text input element: Number plugin
EDIT: See [comment below](https://github.com/openmsupply/application-manager-web-app/issues/768#issuecomment-861087848) for proposed Number plugin

---

I just realised that all shortText responses are saved as strings, so we can't do calculations with them in the evaluator (such as adding multiple responses together).

Suggest adding an additional (optional) booean parameter called "number", which stores an additional "number" field in the response if `true`.

So a standard shortText response is:
```
{
 text: "my response"
}
```

and one with the "number" field activated would be:
```
{
 text: "1000",
 number: 1000
}
```

Then we can use `responses.code.number` to perform additional calculations in the evaluator

The "toNumber" function should first strip all non-digit characters, so it should be able to take in a text input of "$5,000" and store the number `5000`

CarlosNZ:
Fixes #768 

This has been almost done for a while, so I just dusted it off and quickly finished it.

Requires branch from [back-end PR #500](https://github.com/openmsupply/application-manager-server/pull/500) to demonstrate it in action -- see FeatureShowcase, Section 3, pages 5-6

The plugin can be used in two different ways:

- a "simple" mode, which just has a plain number, but it has an incrementing "stepper" clicker in the UI -- useful for selecting numbers in a specific range, like "number of ingredients" or something like that
- with "simple" set to `false`, number can be formatted to a wide range of formats, with correct commas (local dependent), currency options, or even units.

Please see docs for detailed parameters


src/formElementPlugins/number/src/ApplicationView.tsx
```diff
@@ -0,0 +1,169 @@
...
+            : null
+        }
+      />
+    </>
+  )
+}
+
+const parseInput = (
+  textInput: string | null | undefined,
+  numberFormatter: any,
```
nmadruga, line 131: Maybe good to use a type here? I was wondering would this fail if no `numberFormat` is passed?

src/formElementPlugins/number/src/ApplicationView.tsx
```diff
@@ -0,0 +1,169 @@
...
+  simple: boolean,
+  prefix: string,
+  suffix: string
+): [number | null, string | null] => {
+  if (!textInput) return [null, null]
+  const number: number = Number(textInput?.replace(/[^\d\.\-]/g, ''))
```
nmadruga, line 137: For me and others not fluent in Regex - can have a tip comment above explaining what this is doing? In a few words

src/formElementPlugins/number/src/ApplicationView.tsx
```diff
@@ -0,0 +1,169 @@
...
+    if (!textValue) return
+    handleLoseFocus()
+  }, [locale, minValue, maxValue, currency, prefix, suffix, maxSignificantDigits, type])
+
+  function handleChange(e: any) {
+    const text = e.target.value
+    const [number, _] = parseInput(text, numberFormatter, simple, prefix, suffix)
+    setInternalValidation(customValidate(number, type, minValue, maxValue))
+    onUpdate(text)
+    setTextValue(text)
```
nmadruga, line 78: Hmmm not sure I understand this - in case of a change why text isn't parsed?

nmadruga: Also maybe clearer to do similar to other functions:
```
const validation = customValidate(number, type, minValue, maxValue)
    setInternalValidation(validation)
```

src/formElementPlugins/number/src/ApplicationView.tsx
```diff
@@ -0,0 +1,169 @@
...
+  return [number, text]
+}
+
+const customValidate = (
+  number: number | null | undefined,
+  type: NumberType,
+  minValue: number = -Infinity,
+  maxValue: number = Infinity
+): { isValid: boolean; validationMessage?: string } => {
+  if (number === NaN || number === null) return { isValid: true }
```
nmadruga, line 150:
```suggestion
  if (number === NaN || number === null) return { isValid: false }
```

? Not sure why NAN or Null would return a true validation, I'm confused

nmadruga (commented):
Ok. that looks nice. I just made very few suggestions/questions for you and will be starting to test using Front-end PR as well. Great job, so far it's great.

src/formElementPlugins/number/src/ApplicationView.tsx
CarlosNZ: I agree -- any idea how to figure out what the "type" should be?

> I was wondering would this fail if no `numberFormat` is passed?

That won't happen, since numberFormatter is constructed at the top of the module, and uses default values if appropriate parameters are not provided.

src/formElementPlugins/number/src/ApplicationView.tsx
CarlosNZ: Ha ha, I'm not sure anyone is *fluent* in regex -- definitely wasn't designed with human readability in mind!

But yeah sure, I'll add an explanation: ce87618b

`  // Strip out any text that isn't a digit, "-" or "." to create the number`

src/formElementPlugins/number/src/ApplicationView.tsx
CarlosNZ: > in case of a change why text isn't parsed?

While user is typing, we don't want their entered text to change at all.

But once they're finished (i.e. handleLoseFocus), we reformat the entered text according to the plugin settings.

E.g. if the entry is set to display their answer in NZ dollars:

They enter `123` -- nothing changes while they type, but the number "123" is parsed from it.

Once they move out of the field, the displayed "text" changes to "$123.00"

CarlosNZ: > Also maybe clearer to do similar to other functions:
> 
> ```
> const validation = customValidate(number, type, minValue, maxValue)
>     setInternalValidation(validation)
> ```

Not quite sure what you mean here.

src/formElementPlugins/number/src/ApplicationView.tsx
CarlosNZ: Yeah, good point. I was thinking that "NULL" responses are valid by the normal rules of our plugins, but I've updated it now to give more useful error state when a non-number is entered. 37c27c1d

Thanks

src/formElementPlugins/number/src/ApplicationView.tsx
CarlosNZ: > I agree -- any idea how to figure out what the "type" should be?

Got it! bff32532

nmadruga (approved):
Nice. Thanks for the changes

PR #897, opened 2021-08-10
#896 UI selector for template categories

### LINKED ISSUE #896: Template Builder -- category creator/selector should be able to specify the UI area which it appears (using ui_location)

CarlosNZ:
Fixes #896

Quickly added UI in template builder for selecting where template category shows. Just uses @andreievg 's existing UI components (checkbox mainly) -- the only styling is a small addition to the paragraph font in the category block.
Screen Shot 2021-08-10 at 12 33 31 PM

I've also localised all the strings in Categories.tsx -- at some stage someone needs to go through and localise the rest of the Template Builder strings.

**Question**: if anyone knows why I needed to use `as UiLocation` whenever I used the (GraphQL-defined) "UiLocation" type, that'd be good to figure out. I don't get why Typescript required this when all the other fields on `CategoryUpdate` (code, icon, title, etc.) don't need this. 

**Edit**: Also updated most of the page titles (using `usePageTitle` hook)


andreievg (approved):
Look good, thanks.

> if anyone knows why I needed to use as UiLocation whenever I used the (GraphQL-defined) "UiLocation" type, that'd be good to figure out. I don't get why Typescript required this when all the other fields on CategoryUpdate (code, icon, title, etc.) don't need this.

>  value={updateState.uiLocation.includes(key as UiLocation)}

and

>    uiLocation: updateUiLocationArray(key as UiLocation, value),

Object entries would loose the type mapping, if you use an array it works, but would still compile if new ui locations are added (we possibly want it to not compile at this stage to make sure all ui locations have a toggle in builder).

> const newLocations = (updateState?.uiLocation as UiLocation[]).filter((loc) => loc !== location)

don't need to cast it here, just [] would do.

> uiLocation: (selectedCategory?.uiLocation || []) as UiLocation[]

the content of uiLocation array is 'Maybe' so could potentially be null, thus cast is needed

check out this diff: `https://github.com/openmsupply/application-manager-web-app/compare/%23896-ui-selector-for-template-categories...896-types`

CarlosNZ:
> check out this diff: `https://github.com/openmsupply/application-manager-web-app/compare/%23896-ui-selector-for-template-categories...896-types`

Great, thanks for the suggestions -- I used your solution, works nicely 5bc54e13

Cool, won't merge this until the base PRs are also approved.

PR #898, opened 2021-08-11
add version to build

andreievg:
Needs back end branch in this PR: https://github.com/openmsupply/application-manager-server/pull/502

I've added this to demo-08-08-2021 branch (current demo server branch).

Commit should  pretty straight forward.

### DIFF
--- a/src/components/Main/Footer.tsx
+++ b/src/components/Main/Footer.tsx
@@ -1,5 +1,6 @@
 import React from 'react'
 import { Container, Image } from 'semantic-ui-react'
+import config from '../../config'
 import strings from '../../utils/constants'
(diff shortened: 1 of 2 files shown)


CarlosNZ:
> I've added this to demo-08-08-2021 branch (current demo server branch).

I've started a new one from that from today `demo-11-08-2021`. I'll pull these into it before we next update the server.

CarlosNZ (approved):
Cool, thanks. ~~So do have to manually put the version no in the config file?~~ (nvm, I see the back-end build script replaces the string)

Can we set it up so the version field in package.json gets auto-incremented with each build?

Issue #899, opened 2021-08-12
Template export exporting too many tables.


CarlosNZ:
When using the "export" feature of the template builder, the resulting snapshot file also includes the full "filter" and "permissionName" tables. 

Probably the "options" settings for the snapshot command run by the front-end are incorrect? We only want the permissionNames and filters that are actually used by this template.

If this is tricky, then maybe it would be better to not export them at all -- just the references. I'd rather have it fail due to a missing permissionName than import a whole bunch we don't want.

Here is a "user registration" template exported like this: 
[UserRegistration-1.zip](https://github.com/openmsupply/application-manager-web-app/files/6972629/UserRegistration-1.zip)

PR #900, opened 2021-08-12
#899 template export correctly filters permission names and filter linked …

andreievg:
Fixes: #899 

Tables that are hierarchical reference depended to template would automatically inherit template (or their reference) filter. You can check this by adding `console.log(gql)` to:
https://github.com/openmsupply/application-manager-server/blob/136612028b7be0e8e2c1cbfc3aaaf96703a2e188/src/components/exportAndImport/exportToJson.ts#L83-L85

But other tables like permissionName and filter are not filtered accordingly, this is fixed in this PR:

### Changes

* Added further filter criteria  to getFilterBody (which is used in export and duplicate functionality).

### Testing

* Adding a log as per described above, and without this commit run an export on template, see console out put and snapshot.json
* Run export with this commit and note the difference in console output and resulting snapshot.json

### DIFF
--- a/src/containers/TemplateBuilder/shared/OperationContextHelpers.tsx
+++ b/src/containers/TemplateBuilder/shared/OperationContextHelpers.tsx
@@ -249,7 +249,15 @@ export const importTemplate: ImportTemplateHelper =
   }
 
 const getFitlerBody = (id: number) => {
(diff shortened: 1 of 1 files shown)


nmadruga (approved):
LGTM

Issue #901, opened 2021-08-12
Github action fires when pushing to any branch


andreievg:
Notice the following:

![Screen Shot 2021-08-12 at 4 13 15 PM](https://user-images.githubusercontent.com/26194949/129137297-2e9873e9-4b66-4501-9419-f97e4e7ab90e.png)

And look at: build_docs_site.yml

Need to make change to only run action when push to main/master branch

https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-using-multiple-events-with-activity-types-or-configuration

PR #902, opened 2021-08-12
#901 only fire action on master branch push

andreievg:
Fixes: #901 

Haven't tested, would assume it would work when merged to master.

@wlthomson how does this look ?


wlthomson (approved):
Not tested but looks good according to the [docs](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions) 🎉.

PR #906, opened 2021-08-13
#905 Add selectedValuesArray to response

### LINKED ISSUE #905: Checkbox plugin -- need response stored as array as well
Currently, the responses are stored in an object, with keys as number indexes (or custom).

This is fine when we need to refer to specific items, but if the options list is dynamic and we just want to access an array of properties from all selections, we can't really do it with the current evaluator.

Suggestion is to also add a "selectedValuesArray" property to the response, which has exactly that.

(NB. This came up when trying to update the "grantPermissions" template to have a dynamic list of permissionsNames. It's easy to create the options list dynamically, but not to create an array of selected permissions from the response)

CarlosNZ:
Fix #905 

Small additional to the response object to make it easier to query dynamic options lists.

### DIFF
--- a/docs/internal/Element-Type-Specs.md
+++ b/docs/internal/Element-Type-Specs.md
@@ -331,6 +331,9 @@ _One or more checkboxes, any number of which can be selected/toggled_
         <key-name-2> : { text: <text value>, isSelected: <boolean>}
         ... for all checkbox keys
         }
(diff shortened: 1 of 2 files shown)


nmadruga (approved):
LGTM

seems nice, haven't tested. But seems as a pretty straight-forward addition.

Issue #517, opened 2021-03-25
Epic #58 - Edit organisation


nmadruga:
## Description

Edit organisation template that should be only visible for organisation owners - allow to request to changes on organisation. Editing the name, address, etc.

## Documentation
- [ ] TODO: Add as part of [Epic #38](338) Public docs & tutorials

## Issues

PR #910, opened 2021-08-17
#510F Admin UI location category

CarlosNZ:
See complementary [back-end PR #517](https://github.com/openmsupply/application-manager-server/pull/517)

**ALSO**: Updated the "Submission message" string so it's more general (and makes sense for Admin/User applications that are automatic and don't have reviews)


nmadruga (approved):
LGTM

Issue #506, opened 2021-03-25
Epic #47 - Display available actions


nmadruga:
## Description

Each user should see their possible actions to act on each application - based on their role (permissions). The options are visible on the applications list and on each application/review home page. 

Users might have more than one role sometime (e.g. Reviewer, Assigner). And that should result usually in more than one action available.

## Documentation
- [ ] TODO: Add as part of [Epic #38](338) Public docs & tutorials

### Sprint 25:
- [ ] #227 Display actionable in cell on column for Reviewer and Consolidator - Effort 3 ~**In Progress**
- [ ] #189 Display 'changes requested' on application list next to applications - Effort 2 ~**In Progress**

### Sprint 26:
- [ ] #293 Show Action to Assign/Re-assign in applications for Assigner user in the List view (would be good if could be using the current user permission) - Effort 3
- [ ] #310 Document and create new filter to list application assigned/not assigned - **Effort to be decided**
- [ ] #311 Deal with users that have more than on roles: Assigner and others - **Effort to be decided**
- [ ] #474 Documentation on actions displaying on each application for different users - Effort 1

PR #912, opened 2021-08-17
#506 template builder fixes

andreievg:
Partial fixes to https://github.com/openmsupply/application-manager-server/issues/506 

### Changes

see commits, and new crossed out point in https://github.com/openmsupply/application-manager-server/issues/506


CarlosNZ:
Cool, having a look through this now.

Can we do something about the layout for the Actions -- it should at least wrap onto multiple rows:
Screen Shot 2021-08-18 at 9 28 16 AM

CarlosNZ:
The validation expression is not showing up in the GUI. In "Common Properties", there is an "isValid" field, but it's showing as "true" no matter what, even if there should be an expression there.
Screen Shot 2021-08-18 at 9 37 54 AM

CarlosNZ:
This is not a new thing, but just wondering how easy it is to fix -- the "operator" selector disappears off the screen if the component starts too high on the page.
Screen Shot 2021-08-18 at 9 57 37 AM
Can use arrow keys to keep going up, but it's kind of confusing and tricky. Don't worry about it for now if it's too complicated to deal with.

src/containers/TemplateBuilder/evaluatorGui/typeHelpers.ts
```diff
@@ -149,9 +149,11 @@ export const convertTypedEvaluationToBaseType: ConvertTypedEvaluationToBaseType
       }
       const operatorResult: {
         operator: Operator
+        type?: string
         children: EvaluatorNode[]
       } = {
         operator: evaluation.asOperator.operator,
+        type: evaluation.asOperator.type,
```
CarlosNZ, line 156: So does this pass `type: undefined` if type not specified? Does that work?

src/formElementPlugins/listBuilder/src/ApplicationView.tsx
```diff
@@ -243,10 +243,10 @@ const buildElements = async (
     category: field.category as TemplateElementCategory,
     title: field.title as string,
     parameters: field.parameters,
-    validationExpression: field?.validation || true,
+    validationExpression: !!field?.validation,
     validationMessage: field?.validationMessage || '',
-    isVisibleExpression: field?.visibilityCondition || true,
-    isEditableExpression: field?.isEditable || true,
+    isVisibleExpression: !!field?.visibilityCondition,
+    isEditableExpression: !!field?.isEditable,
```
CarlosNZ, line 249: Just wondering -- what's the advantage of this way of expressing it?

CarlosNZ (changes requested):
Cool, thanks -- can you just have a look at those couple of things I mentioned in the main comments? Feel free to make them a new issue if they're not simple.

CarlosNZ:
> The validation expression is not showing up in the GUI. In "Common Properties", there is an "isValid" field, but it's showing as "true" no matter what, even if there should be an expression there.

I think it's just showing the "isValid" value  rather than the validation expression.

src/formElementPlugins/listBuilder/src/ApplicationView.tsx
andreievg: I am not sure why I did that, it's only good for booleans, those are expressions, fixed:
63f304446b6a7204b5549c49087ab7e8ec5c654a
it also fixes
> The validation expression is not showing up in the GUI. In "Common Properties", there is an "isValid" field, but it's showing as "true" no matter what, even if there should be an expression there.

src/containers/TemplateBuilder/evaluatorGui/typeHelpers.ts
andreievg: it seems to work, if you toggle it while switching to show as gui you'll see it disappearing

andreievg:
> Can we do something about the layout for the Actions -- it should at least wrap onto multiple rows:

619c19e4d08cdcdf7fc2955e786cf0f120bf84f3

> This is not a new thing, but just wondering how easy it is to fix -- the "operator" selector disappears off the screen if the component starts too high on the page.

65056d1190a04fb161f4b2651323bfda6f312333

> The validation expression is not showing up in the GUI. In "Common Properties", there is an "isValid" field, but it's showing as "true" no matter what, even if there should be an expression there.

63f304446b6a7204b5549c49087ab7e8ec5c654a

it was treating them as booleans for some reason, i thought i fixed it but not, thanks.

andreievg:
@CarlosNZ made the change and switched base

Issue #524, opened 2021-03-26
Epic #62 - View other's Review (permissionPolicy configurations)


nmadruga:
## Description

Users with permisson to VIEW should be allowed to see other reviews pages done by others. And on the Review Home-page see all the information and make comments to show on internal comments area. 

## Documentation
- [ ] TODO: Add as part of [Epic #38](338) Public docs & tutorials

## Issues
- [x] #356 Show 404 on Review page when changing to user without permission or not assigned
- [ ] Add new option on Template builder to associate VIEW permission
- [ ] Allow users with VIEW permission on the Review home page and Review page

andreievg:
From row **level policies perspective**:

- [ ] Add new policy type `VIEW`, [this branch](https://github.com/openmsupply/conforma-server/compare/Permission-policy-of-type-view)

* Configure policy with these rules (see below for graphql mutation):

```json
{
    "review": {
        "view": {
            "application_id": {
                "$select": {
                    "$from": "review_assignment",
                    "$where": {
                        "$and": [
                            {
                                "template_id": {
                                    "$eq": "jwtPermission_array_bigint_template_ids"
                                }
                            }
                        ]
                    },
                    "$columns": {
                        "application_id": true
                    }
                }
            }
        }
    },
    "application": {
        "view": {
            "id": {
                "$in": {
                    "$select": {
                        "$from": "review_assignment",
                        "$where": {
                            "$and": [
                                {
                                    "template_id": {
                                        "$eq": "jwtPermission_array_bigint_template_ids"
                                    }
                                }
                            ]
                        },
                        "$columns": {
                            "application_id": true
                        }
                    }
                }
            }
        }
    },
    "review_assignment": {
        "view": {
            "reviewer_id": {
                "$isNull": false
            },
            "template_id": "jwtPermission_array_bigint_template_ids"
        }
    }
}
```

Which translates to:

(fyi: `COALESCE(current_setting('jwt.claims.pp123', true), '') != ''` will check that permission policy is linked to user through permission joins -> permission name and `any(string_to_array(COALESCE(current_setting('jwt.claims.pp123_template_ids', true), '0'), ',')::integer[])` would give all template_ids for that policy linked to user through permission joins`)

`can view all reviews with review_assignments for template_id that is linked to this permission policy`

```sql
CREATE POLICY "view_pp123" ON "review" FOR 
SELECT 
  USING (
    COALESCE(
      current_setting('jwt.claims.pp123', true), 
      ''
[...]

PR #914, opened 2021-08-18
#911 What is your emergency?

### LINKED ISSUE #911: Don't allow FDA members to select "No Org" in selector
### Back-end:

- add `isSystemOrg` boolean field to "organisation" table (and set "true" for "Food and Drug agency" org)
- return this field as part of OrgInfo from '/userInfo' endpoint

### Front-end:

- check `isSystemOrg` field in OrgSelector, and only display "No Org" option if false
- in Login screen, don't present  org selection option if `isSystemOrg`

Anything else?

CarlosNZ:
Fixes #911 

Requires [back-end PR #524](https://github.com/openmsupply/application-manager-server/pull/524)

This is kind of an interim solution, we can come up with something else later on if it's a problem. For now, the FDA org never see the "No Org" selection, which is what Craig wants.

- Also tweaked the logic on the login screen, so if a user only has one org (i.e. most people), they'll skip the org selection and automatically be logged in as their only org (so won't see the "No Org" option initially). So only people who go poking around in the Org dropdown will see the "No Org" option

To test, just set the "isSystemOrg" flag on any org to "true" and note that users who are part of this org can no longer see the "No Org" option

Edit: improved it further so there is not even a Dropdown at all when there is only one choice (i.e one company, FDA), it's just a fixed banner)

### DIFF
--- a/src/containers/User/UserArea.tsx
+++ b/src/containers/User/UserArea.tsx
@@ -191,11 +191,13 @@ const OrgSelector: React.FC<{ user: User; orgs: OrganisationSimple[]; onLogin: F
     text: orgName,
     value: orgId,
   }))
-  dropdownOptions.push({
-    key: LOGIN_AS_NO_ORG,
-    text: strings.LABEL_NO_ORG,
-    value: LOGIN_AS_NO_ORG,
-  })
+  // Only add "No Org" option if user not part of "Admin" org (e.g. FDA)
+  if (!orgs.some(({ isSystemOrg }) => isSystemOrg))
+    dropdownOptions.push({
+      key: LOGIN_AS_NO_ORG,
+      text: `> ${strings.LABEL_NO_ORG_SELECT}`,
+      value: LOGIN_AS_NO_ORG,
+    })
   return (
     <div id="org-selector">
       {user?.organisation?.logoUrl && (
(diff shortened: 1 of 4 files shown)


src/containers/User/UserArea.tsx
```diff
@@ -191,22 +191,28 @@ const OrgSelector: React.FC<{ user: User; orgs: OrganisationSimple[]; onLogin: F
...
       )}
       <div>
-        <Dropdown
-          text={user?.organisation?.orgName || strings.LABEL_NO_ORG}
-          options={dropdownOptions}
-          onChange={handleChange}
-        ></Dropdown>
+        {dropdownOptions.length === 1 ? (
+          user?.organisation?.orgName || strings.LABEL_NO_ORG
+        ) : (
```
nmadruga, line 209: What happens when lenght === 0?

It seems like this statement should be <=1

nmadruga (changes requested):
Seems really good. Quick testing now, will let you know if something unexpected happens

src/containers/User/UserArea.tsx
CarlosNZ: The whole "orgSelector" component won't render if it's 0. See line 37:
```
{orgList.length > 0 && <OrgSelector user={currentUser} orgs={orgList} onLogin={onLogin} />}
```

nmadruga: Oh! Great. I tested too and it seems all good.

nmadruga (approved):
LGTM

CarlosNZ:
> LGTM

Thanks -- back-end okay too? (https://github.com/openmsupply/application-manager-server/pull/524)

PR #920, opened 2021-08-20
Fix history with proper template version

nmadruga:
Changes:

- Receives `template.version` in **templateFragment**
- Renamed fragment from Template to templateFragment (a few extra files changes because of that)
- Passes the whole object with `TemplateDetails` to the component `HistoryPanel` that will extract:
  - templateCode
  - templateVersion

What was missing to do (and it was a TODO in the code) was to be passing the correct templateVersion in the query.

### DIFF
--- a/src/utils/graphql/fragments/fullTemplate.fragment.ts
+++ b/src/utils/graphql/fragments/fullTemplate.fragment.ts
@@ -2,7 +2,7 @@ import { gql } from '@apollo/client'
 
 export default gql`
   fragment FullTemplate on Template {
-    ...Template
+    ...templateFragment
     nodeId
     configApplications: applications(filter: { isConfig: { equalTo: true } }) {
       nodes {
--- a/src/components/PageElements/PageElements.tsx
+++ b/src/components/PageElements/PageElements.tsx
@@ -172,11 +172,7 @@ const PageElements: React.FC<PageElementProps> = ({
           })}
         </Form>
         {showHistory && (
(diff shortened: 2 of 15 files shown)


CarlosNZ:
>    a few extra files changes because of that

Screen Shot 2021-08-20 at 3 32 50 PM

A few ! 😆

src/utils/graphql/fragments/fullTemplate.fragment.ts
```diff
@@ -2,7 +2,7 @@ import { gql } from '@apollo/client'
 
 export default gql`
   fragment FullTemplate on Template {
-    ...Template
+    ...templateFragment
```
CarlosNZ, line 5: Is this consistent with how we've been naming other fragments?

CarlosNZ (approved):
Looks good, thanks. If you merge this ASAP, I'll merge develop back into demo again. Thanks.

src/utils/graphql/fragments/fullTemplate.fragment.ts
nmadruga: New ones yes, old ones are like it was before `Template`, but I think it's not very clear where in use. So preferred the new way, do you agree?

CarlosNZ: Yeah sure. Is the convention to use camelCase for these, or would it be "TemplateFragment" ?

PR #930, opened 2021-08-24
Add error-handling for radioChoice and fix dropdown error color

### LINKED ISSUE #784: Improve error handling ui on elements
Update wrapper to have a generic error state to be showing when invalid question or changesRequested to Applicant - for changes requested it needs a second state to show in the comment in grey after changed

### LINKED ISSUE #813: Style -- update dropdown over-rides to match error colour
For example:

Screen Shot 2021-06-24 at 12 57 34 PM

nmadruga:
Fixes #784 and fixes #813 

## Linked PR
- Based in #820 

## Changes

Just adding some error-handling to radio-button to show something is wrong (using validationMessage) when trying to submit with one selected. As mentioned in "Demo server templates/configs": Radio selection on Page 3 (and maybe other radios) need a validation message to show when trying to navigate to next page without one selected - Adam's feedback

Screen Shot 2021-08-24 at 11 57 09 PM
Screen Shot 2021-08-24 at 11 57 26 PM

TODO:
- Add something similar to List-builder once decided

### DIFF
--- a/src/formElementPlugins/dropdownChoice/src/ApplicationView.tsx
+++ b/src/formElementPlugins/dropdownChoice/src/ApplicationView.tsx
@@ -84,13 +84,10 @@ const ApplicationView: React.FC<ApplicationViewProps> = ({
         options={dropdownOptions}
         onChange={handleChange}
         value={selectedIndex}
-        disabled={!isEditable}
         error={!validationState.isValid ? true : undefined}
       />
       {validationState.isValid ? null : (
-        <Label basic color="red" pointing>
-          {validationState?.validationMessage}
-        </Label>
+        <Label pointing prompt content={validationState?.validationMessage} />
       )}
     </>
   )
--- a/src/formElementPlugins/radioChoice/src/ApplicationView.tsx
+++ b/src/formElementPlugins/radioChoice/src/ApplicationView.tsx
@@ -38,7 +38,6 @@ const ApplicationView: React.FC<ApplicationViewProps> = ({
   )
 
   useEffect(() => {
-    onUpdate(currentResponse)
     // This ensures that, if a default is specified, it gets saved on first load
     if (!currentResponse?.text && defaultOption !== undefined) {
       const optionIndex = getDefaultIndex(defaultOption, options)
--- a/src/formElementPlugins/radioChoice/src/ApplicationView.tsx
+++ b/src/formElementPlugins/radioChoice/src/ApplicationView.tsx
@@ -1,5 +1,5 @@
 import React, { useEffect, useState } from 'react'
-import { Radio, Form, Input } from 'semantic-ui-react'
+import { Radio, Form, Input, Label } from 'semantic-ui-react'
 import { ApplicationViewProps } from '../../types'
 import strings from '../constants'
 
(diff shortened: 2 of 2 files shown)


CarlosNZ:
This only works if I try to navigate from the actual radio button page.

i.e. If I'm on the page with the radio buttons and I click "Review & Submit", the error display shows. However, if I'm on the last page and click "Review & Submit", it jumps back to the radio button page, and the progress area shows the error state (as usual), but the radio button elements themselves appear normal.

nmadruga:
> This only works if I try to navigate from the actual radio button page.
> 
> i.e. If I'm on the page with the radio buttons and I click "Review & Submit", the error display shows. However, if I'm on the last page and click "Review & Submit", it jumps back to the radio button page, and the progress area shows the error state (as usual), but the radio button elements themselves appear normal.

Hmm perhaps it would required the validationExpression to run after the validation is triggered this way too. I noticed that is doesn't run the validation sometimes... Will need to investigate further to find out, hope that doesn't take because the logic to validate after the Review and Submit was clicked is very hard to deduce currently... I'll let you know after I've looked at this. Thanks for spotting this out.

nmadruga:
Also - any suggestions for how to display the error on list builder?

CarlosNZ:
> Also - any suggestions for how to display the error on list builder?

I'm about to make some tweaks to list builder (based on your suggestions and a couple of other little problems I've noticed), so I'll add some additional UI for errror display while I do that. I won't have the logic, but it'll be easy to add in once you've finished this.

CarlosNZ:
> Hmm perhaps it would required the validationExpression to run after the validation is triggered this way too. I noticed that is doesn't run the validation sometimes... Will need to investigate further to find out, hope that doesn't take because the logic to validate after the Review and Submit was clicked is very hard to deduce currently... I'll let you know after I've looked at this. Thanks for spotting this out.

What I see happening is:

- the revalidation runs as expected, and the "isValid" state of the elements is "false" as the page renders. But then there is an additional re-render of the radiobutton elements, and its here that the validation state becomes "true" again.

I'm a bit nervous about trying to change this, as I think that changing its render behaviour to fix one thing might break it in other ways. Not sure, I guess have a go, but make sure all the "default" logic and everything continues to work as expected.

nmadruga:
> I'm a bit nervous about trying to change this, as I think that changing its render behaviour to fix one thing might break it in other ways. Not sure, I guess have a go, but make sure all the "default" logic and everything continues to work as expected.

Sure. I'll make more tests to make sure it's kept the same way.



> * the revalidation runs as expected, and the "isValid" state of the elements is "false" as the page renders. But then there is an additional re-render of the radiobutton elements, and its here that the validation state becomes "true" again.

Thaks for the advice, that should be easier to track now. Having a look on this...

nmadruga:
@CarlosNZ is this mergeable? I think I can detach from the base branch since not doing ListBuilder validation here. I'll just check if the Search plugin also needs this

CarlosNZ:
> @CarlosNZ is this mergeable? I think I can detach from the base branch since not doing ListBuilder validation here. I'll just check if the Search plugin also needs this

Sure, why don't you just merge it into base (list builder) and then I'll make sure it works with list builder before I merge it.

PR #932, opened 2021-08-25
#813 fix error handler for dropdown radio search

### LINKED ISSUE #813: Style -- update dropdown over-rides to match error colour

nmadruga:
Fixes #813

Also pulls changes done in PR #930 

## Changes
- Proper color fixes for dropdown
- Add error-handling for Search plugins
Screen Shot 2021-08-25 at 6 28 19 PM

### DIFF
--- a/src/formElementPlugins/search/src/ApplicationView.tsx
+++ b/src/formElementPlugins/search/src/ApplicationView.tsx
@@ -1,7 +1,7 @@
 const DEBOUNCE_TIMEOUT = 350 //milliseconds
 
 import React, { useEffect, useState } from 'react'
-import { Search, Label, Card, Icon } from 'semantic-ui-react'
+import { Search, Label, Card, Icon, Form } from 'semantic-ui-react'
 import { ApplicationViewProps } from '../../types'
 import { useUserState } from '../../../contexts/UserState'
 import strings from '../constants'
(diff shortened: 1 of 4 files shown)


src/formElementPlugins/search/src/ApplicationView.tsx
```diff
@@ -123,20 +123,27 @@ const ApplicationView: React.FC<ApplicationViewProps> = ({
...
-        minCharacters={minCharacters}
-        placeholder={placeholder}
-        results={
-          loading ? [{ title: strings.MESSAGE_LOADING }] : createResultsArray(results, resultFormat)
-        }
-        disabled={!isEditable}
-        input={{ icon: icon, iconPosition: 'left' }}
-        noResultsMessage={strings.MESSAGE_NO_RESULTS}
-      />
+      <Form.Field key={`search-${label}`} error={!validationState.isValid}>
```
CarlosNZ, line 126: Does it have to be wrapped in `<Form.Field>`? There is already a Form.Field wrapper in the ApplicationViewWrapper, which is why most of the plugin components are just wrapped in a `<>` fragment.

You might not be able to avoid it, but if we can avoid `Form.Field`s inside `Form.Fields`, that'd be preferable. 
(I'll check in the morning if it's actually valid HTML)

PR #940, opened 2021-08-27
#692 Auto-assign sections to same assignee + Fix problem with review self-assignment for consolidation

### LINKED ISSUE #692: Improvement: Assigner should assign ALL sections to same reviewer when stage has only one-level review when selecting first one
## Current description
There are 2 problems that are creating the erroneous result as described previously. One (to be fixed as part of #691) is that we don't lock assignment/review of other reviewers when the review submitted makes them invalid reviews. The other - to be addressed later - in this issue is that when the stage has only 1 level, the reviewer is set as `last_level_reviewer` (and can make one decision for the current stage - and therefor make the increment of the stage), but only half of the sections were assigned to them (this is a problem). So the improvement will be _somehow_ (a few options to be discussed in the comments section) we need to set that this stage can be assigned so ALL section belongs to the same Reviewer. For self-assignment that's already implemented, but for assigned by Assigner we need a fix.

## Previous description
Bug: Incrementing application stage after half of sections are approved and submitted as conform by one Reviewer

### To replicate:
Tested using current config (with current assignments of Section 1 to **Reviewer1**, Section 2 to **Reviewer2**) and application http://localhost:3000/application/ABC123/review

- Logged as **Reviewer1** and start review. Approve all responses of Section 1 and click on the `Conform` decision, then submit the review.
- Go back to the review home page and see that the new current stage is `Assessment`, but the review of Section 2 wasn't done yet.
- Log in as **Reviewer2** and change the Filter to Screening to see they can still start their review.

### Screenshots
Review in next stage (no Assignments were created though)
Screen Shot 2021-05-12 at 12 26 05 PM

Previous stage can still be reviewed by **Reviewer2**
Screen Shot 2021-05-12 at 12 25 58 PM

nmadruga:
1 - fixes #692  auto-assign sections to same assigner selected in one row
2 - No issue - [from Demo server config doc](https://docs.google.com/document/d/19CYbrFv2NyV69ggqWRVmbAxktPIwyNRe0W9c9xUikHM/edit#heading=h.wp1x7v9odz8) "Getting an error when trying to submit consolidation -- says “Not all sections have been reviewed” even when they are. Is this a bug?"
Fix self-assignment of consolidation not considering reviews not existing when generating questionAssignments.

## Brief
The problem was related to questions with different configurations, the code logic to give reviewQuestionAssignments was different to the one from `useGetReviewStructureForSection/helpers` and therefor was considering questions with empty responses/review when generating the reviewAssignment for consolidator on Self-assignment. Fix is in `useUpdateReviewAssignment.tsx`

Another problem that started to happen with fix was that only the first sections was being correctly assigned when using self-assignment. Now with this change the assignment will run per Section where the structure is up-to-date.

## Changes
- Uses new state `shouldAssignState` to pass from one section to other with `true` to self-assign another section to the same reviewer, or with the assignee index (on the dropdown) to select the same reviewer selected in one section to others.
- Check if `isLastLevel` to do the auto-assignment otherwise should not auto-assign other sections.
- Move assignment and check for error to inner component so it doesn't re-render each row each time it auto-assign other sections (self-assignment of lastLevel assignment)

### DIFF
--- a/src/containers/Review/ReviewHome.tsx
+++ b/src/containers/Review/ReviewHome.tsx
@@ -17,8 +17,6 @@ const ReviewHome: React.FC<ReviewHomeProps> = ({
   assignmentInPreviousStage,
   fullApplicationStructure,
 }) => {
-  console.log('Render')
-
   const [shouldAssign, setShouldAssign] = useState<number | boolean>(false)
   return (
     <>
--- a/src/containers/Review/ReviewHome.tsx
+++ b/src/containers/Review/ReviewHome.tsx
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, { useState } from 'react'
 import { Segment, Header } from 'semantic-ui-react'
 import { AssignmentDetails, FullStructure } from '../../utils/types'
 import AssignmentSectionRow from './AssignmentSectionRow'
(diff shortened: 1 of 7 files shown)


src/containers/Review/ReviewHome.tsx
```diff
@@ -16,32 +16,39 @@ const ReviewHome: React.FC<ReviewHomeProps> = ({
...
-          {...{
-            assignments: assignmentsByStage,
-            structure: fullApplicationStructure,
-            sectionCode: code,
-          }}
-        />
-        {assignmentsByUserAndStage.map((assignment) => (
-          <ReviewSectionRow
+}) => {
+  console.log('Render')
```
CarlosNZ, line 20: Remove?

CarlosNZ (approved):
Looks good, thanks.

Don't fully understand all the details of this area, but functionality works as expected, and great to have that bug fixed.

PR #948, opened 2021-09-01
Show field "Can Make Decision" in template builder UI

nmadruga:
Partial fix issue https://github.com/openmsupply/application-manager-server/issues/506

- Visible in Permisisons: on each level of a stage
- Rules: Can only be associated to level 1 review and for stage higher than 1.
- Required to get ProdRegistration template working - last stage should be final decision
- set true when selected

### DIFF
--- a/src/containers/TemplateBuilder/template/Permissions/PermissionNameList.tsx
+++ b/src/containers/TemplateBuilder/template/Permissions/PermissionNameList.tsx
@@ -60,6 +60,7 @@ const PermissionNameList: React.FC<PermissionNameListProps> = ({
             {type === PermissionPolicyType.Review && levelNumber && (
               <ReviewTemplatePermission
                 templatePermission={templatePermission}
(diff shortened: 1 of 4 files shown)


CarlosNZ (approved):
Great, works nicely thanks.

I just have one quick question -- why it "Make Decision" only available for Stages higher than 1?

nmadruga:
> I just have one quick question -- why it "Make Decision" only available for Stages higher than 1?

Oh, that's on the interface? Sorry, I think it should be the other way around. My bad....
For now, we would only allow this stage to be looking at the review done in a previous stage, so it should only be allowed for level 1. I'll change that. Good catch!

CarlosNZ:
> For now, we would only allow this stage to be looking at the review done in a previous stage, so it should only be allowed for level 1. I'll change that. Good catch!

Ah level 1! Yes, that makes more sense. Thanks.

PR #949, opened 2021-09-02
#947 search plugin showing changes when it hasn't

### LINKED ISSUE #947: Search element showing as if updated on Changes Required
After the applicant receives changes required (in other elements), if I click on a page with a Search element it is showing as if it had been changed by the applicant. This needs a fix:
photo_2021-09-02_10-04-24

Tested: used `demo_2021_08_27` snapshot on Product Registration template

CarlosNZ:
Fixes #947 

Just updates the default "text" value saved in the response. Previously was just using `JSON.stringify()`, but the output might change as the retrieved response would change the order of the fields in the selection objects.

Added a custom string processor (when "textFormat" is not defined) -- bonus is that the output string is _slightly_ nicer to look at.

### DIFF
--- a/src/formElementPlugins/search/src/ApplicationView.tsx
+++ b/src/formElementPlugins/search/src/ApplicationView.tsx
@@ -53,11 +53,7 @@ const ApplicationView: React.FC<ApplicationViewProps> = ({
 
   useEffect(() => {
     onSave({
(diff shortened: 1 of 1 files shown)


nmadruga (approved):
LGTM - tested working well. Thanks

PR #952, opened 2021-09-02
#951 TB to ignore is config applications that are stuck in processing

### LINKED ISSUE #951: Template builder struck on loading in template config
Found this super annoying when making back end PRs yesterday.

**The cause:**

If server is still starting up, triggers for created applications won't fire, and there will be is_config application with processing trigger for ever, and TB looks at the first is_config application for the template and waits for it becomes trigger null. This shouldn't be a problem anymore but in basic_snapshot there a few cases like this (i've changed this but previously on snapshot load server would restart, and this was more of a problem)

andreievg:
Fixes #951.

As per issue description, changes as per commit.

### Testing.

Try loading basic snapshot (from this repo, haven't tries from other repo), and try to configure org registration template. Do the same with this fix.

### DIFF
--- a/src/utils/generated/graphql.tsx
+++ b/src/utils/generated/graphql.tsx
@@ -30364,7 +30364,7 @@ export const FullTemplateFragmentDoc = gql`
     fragment FullTemplate on Template {
   ...templateFragment
   nodeId
-  configApplications: applications(filter: {isConfig: {equalTo: true}}) {
+  configApplications: applications(filter: {isConfig: {equalTo: true}}, orderBy: ID_DESC) {
     nodes {
       serial
       id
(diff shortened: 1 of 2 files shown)


CarlosNZ (approved):
Cool, thanks. This had been annoying when working on snapshots/templates. I was just manually deleting the stuck applications, but this is a nicer fix.

We probably need to figure out a more global strategy for triggers that get stuck in "PROCESSING" or "ERROR" -- currently they'll never load, as they're waiting to change back to NULL. Something to think about for later. :)

andreievg:
> We probably need to figure out a more global strategy for triggers that get stuck in "PROCESSING" or "ERROR" -- currently they'll never load, as they're waiting to change back to NULL. Something to think about for later. :)

Agree, and noted

PR #953, opened 2021-09-02
Fix start message as null when empty string

nmadruga:
Another partial fix for https://github.com/openmsupply/application-manager-server/issues/506

### Linked PRs
- Based on PR #948 

### Changes
Just enforce starMessage to `null` in case it is set to empty string

### DIFF
--- a/src/containers/TemplateBuilder/template/General/MessagesConfig.tsx
+++ b/src/containers/TemplateBuilder/template/General/MessagesConfig.tsx
@@ -48,7 +48,7 @@ const MessagesConfig: React.FC<MessagesConfigProps> = ({ isOpen, onClose }) => {
   const updateMessage = async () => {
     const { startMessage, submissionMessage } = state
     const result = await updateTemplate(templateId, {
-      startMessage: startMessage && startMessage !== '' ? startMessage : null,
+      startMessage: startMessage ? startMessage : null,
       submissionMessage,
     })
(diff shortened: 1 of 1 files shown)


src/containers/TemplateBuilder/template/General/MessagesConfig.tsx
```diff
@@ -48,7 +48,7 @@ const MessagesConfig: React.FC<MessagesConfigProps> = ({ isOpen, onClose }) => {
   const updateMessage = async () => {
     const { startMessage, submissionMessage } = state
     const result = await updateTemplate(templateId, {
-      startMessage,
+      startMessage: startMessage && startMessage !== '' ? startMessage : null,
```
CarlosNZ, line 51:
```suggestion
      startMessage: startMessage ? startMessage : null,
```
I think this would work, as empty string evaluates to false.

CarlosNZ (commented):
Made one quick review comment, but it looks like all the rest of the changes relate to a different issue. Did you need to change the base branch?

nmadruga:
> Made one quick review comment, but it looks like all the rest of the changes relate to a different issue. Did you need to change the base branch?

Yes. That's actually based on another one. Have changed accordingly now. Thanks

CarlosNZ (approved):
Sweet, LGTM

PR #957, opened 2021-09-03
TB - Nice little feature to add actions at the top or bottom of list

nmadruga:
Partial fix for issues in https://github.com/openmsupply/application-manager-server/issues/506

In Template list of actions - helpful when we need to create one in the beginning of a list

How it looks:
Screen Shot 2021-09-03 at 3 11 34 PM


CarlosNZ (changes requested):
One main problem -- it adds actions even when the template should be uneditable (i.e not in Draft or Disabled).

I'm not sure about the UI. It's very unintuitive. Why not just have a toggle switch next to the "Add action" button, set to either "top" or "bottom", then the "Add" button can just work as before (but respecting the state of the toggle switch)?

nmadruga:
> I'm not sure about the UI. It's very unintuitive. Why not just have a toggle switch next to the "Add action" button, set to either "top" or "bottom", then the "Add" button can just work as before (but respecting the state of the toggle switch)?

Sure. I can change this to be a toggle. Working on this for the next 15 minutes - I don't want to take too long since it's a non-priority. But would be good to shorten this list of PRs today.

nmadruga:
Change done @CarlosNZ 

Have also added some localisation strings to the Action tab and got carried and created a non-priority PR to combine the Action tools vertically aligned in the first column of each row. See if you like this option, not a priority though: #1001

nmadruga:
@CarlosNZ Ok to merge this one now?

CarlosNZ:
> @CarlosNZ Ok to merge this one now?

hang on, I'll have another quick look now.

CarlosNZ:
> > @CarlosNZ Ok to merge this one now?
> 
> hang on, I'll have another quick look now.

Can you update this from develop? Doesn't work with current back-end.

nmadruga:
Done

CarlosNZ (approved):
Looks great, thanks.

One tiny little request if you think it's worth it -- I think the default state for the toggle should be "off", it's more natural to add things to the bottom of a list. Inserting at the top seems more like the "non-default" option.

nmadruga:
> One tiny little request if you think it's worth it -- I think the default state for the toggle should be "off", it's more natural to add things to the bottom of a list. Inserting at the top seems more like the "non-default" option.

Thank you. Agreed, and done.

Issue #959, opened 2021-09-06
Outcomes - Not showing application link in details view


andreievg:
I think i've introduced this bug when refactoring outcomes

Issue #958, opened 2021-09-06
Outcomes - table view not respecting column title


CarlosNZ:
For example, if we have this for outcome_display_table:
<img width="563" alt="Screen Shot 2021-09-06 at 2 01 40 PM" src="https://user-images.githubusercontent.com/5456533/132150245-f2baa2f7-d4cf-43b7-9f1d-be10396ae89c.png">

"column_name" should refer to the column in the "organisation" table, and "title" should be what shows up as the column header in the table UI, so the last column should show the value in field "address_street" in a column named "Street"

However, we get the "column_name" also being used for the column header in the UI:
![Screen Shot 2021-09-06 at 2 03 08 PM](https://user-images.githubusercontent.com/5456533/132150313-8d08d99d-914d-41d8-b696-43655127d010.png)

PR #960, opened 2021-09-06
outcome column header and application list to show correctly

andreievg:
Fixes: #959 
Fixes: #958 

Column title was easy, just made sure it mapped properly.

As for application link, issue was in confirming that table existed, the tried to simplify logic, still not 100% happy with it.

Basically the schema query, is using snake case for tables names, so as a query parameter table name should be singular and snake. useGetSchemaColumns then converts all table names to camel case, and then in OutcomeDisplayInner we check if the table exists (now singular but in camel case). 

### Changes

* Changed `getDBApplicationJoinLinkTableName` to `getApplicationJoinLinkTableName` with extra case parameter
* Use `getApplicationJoinLinkTableName` correctly to provide variable to graphql query (`useGetSchemaColumns`) and to deduce if table name exists in schema (`OutcomeDisplayInner`)

### DIFF
--- a/src/containers/Outcomes/ApplicationLinks.tsx
+++ b/src/containers/Outcomes/ApplicationLinks.tsx
@@ -6,6 +6,7 @@ import { Loading } from '../../components'
 import strings from '../../utils/constants'
 import { useRouter } from '../../utils/hooks/useRouter'
 import { ApplicationLinkQuery } from '../../utils/types'
+import { Link } from 'react-router-dom'
 
 const ApplicationLinks: React.FC<{ applicationLinkQuery: ApplicationLinkQuery; id: number }> = ({
   applicationLinkQuery,
@@ -24,7 +25,6 @@ const ApplicationLinks: React.FC<{ applicationLinkQuery: ApplicationLinkQuery; i
   if (!data) return <Loading />
 
   const applications = applicationLinkQuery.getApplications(data)
-  const navigateToApplication = (serial: string) => push(`/application/${serial}`)
   if (applications.length === 0) return null
 
   return (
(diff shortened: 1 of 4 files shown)


CarlosNZ (changes requested):
Cool, thanks for this.

One quick change if you could: can the link to the "linked applications" open in a new tab? Currently, it's really hard to go "back" to the outcomes page from the application summary (since there's a re-direct in there), which would be made more useable by opening the application in a new tab.

andreievg:
> t's really hard to go "back" to the outcomes page from the application summary (since there's a re-direct in there)

Hmm i can do that, but perhaps the redirect should be a replace rather than push ? (this would be a problem in other places where you navigate to application pages that are not available ?)

CarlosNZ:
> > t's really hard to go "back" to the outcomes page from the application summary (since there's a re-direct in there)
> 
> Hmm i can do that, but perhaps the redirect should be a replace rather than push ? (this would be a problem in other places where you navigate to application pages that are not available ?)

Sure, feel free to change that, too -- but personally I think opening related applications etc. should be in new tab anyway, since it's kind of "supplementary" info (like opening files in the application form). Just my preference though.

andreievg:
> Sure, feel free to change that, too

I'll leave that for now, might check it out a bit later in the week.

>  but personally I think opening related applications etc. should be in new tab anyway, since it's kind of "supplementary" info (like opening files in the application form). Just my preference though.

Fair point 89c4a5c6c088eb9a63e84b1a86b5a198834dab1a

CarlosNZ (approved):
Sweet, thanks.

CarlosNZ:
Cool -- merging this now.

PR #963, opened 2021-09-06
Keep selected page while moving up or down on the list of pages

nmadruga:
Just a small tweak fior Template builder - no issue made in a rush


CarlosNZ (approved):
Cool 👍

I hadn't really noticed this problem before, but I can see how it would be super-annoying if you're moving a lot of pages around :)

Issue #119, opened 2020-11-13
Basic restrictions for existing templates


nmadruga:
More restrictions would be added on the go. This is more as an example for future ones.

What a restriction does different to a basic permission_name associated to a user or user_company join? It restricts when a permission_policy is applicable to a action done by the user. For example: User can only edit a application when status is `Draft` or `Changes requested`.

PR #974, opened 2021-09-09
#285 delete application in draft

### LINKED ISSUE #285: Logic to discard application (in Draft) when Applicant clicks the bin icon on the list

nmadruga:
Fixes #285 

@adamdewey That one is for you : )

### Changes
- Add confirmation and message to show before deletion
- Already displayed under the switch that checks it's a draft application should check anything else?
- [Needs checking] Should be considering the current user's row level permissions - so can only delete if it's yours


adamdewey:
Oh! 👏

src/components/List/ApplicationsList.tsx
```diff
@@ -79,11 +81,27 @@ interface ApplicationRowProps {
...
+    deleteApplication,
+  }
+
+  useEffect(() => {
+    if (data) setIsDeleted(true)
+  }, [data])
+
+  if (error) return <Message error header={strings.ERROR_APPLICATION_DELETE} list={[error]} />
+
+  return isDeleted ? null : (
```
CarlosNZ, line 100: Is there a better way to do this? From what I understand, this just set the local component to "deleted" and just hides it from UI even though it's still actually in the list. What we really want to do is trigger a re-fetch in the list, which means you won't need any state in here.

One way to do it, maybe: 
Use "refetch" in "ListWrapper": `{data, loading, error: applicationsError, refetch } = useGetApplicationListQuery(...`

And pass this down as "refetchList" or something, then in the ApplicationRow component, use the delete mutation's "ON_COMPLETED" property, and call "refetchList" on completion (so no need for useEffect either).

(One side effect of the way you've done it, is if the list is longer than one page, then when you delete one, the list won't re-paginate, which is confusing cos it looks like you've got less applications that there are -- in fact, you could delete all the applications on a page and the user would think they've deleted everything because the existing ones haven't rippled round to the current page like they should have).

CarlosNZ (changes requested):

CarlosNZ:
Actually, now I think about this, the actual deletion should probably be through a back-end endpoint, for security reasons. I know you check for Draft status here, but that's all UI-based -- it's still easy to send "delete" queries to GraphQL for applications not in Draft.

I think we should probably set policy so no user has "delete" permissions on any table directly, only the "isAdmin" JWT in the back-end. So the front-end would make request for deletion, and the back-end does all the checks, then handles the deletion, and returns a confirmation to front-end, which then knows to refetch the list.

Also -- the back-end needs to delete any files associated with the application too.

I realise that makes this a somewhat bigger job, but leaving it like this is quite a big security hole -- so will need to make a new issue and add it to [this issue](https://github.com/openmsupply/application-manager-server/issues/572) please.

(I can probably do the back-end after I finish up on Outcomes, so maybe leave this in PR for now, and come back to it when the back-end API is available)

nmadruga:
> Actually, now I think about this, the actual deletion should probably be through a back-end endpoint, for security reasons. I know you check for Draft status here, but that's all UI-based -- it's still easy to send "delete" queries to GraphQL for applications not in Draft.
> 
> I think we should probably set policy so no user has "delete" permissions on any table directly, only the "isAdmin" JWT in the back-end. So the front-end would make request for deletion, and the back-end does all the checks, then handles the deletion, and returns a confirmation to front-end, which then knows to refetch the list.
> 
> Also -- the back-end needs to delete any files associated with the application too.
> 
> I realise that makes this a somewhat bigger job, but leaving it like this is quite a big security hole -- so will need to make a new issue and add it to [this issue](https://github.com/openmsupply/application-manager-server/issues/572) please.
> 
> (I can probably do the back-end after I finish up on Outcomes, so maybe leave this in PR for now, and come back to it when the back-end API is available)

Yes, good point. What I was thinking is that back-end would already be preventing that only applications related to your applications (already in place) and in the status draft (need to add as part of this-sprint-issue #119), so this check in UI is only for showing the option - not preventing the deletion on the back-end. But you are right, we need to add this restriction on the back-end to make sure there is no hole on this end. If you would like to have a go on changing the policies to check that today I think that should be enough? And the deletion of the file coud be done as part of the garbage-collection issue:
So files that are associated to non-existing applications they get removed?

nmadruga:
@CarlosNZ planning to have a look on adding some restrictions to users as to only delete application in status = DRAFT or WITHDRAWN (if we will still do this one) or EXPIRED. I don't think we should allow changing if status CHANGES_REQUIRED.

Have you and @andreievg started some discussions around security and row-level permissions?

CarlosNZ:
> Have you and @andreievg started some discussions around security and row-level permissions?

Kind of, you'll see I've updated the linked issues on [this summary issue](https://github.com/openmsupply/application-manager-server/issues/572) with more details as to what we discussed. I'll make a comment about deletion on the Policies issue.

The main thing to fix in this PR I think is just my [first comment](https://github.com/openmsupply/application-manager-web-app/pull/974#discussion_r707008534), as I don't think it's good practice to just hide (using CSS) items from the table without actually removing them from the list state -- seems likely to lead to unexpected bugs (such as the one I mentioned in that comment)

src/utils/messages.ts
```diff
@@ -2,6 +2,11 @@ export default {
   APPLICATIONS_LIST_EMPTY: 'No applications found',
   APPLICATION_MISSING_TEMPLATE: '',
   APPLICATION_OTHER_CHANGES_MADE: 'This is a new submission and will require a review.',
+  APPLICATION_DELETION_CONFIRM: {
+    title: 'Delete application',
+    message: 'Please confirm you would like to delete a draft application',
+    option: 'OK',
+  },
```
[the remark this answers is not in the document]
nmadruga: Ok - checked PR #1004 and both files `messages` and `constants are deleted. So I will leave here for now.

src/components/List/ApplicationsList.tsx
nmadruga: Doing this change now. Thanks : )

src/utils/messages.ts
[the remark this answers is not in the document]
CarlosNZ: Yeah, that's cool, I'll need to do a big mop-up before #1004 can merged, so leave any constants stuff for now.

nmadruga:
Ok - Changes done on Front-end.

Regarding security issue, should be dealt as part of issue https://github.com/openmsupply/application-manager-server/issues/589. Have added an comment there about that now.

CarlosNZ:
> Ok - Changes done on Front-end.

Cool, will check out shortly

src/components/List/ApplicationsList.tsx
```diff
@@ -1,6 +1,8 @@
-import React, { Fragment } from 'react'
+import React, { Fragment, useEffect, useState } from 'react'
```
CarlosNZ, line 1: useEffect, useState not used

CarlosNZ (approved):
Great, thanks. I'm pleased the refetch/OnCompleted mechanism worked out.

Works well in the UI now, applications from later pages wrap around correctly when deleting.

Can you just remove those two unused hooks in ApplicationsList.

👍

nmadruga:
> Works well in the UI now, applications from later pages wrap around correctly when deleting.
> 
> Can you just remove those two unused hooks in ApplicationsList.

Sure. Thank you!

PR #979, opened 2021-09-10
Portguese strings v1

nmadruga:
### Linked PR
- Back-end https://github.com/openmsupply/application-manager-server/pull/571

### Changes
- Add new strings in Portuguese
- Plugins have strings add replacing english - needs to fix that before merging!


src/formElementPlugins/fileUpload/constants.ts
```diff
@@ -1,9 +1,9 @@
...
-  ERROR_UPLOAD_PROBLEM: 'Problem uploading file',
-  BUTTON_CLICK_TO_UPLOAD: 'Click to upload',
-  BUTTON_UPLOAD_ANOTHER: 'Upload another',
+  ERROR_FILE_ALREADY_UPLOADED: 'O mesmo arquivo já foi carregado',
+  ERROR_FILE_TYPE_NOT_PERMITTED: 'O tipo do arquivo não é permitido',
+  ERROR_FILE_TOO_BIG: 'O tamanho do arquivo não é permitido',
+  ERROR_TOO_MANY_FILES: 'Não é permitido carregar mais arquivos',
+  ERROR_UPLOAD_PROBLEM: 'Problema ao carregar arquivo',
+  BUTTON_CLICK_TO_UPLOAD: 'Carregar arquivo',
+  BUTTON_UPLOAD_ANOTHER: 'Carregar outro arquivo',
```
CarlosNZ, line 8: Doesn't this need an "_en"/ "_pt" version rather than over-writing the English?

src/utils/messages_en.ts
```diff
@@ -0,0 +1,46 @@
+export default {
```
CarlosNZ, line 1: What the plan for these "messages" files? When we implement the localisation solution, I don't think we want to have two different kinds of strings files hanging around, do we? And will the nested objects work? I think that will make it unnecessarily complicated, don't you? 

(Feel free to reply with just "I have a plan" and we can discuss it later)

src/formElementPlugins/fileUpload/constants.ts
nmadruga: Hi Carl, yes. We could do something more cleaver about these strings later. For now I was just adding then so I can have this on a demo instance that I'm going to build from this PR.

We shouldn't be merging with over-writing.

PR #983, opened 2021-09-15
#976 Outcomes display using API

### LINKED ISSUE #976: Outcomes: Re-factor using new API
According to spec: https://github.com/openmsupply/application-manager-server/issues/556

Re-implement current Outcomes display using new API, and rendering cells with specified formatting options (including defaults for basic types)

### LINKED ISSUE #992: Outcomes -- standardize dataType between back-end and front-end
Current the Outcomes API returns "dataType", which is just the type returned by postGres. We should standardize these so that what the front-end recieves is a closer to javascript type.

E.g "timestamp", "timestamp with timezone" and "date" should all be just "Date".

CarlosNZ:
## From latest PR in this sequence #991 

Adds functionality to handle arrays returned from outomes API. Turns it into a Markdown list, and interprets each item in array in the same way as a single item.

Also uses the same "Substitution" method as used in listBuilder and generateText to populate strings with returned data.

Together this allows drill-down functionality in Outcomes, e.g. Orgs show a click-able list of member users, and vice versa.

Please continue to use back-end https://github.com/openmsupply/application-manager-server/pull/573, and the "outcomes-demo" snapshot in the "#outcomes-demo" branch of templates repo.

Example:

https://user-images.githubusercontent.com/5456533/133909714-c3f2ab31-da53-4c8f-a66c-697f44bd7b9c.mov

---

## From 2nd PR in this sequence #984 :

Based off previous PR #983 (so use same snapshot and back-end)

This was actually a really simple change, but I think it's a lot nicer:

Screen Shot 2021-09-15 at 8 31 15 PM

Screen Shot 2021-09-15 at 8 44 33 PM

You can't even tell the difference between elements that are plain text and those that are rendered with (shortText) SummaryView components, which is nice. :)

So basically, there's no difference between how to configure a table display and a details display element. They're exactly the same, it'll render as a SummaryView component if an element code and parameters are provided, otherwise it'll render as (optionally formatted) text.

UPDATE: Also fix #992 by returning a simplified range of DataTypes from API

---

## From this original PR:

Fixes #976 

Front-end part for back-end PR: https://github.com/openmsupply/application-manager-server/pull/573

Massive PR, sorry, but switching to a different API is not something easily done by halves!

**EDIT**: Added two additional PRs to further improve Outcomes layout and formatting. #984 #991 If you're reviewing UI, please just look at the last one #991. Only go through PRs one by one if you want to inspect code changes more incrementally. 

Please use the snapshot "outcomes_demo". I have put in the "dev/snapshots" folder in the templates repo on a new branch called `#outcomes-demo` (It's basically an export of the recent Angola demo snapshot, with added Outcome configurations)
(Note: you'll need to do a fresh `database_init` on the back-end before loading snapshot, or else you won't even be able to get to the snapshot UI. Or else just load it through the command line)

~~Have mostly tried to keep existing functionality, with just a few small tweaks/improvements -- main job was just simplifying the Outcome querying by using the server API, which also allows it to show different outcome layouts based on permissions.~~ -- I tell a lie -- have made layout improvements [here](https://github.com/openmsupply/application-manager-web-app/pull/991)

### Summary

- Outcome data is returned using three basic hooks -- useOutcomesList, useOutcomesTable & useOutcomesDetails, which share common functionality
- Remove "OutcomeList" component (as outcome list is accessed from UserArea menu)
- Remove wrappers for OutcomeTable and OutcomeDetail, as no longer necessary
- Remove Outcome context as also not necessary
- update OutcomeTable and OutcomeDetail to use the above hooks -- what they render is basically the same
- Improvements to formatting of display elements.
  - Both Table cells and Details elements follow (mostly) the same logic, although table defaults to returning Markdown-rendered text (but can show SummaryView components if needed, e.g. Org logos); and Details elements display SummaryView elements (as it previously did). Any fields that return text get turned into "shortText" elements showing label / value (so all Details elements have the same style)
  - The processing logic for getting the correct display element is basically as follows:
    - if data exists for rendering a SummaryView component, then return that Component (used for Formulations, images, etc.)
    - if not, treat returned value as "text" and pass it through "additional formatting" logic (specified in "additionalFormatting" field). Currently this only has "Date display" and simple "substitution" (to add Markdown formatting, or custom strings, for example), but can be extended as we require.
- Pagination for Outcome tables view -- uses same Pagination component as Application list. Also has sorting, but that's only possible through url parameters currently (no UI), as there are additional complications to deal with in another job.

### Testing

Use snapshot mentioned above. Things to note:

- Dates in Outcomes formatted nicely 👍
- Admin user can see different "Users" table (and details) to "public" users (including Linked Applications only available to Admin) 👍
- Only Admin can see "Licences" at all 👍
- Org logos in Organisations table :) 👍
- Thumbnails missing from snapshot, so the images are broken in most "Documents" details display (issue) 👎
- Most other mistakes in the display (Can't resolve object, [Object, Object] etc) are because the database has bad data 👎
- Org details has a "Members" field, which returns a list of all users that belong to an org. However, the structure of the returned data is not very nice, so haven't got a good way to display it (yet!) 👎

### Additional fixes / layout tweaks

- "Back" link on Details page (copied from "Back to list" in Application Summary page)
- Tried to make Details page more of a "Grid" look. It's far from ideal, but just wanted to mostly keep existing functionality for now
- Small fix for "usePageTitle" hook (so it reacts to changes in input parameters)
- Fixed the problem when going to Summary page the "back" button wouldn't work. Needed to add the "replace (is true)" parameter to the activeSections "updateQuery" function


nmadruga:
Starting to review this one now -- exciting!!

One first not-so-important and more as a usability improvement would be: when not showing the linked application maybe shouldn't also display the Linked Application titles and table on the bottom:
Screen Shot 2021-09-17 at 5 04 44 PM

nmadruga:
Cool error handling. It's simple but looks better as centered and styleless like this - maybe something to add on other places (instead of using a message). Not for now - of course

Screen Shot 2021-09-17 at 5 14 18 PM

nmadruga:
Something extra may be requiried to display the members? Let me know if this is already known and to be done later one : ) I'm just listing here everything that I see to far so we can fell safer about merging this in and adding to the new build next week 🎉 
Screen Shot 2021-09-17 at 6 12 23 PM

CarlosNZ:
Hey @nmadruga for ui related comments, try the next PR first, it's much nicer. This one was mainly to replicate existing functionality using new API, the part 2 one has much better UI stuff.

nmadruga (commented):
> try the next PR first, it's much nicer

Ok, Thanks! I thought I was doing that, but accidentally used the previous one.

CarlosNZ:
> One first not-so-important and more as a usability improvement would be: when not showing the linked application maybe shouldn't also display the Linked Application titles and table on the bottom:

That is what it does. What you're seeing is for an outcome that doesn't have a linked application (which shouldn't happen often in real world). When linked applications are disallowed, you don't see any of it.

CarlosNZ:
> Something extra may be required to display the members? Let me know if this is already known and to be done later one : ) I'm just listing here everything that I see as weird so we can feel safer about merging this in and adding to the new build next week 🎉 
> 
> Screen Shot 2021-09-17 at 6 12 23 PM
> 
> 

I'm working on it for the next PR (#3)

CarlosNZ:
> I'm just listing here everything that I see as weird so we can feel safer about merging this in and adding to the new build next week 🎉

We can just hide that field if it it's not working properly (but it should be by then)

nmadruga:
I think this one is also not right... Any idea of what could be wrong?
Screen Shot 2021-09-17 at 6 20 10 PM

src/utils/constants.ts
```diff
@@ -255,4 +255,10 @@ export default {
   TEMPLATE_BUTTON_SAVE: 'Save',
   TEMPLATE_BUTTON_ADD: 'Add',
   TEMPLATE_APPEARS_IN: 'Appears in',
+  OUTCOMES_TABLE_PAGINATION_TEXT: 'Items per page',
+  APP_LINK_LINKED_APPLICATIONS: 'Linked Application',
```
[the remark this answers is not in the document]
nmadruga: Already fixed in next PR 👍

CarlosNZ:
> I think this one is also not right... Any idea of what could be wrong?
> 
> Screen Shot 2021-09-17 at 6 20 10 PM
> 
> 
> 
> 

Yes, the thumbnails are missing from the snapshot. I made a pr this afternoon that fixes it.

nmadruga:
But should that be the logo? I thought it was weird to be showing on the registration element. I can check maybe that's how it's linked in the organisation application in this template

CarlosNZ:
> But should that be the logo? I thought it was weird to be showing on the registration element. I can check maybe that's how it's linked in the organisation application in this template

I need to double-check it, but I think people were just uploading any old thing (including logos) for documentation questions. 

But all this stuff will be solid by the 3rd PR (I spent an hour or so this arvo "cleaning up" some of the data, such as restoring thumbnails and removed cells with nonsense in them).

nmadruga:
I'm not sure why this title is showing like that, with id and type together
Screen Shot 2021-09-17 at 7 05 42 PM

src/utils/hooks/useOutcomes.tsx
```diff
@@ -0,0 +1,122 @@
...
+type ErrorResponse = {
+  statusCode: number
+  error: string
+  message: string
+}
+
+export const useOutcomesList = ({ templatePermissions }: OutcomeListProps) => {
+  // Note: we don't *need* templatePermissions, we only pass it in so that the
+  // hook can react to changes, since JWT (what we *actually* need) is not in
+  // State (perhaps it should be?)
```
nmadruga, line 39: Not sure what you mean by "not in state" I think this is included in userState. Maybe you can use this here?

src/utils/hooks/useOutcomes.tsx
```diff
@@ -0,0 +1,122 @@
...
+    .then((response) => {
+      if (response?.statusCode) {
+        setState(response, false, undefined)
+        return
+      } else setState(null, false, response)
+    })
+    .catch((error) => {
+      setState(error, false, undefined)
+    })
+}
```
nmadruga, line 122: I checked that different user can still get access to many outcomes that shouldn't be accessible to them, such as other organizations applied by other users and other users in the system. Was wondering if we should add the restrictions as part of this (or maybe in a part-2 PR) to start adding view restrictions to applicant users to only be able to view their outcomes and review users to be able to see all outcomes (for now)

nmadruga (changes requested):
Just a few comments with images that may be relevant + figuring out:
- restrictions should be in place ?

nmadruga:
@CarlosNZ could you have a look on what is pending for this one today? I'll do another review on the top PR: #991 and back-end now : )

src/utils/hooks/useOutcomes.tsx
CarlosNZ: I just mean that nothing changes in response to the JWT changing, as it's not stored anywhere in UserState. So we can't pass it in to the dependency array here. So instead I'm just passing in "templatePermissions", which is derived from JWT, and *is* in UserState, so we can react to it changing.

src/utils/hooks/useOutcomes.tsx
CarlosNZ: The Outcomes API handles column-based restrictions, not row-based, so the reason you're currently seeing what you're seeing is because we don't have these policies applied to user and org table yet. The mechanism is in here -- it passes the user's JWT to the GraphQL query, but the table is not yet restricted.

There is an issue (https://github.com/openmsupply/application-manager-server/issues/589) for Andrei to address this. However, we also realised that (for outcomes that aren't "standard" i.e. user and org) we should implement some row-based restrictions in Outcomes to handle these, since dynamically adding policies to new tables is tricky -- hence [this issue](https://github.com/openmsupply/application-manager-server/issues/590) to deal with that.

src/utils/hooks/useOutcomes.tsx
nmadruga: Ok - I might not be seeing exactly what you would like to change this to be, so you are considering that the UsetState not only store the templatePermissions but also the JWT received. This is because the templatePermissions don't change when we get a new JWT? I'm not really following this part...
Also, you don't access UserState in the `useOutcomes` hook so it's not clear what change you were thinking. Maybe a code example here might help... What I'm understanding is - instead of passing `templatePermissions` as props you could just access it from UserState and trigger `useEffect` based on that. For example:
```
export const useOutcomesList = () => {
  // Note: we don't *need* templatePermissions, we only pass it in so that the
  // hook can react to changes, since JWT (what we *actually* need) is not in
  // State (perhaps it should be?)
  const [error, setError] = useState<ErrorResponse | null>(null)
  const [loading, setLoading] = useState(false)
  const [outcomesList, setOutcomesList] = useState<OutcomesResponse>([])
  const {
    userState: { templatePermissions },
  } = useUserState()

  useEffect(() => {
    const JWT = localStorage.getItem(LOCAL_STORAGE_JWT_KEY)
    if (!JWT) return
    const url = `${serverURL}/outcomes`
    processRequest(url, JWT, setError, setLoading, setOutcomesList)
  }, [templatePermissions])

  return { error, loading, outcomesList }
}
```

But this can be done already, no?

src/utils/hooks/useOutcomes.tsx
nmadruga: Great. Thanks for having already discussed this with Andrei. I'll have a look on these issues now

src/utils/hooks/useOutcomes.tsx
CarlosNZ: Yes, okay, I think it makes sense to get templatePermissions from UserState rather than passing it in. Thanks, I'll make this change shortly.

src/formElementPlugins/listBuilder/src/styles.less
```diff
@@ -16,5 +16,5 @@
 .fit-content {
   width: fit-content;
   width: -moz-fit-content; // Needed for Firefox
-  min-width: 100%;
+  min-width: 250px;
```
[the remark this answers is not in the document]
CarlosNZ: No, I don't think so.

src/containers/Outcomes/OutcomesTable.tsx
```diff
@@ -1,74 +1,108 @@
...
+              {headerRow.map(({ title }: any) => (
                 <Table.HeaderCell key={title} colSpan={1}>
                   {title}
                 </Table.HeaderCell>
               ))}
             </Table.Row>
           </Table.Header>
           <Table.Body>
-            {tableData.map((row: any) => {
+            {tableRows.map((row: any) => {
```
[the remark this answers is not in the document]
CarlosNZ: Okay, cool: bdddc5fe

(My change is actually on the later PR branch, so won't see in this branch)

src/utils/hooks/useOutcomes.tsx
CarlosNZ: Cool, this is much nicer now, thanks for the suggestion 2d4620a0

(Change pushed to later branch though)

nmadruga (approved):

PR #984, opened 2021-09-15
#976(2) Improved details layout.

### LINKED ISSUE #992: Outcomes -- standardize dataType between back-end and front-end

CarlosNZ:
Based off previous PR #983 (so use same snapshot and back-end)

This was actually a really simple change, but I think it's a lot nicer:

Screen Shot 2021-09-15 at 8 31 15 PM

Screen Shot 2021-09-15 at 8 44 33 PM

You can't even tell the difference between elements that are plain text and those that are rendered with (shortText) SummaryView components, which is nice. :)

So basically, there's no difference between how to configure a table display and a details display element. They're exactly the same, it'll render as a SummaryView component if an element code and parameters are provided, otherwise it'll render as (optionally formatted) text.

UPDATE: Also fix #992 by returning a simplified range of DataTypes from API


nmadruga (approved):
LGTM

CarlosNZ:
Closing -- merged to base branch

PR #986, opened 2021-09-15
#964 Number input fix

### LINKED ISSUE #964: Bug - Number plugin element - shows error when value entered only by the stepper

CarlosNZ:
Fix #964 

Quick little fix. Turns out clicking the stepper was only triggering the "onChange" event (which doesn't save), not the "onBlur" event (which does). Added additional "onClick" listener which calls the same function as "onBlur", so we can save input from the stepper alone.

(Also -- fixed the annoying React error about incorrectly nested DOM elements which has been low-key bugging me for ages)


nmadruga (approved):
Great! LGTM

nmadruga (approved):
Great! LGTM

PR #989, opened 2021-09-17
#964 Number plugin fix (2)

### LINKED ISSUE #964: Bug - Number plugin element - shows error when value entered only by the stepper

CarlosNZ:
Fix #964 (again)

Sorry, was a little hasty with yesterday's fix, the "onClick" handler was running validation checks even before any text was entered (the handler is supposed to be just for onLoseFocus, so shouldn't have happened).

Anyway, added tiny extra check to not going any further if the response is still NULL (so it behaves like the other text input plugins again)


nmadruga (approved):
LGTM

PR #991, opened 2021-09-18
#976(3) Outcome display improvements -- drill-down/list handling

CarlosNZ:
Adds functionality to handle arrays returned from outomes API. Turns it into a Markdown list, and interprets each item in array in the same way as a single item.

Also uses the same "Substitution" method as used in listBuilder and generateText to populate strings with returned data.

Together this allows drill-down functionality in Outcomes, e.g. Orgs show a click-able list of member users, and vice versa.

Please continue to use back-end https://github.com/openmsupply/application-manager-server/pull/573, and the "outcomes-demo" snapshot in the "#outcomes-demo" branch of templates repo.

Example:

https://user-images.githubusercontent.com/5456533/133909714-c3f2ab31-da53-4c8f-a66c-697f44bd7b9c.mov

### DIFF
--- a/src/containers/Outcomes/ApplicationLinks.tsx
+++ b/src/containers/Outcomes/ApplicationLinks.tsx
@@ -11,28 +11,28 @@ const displayFields: ApplicationDisplayField[] = [
   {
     field: 'name',
     displayName: strings.APP_LINK_NAME,
-    dataType: 'character varying',
+    dataType: 'string',
     link: '/application/',
     linkVar: 'serial',
   },
   {
     field: 'serial',
     displayName: strings.APP_LINK_SERIAL,
-    dataType: 'character varying',
+    dataType: 'string',
     link: '/application/',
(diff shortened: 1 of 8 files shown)


nmadruga (approved):
Looks great!

I couldn't see the list of organisations associated with the user (when logged as Jane smith) not sure if config was missing - although I haven't investigated this further.

Just to understand how to configure this currently - on other existing snapshots, could you please list what needs to be done somewhere so we can easily add this new config to **demo** and **basic_snapshot** at least when we move this to `develop` branch?

nmadruga:
Something I noticed (might be only related to how front-end deal with the retrieved data), but when a product is unexisting or the user doesn't have access to it, we need some better 404 handling:
Screen Shot 2021-09-23 at 2 36 36 PM

CarlosNZ:
Closing --merged to base branch

PR #994, opened 2021-09-21
#577B Actions triggered on Re-assignment

nmadruga:
## Linked PRs
- Back-end https://github.com/openmsupply/application-manager-server/pull/586
- Templates to use `dev.current` on branch `reassignment-changes` in application-manager-templates

### Changes
- When selecting option "Re-assign" in UI a second dropdown shows for selecting the new reviewer
- After selecting the next reviewer the reassignment is completed by Front-end
- Feature to change re-assignment already implemented partially in #982. Just added the ability to disable the previous review.status linked to the un-assigned reveiwer to have review.status = DISCONTINUED (action that runs ON_REVIEW_UNASSIGN)
- Also allows to revert that by setting an existing review associated with the new ReviewAssignment assigned to have review.status = DRAFT.

### Testing
- Using `dev` snapshot, login as `testAssigner1` and check review home page of application `S-MAS-0004`
  - you can see it's assigned to `assessor1` which is disabled since already submitted
  - and `assessor2` can be re-assigned
-  Log in as `assessor2` and start a review, just add one review to one of the responses
- Log in again as `testAssigner1` and select to reassign section 2 from `assessor2` to `Reviewer2`
  - This would have blocked `assessor2` from continue their review, and allo Reviewer2 to start one
- Log in again as `testAssigner1` and reassign back section 2 from `Reviewer2` to `assessor2`
  - Now if you log in as `assessor2` you can see the previous review is back and can be continued from where stopped

### Improvement
- Also added a better handling to Review subheader, so when is in progress, submitted or locked it show different messages. The discontinued in this case is also treated as locked:
Screen Shot 2021-09-21 at 9 34 46 PM
Screen Shot 2021-09-21 at 9 32 13 PM
Screen Shot 2021-09-21 at 9 34 38 PM

PR #995, opened 2021-09-21
#987 Add Unassignment logic in UI

### LINKED ISSUE #987: Add Un-assignment logic
Maybe would require action to run on back-end to remove other reviewQuestionAssignment, review other associations - or we would keep those and have to decide what to do in case of re-assignment of the same user:
- Show existing assignment and resposes
- Or create new assignment, review and responses
- If keep both need to check for any logic that verify for previous responses (history for example) to also check if review_responses is not from previous review (un-assigned). 

Maybe can add a new status to review_responses to set as LOCKED and revise all the logic of review in status LOCKED - when it's from un-assignment it should get set back to other status. So we might need to flag that differently from a LOCKED due to submission to applicant.

nmadruga:
Fixes #987 

## Linked PRs
- Based on #994 
- Back-end https://github.com/openmsupply/application-manager-server/pull/586
- Templates to use `dev.current` on branch `reassignment-changes` in application-manager-templates

### Changes
- Add mutation to only do the un-assignment of selected user
- TODO: Show confirmation before allowing un-assignment to happen (probably good to add that)
- TODO: Missing core-action to also reactivate one previously un-assigned Review to go back when re-assigning, so ON_REVIEW_ASSIGN will need to run the same Core-action as ON_REVIEW_REASSIGN. Will add to `dev` for testing...


nmadruga:
Warning - still missing new CORE-ACTIONS on `basic_snapshot`

nmadruga:
> Warning - still missing new CORE-ACTIONS on `basic_snapshot`

Ok - Added to `core_templates` in previous PR (how basic_snapshot is called now)
And in snapshot `reassignments` on `dev` snapshots folder in our private repo

CarlosNZ:
Question: Is "Locked" meant to show up in the UI at all? It doesn't as far as I could see.

The thing is, if Admin unassigns a reviewer who has already started their review, there is nothing to tell them what's going on.
Screen Shot 2021-10-04 at 11 19 57 AM
It still shows in the Dashboard as "Draft reviews of applications", but when I click it, there is no Action or other info visible in the List UI. We probably need to have some indication to a reviewer that their review is "Locked", don't you think? Otherwise reviewers will be quite confused, and it will be annoying to have reviews started that they can no longer access but still see in the UI.

Maybe not for this issue, but I just wanted to get some clarify around what is happening vs what you think should be happening?

nmadruga:
> Question: Is "Locked" meant to show up in the UI at all? It doesn't as far as I could see.
> 
> The thing is, if Admin unassigns a reviewer who has already started their review, there is nothing to tell them what's going on. Screen Shot 2021-10-04 at 11 19 57 AM It still shows in the Dashboard as "Draft reviews of applications", but when I click it, there is no Action or other info visible in the List UI. We probably need to have some indication to a reviewer that their review is "Locked", don't you think? Otherwise reviewers will be quite confused, and it will be annoying to have reviews started that they can no longer access but still see in the UI.
> 
> Maybe not for this issue, but I just wanted to get some clarify around what is happening vs what you think should be happening?

That's an interesting point. I was a little confused when you used Locked, because there is 2 types of locking an review:
  - When another reviewer submits back to the applicant
  - When you get un-assigned (which is being set as status = DISCONTINUED)

I think something related to that could be done here. For example, removing from filter Draft if the review is DISCONTINUED. I think that Locked ones can still be visible as Draft, although unsumittable. But when you are un-assigned, it shouldn't really show in the Dashboard anymore.

PR #997, opened 2021-09-22
#587B Quick fix to allow viewing lookup tables inserted with plural names

nmadruga:
Fixes back-end issue https://github.com/openmsupply/application-manager-server/issues/587

Needs to run `yarn` to update new package included 

### Changes
- Just remove added `s` to table names, which are usually a plural name, and fails when trying to query with extra `s`
- Add pluralize package and update tableNmae in lookup import when not plural


CarlosNZ:
Cool, does this solve both problems in https://github.com/openmsupply/application-manager-server/issues/587 ?

If so, can you update that issue and unassign sworup?

nmadruga:
> Cool, does this solve both problems in [openmsupply/application-manager-server#587](https://github.com/openmsupply/application-manager-server/issues/587) ?
> 
> If so, can you update that issue and unassign sworup?

The second problem isn't a real problem. We just need some documentation to go with editing Lookuptables.
I do have a second part for this fix to use the `pluralization` package like you suggested.

src/LookupTable/graphql/getDynamicSingleTable.query.ts
```diff
@@ -3,7 +3,7 @@ import { buildFieldList, capitalizeFirstLetter, toCamelCase } from '../utils'
 
 const getDynamicSingleTable = (structure: any) => gql`
     query getDynamicSingleTable {
-      lookupTable${capitalizeFirstLetter(toCamelCase(structure.name))}s {
+      lookupTable${capitalizeFirstLetter(toCamelCase(structure.name))} {
```
CarlosNZ, line 6: Not a priority right now, but we should probably replace these with Lodash's case conversions, which we're already doing elsewhere. 

Then you can just use "StartCase" (I think) rather than chaining two methods like this

src/LookupTable/hooks/useGetSingleTable.hook.tsx
```diff
@@ -45,7 +45,7 @@ const useGetSingleTable = () => {
     if (structure?.name) {
       const tableName = `${GQL_TABLE_NAME_PREFIX}${capitalizeFirstLetter(
         toCamelCase(structure.name)
```
CarlosNZ, line 47: Same.

CarlosNZ (approved):
LGTM

Maybe replace those case conversion methods, but only if it's a simple thing.

If you merge this to develop, please don't delete the branch, as I'll merge to it demo branch later on today.

src/LookupTable/graphql/getDynamicSingleTable.query.ts
nmadruga: Updating now...

src/LookupTable/hooks/useGetSingleTable.hook.tsx
nmadruga: Thanks! I'm replacing everywhere that was using the local version of startCase

src/LookupTable/graphql/getDynamicSingleTable.query.ts
nmadruga: Hmmm turns out that lodash doesnt have a converter to that already... It also would need to use one function inside the other... So not worth it IMHO

CarlosNZ: Hmm, I thought it did, I used something similar for Outcomes. Never mind, I'll change it at some point if I find it.

CarlosNZ: It's called PascalCase, but yeah it looks like it's not in Lodash :(

PR #998, opened 2021-09-22
#993 Hide "Dev" switcher in production

### LINKED ISSUE #993: Hide users/snapshot switcher in productions builds
Hide this whole bit of UI for builds.

However, we still need a way to load snapshots.

Suggestion: make a new page (`/admin/snapshots`), which is only accessible with an "Admin" JWT that contains the snapshot import/load UI

CarlosNZ:
Fix #993 

Implemented as stated in issue.

- Hide `DevOptions` completely
- Move (and modify layout to fit) the "Snapshots" list component into a page, which is only accessible with Admin JWT
- accessed at `/admin/snapshots` -- note: intentionally omitted from "Admin" menu, we don't really want this accessible to anyone other than devs

### DIFF
--- a/src/containers/Dev/DevOptions.tsx
+++ b/src/containers/Dev/DevOptions.tsx
@@ -4,15 +4,15 @@ import Snapshots from './Snapshots'
 import UserSelection from './UserSelection'
 
 const DevOptions: React.FC = () => {
-  return (
+  return process.env.NODE_ENV === 'production' ? null : (
     <div id="dev-options" style={menuStyle}>
       <Dropdown item icon="user">
         <UserSelection />
       </Dropdown>
 
-      <Dropdown item icon="file video outline">
+      {/* <Dropdown item icon="file video outline">
         <Snapshots />
-      </Dropdown>
+      </Dropdown> */}
     </div>
   )
 }
--- a/src/containers/Dev/Snapshots.tsx
+++ b/src/containers/Dev/Snapshots.tsx
@@ -1,15 +1,5 @@
 import React, { useState, useEffect, useRef } from 'react'
-import {
-  Button,
-  Grid,
-  GridColumn,
-  Icon,
-  Input,
-  Label,
-  Loader,
-  Modal,
-  Popup,
-} from 'semantic-ui-react'
+import { Button, Grid, Icon, Input, Label, Loader, Modal } from 'semantic-ui-react'
 import config from '../../config'
 
 const snapshotsBaseUrl = `${config.serverREST}/snapshot`
@@ -21,22 +11,20 @@ const uploadSnapshotUrl = `${snapshotsBaseUrl}/upload`
 // const diffSnapshotUrl = `${snapshotsBaseUrl}/diff`
 
 const Snapshots: React.FC = () => {
-  const [isOpen, setIsOpen] = useState(false)
   const [compareFrom, setCompareFrom] = useState('')
   const [isLoading, setIsLoading] = useState(false)
-  const [snapshotError, setSnapshotError] =
-    useState<{ message: string; error: string } | null>(null)
+  const [snapshotError, setSnapshotError] = useState<{ message: string; error: string } | null>(
+    null
+  )
 
   const [data, setData] = useState<string[] | null>(null)
 
   useEffect(() => {
-    if (isOpen) {
-      setData(null)
-      setCompareFrom('')
-      setSnapshotError(null)
-      getList()
-    }
-  }, [isOpen])
+    setData(null)
+    setCompareFrom('')
+    setSnapshotError(null)
+    getList()
+  }, [])
 
   const getList = async () => {
     try {
@@ -53,7 +41,6 @@ const Snapshots: React.FC = () => {
 
   const takeSnapshot = async (name: string) => {
     if (!name) return
-    setIsOpen(false)
     setIsLoading(true)
     try {
       const resultRaw = await fetch(`${takeSnapshotUrl}?name=${normaliseSnapshotName(name)}`, {
@@ -70,7 +57,6 @@ const Snapshots: React.FC = () => {
   }
 
   const useSnapshot = async (name: string) => {
-    setIsOpen(false)
     setIsLoading(true)
     try {
       const resultRaw = await fetch(`${useSnapshotUrl}?name=${name}`, {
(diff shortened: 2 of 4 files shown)


CarlosNZ:
Note -- have created new label "#toDemo" for tagging PRs to go into demo server (before merging to develop). Idea is to delete this tag when either:
- merged into demo
- merged into develop

So we can keep track of it.

src/containers/Dev/DevOptions.tsx
```diff
@@ -4,15 +4,15 @@ import Snapshots from './Snapshots'
 import UserSelection from './UserSelection'
 
 const DevOptions: React.FC = () => {
-  return (
+  return process.env.NODE_ENV === 'production' ? null : (
```
nmadruga, line 7: I have seen this today and agree:
https://github.com/openmsupply/application-manager-web-app/blob/ddbd13643747772d01f3b3f867b21059342fba82/src/containers/Application/ApplicationSummary.tsx#L14
I think we could do this now - not sure exactly how. Can we update the config as part of the building process to have the selected environment set in a field on `config.ts` file?

src/containers/Dev/DevOptions.tsx
```diff
@@ -4,15 +4,15 @@ import Snapshots from './Snapshots'
...
     <div id="dev-options" style={menuStyle}>
       <Dropdown item icon="user">
         <UserSelection />
       </Dropdown>
 
-      <Dropdown item icon="file video outline">
+      {/* <Dropdown item icon="file video outline">
         <Snapshots />
-      </Dropdown>
+      </Dropdown> */}
```
nmadruga, line 15: Maybe this line can be removed

nmadruga (changes requested):
Just one tiny thing about using the `production` environment variable

src/containers/Dev/DevOptions.tsx
CarlosNZ: Good idea, not sure how to do that though. You?

nmadruga:
Hmmm - just thought about something while working in a different issue, no having the snapshot dropdown at the top of what you are currently doing (showing within the user switcher) while in development will be annoying since I use this a lot to quickly make a snapshot without having to open another page and then go back to where I was.

Can we still keep this in the same place while on `dev` mode?

CarlosNZ:
> Can we still keep this in the same place while on `dev` mode?

We could, but then we'd have to keep two separate components that are very similar but not the same (since they need different layout on a page to a drop-down, but everything else it the same).

What about just keeping the "/snapshots" page permanently open in a different tab? Would that work for you?

CarlosNZ:
Actually -- one thing this definitely needs before merging is a "Delete snapshot" button. The server is getting filled with a massive list of snapshots, would really help if we could remove them, it's getting ridiculous.

src/containers/Dev/DevOptions.tsx
nmadruga: I believe this was sorted in another PR

nmadruga:
> What about just keeping the "/snapshots" page permanently open in a different tab? Would that work for you?

Sure. I guess that's alright for now

nmadruga:
> Actually -- one thing this definitely needs before merging is a "Delete snapshot" button. The server is getting filled with a massive list of snapshots, would really help if we could remove them, it's getting ridiculous.

hahah good idea! Can you add that to this PR?

CarlosNZ:
> hahah good idea! Can you add that to this PR?

Yes, I'll try and squeeze it in this afternoon.

CarlosNZ:
> > Actually -- one thing this definitely needs before merging is a "Delete snapshot" button. The server is getting filled with a massive list of snapshots, would really help if we could remove them, it's getting ridiculous.
> 
> hahah good idea! Can you add that to this PR?

@nmadruga Done! 7aa9e5eb

But it also needs this [back-end endpoint PR](https://github.com/openmsupply/application-manager-server/pull/602) approved before this one can be merged.

nmadruga:
> > > Actually -- one thing this definitely needs before merging is a "Delete snapshot" button. The server is getting filled with a massive list of snapshots, would really help if we could remove them, it's getting ridiculous.
> > 
> > 
> > hahah good idea! Can you add that to this PR?
> 
> @nmadruga Done! [7aa9e5e](https://github.com/openmsupply/application-manager-web-app/commit/7aa9e5eb2f87a538ffa1c68faad067a3cb901014)
> 
> But it also needs this [back-end endpoint PR](https://github.com/openmsupply/application-manager-server/pull/602) approved before this one can be merged.

Cool!! Will have a look in a momment : )

src/containers/Dev/Snapshots.tsx
```diff
@@ -161,8 +185,7 @@ const Snapshots: React.FC = () => {
                   />
                 </>
               ) : null} */}
```
nmadruga, line 187: Can all the code above not in use be deleted?

src/containers/Dev/Snapshots.tsx
```diff
@@ -193,16 +216,24 @@ const Snapshots: React.FC = () => {
...
-          <Icon className="clickable" name="record" onClick={() => takeSnapshot(value)} />
-        </div>
-      </Grid.Row>
+      <>
+        <Table.Cell>
+          <div className="flex-row-start" style={{ alignItems: 'center' }}>
+            <Input
+              size="mini"
+              onChange={(_, { value }) => setValue(value)}
+              placeholder="New Snapshot"
```
nmadruga, line 225: I guess we are skipping localisations for snapshot currently?

src/containers/Dev/Snapshots.tsx
```diff
@@ -211,45 +242,45 @@ const Snapshots: React.FC = () => {
...
-        />
-      </Grid.Row>
+      <>
+        <Table.Cell colSpan={4} textAlign="right">
+          <Button
+            primary
+            // color="blue"
+            // size="mini"
+            onClick={() => fileInputRef?.current?.click()}
+            // labelPosition="right"
```
nmadruga, line 252: More to go : )

nmadruga (approved):
LGTM

src/containers/Dev/Snapshots.tsx
CarlosNZ: I dunno, this is all stuff that Andrei made and commented out before I even looked at it, so I didn't really take much notice of it. You reckon just get rid of it?

nmadruga: Hmmm this is the diff stuff! Yes, I guess we should leave it for now.

src/containers/Dev/Snapshots.tsx
CarlosNZ: For now, I guess. We should update them if/when we make this tool a bit more user-focused. But for now only devs should be going near it.

PR #1001, opened 2021-09-22
TB - Added segment at the start of each action line combining all action tools

nmadruga:
## Linked PR
- Based on PR #957 

### Changes
- **Not a priority for now so just added to a PR for later review**
- Small change to component `CheckboxIO` to show a slider in a simplified option when receiving the props `basic` 
- Just another layout addition - hope is good and helpful to others - Show all tools in the first column of each action, combining also the `isSequential` option that used to be at the end of the row.

### Images
Screen Shot 2021-09-23 at 11 52 43 AM
Screen Shot 2021-09-23 at 11 52 36 AM
Screen Shot 2021-09-23 at 11 52 23 AM

### DIFF
--- a/semantic/src/site/globals/site.overrides
+++ b/semantic/src/site/globals/site.overrides
@@ -1762,6 +1762,10 @@ a:hover {
 .config-container-alternate {
   .config-container();
   background-color: rgba(0, 0, 0, 0.03);
+
+  .ui.segment {
+    .padding-zero();
+  }
 }
 
 /*******************************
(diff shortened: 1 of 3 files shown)


CarlosNZ:
I don't really get this. I like being able to see the Description on the main panel without opening anything, but the slider seems weird to me. It completely dominates the Segment, which is strange for a switch that is hardly ever used. Probably best to put it back over to the right of the "condition" block where it's more unobtrusive?

nmadruga:
Closed - won't be included

PR #1002, opened 2021-09-23
Localisations in lookup tables UI

nmadruga:
### Linked PR
- Based on #997 

### Changes
- Try to add localised strings in all Lookup tables module


CarlosNZ (approved):
Cool, looks good. Don't think I really need to test -- if it compiles, it's all good, right?

nmadruga:
> Cool, looks good. Don't think I really need to test -- if it compiles, it's all good, right?

Yes, I have done quick testing, looks good to me.

src/utils/constants.ts
```diff
@@ -255,4 +269,29 @@ export default {
...
+  LOOKUP_TABLE_IMPORT_OTHER: 'Import another CSV',
+  LOOKUP_TABLE_DOWNLOAD: "Download '%1' table",
+  LOOKUP_TABLE_FIELD_NAME: 'Database Field-name',
+  LOOKUP_TABLE_LABEL: 'Lookup-table Label',
+  LOOKUP_TABLE_NAME: 'Table name',
+  LOOKUP_TABLE_ADD: 'Add new Lookup-table',
+  LOOKUP_TABLE_TITLE: 'Lookup Table: %1',
+  LOOKUP_TABLE_CONTENTS: 'View individual lookup-table and its contents',
+  LOOKUP_TABLE_SUBTITLE: 'This page contains the list of all lookup-tables',
+  LOOKUP_TABLE_SUCCESS: "Lookup table '%1' has been successfully %2.",
```
CarlosNZ, line 295: I don't think we should use substitutions that replace with other language strings. Isn't that making an assumption about the structure of the language?

I think this would be safer as two complete sentences:
`"Lookup table '%1' has been successfully loaded`
`"Lookup table '%1' has been successfully created`

Because different languages might have different ways to order those sentences, they might not just drop in single words in the same way. Also (and perhaps more importantly) it's easier for the person translating. If they see a string that just has "loaded" or "created" out of context, they might put the wrong word in without knowing the sentence context.

Like, if I'm translating the word "loaded" out of context, I don't know if it's referring to "loaded" as in software, "loaded" as in "stacked up", or "loaded" as in "drunk". If I was translating the whole sentence, it would be clear.

So, as a general rule, I think localised strings should be full phrases/sentences as much as possible, even if there is some repetition in most of the sentence. The translator shouldn't have to cross-check against other strings and perform mental gymnastics to figure out what's being said from multiple strings. Do you agree?

src/utils/constants.ts
```diff
@@ -255,4 +269,29 @@ export default {
...
+  APP_LINK_TYPE: 'Application Type',
+  APP_LINK_COMPLETED: 'Date Completed',
+  LOOKUP_ERROR_NOT_FOUND: 'No lookup-table found!',
+  LOOKUP_ERROR_TITLE: 'Error loading lookup-table',
+  LOOKUP_TABLE_HEADER_LABEL: 'Label',
+  LOOKUP_TABLE_HEADER_NAME: 'Table Name',
+  LOOKUP_TABLE_HEADER_ACTIONS: 'Actions',
+  LOOKUP_TABLE_VIEW: 'View lookup table',
+  LOOKUP_TABLE_IMPORT: 'Import lookup table',
+  LOOKUP_TABLE_IMPORT_INTO: 'Import into Lookup-table: %1',
```
CarlosNZ, line 285: Substitutions (%1) are fine within sentences, but I think we should keep it to a minimum and don't put a substitution in a localised string if possible. In this case, it's unnecessary -- the string should just be "Import into look-up table" and the ": {table name}" part should be purely in the code and not something the translator should have to think about.

src/utils/constants.ts
```diff
@@ -96,13 +98,22 @@ export default {
   LABEL_CONSOLIDATION_DISAGREE: 'Disagree',
   LABEL_CONSOLIDATION_AGREEMENT: 'Agreed',
   LABEL_CONSOLIDATION_DISAGREEMENT: 'Disagreed',
+  LABEL_CREATED: 'created',
```
CarlosNZ, line 101: See comment below about "loaded" and breaking up sentences. I don't think it's a good idea.

CarlosNZ (changes requested):

src/utils/constants.ts
nmadruga: Ok - cool! I'll be changing this as well : )

src/utils/constants.ts
```diff
@@ -284,16 +282,17 @@ export default {
...
   LOOKUP_TABLE_LABEL: 'Lookup-table Label',
   LOOKUP_TABLE_NAME: 'Table name',
   LOOKUP_TABLE_ADD: 'Add new Lookup-table',
-  LOOKUP_TABLE_TITLE: 'Lookup Table: %1',
+  LOOKUP_TABLE_TITLE: 'Lookup table',
   LOOKUP_TABLE_CONTENTS: 'View individual lookup-table and its contents',
   LOOKUP_TABLE_SUBTITLE: 'This page contains the list of all lookup-tables',
-  LOOKUP_TABLE_SUCCESS: "Lookup table '%1' has been successfully %2.",
+  LOOKUP_TABLE_SUCCESS_CREATED: 'has been successfully created.',
+  LOOKUP_TABLE_SUCCESS_UPDATED: 'has been successfully updated.',
```
CarlosNZ: Hmm, I might be slightly contradicting myself here, but I think "full sentences" should take priority over minimising substitutions. 

So perhaps these two strings should be:
- `Lookup Table %1  has been successfully created.`
- `Lookup Table %1  has been successfully updated.`

I think it's easier to read and simpler for a translator to follow, even with the %substitution in it. What do you think?

nmadruga: Not a big difference to me. So ok to do as you suggested.

nmadruga: Done 👍

PR #1003, opened 2021-09-23
#559B Fix list builder size container

nmadruga:
Partial for back-end issue https://github.com/openmsupply/application-manager-server/issues/559

### Changes
- Only fixed for table view (there is also Card and list views... probably good to check)
- Will just auto display a scrollbar when the table is larger than its container
Screen Shot 2021-09-23 at 5 32 16 PM

### DIFF
--- a/semantic/src/site/collections/table.overrides
+++ b/semantic/src/site/collections/table.overrides
@@ -87,3 +87,11 @@
     border-bottom-right-radius: 10px;
   }
 }
+
+// For listBuilder table not overlap the Application/Review
+// Page area - will show scroll when bigger than container
+.ui.table.list-builder {
+  overflow: auto;
+  display: inline-block;
+  max-width: inherit;
+}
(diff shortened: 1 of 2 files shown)


semantic/src/site/collections/table.overrides
```diff
@@ -87,3 +87,11 @@
...
   }
 }
+
+// For listBuilder table not overlap the Application/Review
+// Page area - will show scroll when bigger than container
+.ui.table.list-builder {
+  overflow: auto;
+  display: inline-block;
+  max-width: inherit;
+}
```
CarlosNZ, line 97: We've been putting plug-in specific CSS in the plugin folder. In fact, the list-builder already has a "styles.less" file in it, so that CSS should go in there.

CarlosNZ (changes requested):

CarlosNZ:
Hmm, I dunno. Personally I find a horizontal scrollbar for a small window way more annoying than a overhang. Others may disagree.

Just having a poke around, we need to do something with the max-width of "application-summary-content", but it's really not a trivial thing to only make it grow when inner content (that's several levels deeper)

Maybe ask the others what they think? If they're fine with horizontal scrolling, then cool, we can do this for now. But personally I find horizontal scrollbars really ugly. Reminds of "frames" from the early 2000s.

semantic/src/site/collections/table.overrides
CarlosNZ: Also, pretty sure those last two properties (display, max-width) are not necessary.

CarlosNZ:
You want to close this one now and make a new issue to come up with a more satisfactory solution?

PR #1008, opened 2021-09-26
Quick fix -- outcomes dataTypes

CarlosNZ:
Little thing I noticed while writing docs for outcome configuration.

I had changed the back-end to return "dataType" field with types more consistent with JS basic types, but missed a couple of entries in front-end that use this value.


nmadruga (approved):
LGTM

PR #1010, opened 2021-09-27
#1009 Show template name on applications list

### LINKED ISSUE #1009: Applications List shouldn't show "code" as header
Regular users shouldn't ever see the template "code" in the UI. Header for this page should be template name, not code.

CarlosNZ:
Fix #1009 

Quick little improvement. It did occur to me that calling another loadTemplate hook just to get the name is a bit overkill, but I can't really see a simpler way to do it. I was going to just check the value of the "templateName" in the first item of the "Applications" array, but this won't work if there are no applications returned, and we still want to show a header in that case.

Suggestions welcome, but this is probably fine for now?


nmadruga (changes requested):
Another solution could be passing the `$templateCode` to query `getApplicationsList` and have a second block to only get templates related to that code (also filtering status AVAILABLE). e.g.:
```
templates(
    filter: {
      code: { equalTo: "UserRegistration" }
      status: { equalTo: AVAILABLE }
    }
  ) {
    nodes {
      name
    }
```

CarlosNZ:
> Another solution could be passing the `$templateCode` to query `getApplicationsList` and have a second block to only get templates related to that code (also filtering status AVAILABLE). e.g.:

Good idea, I'll try it out. Thanks.

CarlosNZ:
> > Another solution could be passing the `$templateCode` to query `getApplicationsList` and have a second block to only get templates related to that code (also filtering status AVAILABLE). e.g.:
> 
> Good idea, I'll try it out. Thanks.

Okay, I've done something like your suggestion. See if you prefer it: 6608e083

nmadruga:
Looks good to me - I haven't tried to run though. Will do that later today

nmadruga (approved):

nmadruga:
Have also resolved conflicts with `develop` hope it's good. Please check

Issue #595, opened 2021-04-12
Update layout and use global styles: Application Summary page


nmadruga:
Remember: Before changing styles take a screenshot on current page and add which style to use on each element - post to the issue so others can have a look and agree.

Just change elements with existing style. Anything not defined should be done in another (to be created) issue.

PR #1011, opened 2021-09-27
#594B (1) -- consolidate configs and prefs

CarlosNZ:
Requires  back-end [#595](https://github.com/openmsupply/application-manager-server/pull/595)

Summary, as specified in [this comment](https://github.com/openmsupply/application-manager-server/issues/594#issuecomment-928235451)

- separate "preferences" (i.e. things users might actually want to change) and "config" (i.e settings required for the build and shouldn't be changed while server is running)
- "config" stuff remains in existing config.ts files (front and back)
- "preference" stuff all goes in a common (back-end) preferences.json file, separated into "server" and "web" objects
- back-end config.ts imports prefs and continues as before
- front-end fetches prefs from server (/get-prefs endpoint) right at the top (before LanguageProvider), makes prefs available via `usePrefs` hook (Most things are still in "config" (server addreses etc. Main  thing we want in prefs is default language))
  - have re-configured Localisation a bit so that "languageOptions" list is fetched as part of the same request as prefs, so they'll be passed into LanguageProvider ready to go


src/contexts/Localisation.tsx
```diff
@@ -4,21 +4,25 @@ import strings from '../utils/defaultLanguageStrings'
 import { getRequest } from '../utils/helpers/fetchMethods'
 import { mapValues } from 'lodash'
 
-const { defaultLanguageCode, pluginsFolder } = config
+const { pluginsFolder } = config
 
-const initialLanguageCode = localStorage.getItem('language') ?? defaultLanguageCode
+const savedLanguageCode = localStorage.getItem('language')
```
nmadruga, line 9: Hmm are we always storing on localStorage? I thought this would be included in the user details... So after changing the selected language this would be persisted in the user created or edited in an existing user.

Not sure about implementation wise, do you think this would be much more difficult to do?

nmadruga (approved):
All good - just checking back-end now. I'm getting an error when loading the `core_templates` snapshot:
Screen Shot 2021-10-01 at 11 38 34 AM

CarlosNZ:
> All good - just checking back-end now. I'm getting an error when loading the `core_templates` snapshot:

Yeah, needs jwt secret in .env file: https://github.com/openmsupply/application-manager-server/pull/595#issuecomment-929038183

nmadruga:
> > All good - just checking back-end now. I'm getting an error when loading the `core_templates` snapshot:
> 
> Yeah, needs jwt secret in .env file: [openmsupply/application-manager-server#595 (comment)](https://github.com/openmsupply/application-manager-server/pull/595#issuecomment-929038183)

Oh yea. Thanks!

src/contexts/Localisation.tsx
CarlosNZ: Do you mean saving it in the database on the user table?

I thought about that, then I realised -- that needs them to be logged in before it can be determined. So they're still going to have to use local storage or manually select their language on the Login page. And after that we have it anyway, so no need to save to database!

You get what I mean?

PR #607, opened 2021-04-13
#573 Replace timeCreated with timeUpdated

CarlosNZ:
Fixes #573.

Requires [back end PR #304](https://github.com/openmsupply/application-manager-server/pull/304)

Very minor change, fairly sure it covers everything. Any ideas where else it might be used in front-end? (I dealt with all the compilation errors, so there shouldn't be any more probs, right?)

PR #1012, opened 2021-09-28
#458F Update front-end queries to use secure REST endpoint urls

andreievg:
~~ran out of time to test it, sorry. I'll try to test in the morning, see back end PR~~

Updates to front-end to work with back-end changes to endpoints: [PR #607](https://github.com/openmsupply/application-manager-server/pull/607)

---

**Carl**: Okay, have finished going through this and made a few fixes, as well as checking all the endpoint calls I could find. 

Andrei did most of the initial replacements, I've just fixed a couple of things that were broken and updated a couple of other little things:

- Fix "POST" request method so it works with "fileUpload"
- set JWT in localStorage immediately after user Log in (before loginOrg), so loginOrg has access to it
- set "persistJWT" localstorage key as constant in config.ts
- changed the maximum thumbnail display size in fileUpload and moved it to a config.json file (it's been bothering me for a while that it was too big in most instances)

**UPDATE**: ~~I think this is fine now, but there is a problem with evaluator expressions that use authenticated endpoints now, so this won't be mergeable until I finish https://github.com/openmsupply/application-manager-server/issues/379~~ Done: https://github.com/openmsupply/application-manager-server/pull/610


src/utils/helpers/attemptLogin.ts
```diff
@@ -41,17 +43,14 @@ export const attemptLogin = async ({
 export const attemptLoginOrg = async ({
   orgId,
   sessionId,
-  JWT,
   onLoginOrgSuccess,
   onLoginOrgFailure = () => {},
 }: loginOrgParameters) => {
   try {
-    const authHeader = { Authorization: 'Bearer ' + JWT }
```
CarlosNZ, line 49: Just doing testing and updates on this now. Problem with removing JWT from here and relying on localStorage in the "POST" method is that the initial Login user process didn't set JWT in local storage yet.

I could make it do so, but there could still be a timing issue if loginOrg runs before JWT is written to localStorage. I think it would be better if the GET/POST functions allowed a JWT/auth parameter like they did before and only check the localStorage if not provided, so we can use that mechanism for this initial login and login org.

CarlosNZ: It works if I write to localStorage straight after initial login. Hopefully it won't encounter a situation where the "onLoginOrg" method runs before it's properly in localStorage.

CarlosNZ:
There's currently a problem with the "postRequest" method that means that logging in and/or file upload isn't working properly, so don't test just yet. I'll delete this comment when it's good to go again.

nmadruga (changes requested):
Ok - continued testing here. And I think there is something wrong with thumbnails.
When I load snapshot `core_templates` and go check the organization FDA this is how it's showing on the list of organizations:
Screen Shot 2021-10-07 at 3 36 10 PM

And so I also tried to create an Organisation and after uploading the logo (on previous page) this is what is showing, so seems like a problem to view thumbnails:
Screen Shot 2021-10-07 at 3 37 57 PM

I'll keep testing else where to see if I see something else that needs improving

CarlosNZ:
I'll hold off on merging this until I can also update the core_templates (for outcomes and any other API queries)

CarlosNZ:
> Ok - for me is good to merge when you think it's ready

Cool, I'll try and update the core templates tonight and merge it. Can you approve (and back-end too)?

CarlosNZ:
> Cool, I'll try and update the core templates tonight and merge it. Can you approve (and back-end too)?

Actually, I'll update Outcomes when I merge the Row Restrictions PR, cos this one doesn't have the RowRestrictions field in it.

PR #1014, opened 2021-09-29
Core_templates snapshot v1

nmadruga:
### Linked PR:
- Back-end https://github.com/openmsupply/application-manager-server/pull/601

### Changes
- Updated docs to have a list of existing Core-actions


docs/internal/Home.md
```diff
@@ -4,14 +4,25 @@ Please see [ReadMe](__README.md) for information about maintaining this document
...
 - [Level 1 Review and LOQ](Level-1-Review-And-LOQ.md)
 - [Level 1+ Reviews and Changes Requested](Level-1+-Reviews-and-Changes-Requested.md)
 - [Assignment Process](Assignment-Process.md)
 - [Actions on Applications](Actions-on-Applications.md)
+  List-Core-Actions
+
+### Admin
+
+- [Template Builder](Template-Builder.md)
+- [List Core Actions](List-Core-Actions.md)
```
CarlosNZ, line 25:
```suggestion
- [List of Core Actions](List-Core-Actions.md)
```

docs/internal/admin/List-Core-Actions.md
```diff
@@ -0,0 +1,288 @@
+# List of Core Actions
+
+The snapshot loaded with the system at the start called `core_templates` has a list of all core-actions included in template `Sample template`.
```
CarlosNZ, line 3:
```suggestion
The snapshot loaded with the system at the start, called `core_templates`, has a template called `Sample template`, which contains all the core actions listed here.
```

docs/internal/admin/List-Core-Actions.md
```diff
@@ -0,0 +1,288 @@
...
+
+## Disclaimer
+
+The plugin actions that will be listed here are different from actions that users can act in the sytem, this are only meant for admins and configure how the system will work.
+Core-actions aren't meant to be modified, and only someone with expertise should be dealing with these. Otherwise some of the features that rely on certain actions would stop working.
+
+## Briefing
+
+Core-actions are listed in each template and are configurations that run after certain events happen triggered by changes done by users in Applications linked to the template.
+These actions should be added at the start of a template creation - one siimple way is to duplicate the `Sample template` to start with all of these core-actions in the new template configuration.
```
CarlosNZ, line 13:
```suggestion
These actions should be added at the start of a template creation - one simple way is to duplicate the `Sample template` to start with all of these core-actions in the new template configuration, although it is recommended you duplicate one of the other core templates that mostly closely matches your requirements, and it will also have all the core actions included (with some exceptions for a couple of the simpler templates)
```

docs/internal/admin/List-Core-Actions.md
```diff
@@ -0,0 +1,288 @@
...
+Some of the plugin actions - which aren't core-actions - are also defined per template (and some examples at the end) used mainly for configuring notifications, generatioin of certificates and outcomes.
```
CarlosNZ, line 14:
```suggestion
Some of the plugin actions - which aren't core-actions - are also defined per template (and some examples at the end) used mainly for configuring notifications, generation of certificates, and outcomes.
```

docs/internal/admin/List-Core-Actions.md
```diff
@@ -0,0 +1,288 @@
...
+- User Registration (No review)
+- Company Registration (1 stage)
+  - Review - permission: `reviewGeneral`
+- Join Company (1 stage)
+  - Review: - permissioin: `reviewGeneral`
+- Demo -- Features Showcase (3 stages)
+  - Screening - permission: `reviewScreening` self-assignable
+  - Assemessment: (2 levels)
+    - 1. Review permissions: `reviewGeneral` and assigner: `assignGeneral`
+    - 2. Consolidation permission: `reviewAssessmentLevel2` self-assignable
```
CarlosNZ, line 32: I thought we were going to call this "reviewSenior" or something like that?

docs/internal/admin/List-Core-Actions.md
```diff
@@ -0,0 +1,288 @@
...
+   - Description: "**CORE-ACTION** Change application outcome when last level and last stage - usually a final decision - submits a review as CONFORM or NON_CONFORM. Outcome will be APPROVED or REJECTED accordingly"
+   - Condition: When Reviewer is_last_stage & is_last_level & decision is either CONFORM or NON_CONFORM
+
+9. Change Application Outcome (on Rejection in early stage)
+
+### Other actions
+
+#### Trigger: ON_APPLICATION_SUBMIT
+
+- Asynchronous console log
```
CarlosNZ, line 223: Do we still need this? Probably not in core actions, right?

docs/internal/admin/List-Core-Actions.md
```diff
@@ -0,0 +1,288 @@
...
+8. **TODO** Add example of expiry
+9. **TODO** Add example of certificate
+
+##### Actions
+
+1. Change Application Outcome (on Rejection in early stage)
+
+   - Action: "changeOutcome"
+   - Description: "Example action - Change outcome to REJECT in case of NON_CONFORM decision in first stage of Screening."
+   - Condition: When Reviewer is_last_level & stage is Sreening & decision is NON_CONFORM
```
CarlosNZ, line 252: Maybe add a comment that the database automatically updates the Status to "COMPLETED" when Outcome is changed from "PENDING", so it's clear why we're not changing status here.

CarlosNZ (changes requested):
Great, thanks. Just a few clarifications, typos I think.

Couple of template content suggestions, not super important, but they'll actually need changing in back-end if you go with them.

docs/internal/admin/List-Core-Actions.md
nmadruga: This is just to keep the list of permissions short. But for now I would say let's merge like this and change later if desirable (remembering to update here as well)

docs/internal/admin/List-Core-Actions.md
nmadruga: No - see it's a different section:

## Other actions

docs/internal/admin/List-Core-Actions.md
nmadruga: Yes, will do. Thanks

CarlosNZ (approved):

PR #1019, opened 2021-09-30
#1015 Summary to allow applicant adding new responses

### LINKED ISSUE #1015: Can't edit field in summary section if field was originally left blank when filling out application
Emily noticed that if you don't enter anything in a field while filling out the application, you can't edit those fields in the summary section: 

image

nmadruga:
Fixes #1015

Quick fix to allow adding new responses from Application summary page in first submission

Now on first submission you can click to add a new responses -> clicking will take applicant back to page to incude response:
Screen Shot 2021-09-30 at 2 50 07 PM

When updating from changes-requested option to add doesn't show to keep existing logic:
Screen Shot 2021-09-30 at 2 51 26 PM

### DIFF
--- a/src/components/PageElements/PageElements.tsx
+++ b/src/components/PageElements/PageElements.tsx
@@ -135,10 +135,15 @@ const PageElements: React.FC<PageElementProps> = ({
             } = state
 
             const isResponseUpdated = !!isChangeRequest || !!isChanged
-            // Applicant can edit the summary page when is first submission and a response has been added
-            // Or when changes required for any question that have been updated (isUpdating true)
+
+            // Applicant can add new response to elements not responded in first submission or changes required
+            const canApplicantAddNew = canEdit && !isUpdating && !latestApplicationResponse?.value
+            // And can edit if not an empty response in first submission or
+            // during updated to changes required any question that have been updated (isUpdating true)
             const canApplicantEdit =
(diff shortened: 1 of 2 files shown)


src/components/PageElements/PageElements.tsx
```diff
@@ -268,6 +274,10 @@ export const UpdateIcon: React.FC<{ onClick: Function }> = ({ onClick }) => (
   <Icon className="clickable" name="edit" size="large" color="blue" onClick={onClick} />
 )
 
+export const AddIcon: React.FC<{ onClick: Function }> = ({ onClick }) => (
+  <Icon className="clickable" name="add" size="large" color="blue" onClick={onClick} />
```
CarlosNZ, line 278: Just wonder if the "plus square outline" icon might look more consistent with the existing "edit" icon?
Screen Shot 2021-10-01 at 7 45 54 PM
Screen Shot 2021-10-01 at 7 46 10 PM

CarlosNZ (approved):
Handy to have thanks. See what you think about the icon option.

src/components/PageElements/PageElements.tsx
nmadruga: Yes, sure. That does looks better.

nmadruga: Hmmm I'll leave as is since there is a slightly difference in position with the edit button that I won't investigate now. And it's not as visible with the current item

PR #1022, opened 2021-10-05
#1007 applicationData for Action evaluation previews

### LINKED ISSUE #1007: Template building -- preview Actions outputs
@andreievg -- Nicole and I both think this should be top priority, as it's currently far too easy to mess up configs by not ensuring the action queries are actually pointing to the things we expect them to.

We have the nice little "Preview query" (lightning bolt) function for Form Elements, but we really need it for Actions too. It's currently there, but doesnt' take into account form responses (i.e. applicationData). It's becoming increasingly clear that we need to be able to see the result of the Action while in the Template Builder. 

Hopefully it's not a big job to generate application data for the "dummy" application so we can test Action queries?

(This might require a call to back-end, since usually back-end generates applicationData for Actions. I can quite easily setup an endpoint for "generateApplicationData" if that will help?)

CarlosNZ:
Fix #1007 

Requires back-end https://github.com/openmsupply/application-manager-server/pull/608 in order to fetch server-side applicationData for Actions

This is pretty rudimentary, but it gives us the thing that was most needed for TemplateBuilder -- seeing the result of evaluator queries in Actions editor. In particular, we can check the values going to "modifyRecord" are what we expect in the editor rather than having to test the whole template.

### DIFF
--- a/src/containers/TemplateBuilder/shared/Parameters.tsx
+++ b/src/containers/TemplateBuilder/shared/Parameters.tsx
@@ -1,6 +1,7 @@
 import React, { useState } from 'react'
 import { Accordion, Button, Header, Icon } from 'semantic-ui-react'
 import { EvaluatorNode, FullStructure } from '../../../utils/types'
+import { useActionState } from '../template/Actions/Actions'
 import CheckboxIO from './CheckboxIO'
 import Evaluation from './Evaluation'
 import JsonIO from './JsonIO'
@@ -21,6 +22,7 @@ export const Parameters: React.FC<ParametersProps> = ({
   currentElementCode,
   fullStructure,
 }) => {
+  const { applicationData } = useActionState()
   const [asGui, setAsGui] = useState(true)
   const [isActive, setIsActive] = useState(false)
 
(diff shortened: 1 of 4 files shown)


src/containers/TemplateBuilder/shared/Parameters.tsx
```diff
@@ -1,6 +1,7 @@
 import React, { useState } from 'react'
 import { Accordion, Button, Header, Icon } from 'semantic-ui-react'
 import { EvaluatorNode, FullStructure } from '../../../utils/types'
+import { useActionState } from '../template/Actions/Actions'
```
nmadruga, line 4: Not in use?

src/containers/TemplateBuilder/template/Actions/ActionConfig.tsx
```diff
@@ -9,6 +9,7 @@ import Evaluation from '../../shared/Evaluation'
 import { useOperationState } from '../../shared/OperationContext'
 import { Parameters, ParametersType } from '../../shared/Parameters'
 import TextIO from '../../shared/TextIO'
+import { useFullApplicationState } from '../ApplicationWrapper'
```
nmadruga, line 12: Not in use?

nmadruga (approved):
LGTM

PR #1025, opened 2021-10-06
#611B Allow any template to be made available without disabling another

CarlosNZ:
Front-end component for [back-end PR #612](https://github.com/openmsupply/application-manager-server/pull/612)

All it does is make the "Make Available" button enabled all the time in the Template Builder, as the job of disabling the other versions is now handled by the database. So you can set a template version to AVAILABLE without having to go into the currently available one and disabling it first.


nmadruga (approved):
Very nice improvement

PR #1030, opened 2021-11-11
Quick fix: File upload summary view

CarlosNZ:
I missed one when updating all the urls to use new API paths. Summary view for file Upload wasn't displaying the file images correctly.

This fixes that.


nmadruga (approved):
Thanks for finding this one. 

I guess we will release another beta B-1.0.9 for Monday's upgrade and avoid using B-1.0.8. Since there is a couple of things missing... I'm still worried on what happened with snpashots uploading

PR #1051, opened 2021-11-23
#884 Evaluate submission message after submission

### LINKED ISSUE #884: Evaluate Submission message AFTER submission
Currently the "startMessage" and "submissionMessage" are both evaluated at the same time when loading the template data.

However, if we want to show values from the application responses as part of the submission message (e.g. "Thank you for your application to register product **PRODUCT_NAME**") we'll need to evaluate the submission message separately when loading the Submission page.

Should be a pretty quick fix.

CarlosNZ:
Fix #884 

Had to make "ApplicationSubmission" page fetch the full structure so it can use responses from application in submission message.

Also done here:
-- in ApplicationViewWrapper, fetch JWT from local storage on demand, as it's not always there on load when coming from non-logged in state (which causes problems with some subsequent API calls)

Suggest testing this with new "add internal user" template so you can see the effect on the Submission page ("CreateIntUser-1.zip" in `/demo_angola/templates` in templates repo)

### DIFF
--- a/src/containers/Application/ApplicationPage.tsx
+++ b/src/containers/Application/ApplicationPage.tsx
@@ -66,8 +66,6 @@ const ApplicationPage: React.FC<ApplicationProps> = ({
     },
   } = fullStructure
 
-  console.log(fullStructure)
-
   return (
     <Container id="application-form">
       <Grid stackable>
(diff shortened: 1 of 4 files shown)


nmadruga (approved):
Looks good. Thanks Carl

I was testing the new template and just noticed that after user was created they were not linked to the **system** organisation, just mentioning in case you missed that one.

CarlosNZ:
> I was testing the new template and just noticed that after user was created they were not linked to the system organisation, just mentioning in case you missed that one.

Ah yeah, good catch, thanks. Will fix.

PR #1053, opened 2021-11-23
#1052 Checkbox updates

### LINKED ISSUE #1052: Checkbox updates
Required to make more flexible display of received dynamic data.

As discussed with @nmadruga, agreed solution is:
- new parameter "keyMap", which allows mapping of input objects to the property names expected by the "checkbox objects" expected by the plugin
- add "Reset" button (optional, specified by parameter, default false) to reset selection state to initial values

CarlosNZ:
Fix #1052 

~~Just quickly putting this in Draft PR so @nmadruga can begin using it, still need to thoroughly test it and tweak it.~~

One other little addition -- if the API returns more fields that the required ones, then these will also be passed along and saved with the response. For example, when fetching permission names, values like "id" will be captured and stored with the response even though they're not required for display.

Extra: fixed one more JWT error in `Evaluation.tsx` -- same issue as yesterday's fix, this is just one more I found.


nmadruga (approved):
Great. That looks all good to me now.

The other related problem that we discussed about allowing to display the initialState after something on the dynamic query changes (i.e. selection of user to be used to populate checkboxes) can be done in a separated PR.

PR #1055, opened 2021-11-24
#1054 Quick fix: Admin loading problem fix

### LINKED ISSUE #1054: Bug: Admin pages show 404 page when loading

CarlosNZ:
Fix #1054 

Quick fix for this, since it's been annoying me for a few weeks now. The 404 component shouldn't show when userState is still loading, so this just checks for that before the "isAdmin" check.


nmadruga (approved):
Looks good

PR #1059, opened 2021-11-25
#1058 Handle changing checkbox input parameters correctly

### LINKED ISSUE #1058: Checkbox -- handle dynamic changes correctly

CarlosNZ:
Fix #1058 

This was a bit trickier than I originally thought it would be, as I had to figure out a way for the plugin to know whether the change was a "proper" change, or just the initial loading values. In the former case, the state should be replaced by the new values, but in the latter case (initial load), the checkbox state should be loaded from the saved response. And it's complicated by the fact that we also want to ignore the "Loading" values and not consider them a proper "change".

Anyway, the good thing is I think I can probably apply this technique to other plugins, which might make what I was trying to achieve with proper "default" values somewhat easier. We'll see.

To test (using Nicole's grant permissions template):
- on initial load, all permissions should show unchecked
- select a user, the checkboxes should be populated with the correct permission selections
- delete user from previous question, they should revert to all unchecked
- re-load the user, then select some additional checkboxes
- refresh the page, the extra selections should remain selected

I have one question: currently the "Reset selections" button reverts it to the state it was in *when the page was loaded*. Do you think this is correct, or should it revert to the default selections (current permissions) for the selected user?

### DIFF
--- a/src/formElementPlugins/checkbox/src/ApplicationView.tsx
+++ b/src/formElementPlugins/checkbox/src/ApplicationView.tsx
@@ -34,16 +34,13 @@ const ApplicationView: React.FC<ApplicationViewProps> = ({
   const [checkboxElements, setCheckboxElements] = useState<Checkbox[]>(
     getInitialState(initialValue, checkboxes, keyMap, isFirstRender)
   )
-  const [initialState, setInitialState] = useState<Checkbox[]>()
 
   // When checkbox array changes after initial load (e.g. when its being dynamically loaded from an API)
   useEffect(() => {
     if (checkboxes[0] !== config.parameterLoadingValues.label && isFirstRender) {
       setIsFirstRender(false)
     }
-    const initState = getInitialState(initialValue, checkboxes, keyMap, isFirstRender)
-    setCheckboxElements(initState)
-    setInitialState(initState)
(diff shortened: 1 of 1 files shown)


nmadruga:
> I have one question: currently the "Reset selections" button reverts it to the state it was in _when the page was loaded_. Do you think this is correct, or should it revert to the default selections (current permissions) for the selected user?

Yea, I think it should "revert" to the initial state, in other words, current permissions for the selected user.

nmadruga (approved):
It's working perfectly. Thanks, Carlos 🙈

CarlosNZ:
> > I have one question: currently the "Reset selections" button reverts it to the state it was in _when the page was loaded_. Do you think this is correct, or should it revert to the default selections (current permissions) for the selected user?
> 
> Yea, I think it should "revert" to the initial state, in other words, current permissions for the selected user.

Great, done: ed5f2865

That actually makes the logic and code a bit simpler, so 👍

You want to quickly check it before I merge?

nmadruga:
> That actually makes the logic and code a bit simpler, so 👍
> 
> You want to quickly check it before I merge?

Sure. Will do now.

nmadruga:
Looks great. Let's merge : )

PR #1060, opened 2021-11-29
Quick fix for JWT problem

CarlosNZ:
This should fix the problem Jess was having with the Change Password. Weird that it wasn't happening in Firefox, but just to be safe I've moved all the JWT fetches from local storage that I could find into function, so it's refetched every time rather than loaded once for each module.

### DIFF
--- a/src/formElementPlugins/SummaryViewWrapper.tsx
+++ b/src/formElementPlugins/SummaryViewWrapper.tsx
@@ -11,7 +11,6 @@ import globalConfig from '../config'
 import { TemplateElementCategory } from '../utils/generated/graphql'
 
 const graphQLEndpoint = globalConfig.serverGraphQL
-const JWT = localStorage.getItem(globalConfig.localStorageJWTKey)
 
 const SummaryViewWrapper: React.FC<SummaryViewWrapperProps> = ({
   element,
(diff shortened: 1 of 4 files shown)


nmadruga (approved):
LGTM. Add internal user's looks fine to check for existing usernames!

Although I'm never being able to get to the ResetPassword application when I click the [here]() link:
Screen Shot 2021-11-30 at 4 33 37 PM
It goes to the login page instead:
Screen Shot 2021-11-30 at 4 32 18 PM

CarlosNZ:
> Although I'm never being able to get to the ResetPassword application when I click the here link:

I'll check this again, but I think it's because you're already logged in as Admin. But for a normal user, they won't have logged in at all yet, so it should work.

CarlosNZ:
> I'll check this again, but I think it's because you're already logged in as Admin. But for a normal user, they won't have logged in at all yet, so it should work.

Yup, as I thought. If you explicitly log out after creating the user (as Admin), then when you click the link (as the new user), you'll be a logged-out user, so the link to change password will work.

nmadruga:
> Yup, as I thought. If you explicitly log out after creating the user (as Admin), then when you click the link (as the new user), you'll be a logged-out user, so the link to change password will work.

Nice

PR #1066, opened 2021-12-06
#1064 Template builder layout/style improvements

### LINKED ISSUE #1064: Template builder layout improvements
As discussed [here](https://github.com/openmsupply/application-manager-web-app/issues/1063)

- Improve wrapper containers (reuse existing css containers)
- blocks should be vertically flush at the top (not centred vertically)
- Modals:
  - Text shouldn’t be centered, it should be left justified in a centered block like the main application content
- Misc tidy-ups...

CarlosNZ:
Fix #1064 

~~(Work in progress -- haven't addressed Modals yet)~~

Not totally happy with this, but made it a bit more consistent with the rest of the app, and tidied up quite a lot of alignment / spacing stuff.  Colour palette could really do with some more design work.

I would have liked to turn the "Input" boxes in to just standard Semantic form inputs, but they're too tightly integrated with the functionality in TextIO etc, so have just tweaked them a touch.

### DIFF
--- a/semantic/src/site/globals/site.overrides
+++ b/semantic/src/site/globals/site.overrides
@@ -1876,8 +1876,8 @@ a:hover {
   padding: 5px;
   margin: 2px;
   border-radius: @borderRadius;
-  border: 2px solid rgba(0, 0, 0, 0.2);
-  background-color: rgba(50, 252, 50, 0.01);
+  border: 1px solid @surfacesMed;
+  // background-color: rgba(50, 252, 50, 0.01);
 }
 
 .config-container {
--- a/semantic/src/site/globals/site.overrides
+++ b/semantic/src/site/globals/site.overrides
@@ -192,6 +192,14 @@ a:hover {
   margin-right: 10px !important;
 }
 
+.flex-gap-10 {
+  gap: 10px;
+}
+
+.flex-gap-20 {
+  gap: 20px;
+}
+
 /*--------------------------------
    Login page
 ---------------------------------*/
--- a/semantic/src/site/globals/site.overrides
+++ b/semantic/src/site/globals/site.overrides
@@ -121,6 +121,12 @@ a:hover {
   .flex-row-start-center;
   flex-wrap: wrap;
 }
+
+.flex-row-end {
+  .flex-row();
+  width: 100%;
+  justify-content: flex-end;
+}
 .flex-column {
   display: flex;
   flex-direction: column;
@@ -1638,7 +1644,7 @@ a:hover {
   .key {
     border-end-end-radius: 0px;
     border-start-end-radius: 0px;
-    background-color: @indicatorColor;
+    background-color: @surfacesLight;
     a {
       text-decoration: underline;
     }
@@ -1821,12 +1827,11 @@ a:hover {
   border-radius: @borderRadius;
   background-color: @surfacesWhite;
   margin: 5px;
-  background-color: @surfacesLight;
 }
 
 .config-container-alternate {
   .config-container();
-  background-color: rgba(0, 0, 0, 0.03);
+  background-color: @surfacesLight;
 }
 
 /*******************************
--- a/semantic/src/site/globals/site.overrides
+++ b/semantic/src/site/globals/site.overrides
@@ -174,6 +174,14 @@ a:hover {
   margin: 0px !important;
 }
 
+.full-width-container {
+  width: 100%;
+}
+
+.text-block-width {
+  max-width: 800px;
+}
+
 /*--------------------------------
    Login page
 ---------------------------------*/
@@ -1813,6 +1821,7 @@ a:hover {
   border-radius: @borderRadius;
   background-color: @surfacesWhite;
   margin: 5px;
+  background-color: @surfacesLight;
 }
 
 .config-container-alternate {
--- a/semantic/src/site/globals/site.overrides
+++ b/semantic/src/site/globals/site.overrides
@@ -1603,11 +1603,11 @@ a:hover {
 .io-wrapper {
   .flex-row();
   align-items: stretch;
-  margin: 2px;
+  margin: 2px 2px 8px;
   .io-component {
     flex-grow: 1;
     margin: 0px;
-    text-align: center;
+    // text-align: center;
     white-space: nowrap;
     border: 2px solid @indicatorColor;
     border-radius: @borderRadius;
@@ -1661,7 +1661,7 @@ a:hover {
 .longer {
   input,
   textarea {
-    min-width: 600px !important;
+    min-width: 500px !important;
   }
 }
 .long {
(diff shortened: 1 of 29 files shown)


semantic/src/site/globals/site.overrides
```diff
@@ -77,6 +77,11 @@ a:hover {
   justify-content: space-between;
 }
 
+.flex-row-space_evenly {
```
nmadruga, line 80: I think for consistency it would be better called `flex-row-space-evenly` and elsewhere too

semantic/src/site/globals/site.overrides
```diff
@@ -1603,18 +1643,29 @@ a:hover {
...
+  margin: 2px 2px 8px;
+  min-height: 25px;
   .io-component {
     flex-grow: 1;
     margin: 0px;
-    text-align: center;
-    white-space: nowrap;
-    border: 2px solid @indicatorColor;
+    // text-align: center;
+    // white-space: nowrap;
```
nmadruga, line 1652: Can be removed

semantic/src/site/globals/site.overrides
```diff
@@ -1624,18 +1675,25 @@ a:hover {
     padding: 0px !important;
     font-size: 12px !important;
     width: none !important;
-    min-width: 200px;
+    // min-width: 200px;
```
nmadruga, line 1678: Remove please 😇

src/containers/TemplateBuilder/template/General/General.tsx
```diff
@@ -34,19 +37,18 @@ const General: React.FC = () => {
...
   return (
     <div className="flex-column-center-start">
-      <div className="flex-row">
-        <div className="spacer-10" />
+      <div className="flex-row flex-gap-10">
         <ButtonWithFallback
-          title="Make Available"
-          disabledMessage="At least one template with the same code is already available, or this template already available"
+          title={strings.TEMPLATE_GEN_BUTTON_AVAILABLE}
+          disabledMessage={strings.TEMPLATE_GEN_BUTTON_AVAILABLE_DISABLED}
```
nmadruga, line 43: Nice. Thanks!

nmadruga (commented):
Ok - I didn't look in much detail on the changes but had a quick look overall looks good to me. I can see some styles are entered directly on the component, which I think it's fine. Mostly margins... And you added many reusable less styles which is great.

What I have checked, which might be useful (or not, maybe just leave that if you think it's too much work) is that trying on a responsive mode - for example `iPad` the form is still not wrapping in some places. Also I notices that the TopArea is showing on top of the form... I don't think is related to latest changes though.

Check out my prints when checking using `iPad` default sizes (and comments) below:

#### General tab looks good! 
**Note 1** The top area is overridding the form header
**Note 2** Maybe the header would be better in 2 rows there, on for the `back` button (Templates/Procedures) and another for the version, name... 
Screen Shot 2021-12-07 at 10 11 15 AM

#### Form is looking almost good
**Note 1** Missing better wrapping around the Section header
**Note 2** I personally would prefer other style on sections and pages displayed, It could be on the same liine as the + button and maybe using another style. What do you think?
Screen Shot 2021-12-07 at 10 10 27 AM

#### The Element config is looking perfect before opening parameters
Screen Shot 2021-12-07 at 10 10 49 AM

#### After opening parameters it doesn't wrap well
Screen Shot 2021-12-07 at 10 10 43 AM

#### Permissions tab - almost perfect 
**Note** Just one missing style to also wrap the name of the stage...
Screen Shot 2021-12-07 at 10 10 17 AM

#### 1
Screen Shot 2021-12-07 at 10 10 06 AM

#### 2
Screen Shot 2021-12-07 at 10 10 01 AM

#### 3
Screen Shot 2021-12-07 at 10 09 28 AM

#### 4
Screen Shot 2021-12-07 at 10 06 26 AM

#### 5
Screen Shot 2021-12-06 at 9 40 14 AM

#### 6
Screen Shot 2021-12-06 at 9 39 16 AM
Screen Shot 2021-12-03 at 10 28 46 AM

CarlosNZ:
> What I have checked, which might be useful (or not, maybe just leave that if you think it's too much work) is that trying on a responsive mode - for example `iPad` the form is still not wrapping in some places

Yeah, the Template Builder has never worked properly on mobile devices -- we've kind of decided that Template config should be an "at your computer" activity. But there's lots of other stuff that doesn't actually work on mobile, too -- will need to revisit that at some point.

Having said that... most of the things you pointed out should be quite easily fixable, at least in terms of simple wrapping, so I'll do what I can now.

nmadruga:
Yes, I understand exactly that we never intended it to work on mobile, although if is that close to, why not right?

> Having said that... most of the things you pointed out should be quite easily fixable, at least in terms of simple wrapping, so I'll do what I can now.

Cool. Sorry I submitted before coming on last screenshots. I'll add my comments there now and move on to reply to reviews on other PRs

nmadruga:
Updated comments on my review now : )

semantic/src/site/globals/site.overrides
CarlosNZ: Yup, done.

CarlosNZ:
> **Note 1** The top area is overridding the form header

Yeah, this is an existing issue with the UserArea component

> **Note 2** Maybe the header would be better in 2 rows there, on for the `back` button (Templates/Procedures) and another for the version, name...

Yeah, good idea -- done, with better wrapping too: f2c6a84a

> Missing better wrapping around the Section header

Done

> I personally would prefer other style on sections and pages displayed, It could be on the same liine as the + button and maybe using another style. What do you think?
> Control for page seems mix with elements controls. Maybe a border around or show on a pop-up (your choice if can be done now or later)

Okay, I've made some improvements to this page. I'm not particularly happy with it, but I'm not sure what else to do without re-creating the actual content in a nicer way. 2cce22d5

> After opening parameters it doesn't wrap well

Yeah, this is just one of those areas that can't really look good on a small screen. The parameter editing gets really complex with all the nested evaluator elements, so hard to account for every case. (Again, not without a significant re-design)

>  Just one missing style to also wrap the name of the stage...

Seems okay to me:
Screen Shot 2021-12-07 at 1 08 50 PM
I dunno, maybe something else I changed fixed it 🤷‍♂️

> The isSequential field isn't showing aligned among all children in the list

The sequential button and the close button are their own "block", separate from the rest (with space in between). But as the screen gets smaller that becomes less apparent and they look like one thing, which is a bit weird, I grant  you. However, I've tidied up the wrapping and spacing for these a bit now. 6e44c148

> Personally I don't like how the controls are displaying on the left (and the remove ont he right) - I think they could be combined in a nice vertical area something similar to what was done in PR: #1001

I haven't seen a layout for this that looks good. If you bunch them all together it just looks like a clutter (and the 'isSequential' button is rarely used, so it makes sense to keep it far away from the rest. And the "close" icon is traditionally in the top-right corner)

CarlosNZ:

Issue #1063, opened 2021-12-03
Misc Snapshot/Template Builder bugs 2


nmadruga:
There was already an issue to fix some template builder bugs with same title [Misc Snapshot/Template Builder bugs](https://github.com/openmsupply/application-manager-server/issues/506) and this is just a list of remaining or other existing bugs to be used for priority and fix before presenting the Template builder on next week.

Important to flag: We would like to only be fixing very obvious or easy fix bugs - leave the more complicated/refactoring for now...

## Selected to be done now:
- [x] Few improvements of layout  #1064 
 - blocks should be vertically flush at the top (not centred vertically)
 - Improve wrapper containers (reuse existing css containers)
- [x] #1045
  - Bug: click to add element and app breaks - isFirstElement check (it's ok after reloading the page)
- [x] ApplicationData in Query evaluation should be same as it would be in actual workflow (Form and Actions) #1068
- [x] Required parameters / Optional parameters not accurate (where's it getting them from?) #1072
- [x] Pressing “SAVE” shouldn’t close modal. Should be a separate “Close” button/icon (And shouldn’t close from clicking outside the modal) PR https://github.com/openmsupply/application-manager-web-app/pull/1065
  - Modal: Text shouldn’t be centered, it should be left-justified in a centered block like the main application content
- [x] Should load with S1, Page 1 by default PR https://github.com/openmsupply/application-manager-web-app/pull/1065
- [x] Stop sorting parameters after renaming... do only on page load to avoid missing the element wanted to be configuring. PR https://github.com/openmsupply/application-manager-web-app/pull/1070

## Bugs and issues
- [ ] [Create issue] Bug showing error to load plugin after changing/importing needs to change tabs to see it
- [ ] When selecting an element type show link to the relevant documentation page - Add to manifest page
  - Connect the config.json files for each plugin to be passed in the manifest
- [ ] #934
  - Maybe that should be for any array element...
  - At least the ability to add on the top instead of the bottom of the array

## Needs more discussion - Future work!
- [ ] #864
  - We probably need a simpler way to send notifications to the reviewer...
- [x] #908
  - https://github.com/openmsupply/application-manager-server/issues/488
  - Fix hack with addition of empty string to use objectProperty for using in API (Get an example) 
- [ ] #1043
- [ ] [Create issue] Front-end plugin config.json files should have the requiredParameters and optionalParameters objects (with label and type - idea of enum with selections i.e. Table, view, card) (Also covered by #1072)

[...]

PR #1067, opened 2021-12-06
#1063B Stop sorting while adding parameters

nmadruga:
Partial fix for #1063 

Fixing: 

## Changes
- only sorts when openning modal
- ~~Fix from using array to object when storing `parameters` object~~ **DONE**


CarlosNZ (changes requested):
The parameters are now showing up in the UI as index numbers rather than names:
Screen Shot 2021-12-07 at 9 03 17 AM

nmadruga:
> The parameters are now showing up in the UI as index numbers rather than names:
> Screen Shot 2021-12-07 at 9 03 17 AM

Yes... that's the missing TODO I flagged there. Hope to be finishing this before lunch

nmadruga:
@CarlosNZ I was having trouble to merge this one with previous, so created a new PR based from `develop`: https://github.com/openmsupply/application-manager-web-app/pull/1070

PR #1069, opened 2021-12-06
#1068 Application Data fix

### LINKED ISSUE #1068: ApplicationData in Query evaluation should be same as it would be in actual workflow
From [here](https://github.com/openmsupply/application-manager-web-app/issues/1063)

CarlosNZ:
Fix #1068 

- Make sure the data shown "Object Properties" panel matches what will actually be used when the template is used
- Show "Object Properties" panel for Actions as well
- Limit width of "ObjectProperties" panel to 500px
- Disabled the "Info" panel displayed from the "i" icon in Permission Names, cos it's an absolute mess and unusable without a lot of work. (Have just commented out the icon so it can't be launched). Will have to revisit this with a better UI at some point

### DIFF
--- a/semantic/src/site/globals/site.overrides
+++ b/semantic/src/site/globals/site.overrides
@@ -1876,6 +1876,7 @@ a:hover {
 .object-properties-container {
   .flex-column-start-center();
   margin-left: 30px;
+  max-width: 500px;
 }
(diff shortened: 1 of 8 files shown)


CarlosNZ:

PR #1070, opened 2021-12-07
#1063 b stop sorting while adding parameters

nmadruga:
Partial fix for #1063 

Fixing: 

## Changes
- only sorts when openning modal
- ~~Fix from using array to object when storing `parameters` object~~ **DONE**

### DIFF
--- a/src/containers/TemplateBuilder/shared/Parameters.tsx
+++ b/src/containers/TemplateBuilder/shared/Parameters.tsx
@@ -30,8 +30,8 @@ export const Parameters: React.FC<ParametersProps> = ({
     // Sort elements only when opening modal
     const sortedParameters = Object.entries(parameters)
       .sort(([key1], [key2]) => (key1 > key2 ? -1 : key1 === key2 ? 0 : 1))
-      .map(([key, element]) => ({ [key]: element }))
-    setParameters(Object.assign(sortedParameters))
+      .reduce((sortedParameters, [key, element]) => ({ ...sortedParameters, [key]: element }), {})
+    setParameters(sortedParameters)
   }, [])
 
   return (
@@ -41,10 +41,9 @@ export const Parameters: React.FC<ParametersProps> = ({
         active={isActive}
         onClick={() => setIsActive(!isActive)}
       >
-        <Header
-          as="h4"
-          className="no-margin-no-padding"
-        >{`Plugin Specific Parameters (${parameters.length})`}</Header>
+        <Header as="h4" className="no-margin-no-padding">{`Plugin Specific Parameters (${
+          Object.keys(parameters).length
+        })`}</Header>
         <Icon size="large" name={isActive ? 'angle up' : 'angle down'} />
       </Accordion.Title>
       {isActive && (
--- a/src/containers/TemplateBuilder/shared/Parameters.tsx
+++ b/src/containers/TemplateBuilder/shared/Parameters.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from 'react'
+import React, { useEffect, useState } from 'react'
 import { Accordion, Button, Header, Icon } from 'semantic-ui-react'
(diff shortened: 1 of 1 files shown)


src/containers/TemplateBuilder/shared/Parameters.tsx
```diff
@@ -26,9 +26,14 @@ export const Parameters: React.FC<ParametersProps> = ({
...
-  const sortedParameters = Object.entries(parameters).sort(([key1], [key2]) =>
-    key1 > key2 ? -1 : key1 === key2 ? 0 : 1
-  )
+  useEffect(() => {
+    // Sort elements only when opening modal
+    const sortedParameters = Object.entries(parameters)
+      .sort(([key1], [key2]) => (key1 > key2 ? -1 : key1 === key2 ? 0 : 1))
+      .reduce((sortedParameters, [key, element]) => ({ ...sortedParameters, [key]: element }), {})
+    setParameters(sortedParameters)
+  }, [])
```
CarlosNZ, line 35: Just wondering why we need to sort them at all? What's wrong with using the "natural" order?

CarlosNZ (commented):
Also -- noticed the "Close" button still shows warning when nothing changed. Probably just needs your other branch merged into here, right?

src/containers/TemplateBuilder/shared/Parameters.tsx
CarlosNZ: Yeah, cos it re-sorts when you re-open the modal, so the parameter you last added is no longer on the bottom of the list, which is kind of annoying.

nmadruga:
> Also -- noticed the "Close" button still shows warning when nothing changed. Probably just needs your other branch merged into here, right?

Yes, that should be sorted on the other one. Because this is no longer based on that (it was just trouble to keep updated)

src/containers/TemplateBuilder/shared/Parameters.tsx
nmadruga: I think that can be sorted later... For now seems ok to me to sort the parameters... It helps with finding things when you have a longer list of elements

CarlosNZ: When you're making a new element, you're going to add parameters in a particular order that makes sense to you, probably starting with the simple ones, then adding the more complicated or niche parameters. Why would you want that order to change? When you have a list of 10 or so parameters, you want them to appear _in the order you entered them_, not get re-alphabetized when you re-open the modal. That's why we did this issue in the first place, we even agreed that's how it would work. This feels like we haven't actually solved the problem, just made it happen less often.

CarlosNZ: Just an example from my own usage -- when I'm using the Modify Record action, there's usually loads of parameters because of all the fields that get updated. I find it really tricky to make sense of what's what. So what I'd normally do is add "tableName" first, then "MatchField", "matchValue", and then (finally) all the individual fields that need to be written to. It would be super annoying if, next time I open that Action, the parameters have all been sorted alphabetically instead of the logical, helpful order I added them. Why would we want to do this?

CarlosNZ:

src/containers/TemplateBuilder/shared/Parameters.tsx
CarlosNZ: I just spent a few minutes playing round with this and what I hoped doesn't actually work either. We can't really have *any* control over the key order, as the database has its own way of handling it when it comes to JSONB fields.

It's basically an arbitrary order, but I'd argue that that's no better or worse than alphabetical -- they're both pretty useless.

If we *did* want to change the ordering, then we can just to a simple sort on the "parameters" prop as it's passed into the "Parameters" component, but doesn't seem necessary right now.

What would be good is if it would show them in the order they're listed in the "plugin.json" config file -- so we should remember to revisit this when we do the issue to pre-fill parameters from the required list.

PR #1076, opened 2021-12-15
Max width on dashboard container

CarlosNZ:
Oops -- something we neglected previously (I thought we'd already done it). The Dashboard needs to have a max-width so it doesn't show more than three template columns across, otherwise it looks really bad:
Screen Shot 2021-12-15 at 1 19 27 PM

Screen Shot 2021-12-15 at 1 20 16 PM

Noticed when adding admin permissions.


nmadruga (approved):
Yes, It's nice. Thanks

PR #1077, opened 2021-12-15
Template builder -- show page 1 by default

CarlosNZ:
We missed this the other day -- when the "Form" tab loads, it opens with Section 1, Page 1, however when selecting a new Section, there was no page elements shown again.

Fix for that -- now when selecting new section the first page of that section is loaded.


nmadruga (approved):
Nice. Thanks Carl

PR #1086, opened 2022-01-20
#1085 Review Page tab structure

### LINKED ISSUE #1085: New Review home page: Overall tab structure
Add tab layout to `/review` page

Put current review page (assignments) into "Assignments" tab, leave placeholders for other tabs, placeholder for Progress bar.

Make sure routing works with url queries to each tab

CarlosNZ:
Fix #1085
Screen Shot 2022-01-21 at 12 43 05 PM

- What was previously the "main" review home page has been moved into Assignments tab
- All other tabs just have placeholder components (and Progress Bar)
- Some slight restructuring of the hook order: the whole wrapper (ReviewWrapper, outside the tabs and ReviewPage) gets the full structure and passes it into each tab component. The "useGetReviewInfo" has been moved into the Assignments tab itself. Previously it was the other way round, but I figured all tabs would probably use the full structure, whereas assignment info is probably only useful for the Assignment tab.
_One thing to consider_: there was previously a slight difference in the way the ReviewPageWrapper (wrapper for actual *review*) and the Assignment container called "useGetApplicationStructure" -- The Review page used `shouldGetDraftResponses: false` whereas the Assignments used the default (`true`). Now they're both using `false` as they share this now. I don't think this should be a problem, but @nmadruga you might know better?
- moved tab components into subfolder for better organisation


src/containers/Review/assignment/Assignment.tsx
```diff
@@ -1,7 +1,7 @@
 import React, { useState } from 'react'
 import { Segment, Header } from 'semantic-ui-react'
-import { AssignmentDetails, FullStructure } from '../../utils/types'
-import AssignmentSectionRow from '../Assign/AssignmentSectionRow'
+import { AssignmentDetails, FullStructure } from '../../../utils/types'
+import AssignmentSectionRow from '../../Assign/AssignmentSectionRow'
 import ReviewSectionRow from './ReviewSectionRow'
 
 interface ReviewHomeProps {
```
nmadruga, line 7: Could change this name also, please?

Not sure, because we already have other components called `Assignment` Should this be called AssignmentLevel or something - when we add the other filter?

src/containers/Review/index.ts
```diff
@@ -1,5 +1,18 @@
-// import ReviewOverview from './ReviewOverview'
-// import ReviewPageWrapper from './ReviewPageWrapper'
```
nmadruga, line 2: Better to remove, please

nmadruga (approved):
Very nice. 🎉 🎉 

Sorry it took me so long to review 😝

CarlosNZ:
Thanks Nicole. I've just noticed an additional bug that I've introduced -- the Assignment page for non-self-assigned users isn't appearing correctly, so I'll need to sort that out now.

src/containers/Review/assignment/Assignment.tsx
CarlosNZ: Would it be okay if I move the whole "Assign" folder (in containers) into the "assignment" tab folder?

src/containers/Review/index.ts
CarlosNZ: This hasn't been like this since about 2 commits back -- have you got the latest? 

Should be:
```
import ReviewWrapper from './ReviewWrapper'
// Tabs:
import OverviewTab from './overview/OverviewTab'
import AssignmentTab from './assignment/AssignmentTab'
import SummaryTab from './summary/SummaryTab'
import NotesTab from './notes/NotesTab'
import DocumentsTab from './documents/DocumentsTab'
import ReviewProgress from './ProgressBar'

export {
  ReviewWrapper,
  OverviewTab,
  AssignmentTab,
  SummaryTab,
  NotesTab,
  DocumentsTab,
  ReviewProgress,
}
```

Weird that it's showing up in github.

CarlosNZ:
> I've just noticed an additional bug that I've introduced -- the Assignment page for non-self-assigned users isn't appearing correctly, so I'll need to sort that out now.

Okay, I fixed that -- was accidentally only passing the partial structure, not "fullStructure" into the tab components. Sorted now: 17585622

src/containers/Review/assignment/Assignment.tsx
nmadruga: Yes. All good

nmadruga:
> Okay, I fixed that -- was accidentally only passing the partial structure, not "fullStructure" into the tab components. Sorted now: [1758562](https://github.com/openmsupply/application-manager-web-app/commit/17585622088a9c0c487ec995c3eb5e8236a343eb)

Cool. Thanks for noticing that. I'll have another go and say if looks good to me

nmadruga:
Looking great

src/containers/Review/assignment/Assignment.tsx
CarlosNZ: Cool -- 65cca0b2

PR #1088, opened 2022-01-24
Review decision -- include comment in submit

CarlosNZ:
Very small change, something that was causing a small problem developing the "activity_log" for back-end. The "comment" field for "review_decision" was being written in a separate mutation, not in the review submission mutation. This just includes the comment in the main mutation as well, just in case the comment mutation hasn't finished before the submission one (which is what was causing the issue)


nmadruga (approved):
All good - as discussed via telegram

Issue #850, opened 2021-07-22
Bug: Show correct stage label on a previous stage review with - current one is showing


nmadruga:

CarlosNZ:
Doing this as part of #1084 now.

CarlosNZ:
> Doing this as part of #1084 now.

Okay, this is proving a bit trickier than anticipated, will need to leave it for its own issue I think.

**Intended solution:**

Run "refetchQueries" on 'getApplication' as part of the "submitReview" mutation, which will force the structure (and therefore stage info) to reload

**The problem:**

Problem is the trigger on the application table is never changed, so the data fetched as part of refetchQueries still contains the previous stage value. The "submit review" mutation sets the trigger on the REVIEW table, but not the APPLICATION table, so the refetch query won't wait till the trigger is reset (since it never gets set in the first place)

I then thought that it should be possible to ALSO set the trigger on the application table as part of the review submission. However, that's not possible either, as the reviewers don't have permission to write to the application table, so that doesn't work either. (It works when permissions is disabled)

Open to suggestions?

jagoje:
@CarlosNZ Hmmm complex one - thinking about this now. I have some thoughts, will be good to chat on our meeting on Friday.

CarlosNZ:
Apologies, I was commenting on the wrong issue. Have deleted my comments and started a new issue.

PR #1092, opened 2022-01-26
#1084 Progress bar

### LINKED ISSUE #1084: New Review home page: Progress bar
Separate work into building the query of Graphql that gives all stages of an application (regardless of user's permission to review/assign) and defining the status of each stage (considering multiple levels) in a single query.

Then use this information to draw the Progress component at the top
For each stage should display the icon & label (of the Stage) in color, and follow the same icon/colour key as the application progress bar (left side bar):
Screen Shot 2022-01-20 at 3 04 38 PM
- Green (with Tick ✅ icon): In progress or Missing levels to complete
- White (hollow): Stage not yet reached
- Blue dot (in circle): In progress (i.e. we're at this stage, either applicant or reviewer could be working on it)
- Pink (with X icon): application was rejected at this stage (this is different to application progress bar, but still implies a "negative" state)

Use this as a reference for the progress component:
Screen Shot 2022-01-19 at 3 04 49 PM

CarlosNZ:
Fix #1084 

Took a bit of fiddling to figure out how to override all the styles to get the desired look, but I think it's working quite well now.

Uses this npm package (so don't forget to `yarn install`): https://www.npmjs.com/package/react-progress-stepper

I tried several, this was the one that was the most suitable to our requirements and is fairly lightweight and customisable.

A completed/approved application:
Screen Shot 2022-01-27 at 9 10 24 AM

An application rejected at the screening stage:
Screen Shot 2022-01-27 at 9 10 41 AM

An in-progress application at the Screening stage:
Screen Shot 2022-01-27 at 9 10 51 AM

An in-progress application at the Assessment stage, with LOQ sent to applicant:
Screen Shot 2022-02-02 at 10 44 52 AM


src/containers/Review/ProgressBar.tsx
```diff
@@ -1,22 +1,57 @@
...
+  } = fullStructure
+
+  const [currentStageClass, currentStageIcon]: [string, SemanticICONS] =
+    outcome === ApplicationOutcome.Rejected ? ['rejected', 'remove'] : ['', 'circle']
+
+  const isApproved = outcome === ApplicationOutcome.Approved
+
+  const steps = [
+    <Step key="first">
+      <StepTitle>Submitted</StepTitle>
```
nmadruga, line 26: Please use localization

src/containers/Review/ProgressBar.tsx
```diff
@@ -1,22 +1,57 @@
...
+    ),
+  ]
 
   return (
     <div id="review-progress">
-      <Message size="huge">
-        <Message.Header>Placeholder for PROGRESS BAR</Message.Header>
-      </Message>
+      <div id="stepper-wrapper">
+        <Stepper id="stepper" step={currentStage.number + Number(isApproved)} numbered={false}>
```
nmadruga, line 51: Still trying to understand what this means: `step={currentStage.number + Number(isApproved)}`

yarn-error.log
```diff
@@ -1,17 +1,30 @@
 Arguments: 
-  /Users/nicole/.nvm/versions/node/v14.8.0/bin/node /usr/local/bin/yarn push_docs
+  /Users/carl/.nvm/versions/node/v14.18.1/bin/node /Users/carl/.nvm/versions/node/v14.18.1/bin/yarn add --dev @types/react-progress-stepper
```
nmadruga, line 2: Should we remove this file?

nmadruga (commented):
That should have been fun to do : )

nmadruga:
Not sure what was missing but right after submitting a review (as Screener) and going back to the Review home page the progress bar was still showing as the previous stage:

nmadruga:
I would prefer to have something clearer on this UI about when LOQ was sent back to the Applicant. I think in this case it should also show the red for this stage. What do you think @jagoje @adamdewey ?

This is how it usually shows after sending the LOQ to the applicant. Not ver clear...
Screen Shot 2022-02-01 at 10 29 34 AM

adamdewey:
@nmadruga hmm, red might be a bit aggressive, maybe we need a colour scheme for 'pending' items - yellow? Or light orange?

src/containers/Review/ProgressBar.tsx
CarlosNZ: Yup! ca60091a

nmadruga:
Jessica Jago, [2/02/22 10:06 AM]
I cant reply in Github

Jessica Jago, [2/02/22 10:07 AM]
but I agree it should be marked with the same colour (that pinky red) to show it is at the List of questions

Jessica Jago, [2/02/22 10:07 AM]
this view is only internal so I think that is fine and not aggressive :)

src/containers/Review/ProgressBar.tsx
CarlosNZ: `step` defines the "current" step for the progress bar, so mostly the step number is the stage number. However, in the case where the application is completed, we want the "final" step to show as "done", so we make the progress bar "step" a higher number than there is available, so it shows ALL steps as done. So the `Number(isApproved` just adds 1 to the step number if the application is completed in order to achieve this.

yarn-error.log
CarlosNZ: Yeah good idea: 5d288001

CarlosNZ:
> I would prefer to have something clearer on this UI about when LOQ was sent back to the Applicant. I think in this case it should also show the red for this stage. What do you think @jagoje @adamdewey ?

As long as it's *obvious* what the different colours mean. We don't want to have to provide a key to explain each icon.

> Jessica Jago, [2/02/22 10:07 AM]
> but I agree it should be marked with the same colour (that pinky red) to show it is at the List of questions

So to summarise:
- Done -> Green tick
- In progress (review) -> blue dot
- In progress (applicant) -> pink (something?)
- Not reached -> empty circle

I'll try something with the pink LOQ option, see what we think.

nmadruga:
> * Done -> Green tick
> * In progress (review) -> blue dot
> * In progress (applicant) -> pink (something?)
> * Not reached -> empty circle

I think there is also these 2 variants, right? Considering the `outcome` isn't pending:
* Done (approved) -> Green tick
* Done (rejected) -> Cross pink

CarlosNZ:
> I think there is also these 2 variants, right? Considering the `outcome` isn't pending:
> 
>     * Done (approved) -> Green tick
> 
>     * Done (rejected) -> Cross pink

Yes, correct. (But those are already done). Have updated my previous comment. Thanks

CarlosNZ:
Okay, have updated to show the "LOQ" state. 4969856f

Have added a new screenshot in the original post (top of this thread) -- see what you think @nmadruga @jagoje @adamdewey

adamdewey:
Looks great @CarlosNZ !

CarlosNZ:
> Not sure what was missing but right after submitting a review (as Screener) and **going back to the Review home page** the progress bar was still showing as the previous stage:

This is because the front-end doesn't know to re-load the application from the database. The change of stage happens as an Action, so the front-end doesn't know it should refetch the full structure when going back to the page with the Progress Bar. It just keeps using the existing "fullStructure" value.

Any thoughts on how to achieve this? Let's discuss.

CarlosNZ:
> Any thoughts on how to achieve this? Let's discuss.

This is proving trickier that anticipated so will have to leave it for now until we can find an adequate solution. Have summarised the problem in [this comment](https://github.com/openmsupply/application-manager-web-app/issues/850#issuecomment-1027505690).

So, putting that aside for now -- anything else to address in this PR, @nmadruga ?

nmadruga (approved):
No looks all good now. Thanks for amazing work : )

PR #1091, opened 2022-01-26
#1089A Timeline hook

CarlosNZ:
First version of a timeline data wrangler for #1089.

Requires back-end branch from: https://github.com/openmsupply/application-manager-server/pull/647

Have generated timeline events according to the general categories listed in [this document](https://docs.google.com/document/d/1LWCvv42dHwcv4yCWKG26MJI8yZ6QGHAnSwpY7-rWblQ/edit#) (final page). 

I think we can still have some further discussion as to the trade-off between back-end and front-end logic. Obviously there is quite a bit in both currently, with some of the limitations I mentioned in the above PR. We can talk more about this balance if we need too. Personally I think it might be easier to just put everything in "details" (there's enough there already) and let the front-end work it out from there, but it can work either way. We can discuss.

Currently this just loads a raw JSON readout onto the Overview tab -- haven't done any UI work, that'll happen in #1083 , just want to make sure we standardise a format before progressing further.

Have abstracted all the messy logic away into a `useTimeline` hook, so that the Overview page will receive all the data nicely formatted for easy display.

To test, use the Structure mentioned in the back-end PR, start going through the application process, and note the JSON display which appears on the "Overview" tab (review page) as you go through the various steps.

Issue #1083, opened 2022-01-20
New Review home page: Overview tab


nmadruga:
Includes the timeline

Discussed the idea of creating a new "activity" table in DB to log all events. Will create back-end issue to discuss this in more detail.

CarlosNZ:
Back-end PR: https://github.com/openmsupply/application-manager-server/pull/647

Pre-requisite issue: #1089

Issue #1087, opened 2022-01-24
Assignment bug: Trigger running 5 times!


CarlosNZ:
When selecting "Self-assign", somehow the trigger is being activated five times instead of once.

Haven't tried for regular assignment, will check.

nmadruga:
While chatting today - Carl and I when realized that the reason for that is that the action would run once for each section. That's how the current action to self-assign (or assign) works. 

#### Current assignment actions logic
When a reviewer self-assigns by clicking in one section `self-assign` option, if it needs to be assigned to all the sections (e.g. consolidation or review with only 1 level), the action ON_REVIEW_SELF_ASSIGN will run for each section in sequence.

A similar thing happens when an Asdsigner selects a reviewer to review a section. If all the sections of the review should be done by the same reviewer all other sections will be assigned.  

One noticeable thing is that the page reloads after each action runs - which is ugly. Also now there is a problem with the log that is showing more than once that the review has been self-assigned by a reviewer on the Overview tab.

#### Desirable assignment logic
Changing to run all section-assignment in a single-mutation
Will require a UI change to show a `Submit` button that will do the assignment/re-assignment of all selected reviewers in the sections dropdown. So the action would stop to be running **on change** of the dropdown to only run on button click. 

Disadvantages - when the assigner selects reviewers but doesn't click the submit button it wouldn't apply the assignment.
Advantages - The page won't be flickering anymore and there would only be one assignment action for all sections.

Extra: would be good to show one `assign to all sections` after the reviewer is selected in one section and isn't selected on the other sections.

CarlosNZ:
Nice summary -- thanks :)

PR #1093, opened 2022-01-27
#1083 Overview tab / timeline

CarlosNZ:
Implements #1083 

This is not right yet as you can see from the screenshots (some parsing issues down the bottom -- but that's more of an issue with #1091). But I think we should sit down as a group and figure out the exact content requirements for the timeline now that we have a skeleton, rather than making a bunch of changes now that will probably get changed again.

Also -- note that the timeline would look a bit more structured in a real application as there would be a lot less entries in a single day.

Screen Shot 2022-01-27 at 3 58 39 PM
Screen Shot 2022-01-27 at 3 58 51 PM


nmadruga:
Something that bothered me is that the Overview is always starting closed. Not sure if it should be opened instead.

Also on a chat with Jess and Adam today I said we could finish up the current work:
- Timeline fixes
- Internal comments tab
- Generated documents tab
- Permission for user to view without being the Assigner/Review

So by the end of the week we could sit down and show the new tabs, and the original idea of showing the link to each review from the Overview page and use the time to discuss next deliverables for LOQ and any improvements to this work

nmadruga:
Oh we also talked today about the need to initiate 2 instances one to be a Public demo server and another one for the demo presentation, both to be including this new work + being stable and adding new templates. So challenges for next week.

src/containers/Review/overview/OverviewTab.tsx
```diff
@@ -1,27 +1,26 @@
...
-      <Message>
-        <Message.Header>Placeholder for OVERVIEW tab</Message.Header>
-      </Message>
-      {loading && <p>Loading...</p>}
-      {timeline && <pre>{JSON.stringify(timeline, null, 2)}</pre>}
+      {loading && <Loading />}
+      {timeline && <Overview structure={fullStructure} activityLog={timeline.rawLog} />}
+      <div id="timeline">
+        <Header as="h2" textAlign="center">
+          Activity
```
nmadruga, line 20: Localization please

src/containers/Review/overview/Overview.tsx
```diff
@@ -0,0 +1,79 @@
...
+              <p className="right-item">
+                <strong>Outcome: </strong>
+                {Outcome[outcome as ApplicationOutcome]}
+              </p>
+            </div>
+            <div className="flex-row wrap">
+              <p className="left-item">
+                <strong>Serial: </strong>
+                {serial}
+              </p>
```
nmadruga, line 64: All localization missing. Sorry to be picking on this, but it would be good to already add there as well.

src/containers/Review/overview/Overview.tsx
```diff
@@ -0,0 +1,79 @@
...
+      </Segment>
+    </div>
+  )
+}
+
+const getDates = (activityLog: ActivityLog[]): { started: string; completed: string } => {
+  console.log(activityLog)
+  const startEvent = activityLog.find((e) => e.type === 'STATUS' && e.value === 'Started')
+  const endEvent = activityLog.find((e) => e.type === 'OUTCOME' && e.value !== 'PENDING')
+  console.log('endEvent', endEvent)
```
nmadruga, line 77: Please remove log

nmadruga (commented):
2 overall comments:

- We need to fix the duplication of "Reviewer X assigned themselves" - I"ll see if I can help with that, also to help me understand this work better
- When all sections can we instead of list them say **All sections**?

CarlosNZ:
> Something that bothered me is that the Overview is always starting closed. Not sure if it should be opened instead.

I've changed it so that the current stage starts open, other stages start closed. c54f69c7

src/containers/Review/overview/OverviewTab.tsx
CarlosNZ: Yup, and a few others. 6bc58dc2

CarlosNZ:
Okay, I've significantly re-factored this as discussed, with all the "logging" done in the back-end, and all the "interpretation" done here. It's much more straightforward now, although there's still some complicated aspects.

Added a generic "waiting for applicant/reviewer to make changes" message that only shows when it's the current event (i.e. nothing has happened since), as we discussed.

Also added logic to ensure that the final outcome always shows up last (it might not be in the log)

Should be pretty solid now, so I'll show this version at Friday's meeting.

src/containers/Review/overview/Overview.tsx
```diff
@@ -0,0 +1,78 @@
...
+        <Message info icon>
+          <Icon name="info circle" color="teal" />
+          <Message.Content>
+            <div className="flex-row wrap">
+              <p className="left-item">
+                <strong>{strings.REVIEW_OVERVIEW_APPLICANT}: </strong>
+                {applicant}
+              </p>
+              <p className="right-item">
+                <strong>{strings.REVIEW_OVERVIEW_OUTCOME}: </strong>
```
nmadruga, line 37:
```suggestion
                <strong>{strings. REVIEW_OVERVIEW_ORG}: </strong>
```

nmadruga (commented):
I agree now with moving the logic, seems more straightforward

nmadruga:
problem found while working of issue #1087 

Log aren't showing correctly after un-assignment (or re-assignment that would also run unassign first and then assign).
Screen Shot 2022-02-03 at 3 22 47 PM

Snapshot: 
[unassigned_missing_log.zip](https://github.com/openmsupply/application-manager-web-app/files/7991360/unassigned_missing_log.zip)

CarlosNZ:
> Log aren't showing correctly after un-assignment (or re-assignment that would also run unassign first and then assign).

Okay, so as I mentioned IRL, your log is out of date so the value "STATUS = Started" is no longer used (or interpreted):
Screen Shot 2022-02-03 at 4 32 47 PM

However, the "Unassigned" one was a typo on my part (I had "UnAssigned" in front-end), so I've fixed that. 14020929

Good spotting, thanks.

PR #1097, opened 2022-02-03
#1087 assignment with submit button

### LINKED ISSUE #1087: Assignment bug: Trigger running 5 times!

nmadruga:
Fixes #1087 

Based on PR #1093 

#### Changes
- Remove Assignment and Reassignment events from the change of dropdowns on each section
- Stores new state `assignedSections` in each section on dropdown change and already apply to all sectiion if `isLastLevel`
- When clicking on the new **Submit** button will actually run **a single** event for all sections assigned/reassigned to the same user. The button function will get if it should run the updateReviewAssignment or reassignReview logic.

#### Not sorted
- Error displayed on the dropdown on failures
- When reassignment is done **from more than one** reviewer to a new one that's not covered correctly. Need to rethink the whole logic of re-assignment. If we want to use separated the call to unassign before reassigning.
- Not implemented option to assign all sections (when it doesn't happen automatically) - this is getting too bug, and that's not really a priority so I suggest to be separated on a new issue.


CarlosNZ:
Just testing, will note down comments as I come across them:...

- We were going to remove the "Stage" dropdown selector and just show the current one -- can that be done in this PR?
- For first stage self-assign, the UI is confusing after assigning (shows not-assigned):
Screen Shot 2022-02-04 at 9 25 18 AM
 If the user can no longer re-assign or make changes to this, then the options should be now shown (including Submit button). Although a level 1 self-assigner *should* be able to unassign themselves, right? (They can't with this interface)
- 2nd stage assigner (assigning others). The UI looks exactly the same after "Submit" as it did before, its a bit weird how it just flashes and then looks the same without any feedback. We should probably have a discussion about what should happen here, cos it's not obvious, but needs to be something (confirmation modal? I dunno) I think the main thing is the UI needs to distinguish between what *is* and what you're *going to change it too* Currently they're the same thing, which is why it's confusing. Let's discuss!

nmadruga:
> * We were going to remove the "Stage" dropdown selector and just show the current one -- can that be done in this PR?

It will be done in another PR - which I'm planning to work on next week: #1098

nmadruga:
> * For first stage self-assign, the UI is confusing after assigning (shows not-assigned):
>   Screen Shot 2022-02-04 at 9 25 18 AM
>   If the user can no longer re-assign or make changes to this, then the options should be now shown (including Submit button). Although a level 1 self-assigner _should_ be able to unassign themselves, right? (They can't with this interface)

as discussed in person: that's showing very wrong things - so yes - to be fixed in this PR. It should be only displaying in one line as this user is the (self)-assigner.

nmadruga:
> * 2nd stage assigner (assigning others). The UI looks exactly the same after "Submit" as it did before, its a bit weird how it just flashes and then looks the same without any feedback. We should probably have a discussion about what should happen here, cos it's not obvious, but needs to be something (confirmation modal? I dunno) I think the main thing is the UI needs to distinguish between what _is_ and what you're _going to change it too_ Currently they're the same thing, which is why it's confusing. Let's discuss!

After out meeting today what we decided would be best - which I can start working on it after approval from @jagoje 

### Show the Assignment dropdown 
- When Assigning (when not assigned to someone)
- When Re-assigning

### Show Label
- After clicking on the **Submit** button should change from dropdown to label
Screen Shot 2022-02-04 at 1 07 38 PM
- After the user clicks on Re-assign would change to show the dropdown
Screen Shot 2022-02-04 at 1 08 36 PM

- After clicking in Un-assign button same confirmation as currently is displayed

#### Changes
No longer shows option Re-assign or Un-assign in dropdown... 
After submitting or when loading the page after you were assigned shows as label (no longer as Dropdown)

CarlosNZ:
> * After clicking on the **Submit** button should change from dropdown to label
> 
> 
> Screen Shot 2022-02-04 at 1 07 38 PM

Looks great. Only suggestion would be to include the date (if possible, it's currently in the row below), and to make the "Reassign/Unassign" links not all the way over on the right, just a wee way from the Reviewer element. Also, for consistency with the rest of the app, links next to each other like that we usually separate with a bullet, like: [Re-assign](url) • [Unassign](url) (or a pipe: [Re-assign](url) | [Unassign](url) )

nmadruga:
Just 2 things - one may be related - another is just something to fix that might be good to do a fix (since now is misleading).
- The assignment tab continues flickering while the assignment is running
- Sections without assignable questions are still showing the dropdown (as Not assigned) and allow changing to the **new** assigner but after shows again as Not assigned. Since they can't be assigned now. Show we just add assignment to those?

CarlosNZ (commented):
Cool, basic mechanism is good, the UI works nicely.

Here's just a few initial testing comments, some of these you'll probably already be addressing, and there's a couple of finicky formatting things which you can leave till last to focus on the functionality.

There's still the nasty bug (see below), which we really need to fix. Do you want to discuss what might be the problem there, and how we might fix it?

- Shouldn't see "Re-assign" link if it's not appropriate i.e Screening reviewer who can only self-assign -- they should see "Unassign" but not "Reassign"
- Why does the "Unassign" work the old way (immediate change) but the "Reassign" has the new (better) way? Can we make "Unassign" the same, or is there a reason why it should be this way?
-  Can we disable the button if there's nothing changed yet (it's weird being able to click it and nothing happening)
- You haven't done the "Levels" UI yet? Is that meant to be for this PR too?
- Bug: see screenshot. Senior Reviewer has *completed* their review of first two sections. I'm logged in as Samantha, but the first two sections is saying contradictory info (correct reviewer, but shouldn't be "assigned to me", or "start"-able:
Screen Shot 2022-02-09 at 12 24 00 PM
    I think this is part of a deeper bug that has come up before -- when I try and submit Samantha's review, I can't because it tells me I haven't reviewed the first two sections yet, even though I'm not assigned to them. Perhaps this is a bug for another issue, but it's quite a nasty one.

- Formatting: 
  - When there is no one assigned yet, should probably say "Reviewer: Not assigned" or something perhaps?
  - Smaller gap between "Reviewer" and name, bigger gap between that block and the "Reassign|Unassign" block (so it's clear which ones go together)

nmadruga:
Cool. I think I have addressed the ones we agreed in @CarlosNZ. In summary:
- Reviewer selected label now showing in single-label
- Removed `re-assign` when reviewer can only self-assign (so only option is unassign)
- Disabled submit button before some action is done
- Check that unassignment runs only once

Now I'll create another issue to fix the assignment per section. I'll try to replica while creating so is easy to follow back the bug and fix iin the future. I'll tag to 2.0.1 anyway cheers 🍻

CarlosNZ:
Cool, this is looking really good now and working well. I'm just going to run through a few more tests and look over the code.

CarlosNZ (changes requested):
Bummer - still one functionality bug: assigning different people is not working.

I start with one that is assigned all to the same person:
Screen Shot 2022-02-10 at 10 04 42 AM
Re-assign *some* sections to a different person:
Screen Shot 2022-02-10 at 10 05 46 AM

After "Submit", it ends up like this:
Screen Shot 2022-02-10 at 10 06 42 AM

🙁

Haven't looked at the code, but is the mutation perhaps updating ALL assignments, even the ones that haven't changed?

nmadruga:
In reply to [this comment](https://github.com/openmsupply/application-manager-web-app/pull/1097#pullrequestreview-878046732) re a bug on re-assign for different reviewers

I'll have a look on that now 😺

CarlosNZ:
Thanks Nicole, I'll make a new issue for that remaining bug.

CarlosNZ:
> Thanks Nicole, I'll make a new issue for that remaining bug.

Done: https://github.com/openmsupply/application-manager-web-app/issues/1109

PR #1102, opened 2022-02-05
#1079 Notes tab

### LINKED ISSUE #1079: New Review home page: Internal notes tab
Page to allow uploading documents and adding new internal comments. This page should combine also comments made on the `review_decision` table, although how the permissions work should be checked so we make sure reviewers of the final stages can see previous reviews comments.

Any user with permission to review or assign should be able to see all of the tabs on the Review home page and create comments.

New comments added are saved only on submission of records in new table `application_note`.
Before submitting the comments, the user can upload files using the "Upload document" plugin and all files uploaded should be combined in a JSON that gets saved within the `application_note` record.  

Q1: If the reviewer doesn't have permission to view other `review_decision` they would just be hidden from the Internal comments page?

Q2: Should we differentiate somehow comments of application_note from review_decision?

---

## Implementation overview:

From: https://docs.google.com/document/d/1LWCvv42dHwcv4yCWKG26MJI8yZ6QGHAnSwpY7-rWblQ/edit#heading=h.u7k1iehvx0aw

### Back-end:
Done : https://github.com/openmsupply/application-manager-server/pull/646

- New table “application_note”
  Fields:
  - application_id (FK)
  - user_id (FK)
  - org_id (FK)
  - timestamp
  - comment (markdown text)
  - files ~~(JSON, same format as FileUpload response)~~ just an array of file ids will do
  What permissions do we need for this table?

### Front-end:

- Query “application_note” table for current applicationId AND review_decision table (for “overall comments”)
- Display returned data as per design
- Create UI for adding new comment/file

CarlosNZ:
Fix #1079 

Don't be put off by the large number of files changed -- much of it is just language strings which I tidied up a bit.

Requires branch from back-end PR https://github.com/openmsupply/application-manager-server/pull/653

Screen Shot 2022-02-05 at 2 38 29 PM
 
### Summary of Notes functionality

- Can add new notes, attach files
- Own comments show highlighted
- Can sort by oldest or newest, and also filter for comments with files attached
- Can delete own comments for 5 minutes after posting (This is a really crude implementation of the time limit -- it doesn't update live on the page, and its super-easy to circumvent just by changing your local clock, so not particularly safe, but will do for now. Ideally it should be enforce by the back-end
- Files are deleted along with notes due to back-end trigger

### Additional changes

- Added package "react-string-replace" -- the only straightfoward way to get a hyperlink in the middle of a localised string (The "Markdown supported comment), so need `yarn install`
- Moved `changeable-background` css class into main scope and re-named to `highlight-background` so it can be re-used (used here for highlighted own comments)
- Fixed a few "missing unique key" errors
- Tidied up language strings a bit to make "basic" ones (like "OK", "Save", "Cancel" etc) less repetitious and more re-usable


src/containers/Review/ReviewWrapper.tsx
```diff
@@ -112,6 +123,7 @@ const ReviewWrapper: React.FC<ReviewWrapperProps> = ({ structure }) => {
                 panes={tabPanes}
                 onTabChange={handleTabChange}
                 activeIndex={getTabFromQuery(tab)}
+                // renderActiveOnly={false}
```
nmadruga, line 126: Remove?

src/containers/Review/notes/NotesTab.tsx
```diff
@@ -1,25 +1,221 @@
...
+}
+
+const canDelete = (timestamp: string): boolean => {
+  // This is pretty insecure -- all someone needs to do is change their local
+  // time on their computer and they can delete whatever they want!
+  return (
+    Interval.fromDateTimes(DateTime.fromISO(timestamp), DateTime.now()).length('minutes') <
+    COMMENT_DELETION_LIMIT
+  )
+}
```
nmadruga, line 221: Have you added an issue to the future? I think it's better to do

nmadruga (commented):
General comments:
- Tried to re-upload same file after deleted - nothing happened
- Just nice-to-have would be to show a clickable delete icon on each item when hoover. It seems not very nice on click, but that can be done later

src/containers/Review/notes/NotesTab.tsx
CarlosNZ: We should do it, but we haven't decided what the delete policy should be.

Have made an issue to discuss and plan: #1105

CarlosNZ:
> * Tried to re-upload same file after deleted - nothing happened

Not sure what you mean. I can't see any unexpected behaviour. Can you describe the steps?

> * Just nice-to-have would be to show a clickable delete icon on each item when hoover. It seems not very nice on click, but that can be done later

Oh, okaaay, since you insist: f6ae516d

src/containers/Review/ReviewWrapper.tsx
CarlosNZ: 0e0e7702

nmadruga:
> > * Tried to re-upload same file after deleted - nothing happened
> 
> Not sure what you mean. I can't see any unexpected behaviour. Can you describe the steps?

Sorry forgot to reply to this:
- I uploaded a file
- Removed the file
- Tried to upload the same file
- It doesn't show again on the list

CarlosNZ:
> Sorry forgot to reply to this:
> 
>     * I uploaded a file
> 
>     * Removed the file
> 
>     * Tried to upload the same file
> 
>     * It doesn't show again on the lis

Oh that's weird. It's working as expected in Firefox, but when I try in Chrome or Safari I see the problem you're describing. I'll look into it further...

CarlosNZ:
> Oh that's weird. It's working as expected in Firefox, but when I try in Chrome or Safari I see the problem you're describing. I'll look into it further...

Yeah, no idea what's going on there. In those browsers there is just no response to the event at all -- the handler function is just not running. I don't think it's something we can fix, seems like a browser bug (although weird that it's doing it in two different browsers -- you wanna check in Opera or whatever other browsers you have installed?) 🤷‍♂️

nmadruga (approved):
I'm happy to merge this for now and we can work on this problem in the future. It's not big thing, but it worth creating an issue about that. I'm using Chrome too

Issue #1100, opened 2022-02-03
Bugs in Timeline messages


CarlosNZ:
- Need better fallback when no decision comment:
![MISSING COMMENT](https://user-images.githubusercontent.com/5456533/152446842-11509dde-d0e7-436c-9e30-d030ced9e073.png)
- Replacement parameters and replacements don't properly match up with consolidations:
![CONSOLIDATION BUGGY](https://user-images.githubusercontent.com/5456533/152446917-ae682d1d-f6d7-4ea3-8393-e0e5302c9118.png)

(Should consolidation messages include Section info?)

PR #1104, opened 2022-02-08
#1099 #1100 #850 Review overview improvements

### LINKED ISSUE #1099: Show links to existing reviews on Review Overview tab
~~Should show list of links in Timeline, *below* each Stage heading, but above the timeline activity.~~ (Too much duplication, probably

Also should show active links on timeline messages. e.g. "Reviewer X submitted [their review ](/)of Sections 1,2,3" -- but only ONCE per review, and only if "submitted"

Small extras to do as part of this issue:
- [x] Remove Summary tab
- [x] Add Template name to Overview info box
- [x] Make Overview info box a little tighter (less space between rows)
- [x] Use enum->string map for Decision names (LIST_OF_QUESTIONS, CONFORM, NON_CONFORM)
- [x] Fix up this: https://github.com/openmsupply/application-manager-web-app/issues/1100
- [x] Fix this: https://github.com/openmsupply/application-manager-web-app/issues/850

CarlosNZ:
Fix #1099 #1100 #850 as outlined in #1099 issue.

- Adds hyperlink to review page for *last submission* of each review -- review opens in new tab with the correct sections opened up (i.e. the assigned ones only)
- Simplified and fixed the Review/Consolidation messages so as to only show Section information on Assignment events (as discussed in Telegram)

### Other tweaks (not mentioned in original issue #1099)

- Fixed (another) "nested heading" React error -- problem caused by placing component into `subheader` prop of Semantic's "Header" component
- Changed the "isCompleted" value in ApplicationSummary to "isSubmitted", as the "Review & Submit" message should only be visible in "Draft", not just when completed.
- Updated the "activeSections" query handler to also take `activeSections=all` value. This makes it easier to open a review page with all sections open without the link source knowing how many sections there are in total
- "Default tab" check moved to Route that doesn't include the "ReviewPage" component, otherwise the tab query was showing up in a regular review page
- Fixes to semanticReactMarkdown component -- it wasn't handling external links or new tab links correctly. 
- Added "logType" field to "TimelineEvent" interface -- it's useful to have the *original* activity log type kept along with the newly generated "eventType" type (though open to suggestions for better name?)
- Removed "organisation" column from Application List in "reviewer" mode, since org is already shown in Name column


src/utils/hooks/useTimeline/useTimeline.tsx
```diff
@@ -40,7 +50,8 @@ export default useTimeline
 const buildTimeline = (
   activityLog: ActivityLog[],
   structure: FullStructure,
-  strings: LanguageStrings
+  strings: LanguageStrings,
+  desicionStrings: { [key in Decision]: string }
```
nmadruga, line 54:
```suggestion
  decisionStrings: { [key in Decision]: string }
```

src/utils/hooks/useTimeline/useTimeline.tsx
```diff
@@ -56,15 +67,26 @@ const buildTimeline = (
...
         timestamp: event.timestamp,
         details: event.details,
-        ...generateTimelineEvent[event.type](event, activityLog, structure, index, strings),
+        ...generateTimelineEvent[event.type](
+          event,
+          activityLog,
+          structure,
+          index,
+          strings,
+          desicionStrings
```
nmadruga, line 76:
```suggestion
          decisionStrings
```

nmadruga (changes requested):
Overall comments:
- Really cool that it was that easy to grab the current stage 🎉 
- Also nice that you have fixed those annoying warnings and css looks tider now
- I'm getting this error in ApplicationSummary after uploading a file, not sure if related:
Screen Shot 2022-02-09 at 6 41 14 PM

nmadruga:
Also the links are working nicely! I'll test now after changes are done by the applicant and see what difference it makes. But looks pretty awesome so far : )

Screen Shot 2022-02-09 at 6 45 11 PM

nmadruga:
Quick comment: when there is no organisation associated the field is empty - should we state NO ORGANISATION?

nmadruga:
So, after changing and reviewer-approving the links jumps to the last submitted review and looks great!
Screen Shot 2022-02-09 at 6 50 26 PM

CarlosNZ:
> I'm getting this error in ApplicationSummary after uploading a file, not sure if related:

The GET errors are just missing files, usually logos. The other 2 errors have been around for a while, not sure exactly what they're caused by. I've been keeping an eye on them, but they're not particularly consistent. I think it's something to do with not cleaning up properly when unmounting components, but needs further investigation.

src/utils/hooks/useTimeline/useTimeline.tsx
CarlosNZ: 😳

CarlosNZ:
> Quick comment: when there is no organisation associated the field is empty - should we state NO ORGANISATION?

Yes, good point. 827b70f9

Screen Shot 2022-02-09 at 8 55 35 PM

Also added a subtle fallback for Completed field

CarlosNZ:
FYI -- I've just pushed an update to address the remaining notes from Jess 6705adba

I realised that "level_number" was not being passed as camelCase from the back-end for assignment events, so I've just made a very small tweak to back-end in `#1079F-notes-schema-changes` branch, so make sure you switch to that if you're testing this again. (and any events already in your activity log won't be correct for consolidation assignments)

PR #1106, opened 2022-02-09
#1095 Wait for triggers, reload when appropriate

### LINKED ISSUE #1095: Review Progress bar needs to update to show the current stage after the stage changes
Original problem described in this comment: https://github.com/openmsupply/application-manager-web-app/pull/1092#issuecomment-1026225386

The issue is that there when a review is submitted, the stage change happens in the back end and there's nothing to tell the front-end to re-fetch the application structure, so the progress bar doesn't get the new stage info. 

**(Original) Intended solution:**

Run "refetchQueries" on 'getApplication' as part of the "submitReview" mutation, which will force the structure (and therefore stage info) to reload

**The problem:**

Problem is the trigger on the application table is never changed, so the data fetched as part of refetchQueries still contains the previous stage value. The "submit review" mutation sets the trigger on the REVIEW table, but not the APPLICATION table, so the refetch query won't wait till the trigger is reset (since it never gets set in the first place)

I then thought that it should be possible to ALSO set the trigger on the application table as part of the review submission. However, that's not possible either, as the reviewers don't have permission to write to the application table, so that doesn't work either. (It works when permissions is disabled)

Open to suggestions?

CarlosNZ:
Fix #1095 

Basically works as outlined in [this comment](https://github.com/openmsupply/application-manager-web-app/issues/1095#issuecomment-1032311757), but with a couple of extra points:

- the function to reload the application is attached to the structure in "useLoadApplication", so it can be called from anywhere with `structure.reload()`
- timeline page refetches activity log whenever that tab is re-activated
- added the "useTriggers" hook to "useLoadApplication" and "useGetReviewInfo" rather than them doing their own trigger checking. The other hook that waits for triggers is "useGetApplicationSerial", which I've left as is, since it needs application Id, not serial (since it doesn't know the serial yet), plus it's not really necessary to wait for ALL triggers for that one
- the useTriggers hook reports three different types of error: "trigger" (meaning one of the triggers is in error state), "timeout" (still processing after maximum refetch), or "network" (the request to the endpoint failed somehow) -- the details are printed to the console

You should be able to check the following works:
- Any assignments or review changes (start, submit) are reflected immediately on the Timeline
- Submitting a review that increments the stage should show immediately on the Overview page
- Is there anything else?


nmadruga (approved):
LGTM

PR #1108, opened 2022-02-10
#1098 Review level dropdown

### LINKED ISSUE #1098: Review home page: Add new Filter per  Review level on Assignment tab
Currently when Re-assigning isn't clear which level it is the assignment for. We should:
* Remove the option to change the stage selection (not required since should always show the current)
* Add an option to filter the assignments showing per level (or see all - but would only allow changing one assignment?)
* Allow Assigning someone on the next level even before the Review has started for the previous level (This could be done in separate issue if too big for now)

nmadruga:
Horray! Happy to view this PR out : )

Fixes #1098 

Based on PR #1097 

Preview of filter showing ALL_LEVEL:
Screen Shot 2022-02-10 at 1 05 25 PM

### Changes
- Kept the logic to display assignment per stage in case we need to move back to it (or add other filters) in the future
- Removed the dropdown to select the **Stage**, this is now fixed on the current one
- Changed previous dropdown to filter by reviewer to be filtering by level
- Needed to also group all assignments (in `Assignment` component) to send only assignment of a certain stage to be included in each `AssignmentSectionRow`. This is working well for showing all levels, but there is still a problem when selecting the **next level** when it shows the assignment as per previous one. See screenshot at the end.
- Added the reviewLevel name on the query and structure to be used on the filter droodown
- Show the level name on each Assignment row as well. Attempted to layout this better... Failed.

### Snapshot
Tested using snapshot `dev/snapshots/review_structure_testing` and Test Review template.

The screenshot showing problem (Level: Consolidate Assessment) isn't assigned yet. Something to do with re-rendering component `AssignmentSectionRow`:
Screen Shot 2022-02-10 at 1 16 52 PM


CarlosNZ (changes requested):
Looking great, Nicole. So much nicer to use this interface than the old one.

The [bug from the previous PR](https://github.com/openmsupply/application-manager-web-app/pull/1097#pullrequestreview-878046732) is still there but you can fix that over there.

Only one functionality issue I've noticed:
Screen Shot 2022-02-10 at 1 44 19 PM
We shouldn't see the "Reassign | Unassign" links when that level is not the current level. In this case the previous level review is completed, so doesn't make sense to be able to change it. (I was able to actually "unassign" the reviewer, which doesn't really make sense)

All my comments are really just nitpicky layout things:

- Do you think the default view should be the current level rather than all levels? Seems like that would be a simpler thing to see first.
- I'd put the "Stage: Screening" label back on the far right where it was, since it's non-interactive, and it'll save vertical space by sharing the same row as the level selector
- Some improvements to the spacing of the main row would be nice, as per my comment on the last PR.

I can have a little muck around with those style things and send you a suggestion diff if that helps -- then you can focus on the functionality stuff.

nmadruga:
> The [bug from the previous PR](https://github.com/openmsupply/application-manager-web-app/pull/1097#pullrequestreview-878046732) is still there but you can fix that over there.

Cool, I can have a look and fix in this one then. I think we should merge the other one now.

> I can have a little muck around with those style things and send you a suggestion diff if that helps -- then you can focus on the functionality stuff

Yes. Thanks, that would be helpful

src/containers/Review/assignment/AssignmentSectionRow.tsx
```diff
@@ -122,9 +125,20 @@ const AssignmentSectionRow: React.FC<AssignmentSectionRowProps> = ({
...
-    <Grid className="section-single-row-box-container">
+    <Grid columns={2} className="section-single-row-box-container">
       <Grid.Row>
+        <Grid.Column className="review-level">
+          <Label
+            className="simple-label"
+            content={`${strings.REVIEW_FILTER_LEVEL}: ${levelName}`}
+          />
+        </Grid.Column>
         <Grid.Column className="centered-flex-box-row">
```
CarlosNZ, line 142:
```suggestion
    <Grid columns={2} stackable className="section-single-row-box-container">
      <Grid.Row>
        <Grid.Column className="review-level" width={5}>
          <Label className="simple-label">
            {strings.REVIEW_FILTER_LEVEL}:&nbsp;&nbsp;<strong>{levelName}</strong>
          </Label>
        </Grid.Column>
        <Grid.Column className="centered-flex-box-row" width={11}>
```
I think these proportions are a little nicer. (FYI I told you wrong, it's the "width" prop you needed, "colspan" is just HTML)
Also, **bolding** of the Level name for consistency with my comment on the last PR (bold the review name)

CarlosNZ (changes requested):
Okay, 2 small layout suggestions added inline, as per my last review comment

semantic/src/site/globals/site.overrides
```diff
@@ -546,6 +546,14 @@ a:hover {
 .review-section-row {
   .section-single-row-box-container();
 
+  .ui.grid > .row {
+    justify-content: stretch;
+  }
+
+  // .ui.grid > .column.review-level {
+  //   width: 10%;
+  // }
```
CarlosNZ, line 555: Remove commented

src/utils/types.ts
```diff
@@ -251,7 +252,7 @@ interface EvaluatorParameters {
 type ElementsById = { [templateElementId: string]: PageElement }
 
 interface Filters {
-  selectedReviewer: number
+  selectedLevel: number
   selectedStage: number
```
CarlosNZ, line 256: Maybe change name of this field to "currentStage" since it's no longer selectable?

src/containers/Review/assignment/ReviewLevel.tsx
```diff
@@ -0,0 +1,81 @@
...
+          options={getLevelsOptions(stage.levels)}
+          value={filters?.selectedLevel}
+          onChange={changeFilters('selectedLevel')}
+        />
+      </div>
+    </div>
+  )
+}
+
+const useHelpers = () => {
```
CarlosNZ, line 58: Does this need to be a hook rather than just a static function? Is it just because you have to use "strings" in there, in which case it has to be a hook to call another hook?

I've come across this issue too quite a bit -- I've just been calling the "useLanguageProvider" at the top level and passing in the strings object to static functions as a parameter. I guess either way is fine, don't suppose you have any idea about the performance differences?

CarlosNZ (commented):
Sorry about my piece-meal reviews. Just wanted to give you something to go on with while I looked through the code.

This is all looking really solid though. Comments all fairly minor.

src/containers/Review/assignment/AssigneeLabel.tsx
```diff
@@ -4,6 +4,7 @@ import { useLanguageProvider } from '../../../contexts/Localisation'
 
 interface AssigneeLabelProps {
   assignee: string
+  isSubmitted: boolean
```
CarlosNZ, line 7: Sorry to be a pain -- just one comment about this:
There is the case where a consolidator requested changes from the lower level reviewer, but that reviewer is no longer available so they need to be re-assigned. By restricting to "isSubmitted" we miss that case. I was thinking a better prop would be `isCurrentLevel` (the current levelName is already in your "AssignmentSectionRow" component), so we can do re-assignments if it's gone back to that level, not just if it's submitted or not.

src/containers/Review/assignment/AssignmentTab.tsx
```diff
@@ -63,16 +54,29 @@ const AssignmentTab: React.FC<{
...
+      <div className="flex-column-start-center">
+        <Label className="uppercase-label" content={strings.REVIEW_OVERVIEW_STAGE} />
+        <Stage name={stageName} colour={stageColour || ''} />
+        <ReviewLevel
+          filters={filters}
+          setFilters={setFilters}
+          structure={fullStructure}
+          assignments={assignments}
+        />
+      </div>
```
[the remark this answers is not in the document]
nmadruga: I've just made a change on this one... See if that's good:
Screen Shot 2022-02-10 at 2 40 22 PM

nmadruga:
> * Do you think the default view should be the current level rather than all levels? Seems like that would be a simpler thing to see first.

I don't think we currently have such thing as the `currentLevel`... So, also replying to this [comment](https://github.com/openmsupply/application-manager-web-app/pull/1108/#discussion_r803231897) I don't think we can do that without more fiddling...

src/containers/Review/assignment/AssignmentTab.tsx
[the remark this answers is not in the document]
CarlosNZ: Great thanks.

src/containers/Review/assignment/ReviewLevel.tsx
nmadruga: Not sure... I only kept it there since it was already being used. But it could be a function inside the component I guess...

src/utils/types.ts
nmadruga: Sure

semantic/src/site/globals/site.overrides
nmadruga: Cool. I was just trying to make the first column smaller. But let's leave it for a future layout improvement.

CarlosNZ:
> I don't think we currently have such thing as the `currentLevel`... So, also replying to this [comment](https://github.com/openmsupply/application-manager-web-app/pull/1108/#discussion_r803231897) I don't think we can do that without more fiddling...

You already have this in AssignmentSectionRow, which retrieves the current level:
```
  const levelName =
    structure.stages
      .find(({ stage: { number } }) => number === structure.info.current.stage.number)
      ?.levels.find(({ number }) => reviewLevel === number)?.name || strings.LEVEL_NOT_FOUND
```
So it's definitely available for the Assignee label, and you can just move that statement up  a level (or two?) and it will be available for the whole tab.

src/containers/Review/assignment/ReviewLevel.tsx
CarlosNZ: All good, it's fine. Just something to keep in mind for future. For some reason I'm reluctant to create new hooks unless they're really necessary -- but I don't really have any justification for that 😆

nmadruga:
I'll fix the conflict by running the branch `update-core-templates` with yarn database_init and merge this

PR #1112, opened 2022-02-10
#655B add assigned sections to review assignment table

nmadruga:
Front-end PR for back-end issue [PR#655](https://github.com/openmsupply/application-manager-server/pull/656)

Based on PR #1106 

### Changes
- Removed parameter not in use on `getReviewOptions`...
- Renamed variable to be same as back-end `allowedSections`
- Fetch new field in assignment table `assignedSection` - remember to run `yarn generate` after reloading the back-end
- Migration code to run once the review is fetched and update the assignedSections to list all the ones found in `reviewQuestionAssignments`

### DIFF
--- a/src/utils/hooks/useGetReviewInfo.tsx
+++ b/src/utils/hooks/useGetReviewInfo.tsx
@@ -3,6 +3,7 @@ import { AssignmentDetails } from '../types'
 import {
   Review,
   ReviewAssignment,
+  ReviewAssignmentStatus,
   ReviewQuestionAssignment,
   ReviewStatus,
   useGetReviewInfoQuery,
@@ -84,6 +85,7 @@ const useGetReviewInfo = ({ applicationId, serial }: UseGetReviewInfoProps) => {
         reviewer,
         reviewAssignmentAssignerJoins,
         allowedSections,
+        assignedSections,
         isFinalDecision,
         isLastLevel,
         isSelfAssignable,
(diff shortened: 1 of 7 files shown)


src/utils/hooks/useGetReviewInfo.tsx
```diff
@@ -152,4 +157,26 @@ const useGetReviewInfo = ({ applicationId, serial }: UseGetReviewInfoProps) => {
...
+          const code = templateElement?.section?.code as string
+          if (!assignedSections.includes(code)) assignedSections.push(code)
+          return assignedSections.sort()
+        },
+        []
+      )
+    }
+  }
+}
+
```
CarlosNZ, line 181: I don't think we need to start putting migration code in yet do we? It's still fairly trivial to just manually run an SQL command on existing databases.

Plus, this would be easier done in the back-end wouldn't it? Seems like it would be tidier to manually mutate the data ONCE rather than modify local state in front-end every time it's run.

I think we need to have a plan for adding migration code, I'm a bit reluctant to start adding more complexity to already complicated existing components.

For migration code, I'd probably want to have:
- as little as possible
- all in one script, preferably in the back-end
- mutates data once, so it doesn't have to make state conversions at runtime
- doesn't run as part of running app -- should be a start-up or manual thing only

Anyway, let's chat about it early next week?

CarlosNZ (changes requested):
Cool. So this is just to fetch the relevant section data - it doesn't actually use it at all yet, right?

src/utils/hooks/useGetReviewInfo.tsx
nmadruga: Totally agreed. So, I'll be removing this now and checking out [your PR](https://github.com/openmsupply/application-manager-server/pull/659) that moves it to back-end

nmadruga:
Closing so all changes can be checked and merged as part of PR #1116

Issue #1081, opened 2022-01-19
Misc bugs during review process


CarlosNZ:
Jess noticed some issues yesterday, so I'm going through the review process in detail and I'll note any issues I find here:

- "NOT FOUND" stage in History:
![Screen Shot 2022-01-20 at 12 05 21 PM](https://user-images.githubusercontent.com/5456533/150232602-ac6666d8-233b-43b0-bc3d-532a80b77e70.png)
What is this for? What should it be? (Note: This disappears after the application had moved into Assessment stage)

- Self-assigned a section, but shows as "Not Assigned":
![Screen Shot 2022-01-20 at 12 12 29 PM](https://user-images.githubusercontent.com/5456533/150233435-5e27fb5d-e17b-4748-bc22-a6936d6bf5f2.png)

- Not showing who the other Section is assigned to (I'm logged in as Senior Reviewer, other section has been assigned to Admin)
![Screen Shot 2022-01-20 at 12 18 21 PM](https://user-images.githubusercontent.com/5456533/150234018-783cf414-55f8-4dde-8c4c-587a39032c8a.png)
(Also, having it in blue is confusing, since it's not actually a link)

UPDATE on 11/02/2022: Next 3 issues should be fixed in PR #1113 👇 

- Senior Reviewer has completed their application. But now "Admin" can see a "start" link for reviews that are already completed:
![Screen Shot 2022-01-20 at 12 24 57 PM](https://user-images.githubusercontent.com/5456533/150234559-bf196836-bfae-4ab5-97c7-ac2e59b272bc.png)
It might be because Admin has a section assigned to them which they haven't yet started, so it's displaying "Start" on all sections erroneously. Don't think the self assign link should be visible though. (Note: doesn't show up in the Dashboard "available for self-assignment" filter, so it's being filtered correctly there)

- Following on from previous, when I (as Admin) "start" one of the sections I'm not assigned to, it looks like the review that's created is correct (i.e. can only review the section that I've been assigned). However, all the other sections still say "Assigned to you" even though they aren't:
![Screen Shot 2022-01-20 at 12 30 23 PM](https://user-images.githubusercontent.com/5456533/150235038-52d216cb-954f-4e94-bf3c-00bd74a78fd4.png)

- Continuing -- when I try and submit review (as Admin), I get a warning that I haven't completed Section 1 -- even though I'm not assigned to it, and it's already been completed (by Senior Reviewer)
[...]

PR #1113, opened 2022-02-11
#1083 Part1 fix partial reviews not working

nmadruga:
Part 1 - since there are a few other with minor priority

Based on PR #1112 

Fixing issues listed in issue #1081:

> - Senior Reviewer has completed their application. But now "Admin" can see a "start" link for reviews that are already 
> completed:
> Screen Shot 2022-01-20 at 12 24 57 PM
> It might be because Admin has a section assigned to them which they haven't yet started, so it's displaying "Start" on all sections erroneously. Don't think the self assign link should be visible though. (Note: doesn't show up in the Dashboard "available for self-assignment" filter, so it's being filtered correctly there)
> 
> - Following on from previous, when I (as Admin) "start" one of the sections I'm not assigned to, it looks like the review that's created is correct (i.e. can only review the section that I've been assigned). However, all the other sections still say "Assigned to you" even though they aren't:
> Screen Shot 2022-01-20 at 12 30 23 PM
> 
> - Continuing -- when I try and submit review (as Admin), I get a warning that I haven't completed Section 1 -- even though I'm not assigned to it, and it's already been completed (by Senior Reviewer)
> Screen Shot 2022-01-20 at 12 31 44 PM
> But there's no button to actually review it, so I'm stuck now.
> 
> There's probably more, but I'll leave it at that for now. I think a lot of this is caused by having a user who has permission for both levels of review -- the UI gets very confused. I'll try again and make sure level permissions are separated out and see what happens

### Changes
- This used to work before the latest changes were added to Re-assignment and Final decision... But since the code is very much non-trivial with all concepts of Reviewable and Active is hard to get to what exactly changed the original behaviour.
- Fixed by using the `assignedSections` to only set Reviewable for these and Review or Consolidation progress also uses that to only be setting progress on pages and sections assigned to the current user.

### DIFF
--- a/src/containers/Review/assignment/ReviewSectionRow.tsx
+++ b/src/containers/Review/assignment/ReviewSectionRow.tsx
@@ -48,7 +48,9 @@ const ReviewSectionRow: React.FC<ReviewSectionRowProps> = ({
   if (!section) return null
 
   const thisReview = fullReviewStructure?.thisReview
-  const isAssignedToCurrentUser = !!section?.assignment?.isAssignedToCurrentUser
+  const isAssignedToCurrentUser =
+    !!section?.assignment?.isAssignedToCurrentUser &&
+    assignment.assignedSections.includes(section.details.code)
 
   const props: ReviewSectionComponentProps = {
     fullStructure: fullReviewStructure,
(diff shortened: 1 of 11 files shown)


nmadruga:
Closing so all changes can be checked and merged as part of #1116

Issue #1110, opened 2022-02-10
Assignment dropdown not correctly updated when changing to another level


nmadruga:
I forgot about this problem, it should be simple to fix. I might have a look today. But before I forget is here:

Preview of filter showing ALL_LEVEL:
![Screen Shot 2022-02-10 at 1 05 25 PM](https://user-images.githubusercontent.com/16461988/153312702-f9e65182-dbfa-4009-ae12-22001b4fc221.png)

Then change to the level without assignments (Level: Consolidate Assessment) but it is showing the same as previous level assigned:
![Screen Shot 2022-02-10 at 1 16 52 PM](https://user-images.githubusercontent.com/16461988/153312694-40112e12-4b29-4d1e-bb72-43abfdae48b1.png)

nmadruga:
This is basically fixed now in PR #1116 but there is just a couple of extra things to fix and make the Assignment of Consolidator as it should be:
- The option to un-assign is still available for Submitted reviews... not sure how this got undone
- When assigning a Consolidator over a partial review (Only S3 & S4 submitted for example) the result is now different than what it used to be: I think it's due to the new use of **AssignedSections**. We need to assign all sections in this case anyway.

An image is more worth than words:
![Screen Shot 2022-02-17 at 1 06 34 PM](https://user-images.githubusercontent.com/16461988/154378567-03753aac-3609-4d59-a5dc-2b1078e0f4c8.png)

Issue #1109, opened 2022-02-10
Assignment bug -- problem when reassigning not all the sections to a reviewer who previously was assigned to all sections


CarlosNZ:
See this comment for description of the problem: https://github.com/openmsupply/application-manager-web-app/pull/1097#pullrequestreview-878046732

The problem is that the process/mutation to change the assignment doesn't check sections. So when reviewer is re-assigned for some sections, it just marks their whole review_assignment record as "AVAILABLE" again.

What it should do is check if some sections are still assigned, and reallocate review-question-assignments accordingly.

nmadruga:
Something else - noted while working on un-assignment mutation update PR #1124 - is that we previously were not removing `reviewQuestionAssignments` which were linked to the `reviewAssignment` being unassigned. That has been done now as part of the refered PR, but would need some added logic when we want to do a partial unassignment, so only reviewQuestionAssignments belonging to certain sections should be removed also.

PR #1116, opened 2022-02-16
#1115 refactor assignments full structure with review structures state

### LINKED ISSUE #1115: Restructure required to unify each assignment review structures in one
Noted while trying to fix #1110 ....  

### The problem found
The problem was caused when we changed how the assignment is done. The change done in 
#1097 started to be using only one of the structures (available in one of the sections) to generate the reviewAssignemnt. This wasn't a problem for reviews level 1, although for review level 1+ was required to get the `lowerLevelReviewLatestResponse` in `constructAssignSectionPatch` on the hook to update assigment `useUpdateReviewAssignment`. But it was not set correctly, **because** we were generating one structure: FullStructure with ALL the elements in it, but only linked properly other reviews belonging to that section. This means we are creating one strucutre per assignment per section. That's a lot of structures!

### The solution
Currently, the FullStructure not only keeps things related to the Application itself (validation and details about each element), but also stores things such as `thisReview`,  `assignment` and all the history of reviews done for each element inside `sections: SectionState`. This isn't required and in cases that we need to have more than one structure (ReviewHomePage) it gets very confusing, specially with the multiple structures that we are building (per section).

The proposed plan is to keep things separated:
#### applicationStructure: FullStructure
```
interface FullStructure {
  attemptSubmission: boolean
  canApplicantMakeChanges: boolean
  details: ApplicationDetails
  elementsById?: ElementsById
  firstIncompleteReviewPage?: SectionAndPage
  lastValidationTimestamp?: number
  reload: () => void
  responsesByCode?: ResponsesByCode
  sortedSections?: SectionState[]
  sortedPages?: Page[]
  stages: {
    stage: StageDetails
    levels: LevelDetails[]
  }[]
}
```

##### reviewStructures: ReviewStructure[]

Each `reviewStructure` element contains:
```
interface ReviewStructure {
  assignment: ReviewAssignment
  details: ReviewDetails | null
  sections: SectionState[]
}
```

Just for clarity, the `SectionState` and `Page` are as following:
```
interface SectionState {
  details: SectionDetails
  progress?: ApplicationProgress
  reviewProgress?: ReviewProgress
  consolidationProgress?: ConsolidationProgress
  assignment?: SectionAssignment
  changeRequestsProgress?: ChangeRequestsProgress
  pages: {
    [pageNum: number]: Page
  }
}
```

```

interface Page {
  number: number
  sectionCode: string
  name: string
  progress: ApplicationProgress
  reviewProgress?: ReviewProgress
  consolidationProgress?: ConsolidationProgress
  changeRequestsProgress?: ChangeRequestsProgress
  state: PageElement[]
}
```

### Summary of changes
- Separate into new strctures
- Change all following components/helpers to start using ReviewStructure instead: 
> If I remove thisReview, assignment from the FullStructure is breaking basically these components:
> - ReviewPage
> - ReviewSubmit
> - usReviewSubmit
> - ReviewSectionRowAction
> - useGetReviewStructureForSection/helpers
> - TemplateBuilder/template/Form/Elements.
> 
> If I also remove sections with the progress for each section:
> - utils/helpers/structure/generateReviewValidity
> - utils/helpers/structure/generateReviewerResponsesProgress
> - utils/helpers/structure/generateFinalDecisionProgress
> - utils/helpers/structure/generateConsolidationValidity
> - utils/helpers/structure/generateConsolidatorResponsesProgress
> - utils/helpers/structure/generateApplicantResponsesProgress
> - utils/helpers/structure/checkPageIsAccessible
> - utils/helpers/structure/addSortedSectionsAndPages
> - utils/helpers/structure/addEvaluatedResponsesToStructure
> - utils/helpers/structure/addElementsById
> - utils/helpers/structure/addChangeRequestForReviewer
> - Application/ProgressBar
> - Application/ApplicationSections
> - TemplateBuilder/template/Form/Pages
> - TemplateBuilder/template/Form/Sections
- Pass all sections to the correctly generate a single-structure per ReviewStructure in `useGetReviewStructureForSection/helpers`
- Check if assignments woud be working correctly now

nmadruga:
Fixes #1115

~~Based on PR #1113 (which also uses [Back-end PR](https://github.com/openmsupply/application-manager-server/pull/656)~~

UPDATED 17/02
Using backend [#642-Migration script](https://github.com/openmsupply/application-manager-server/pull/659)

Please use the snapshot to test `dev/snapshots/assignment_fixes` (in Templates repo)

### Changes
~~- Sorry about so many changes! Maybe I shouldn't have renamed `fullApplicationStructure` to `fullReviewStructure` in the types, since a lot of the files changes are because of that... Let me know if better to revert.~~
- Basically, I wanted to be able to call the hook `useGetReviewStructureForSections` for ALL sections on each reviewAssignment to be out of the ReviewSectionRow (which is already for a specific **section**), but it can't be called in a loop statement - hooks' rules - and create one `reviewStructure` of type FullStrucutre. 
So I have adapted to have the data flow in the following model:
```
AssignmentTab
   Assignment (per reviewAssignment) - not rendered only used to call the hook
   AssignmentRows
      AssignmentSectionRow (per section/level)
      ReviewSectionRow (per section/level with existing review started or submitted)
```
- New context `ReviewStructuresState` that stores one `reviewStructure` per assigment.id on a map (so it's easily accessible to be passed to ReviewSectionRow) that gets set after the hook is called and generates a single **reviewStructure** (starting from the fullReviewStructure that doesn't have the reviewResponses) for each reviewAssignment. 
~~- Required to add the field `syncToken` in the context to avoid `Assignmment` component to try to update the component many times and not allow the Page to render. That's not ideal and tehre is still a warning, so a better solution should be added soon.~~ Solution by Carl added in [d13a622](https://github.com/openmsupply/application-manager-web-app/pull/1116/commits/d13a622d0ed9f63226c2acb64c6b9423bb741c1c)
- Moved all the logic to define if is `assignment` or `reassignment` to component that renders the Submit button: `AssignmmentSubmit` which also makes `AssignmentTab a lot simpler.


nmadruga:
Ok - just realised that a lot of changes haven't been explained. I removed `assignment` and `thisReview` from some calls to reduce the amount of components passed, there is already one Assignment object and review which are included in the `fullReviewStructure`.

CarlosNZ:
> Sorry about so many changes! Maybe I shouldn't have renamed `fullApplicationStructure` to `fullReviewStructure` in the types, since a lot of the files changes are because of that... Let me know if better to revert.

No, that's good -- I'm always in favour of more meaningful names. Could is just be "ReviewStructure" though? Is there a difference between 'ReviewStructure' and 'fullReviewStructure'?

src/containers/Review/assignment/Assignment.tsx
```diff
@@ -1,162 +1,48 @@
...
+      setReviewStructures({ type: 'getSyncToken' })
 
-    const firstAssignedSections = Object.entries(assignedSections).filter(
-      ([_, { previousAssignee }]) => previousAssignee === undefined
-    )
+      setReviewStructures({
+        type: 'addReviewStructure',
+        reviewStructure: fullReviewStructure,
+        assignment,
+      })
```
CarlosNZ, line 39: Oooh, I think I've figured it out -- `useEffect` to the rescue!

Don't need the tokens at all.

The whole "Assignment" component can just be:
```ts
// Component only used to update the contxt with reviewStructuresState (run one instance of hook)
const Assignment: React.FC<ReviewHomeProps> = ({ assignment, structure }) => {
  const {
    reviewStructuresState: { syncToken, structures },
    setReviewStructures,
  } = useReviewStructureState()

  const shouldUpdate = !Object.keys(structures).some((key) => Number(key) === assignment.id)

  const { fullReviewStructure, error } = useGetReviewStructureForSections({
    reviewAssignment: assignment,
    fullReviewStructure: structure,
    awaitMode: !shouldUpdate,
  })

  useEffect(() => {
    if (fullReviewStructure && shouldUpdate) {
      setReviewStructures({
        type: 'addReviewStructure',
        reviewStructure: fullReviewStructure,
        assignment,
      })
    }
  }, [fullReviewStructure])

  if (error) return <Message error title={strings.ERROR_GENERIC} list={[error]} />

  return null
}
```

Loads properly, seems to work and no React error! (there's still a missing key error, but I've mentioned that elsewhere)

CarlosNZ (changes requested):
I think I have a solution -- hopefully it works, see inline comment.

There's a bit of formatting weirdness that's been introduced, but we can address that seperately (I can take a look at it if you want)

src/containers/Review/assignment/AssignmentRows.tsx
```diff
@@ -0,0 +1,67 @@
...
+  } = useReviewStructureState()
+
+  return (
+    <>
+      {Object.values(fullStructure.sections).map(({ details: { id, title, code } }) => (
+        <Segment className="stripes" key={id}>
+          <Header className="section-title" as="h5" content={title} />
+          {Object.entries(assignmentGroupedLevel).map(([level, assignments]) => (
+            <>
+              <AssignmentSectionRow
```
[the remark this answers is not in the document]
nmadruga: Yes, not reason. Usually it's nice to have all in on props object, which isn't the case here. I'll change to the conventional approach.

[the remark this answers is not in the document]
nmadruga: Thanks added. As shoed IRL the key on the outter block `<>` does mess with the layout, so leaving that one.

src/containers/Review/assignment/AssignmentSectionRow.tsx
```diff
@@ -172,7 +171,6 @@ const AssignmentSectionRow: React.FC<AssignmentSectionRowProps> = ({
           <Reassignment
             assignments={assignments}
             sectionCode={sectionCode}
-            elements={elements}
             isLastLevel={isLastLevel}
             previousAssignee={assignmentOptions.selected}
             assignedSectionsState={assignedSectionsState}
```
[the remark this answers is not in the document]
nmadruga: Cool. Thank you I'll do it now

src/containers/Review/assignment/Assignment.tsx
nmadruga: AMAZING!!! I'll add this right away and test. I guess after that we can probably start workiing on merging all this together? I still would like to add a solution for issue #1109 and #1110, but I would like more if this can be done in the next version, based from `develop`.

nmadruga:
> > Sorry about so many changes! Maybe I shouldn't have renamed `fullApplicationStructure` to `fullReviewStructure` in the types, since a lot of the files changes are because of that... Let me know if better to revert.
> 
> No, that's good -- I'm always in favour of more meaningful names. Could is just be "ReviewStructure" though? Is there a difference between 'ReviewStructure' and 'fullReviewStructure'?

Sure, I can change to reviewStructure no problem. The `full` in it was just to emphatize this is **still** the same structure as used for ApplicationStructure, just adding the reviews elements.

nmadruga:
Ok - this should be ready for a re-review now. 

I'm going to add a few missing bits to #1110 seems which is particularly fixed:
- The option to un-assign is still available for Submitted reviews... not sure how this got undone
- When assigning a Consolidator over a partial review (Only S3 & S4 submitted for example) the result is now different than what it used to be: I think it's due to the new use of **AssignedSections**. We need to assign all sections in this case anyway.

An image is more worth than words:
Screen Shot 2022-02-17 at 1 06 34 PM

src/containers/Review/assignment/Assignment.tsx
```diff
@@ -1,162 +1,42 @@
...
+  assignment: AssignmentDetails
+  structure: FullStructure
 }
 
-const Assignment: React.FC<ReviewHomeProps> = ({
-  assignmentsByStageAndLevel,
-  assignmentInPreviousStage,
-  fullApplicationStructure,
-}) => {
+// Component only used to update the contxt with reviewStructuresState (run one instance of hook)
```
CarlosNZ, line 12:
```suggestion
// Component only used to update the context with reviewStructuresState (run one instance of hook)
```

src/containers/Review/assignment/Assignment.tsx
```diff
@@ -1,162 +1,42 @@
...
-          reassignmentGroupedSections.push({
-            sectionCodes: [sectionCode],
-            reassignment,
-            unassignmentId: unassignment.id,
-          })
-      }
-    })
-    reassignmentGroupedSections.forEach(({ sectionCodes, reassignment, unassignmentId }) =>
-      reassignSections({ sectionCodes, reassignment, unassignmentId })
-    )
```
CarlosNZ, line 79: I love seeing huge chunks get replaced with something smaller and simpler! 👍🏼

src/containers/Review/assignment/AssignmentRows.tsx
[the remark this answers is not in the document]
CarlosNZ: Keys fixed in #1120 -- by removing the "stripes" and just assigning a different background to each type of row.

src/containers/Review/notes/NewCommentForm.tsx
```diff
@@ -31,7 +31,7 @@ const NewCommentForm: React.FC<{
       setError(strings.REVIEW_NOTES_COMMENT_ERROR)
       return
     }
```
CarlosNZ, line 33: This shouldn't be in here. It's reverting changes that I've already done and merged. Weird.

Anyway, can you just copy and paste this whole file from develop into here to make it correct again?

Or maybe you can merge from develop and that might fix it?

CarlosNZ (changes requested):
Wow -- what a biggie! I'm sure we've missed things with all the changes, so we'll just have to keep a close eye on it in the next while. And you already know about and can fix some of the bugs so that's good 👍🏼 

Can you just fix those incorrect changes in the NewCommentForm please and then we can merge this.

src/containers/Review/notes/NewCommentForm.tsx
nmadruga: Oh, it's due to my rebase.... I was trying to make this up-to-date with latest changes from other PRs...

PR #1118, opened 2022-02-16
#1074 List text adjustment

### LINKED ISSUE #1074: Application list -- layout problem in "reviewer" mode
Needs further investigation, but table columns are a bit messed up in "reviewer" mode. The addition of the extra "organisation" column is making the other stuff too squeezed up. Also -- should there actually be organisations listed in that column?

Screen Shot 2021-12-08 at 10 51 13 AM

Suggest making the text for the "2 weeks ago" smaller for starters. Also, probably want to set a width limit on the Name column.

CarlosNZ:
Fix #1074 

The organisation column has already been done elsewhere, so all I've done is make the text for the date strings a bit smaller so it can fit on one line.
Screen Shot 2022-02-17 at 10 00 09 AM


nmadruga (approved):
Nice

PR #1119, opened 2022-02-16
#1032 Restore text element default

### LINKED ISSUE #1032: "Standard" defaults for Text input plugins
Currently the "default option" is only evaluated as a stored response when the application is created. This isn't good for elements where we want them to be pre-filled with other responses from the current application. 

Suggest restoring the original "dynamic" default parameter for text input elements. It's already there for most of the others (dropdown etc.), should really be available for text as well.

CarlosNZ:
Fix #1032 

Restored the "default" field of shortText and longText elements, so templates have the option to use this parameter, or the shared "defaultValue" parameter -- since they behave quite differently. I've added an explanation in the docs as to what the difference is.

I have a made a snapshot: `/dev/snapshots/text_field_demo` in the templates repo. There is a template type called "Short Text Demo", which is accessible from the User menu dropdown. You can play around with the values and see the behaviour of the fields with dynamic defaults (Can log in as "Carl" and start a new form if you want)

Expected behaviour:
- Fields with dynamic defaults should change as their source values (other fields) change
- Once the user has edited the value, it will no longer change in response to changes in the source fields

### DIFF
--- a/docs/internal/Element-Type-Specs.md
+++ b/docs/internal/Element-Type-Specs.md
@@ -37,7 +37,8 @@ These fields are common to all element types and have their own field in the `te
   - default: `{"value": true}`
 - **is_editable**: `JSON` -- dynamic query determining whether can be edited (Would only be false in rare circumstances)
   - default: `{"value": true}`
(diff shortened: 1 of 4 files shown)


nmadruga (approved):
All good

PR #1120, opened 2022-02-17
#1115B Layout tweaks

CarlosNZ:
I'm not super happy with this, but there's just too many combinations of assignment rows and review rows with so many different possible things inside them, that it's too hard to make them *all* look good.

Right-justifying the actions at the end of each type of row is not possible without overhauling the whole layout structure which would be a big job.

I think this will have to do for now.

Screen Shot 2022-02-17 at 3 40 29 PM

### DIFF
--- a/src/containers/Review/assignment/AssignmentRows.tsx
+++ b/src/containers/Review/assignment/AssignmentRows.tsx
@@ -34,11 +34,7 @@ const AssignmentRows: React.FC<AssignmentRowsProps> = ({
         <Segment key={id}>
           <Header className="section-title" as="h5" content={title} />
           {Object.entries(assignmentGroupedLevel).map(([level, assignments]) => (
-            <div
-              className="flex-column"
-              key={`assignment-group-level-${level}`}
-              // style={{ gap: 10 }}
-            >
+            <div className="flex-column" key={`assignment-group-level-${level}`}>
               <AssignmentSectionRow
                 assignments={assignments}
                 sectionCode={code}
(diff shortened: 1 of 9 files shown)


src/containers/Review/assignment/AssignmentRows.tsx
```diff
@@ -31,12 +31,15 @@ const AssignmentRows: React.FC<AssignmentRowsProps> = ({
...
           <Header className="section-title" as="h5" content={title} />
           {Object.entries(assignmentGroupedLevel).map(([level, assignments]) => (
-            <>
+            <div
+              className="flex-column"
+              key={`assignment-group-level-${level}`}
+              // style={{ gap: 10 }}
+            >
               <AssignmentSectionRow
-                key={`assignment-row-section-${code}-level-${level}`}
```
nmadruga, line 39: Oh, it doesn't need this then? Maybe also not required in `ReviewSectionRow` below?

nmadruga (changes requested):
Cool. Thanks for doing this, I can see some improvements, and I agree it could be better but will require quite a reshuffle around. For now, only 2 changes if possible please:

I didn't like this rigght-arrow showing on the middle of the row, it kind of disapears. Personally I would rather have that (and the `Review not started` at the end)
Screen Shot 2022-02-18 at 3 17 08 PM

Space right below the Section title, could be half the high:
Screen Shot 2022-02-18 at 3 16 04 PM

nmadruga:
> Screen Shot 2022-02-18 at 3 35 45 PM
> Sure -- will just remove what you put there originally! 😜

Yikes! 🤣

PR #1124, opened 2022-02-18
Update to consider assignedSections: null

nmadruga:
Front-end to be used with Back-end [PR](https://github.com/openmsupply/application-manager-server/pull/660)

### Changes
- Update `unassignReviewAssignment` to also remove the reviewQuestion assignment connect to the Assignment.id
- Stopping checking if assigned per `reviewQuestionsAssignments` and check per `assignedSections` instead
- Regenerate graphql.ts to be using same back-end schema changes

#### Screenshots
UI isn't showing very nicely when seeing `all levels` with assignments available on each. Maybe we need a tweak there...
But the logic seems to be in place now:

After unassignment runs:
Screen Shot 2022-02-21 at 12 52 38 PM

After assigned the sections again to same user:
Screen Shot 2022-02-21 at 12 53 01 PM

### DIFF
--- a/src/containers/Review/assignment/useGetAssignmentOptions.ts
+++ b/src/containers/Review/assignment/useGetAssignmentOptions.ts
@@ -40,12 +40,10 @@ const useGetAssignmentOptions = () => {
       ({ isCurrentUserAssigner, isSelfAssignable }) => isCurrentUserAssigner || isSelfAssignable
     )
 
-    const currentlyAssigned = assignments.find((assignment) =>
-      assignment.current.assignmentStatus === ReviewAssignmentStatus.Assigned &&
-      // Restriction only apply to level one that isn't lastLevel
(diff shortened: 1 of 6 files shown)


CarlosNZ:
I think this PR demonstrates the point I was making [here](https://github.com/openmsupply/application-manager-server/pull/660#pullrequestreview-886875670) Look how much extra work you're having to do just to handle the null case!

nmadruga:
> I think this PR demonstrates the point I was making [here](https://github.com/openmsupply/application-manager-server/pull/660#pullrequestreview-886875670) Look how much extra work you're having to do just to handle the null case!

Yes. Definitely! All reverted now. Thanks

CarlosNZ (approved):
LGTM! :) Let's get these two merged

CarlosNZ:
> Maybe we need a tweak there...

That overlap *should't* be happening with my CSS tweaks in. Have you pulled in the latest develop into here?

PR #1127, opened 2022-02-23
Urgent Fix -- missing responses on review pages

CarlosNZ:
Oops -- can't believe we didn't notice this regression for so long. Looks like it slipped in when I was building the Tab structure. Send the smaller structure instead of fullStructure into the ReviewWrapper.

Luckily Adam picked it up this afternoon before tomorrow's demo:
photo_2022-02-23 19 57 06

Fortunately a really small fix.

@nmadruga I will do a new build tonight using this branch and update the demo. So you can review this PR in your own good time in case there's any other related things that I might have missed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "application-manager-web-app",
-  "version": "0.2.0-0",
+  "version": "0.2.0-1",
   "main": "index.tsx",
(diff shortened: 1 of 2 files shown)


nmadruga (approved):
LGTM

PR #1129, opened 2022-02-24
#1111 generic Toast hook

### LINKED ISSUE #1111: Generic "toast" solution
Screen Shot 2022-02-11 at 8 13 53 AM

The little "toast" Message we used in the Template builder is quite good, and we should probably use it more often throughout the app.

However, it needs to be simpler to implement. We should make a hook called "useToast" which returns a function that can be called whenever required.

```
const toast = useToast({props})

.... And then in place, just:
toast()
```

This is just a modified "Message" component, so the standard Message styles (success, error, etc) should be passed in as props.

Ideally it should slide in and out, but only if it's not too big a job.

CarlosNZ:
Fix #1111 

Not strictly on the agenda but I needed it for something else I was doing, so I just whipped this up in the weekend. 

Does pretty much what the issue says -- a simple, re-useable toast notification module that is easy to implement.

This PR only has the hook (and a small implementation in ActionConfig), so I've made another branch with a Demo component so you can try it out easily with all the variations.

### Demo branch:

`#1111-toast-hook-demo`

Go to [http://localhost:3000/toast-demo-auth](http://localhost:3000/toast-demo-auth) in that branch for a demo UI.

Screen Shot 2022-02-24 at 7 53 17 PM

Also -- if we're happy with this, do you think it could be useful to do something similar for Modals? They always seem a bit of a hassle to implement, would a generic "useModal" hook, with all the boilerplate abstracted away be helpful?


nmadruga (commented):
Nice one

PR #1131, opened 2022-02-27
UI change for evaluator fallback parameter

CarlosNZ:
Front-end complement to [back end PR #672](https://github.com/openmsupply/application-manager-server/pull/672)

Adds UI element to enable and define the Fallback for every evaluator instance.

Don't love the UI, but it's just re-using existing elements, so significant improvements not really do-able without a larger overhaul.

Make sure you `yarn install` to get the latest version of evaluator in this branch.

Note: nothing actually changed in `Evaluation.tsx`, seems like auto-formatting changes.

### DIFF
--- a/src/containers/TemplateBuilder/evaluatorGui/semanticComponentLibrary.tsx
+++ b/src/containers/TemplateBuilder/evaluatorGui/semanticComponentLibrary.tsx
@@ -32,13 +32,14 @@ const ComponentLibrary: ComponentLibraryType = {
 
     return (
       <div className="io-wrapper">
-        <div className="io-component key">{title}</div>
+        {title && <div className="io-component key">{title}</div>}
         <Input
           value={innerValue}
+          // type="number"
           className="io-component value"
           size="small"
           onChange={async (_, { value }) => {
-            if (!value.match(/^[\d]+$/)) return
+            if (!value.match(/^-?[\d]+$/)) return
             setInnerValue(value)
           }}
           onBlur={() => {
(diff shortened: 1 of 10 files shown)


src/containers/TemplateBuilder/evaluatorGui/semanticComponentLibrary.tsx
```diff
@@ -32,13 +32,14 @@ const ComponentLibrary: ComponentLibraryType = {
 
     return (
       <div className="io-wrapper">
-        <div className="io-component key">{title}</div>
+        {title && <div className="io-component key">{title}</div>}
         <Input
           value={innerValue}
+          // type="number"
```
nmadruga, line 38: TO be removed please

nmadruga (commented):
I trust your code, doesn't seem there is any clear issue with that, but I don't really know how to test this. I can see this is part of the Template Builder. But could you give me some guidance on how to test this work please?

CarlosNZ:
> I can see this is part of the Template Builder. But could you give me some guidance on how to test this work please?

In terms of "does it work", the back-end PR (https://github.com/openmsupply/application-manager-server/pull/672) has additional tests that show the "fallback" working as expected. 

In this PR, the only thing that needs to test is to check that the selected "fallback" values are actually making it into the query expression. Which you can do by either or all of:

- turn off "show as GUI" and check the JSON is correct
- check that the "fallback" value persists after exiting then re-entering the template editor
- run one of the queries using the Arrow icon in the template builder and confirm the fallback is working as expected
- check the database to see that the "fallback" is included in the saved query (this shouldn't be necessary if the previous three are all working correctly).

src/containers/TemplateBuilder/evaluatorGui/semanticComponentLibrary.tsx
CarlosNZ: Can we leave this one in? It strictly *should* be active to let the native UI enforce number entry correctly, but Andrei made so many modifications to the behaviour of these components that by turning it on it was causing other issues. But I'd like to leave it there as a reminder that it should be set to `type=number` for a NumberInput component.

Issue #1134, opened 2022-03-05
Fix useToast


CarlosNZ:
The problem with the current "useToast" hook is that it shows the toast notification relative to the parent container where it's called from. Most of the time this is okay, but when it's inside a small component, it can look quite ugly (and sometimes has z-index issues as well).

~~I think the solution is to use React Portals (https://reactjs.org/docs/portals.html), which would allow us to always attach the Toast component to the top-level "body" tag, regardless of where it's called from, so it's always positioned absolutely using the whole page as reference frame.~~ Context is better -- see below

CarlosNZ:
Other option (which might be simpler) is to create a ToastContext to store all toast state (with a "queue" for multiple toasts), and we inject the component somewhere at the top of the tree so it can easily be called from anywhere and sits outside the main HTML containers.

CarlosNZ:
Also, I noticed a small regression in the styling of the Message boxes throughout the app, I suspect it might have occured when this was first implement, so try and fix that in here too if possible.

PR #1135, opened 2022-03-05
#1132 Assignment fixes

### LINKED ISSUE #1132: Bug: Re-assignment shouldn't be allowed for changing 2 assignments at the same time
The problem started to happen after the changes were made to add a new **Submit** button on the Reassignment tab to make a single-mutation with all the sections re-assigned. The problem is that we only deal with unassigning from **one** review_assignment, and not many - although we allow this on the Front-end and the result is the bug described below:

This example uses the snapshot `dev/snapshots/__just_before_assignment` to make this easy to recreate. And the problem here has been mentioned on a [comment in the PR](https://github.com/openmsupply/application-manager-web-app/pull/1128#discussion_r818167355) for removing the reviewQuestionAssignments table and allowing for actually unassigning only sections (instead of the whole review) from a review_assignment.

In this example I have assigned section 1 to **Samanta** and Sections 2 and 3 to **Senior reviewer**:
Screen Shot 2022-03-03 at 5 09 08 PM

Now if we try to re-assign both sections 1 and 2 to **Admin** (note these are different review_assignments):
Screen Shot 2022-03-03 at 5 09 23 PM

The assignment doesn't show any problems... 
If you are logged as Admin you can see you are assigned to the section and Samanta has been correctly unassigned:
Screen Shot 2022-03-03 at 5 08 50 PM

Now the problem is, if you logiin as **Senior reviewer** it is **ALSO** showing as they are assigned to Sectioin 2!
Screen Shot 2022-03-03 at 5 08 37 PM

CarlosNZ:
Fix #1132

Much simpler now -- one hook with a couple of small helpers and one mutation to handle ALL assigning/reassigning/unassigning cases,  replacing two hooks and three mutations, and the logic is less complicated. 

Have tried many combinations of assigning, unassigning and re-assigning, including swapping, reassigning from multiple assignments at once, etc., and all seem to be working nicely.

Also tidied up the error handling a bit, since there was (for example), "setAssignmentError" statements that weren't actually being handled. Now they're all consolidated under one "assignment error". I don't love the UI when the error shows, but it's better than nothing (and hopefully something the user will rarely see). (I was going to use the "useToast" hook to display the error, but I realised there are [some problems with that that need fixing](https://github.com/openmsupply/application-manager-web-app/issues/1134) -- but once we've done that we can use Toast notifications to display errors.)

Also -- I've kept the unassignment behaviour the same (unassign _all_ sections from reviewer at once), but it can easily be the other way (one at a time, like assignment), and have left some commented-out code that can easily be swapped out to do it the other way. What do we think is best? (Ideally, it would be user-selectable, but I couldn't really think of a nice way to handle it in the UI without it becoming messy and cluttered, so we can perhaps re-visit this at some point).


src/utils/hooks/useUpdateAssignment.tsx
```diff
@@ -0,0 +1,86 @@
...
+interface AssignmentPatch {
+  status: ReviewAssignmentStatus
+  assignedSections?: string[]
+}
+
+const createAssignmentPatch = (
+  assignment: AssignmentDetails,
+  assignedSections: SectionAssignee | null
+) => {
+  if (!assignedSections) return { status: ReviewAssignmentStatus.Available }
```
nmadruga, line 63: Is this one to unassign a complete review_assignment? And is it only setting the status to `Available` since the back-end will make sure to set the array to empty/null?

nmadruga (commented):
Seems very siimple to read. II liked it a lot. Now testing...

src/utils/hooks/useUpdateAssignment.tsx
CarlosNZ: This line is just a guard for the case when "assignedSections" is null. Originally I had it that assignedSections couldn't be `null`, but when I was doing the unassignment, I needed a way to unassign from all sections without having to add a bunch of additional logic. 

If you look at "AssignmentSectionRow", you'll see my commented out code that would unassign one section at a time. If we went that way, we wouldn't need this line here either.

nmadruga:
Running into 2 problems that might require a little investigation - to make sure it's not a regretion. But if possible (even if not a regression) fixed as part of this PR, since the work will be paused for further issues...

Problem one. After re-assigning and un-assigning for tests. Looking on the **Overview tab** shows that the re-assigner/unassigner is missing from the review_assignment probably:
Screen Shot 2022-03-07 at 12 22 50 PM

Problem two. What cause is not clear. It just happened after re-assigning, unassigning and re-assigning for tests. Not the user **Samanta** have this option: only **Not assigned** to choose from the dropdown. I'll try to replicate 🤭  
Screen Shot 2022-03-07 at 12 24 14 PM

nmadruga:
> I'll try to replicate 🤭

Ok - Following up, this happens un-related to trying to re-assign or unassign. If you just logs in as Samanata or Senior Reviewer at the start you can see the problem. Not sure why its showing the assignment to these users before someone is assigned.

CarlosNZ:
> Ok - Following up, this happens un-related to trying to re-assign or unassign. If you just logs in as Samanata or Senior Reviewer at the start you can see the problem. Not sure why its showing the assignment to these users before someone is assigned.

Isn't this what we expect? "Samantha" doesn't have assigning permissions for this template, so she shouldn't be able to assign anyone. Are you saying that the Dropdown menu shouldn't even be there? What should be, can you remember?

nmadruga:
> Are you saying that the Dropdown menu shouldn't even be there?

Yes, that's the problem. It should be showing anything to a user that can be a reviewer but isn't assigned.

CarlosNZ:
> Problem one. After re-assigning and un-assigning for tests. Looking on the **Overview tab** shows that the re-assigner/unassigner is missing from the review_assignment probably: Screen Shot 2022-03-07 at 12 22 50 PM

Weird that you are getting parsing errors in the Timeline, I can't replicate that at all. Are you using a fresh snapshot and checked there isn't stale records in the activity log?
Screen Shot 2022-03-07 at 12 52 04 PM

If you can replicate this, can you save a snapshot in the error state and send it to me?

CarlosNZ:
> Yes, that's the problem. It should be showing anything to a user that can be a reviewer but isn't assigned.

Hmm, if that's changed it probably changed while doing the review_question_assignment removal. Can you remember what logic/state that component uses to check whether it should show the Dropdown?

nmadruga:
> > Yes, that's the problem. It should be showing anything to a user that can be a reviewer but isn't assigned.
> 
> Hmm, if that's changed it probably changed while doing the review_question_assignment removal. Can you remember what logic/state that component uses to check whether it should show the Dropdown?

I'll have a look

CarlosNZ:
Okay, have added guard to not show Dropdown or Unassign link when user doesn't have permission: 625b57b5

Note, this now means that a user can't even unassign themselves if they don't have assigning permissions -- which I guess is okay?

nmadruga:
> Okay, have added guard to not show Dropdown or Unassign link when user doesn't have permission: [625b57b](https://github.com/openmsupply/application-manager-web-app/commit/625b57b58bd77388206e3453be5872ec8e00ad80)
> 
> Note, this now means that a user can't even unassign themselves if they don't have assigning permissions -- which I guess is okay?

Not sure... What do you meant by assigning permission? If they have been assigned to a permission on that template that is `selfAssignable` they have permission and should see the dropdown.

I'm noticing a poblem now after self-assigning the "selected" reviewer isn't showing (before I click Submit)

CarlosNZ:
> Not sure... What do you meant by assigning permission? If they have been assigned to a permission on that template that is `selfAssignable` they have permission and should see the dropdown.

Yes, that's right. But before we had it that anyone could *UN*assign themselves from _any_ assignment regardless of permissions.

> I'm noticing a poblem now after self-assigning the "selected" reviewer isn't showing (before I click Submit)

What do you mean, exactly? When self-assigning, it says "Yourself" as the selected option before you Submit:
Screen Shot 2022-03-07 at 2 49 49 PM
Is that not what you're seeing?

nmadruga:
> Is that not what you're seeing?

Exactly, I wasn't seen that! I'll try again later today : )

CarlosNZ:
> > Is that not what you're seeing?
> 
> Exactly, I wasn't seen that! I'll try again later today : )

Okay, the problem was that the "Not assigned" option was disappearing from the options list after making a selection, and my check was checking for more than one item in the list.

I've updated it so that "Not Assigned" is always in the options list, and changed the "isLastLevel" check to just look at the first assignment and not depend on who is selected: 6b8df28c

CarlosNZ:
Also -- I realised that the assigner_id wasn't being updated in review_assignments, which is what was causing some of the Timeline parsing errors.

I've updated it so that the assigner_id is set to whoever *last* changed the review_assignment. So if Supervisor A assigns three sections, then Supervisor B  comes along later and unassigns it, the "assigner_id" will be for Supervisor B. d8819c56

Even if Supervisor B re-assigns one section (and leaves the other two as is), the assigner_id will be updated with Supervisor B's id. I don't see any other option, since there's only one assigner_id field per review_assignment. (The activity log will keep track of changes though.)

@nmadruga does that sound okay? Any other potential problems with the assigner_id field?

nmadruga (approved):
It looks good to me now Carl. I still find one not-expected behaivour when the re-assign/unassign are showing up on the previous level assignment after you select another Reviewer to re-assign on the Consolidation. And things are a bit buggy if I try to actually re-assign the Reviewer of previous level as well. But I think we can just add another bug fix for the future - not a very common path anyway.

PR #1136, opened 2022-03-07
#1122 Fix bug with Consolidation not correctly setting LOQ to applicant

### LINKED ISSUE #1122: Bug: Replying to list of question from Consolidation isn't working properly
After submitting Section 1 with Changes required from Applicant the action show correctly that an update is required from the Applicant. It is correctly showing the action to **Update**:
Screen Shot 2022-03-04 at 10 45 26 AM

After clickig on the **Update** goes to:
Screen Shot 2022-02-18 at 1 31 06 PM

This is how it's showing on the **Consolidator** Review home page (See Section 1 has been declined):
Screen Shot 2022-03-04 at 10 46 00 AM

And on the Review page (for the **Consolidator**):
Screen Shot 2022-03-04 at 10 48 35 AM

nmadruga:
Fixes #1122

Very small bug, caused by using the mutation instead of our Hook - where we pass the especific field required to update the original review response review_response.is_visible_to_applicant in `action_update_review_visibility`.

The bug only happened during MassApprove, when clicking on the "Agree with all" button on the consolidation.

### DIFF
--- a/src/containers/Review/ReviewPage.tsx
+++ b/src/containers/Review/ReviewPage.tsx
@@ -274,7 +274,6 @@ const ApproveAllButton: React.FC<ApproveAllButtonProps> = ({
   page,
 }) => {
   const { strings } = useLanguageProvider()
-  // const [updateReviewResponse] = useUpdateReviewResponseMutation()
   const updateReviewResponse = useUpdateReviewResponse()
 
   const reviewResponses = page.state.map((element) => element.thisReviewLatestResponse)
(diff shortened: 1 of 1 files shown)


CarlosNZ (approved):
LGTM!

Thanks for finding this. Did you say there was another one that was using the mutation directly?

nmadruga:
> LGTM!
> 
> Thanks for finding this. Did you say there was another one that was using the mutation directly?

Yes -- I though there was something called by `ApplicationView` but Ii can't find now when I do a global search for `useUpdateReviewResponseMutation` which was the one called by mistake

PR #1147, opened 2022-04-11
#1146 File upload template builder fix

### LINKED ISSUE #1146: Problem uploading files in template builder dummy applications
This is more of an annoyance than a bug, since technically it's behaving as it's supposed to.

The problem: in the template builder, when filling out the "config" application, the file upload plugin fails. 

Why: The "serial" parameter of the upload query is undefined, which causes a database error. This is because the fileUpload plugin gets the serial from the URL using "useRouter", but in the template builder the application serial is not part of the URL.

Possible solution: Might be able to get the "serial" from applicationData instead in the plugin. Need to check that this doesn't cause other problems, but I think it should be fine.

CarlosNZ:
Fix #1146 

Very simple -- just use "serial" from applicationData rather than URL parameters. No idea why I didn't do it like this the first time. 🙄


nmadruga (approved):
LGTM

PR #1148, opened 2022-04-11
#1143 Trigger waiting bug

### LINKED ISSUE #1143: Bug: Not displaying the Submission page after submitting (Trigger problem)
Only happens sometimes -- when the application is re-queried before Actions finish and update the status.

CarlosNZ:
Fix #1143 (kinda)

Hmm, after all that time spent trying to deal with this, it looks like we can get away with a very simple fix. 🙄

This is still a bit hacky, as it's causing more re-renders than it should (as the "Submit" mutation is also causing a refresh of the structure), but at least it works as far as I can tell.

There is a back-end branch called `__delayed_actions` which you can use to test this against -- it puts a delay between each action so the trigger will be processing for a significant amount of time, therefore any "trigger waiting" errors will become apparent.

Have a think about whether this is a good enough solution to #1143, or whether we should leave that open to deal with more thoroughly at a later date.

### DIFF
--- a/src/containers/Application/ApplicationSummary.tsx
+++ b/src/containers/Application/ApplicationSummary.tsx
@@ -63,6 +63,7 @@ const ApplicationSummary: React.FC<ApplicationProps> = ({
           } else {
             try {
               await submit(fullStructure)
+              fullStructure.reload()
               push(`/application/${fullStructure?.info.serial}/submission`)
             } catch (e) {
               console.log(e)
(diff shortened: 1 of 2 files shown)


nmadruga (approved):
I couldn't find the other branch to test more... But so far this seems to fix the UserRegistration one. I'm happy for you to merge and we can include on the server

CarlosNZ:
Okay, merged. I think I'll keep the original issue open until we've seen this in action for a bit longer.

Issue #719, opened 2021-05-18
Bug: Problem when using yarn dev_pg_permissions script that doesn't correctly change to current user permissions when loading the list of applications


nmadruga:
Also - I stater noticing a bug. Not sure if it's related to this PR or not.

Steps to recreate:
1. Login as **Reviewer1**, click to see list of `ReviewTest` applications
2. Change to another user **Nicole** for example. Now you see this error:
<img width="796" alt="Screen Shot 2021-05-19 at 10 48 00 AM" src="https://user-images.githubusercontent.com/16461988/118733049-b3203a80-b88f-11eb-8577-58a1e5cf7232.png">
3. Change back to **Reviewer1** and the error - which shouldn't show is showing. Even after refreshing the page.

PR #1150, opened 2022-04-13
#1033 #1151 Lookup table fixes/improvements

### LINKED ISSUE #1033: Lookup tables problems
- [x] Should allow `_` characters in table name (or be more clear about how name is transformed)
- [x] Keep getting `Cannot read property 'value' of undefined` error (every 2nd time or so)
- [ ] Should interpret non-text fields as correct type (e.g. boolean or number)
- [x] Why is pluralising the table name entered -- this is inconsistent with other table naming, and confusing
- [x] Gets stuck importing large lists (> 2000, not sure exactly)
...

CarlosNZ:
Fix #1033 (mostly)

Requires back-end branch from [PR #719](https://github.com/openmsupply/application-manager-server/pull/719)

Addresses all the items listed in #1033, except:
>  Should interpret non-text fields as correct type (e.g. boolean or number)

Need to be a bit careful with that. Even though I've already written a function to detect the Postgres type based on the content, I noticed some lookup tables had a mix of types (code for example is sometimes a number and sometimes text), so if we assumed that the first row was representative of the whole column, we'd cause problems. So will need to leave that for now until we can think it through a bit better, but it's not high priority for now.

The main importing error is fixed in the back-end.

I'll annotate my code with some comments to point out specific things.

Edit: Also chucked in a quick fix for #1151 cos it's so tiny 93343458

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,6 +12,7 @@
   },
   "scripts": {
     "dev": "webpack-dev-server --env.env=dev",
+    "start": "yarn dev",
     "prebuild": "rimraf dist",
     "build": "cross-env NODE_ENV=production webpack -p --env.env=prod",
     "serve": "serve dist",
--- a/src/LookupTable/components/DownloadButton.tsx
+++ b/src/LookupTable/components/DownloadButton.tsx
@@ -3,12 +3,14 @@ import config from '../../config'
 import React, { Fragment, useState } from 'react'
 import { Popup, Button, Icon, Message } from 'semantic-ui-react'
 import { useLanguageProvider } from '../../contexts/Localisation'
+import { DateTime } from 'luxon'
 
 const DownloadButton = ({
   id,
   open: openPopup = false,
   popUpContent = '',
   content = '',
+  name = 'lookup_table',
   ...props
 }: any) => {
   const { strings } = useLanguageProvider()
@@ -47,7 +49,7 @@ const DownloadButton = ({
         const url = window.URL.createObjectURL(new Blob([response.data]))
         const link = document.createElement('a')
         link.href = url
-        link.setAttribute('download', 'file.csv')
+        link.setAttribute('download', `${name}_${DateTime.now().toISODate()}.csv`)
         document.body.appendChild(link)
         link.click()
       })
--- a/src/LookupTable/components/ImportCsvModal.tsx
+++ b/src/LookupTable/components/ImportCsvModal.tsx
@@ -14,7 +14,6 @@ import { LookUpTableImportCsvContext } from '../contexts'
 import config from '../../config'
 import axios from 'axios'
 import { useLanguageProvider } from '../../contexts/Localisation'
-import pluralize from 'pluralize'
 
 const ImportCsvModal: React.FC<any> = ({
   onImportSuccess,
@@ -50,9 +49,7 @@ const ImportCsvModal: React.FC<any> = ({
     let formData: any = new FormData()
     formData.append('file', file)
 
-    const pluralTableName = pluralize.isPlural(tableName) ? tableName : pluralize.plural(tableName)
-
-    if (!tableStructureID) formData.append('tableName', pluralTableName)
+    if (!tableStructureID) formData.append('tableName', tableName)
 
     const JWT = localStorage.getItem(config.localStorageJWTKey || '')
     const authHeader = JWT ? { Authorization: 'Bearer ' + JWT } : undefined
--- a/src/LookupTable/components/ImportCsvModal.tsx
+++ b/src/LookupTable/components/ImportCsvModal.tsx
@@ -56,9 +56,9 @@ const ImportCsvModal: React.FC<any> = ({
 
     await axios
       .post(
-        config.serverREST +
-          '/admin/lookup-table/import' +
-          (tableStructureID ? '/' + String(tableStructureID) : ''),
+        `${config.serverREST}/admin/lookup-table/import${
+          tableStructureID ? '/' + String(tableStructureID) : ''
+        }?tableName=${tableName}`,
         formData,
         {
           headers: {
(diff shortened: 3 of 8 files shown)


src/LookupTable/components/DownloadButton.tsx
```diff
@@ -47,7 +49,7 @@ const DownloadButton = ({
         const url = window.URL.createObjectURL(new Blob([response.data]))
         const link = document.createElement('a')
         link.href = url
-        link.setAttribute('download', 'file.csv')
+        link.setAttribute('download', `${name}_${DateTime.now().toISODate()}.csv`)
```
CarlosNZ, line 52: Downloaded file now has the table name and date in the name.

src/LookupTable/components/ImportCsvModal.tsx
```diff
@@ -50,18 +49,16 @@ const ImportCsvModal: React.FC<any> = ({
     let formData: any = new FormData()
     formData.append('file', file)
 
-    const pluralTableName = pluralize.isPlural(tableName) ? tableName : pluralize.plural(tableName)
-
-    if (!tableStructureID) formData.append('tableName', pluralTableName)
+    if (!tableStructureID) formData.append('tableName', tableName)
```
CarlosNZ, line 52: Got rid of all that plural nonsense -- tables are now saved as singular names, just like all the other tables in the database. However -- you *can* still name it a plural name if you want, just remember that Postgraphile expects singular named tables when generating the GraphQL schema.

src/LookupTable/components/ImportCsvModal.tsx
```diff
@@ -50,18 +49,16 @@ const ImportCsvModal: React.FC<any> = ({
...
     const authHeader = JWT ? { Authorization: 'Bearer ' + JWT } : undefined
 
     await axios
       .post(
-        config.serverREST +
-          '/admin/lookup-table/import' +
-          (tableStructureID ? '/' + String(tableStructureID) : ''),
+        `${config.serverREST}/admin/lookup-table/import${
+          tableStructureID ? '/' + String(tableStructureID) : ''
+        }?tableName=${tableName}`,
```
CarlosNZ, line 61: As mentioned in back-end PR, I'm passing the `tableName` as a query parameter since I couldn't extract it from `formData` with the changes to the file handling.

src/LookupTable/components/list/ListTable.tsx
```diff
@@ -82,10 +82,11 @@ const ListTable: React.FC<any> = ({
                     <DownloadButton
                       popUpContent={strings.LOOKUP_TABLE_DOWNLOAD.replace('%1', lookupTable.label)}
                       id={lookupTable.id}
+                      name={lookupTable.label}
                     />
                   </Button.Group>
                 </Table.Cell>
-                <Table.Cell icon="chevron down" collapsing />
+                {/* <Table.Cell icon="chevron down" collapsing /> */}
```
CarlosNZ, line 89: No idea what this was for. I'm assuming it was intended as a possible Accordion opener for some future implementation, so just hidden for now.

src/LookupTable/contexts/LookUpTableImportCsvContext.tsx
```diff
@@ -49,8 +48,8 @@ const LookUpTableImportCsvReducer = (
     case LookUpTableImportCsvActions.ImportCSV:
       return { ...state, file: action.payload }
     case LookUpTableImportCsvActions.SetTableName: {
-      const removeNumbers = action.payload.replace(/([0-9])/gm, '')
-      return { ...state, tableName: camelCase(removeNumbers) }
+      const restrictCharacters = action.payload.replace(/[^A-z_]/gm, '')
```
CarlosNZ, line 51: Name can now be any alphabetical character or `_`. So you can write names in camelCase or snake_case as you see fit -- the database table name will always be converted to snake_case, but the "label" will be exactly what you enter.

src/LookupTable/hooks/useGetSingleTableStructure.hook.tsx
```diff
@@ -28,7 +29,11 @@ const useGetTableStructure = (): LookupTableStructureType => {
 
   useEffect(() => {
     if (!loading && !error && data?.lookupTable) {
-      setStructure(data.lookupTable as any)
+      const lookupTable = data.lookupTable
+      const namePlural = pluralize.isPlural(lookupTable.name)
+        ? lookupTable.name
+        : pluralize.plural(lookupTable.name)
+      setStructure({ ...lookupTable, name: namePlural })
```
CarlosNZ, line 36: I think the reason why table names were converted to plural is because the GraphQL schema has a pluralised node for returning multiple entities of a type (as in `application` for a single one and `applications`). However, forcing the table name to be plural actually *restricted* the GraphQL schema, cos you couldn't easily differentiate between a single item and a list of items, and being able to search by a single item is probably more likely within application forms.

So this addition here just converts the table name to plural form for querying the "multiple" node for that entity (e.g `products`)

package.json
```diff
@@ -12,6 +12,7 @@
   },
   "scripts": {
     "dev": "webpack-dev-server --env.env=dev",
+    "start": "yarn dev",
```
CarlosNZ, line 15: Unrelated, but I keep getting mixed up between `yarn start` and `yarn dev` as different projects use different ones to start the dev app. So let's just allow both! 😄

src/containers/Outcomes/helpers.tsx
```diff
@@ -87,7 +87,7 @@ export const getElementDetails = (value: any, displayDefinition: DisplayDefiniti
   const isAlreadyElement = !!formatting?.elementTypePluginCode
   const elementTypePluginCode = isAlreadyElement ? formatting?.elementTypePluginCode : 'shortText'
   const elementParameters = isAlreadyElement
-    ? formatting?.elementParameters
+    ? formatting?.elementParameters ?? {}
```
CarlosNZ, line 90: Solves #1151

PR #1155, opened 2022-04-24
Fix User Area to be on top of other ui

nmadruga:
Tiny fix, hope it's good for other things that were also showing on top of UserArea...

Example of previous bug:
Resultados


CarlosNZ (approved):
Thanks

CarlosNZ:
Oops -- I just realised that this makes out "dev menu" (with the user quick switch) sit underneath the UserArea. I've just changed their z-index to 1001 (1 higher than this) and pushed straight to `develop`.

PR #1166, opened 2022-04-27
Fix again consolidation

nmadruga:
It was considering all questions that didn't have responses...

### DIFF
--- a/src/utils/helpers/structure/generateConsolidatorResponsesProgress.ts
+++ b/src/utils/helpers/structure/generateConsolidatorResponsesProgress.ts
@@ -56,8 +56,8 @@ const generatePageConsolidationProgress = (page: Page, assignedSections: string[
   if (!assignedSections.includes(page.sectionCode)) page.consolidationProgress = initial
   else {
     const totalReviewable = page.state.filter(
-      ({ isAssigned, thisReviewLatestResponse, element }) =>
-        (isAssigned || !!thisReviewLatestResponse) && element.isVisible
+      ({ isAssigned, isActiveReviewResponse, thisReviewLatestResponse, element }) => 
+        ((isAssigned && isActiveReviewResponse) || !!thisReviewLatestResponse) && element.isVisible
     )
 
     const totalActive = totalReviewable.filter(activeThisReview)
--- a/src/utils/helpers/structure/generateReviewSectionActions.ts
+++ b/src/utils/helpers/structure/generateReviewSectionActions.ts
@@ -51,11 +51,11 @@ const actionDefinitions: ActionDefinition[] = [
   {
     // Show "start" when is new review submitted to consolidation (even for partial reviews)
     action: ReviewAction.canStartReview,
-    checkMethod: ({ reviewStatus, reviewLevel, isSecondReview, isPendingReview }) =>
-      reviewStatus === ReviewStatus.Pending &&
+    checkMethod: ({ reviewAssignmentStatus, reviewLevel, isSecondReview, reviewStatus }) =>
+      reviewAssignmentStatus === ReviewAssignmentStatus.Assigned &&
       reviewLevel > 1 &&
       !isSecondReview &&
-      isPendingReview,
+      !reviewStatus,
   },
   {
     action: ReviewAction.canReReview,
@@ -131,10 +131,13 @@ const generateReviewSectionActions: GenerateSectionActions = ({
 
     const totalNewReviewable = section?.consolidationProgress?.totalNewReviewable
 
+    const reviewStatus = thisReview?.current.reviewStatus
+
     const isReviewable =
-      (totalReviewable || 0) > 0 &&
       assignmentStatus === ReviewAssignmentStatus.Assigned &&
-      assignedSections.includes(section.details.code)
+      assignedSections.includes(section.details.code) &&
+      // Only check totalReviewable if review already exists
+      (reviewStatus ? (totalReviewable || 0) > 0 : true)
 
     const isAssignedToCurrentUser =
       assignee?.id === currentUserId && (isReviewable || isFinalDecision)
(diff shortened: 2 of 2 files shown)


CarlosNZ (changes requested):
Found a problem with this.

Can reproduce with the Angola snapshot. Product registration, level2 stage2 (consolidation). Once the level 2 reviewer is assigned, I log in as that reviewer. The list and dashboard shows the application I can review, but when I go the review/assignment page, there is no "start" link. As soon as I switch the branch back to `develop` the link re-appears, so it's definitely something to do with this.

I'll have a look at it if I get a chance in the next hour or so, otherwise I'll have to remove the 2nd level review from ProdReg until we sort it out.

src/utils/helpers/structure/generateConsolidatorResponsesProgress.ts
```diff
@@ -56,8 +56,8 @@ const generatePageConsolidationProgress = (page: Page, assignedSections: string[
   if (!assignedSections.includes(page.sectionCode)) page.consolidationProgress = initial
   else {
     const totalReviewable = page.state.filter(
-      ({ isAssigned, thisReviewLatestResponse, element }) =>
-        (isAssigned || !!thisReviewLatestResponse) && element.isVisible
+      ({ isAssigned, isActiveReviewResponse, thisReviewLatestResponse, element }) => 
+        ((isAssigned && isActiveReviewResponse) || !!thisReviewLatestResponse) && element.isVisible
```
CarlosNZ, line 60: Oh yeah, I think I can see the problem -- you're using `isActiveReviewResponse` but on the Assignment page, there are no active review responses yet (cos it's not started).

2 options:
- fix this logic to not use `isActiveReviewResponse` but still fix the problem
- use a different method to determine reviewer action on Assignment page

Thoughts?

CarlosNZ: I've just realised that `assignmentGroupedLevel` that is passed to the assignment rows has this info in it:
```
{
  "2": [
    {
      "id": 26,
      "reviewer": {
        "__typename": "User",
        "id": 5,
        "firstName": "Jessica",
        "lastName": "Jago"
      },
      "level": 2,
      "current": {
        "stage": {
          "id": 153,
          "name": "Avaliação técnica",
          "number": 2,
          "colour": "#E17E48"
        },
        "assignmentStatus": "ASSIGNED",
        "timeStageCreated": "2022-04-28T16:21:43.735+12:00",
        "timeStatusUpdated": "2022-04-28T16:22:52.765333+12:00"
      },
      "isCurrentUserReviewer": true,
      "isCurrentUserAssigner": false,
      "isFinalDecision": false,
      "isLastLevel": true,
      "isSelfAssignable": true,
      "isLocked": false,
      "allowedSections": [],
      "assignedSections": [
        "M1",
        "M2",
        "M3",
        "M4"
      ],
      "review": null
    }
  ]
}
```

So we should be able to use ` "isCurrentUserReviewer": true,` to determine the action rather than relying on `generateConsolidatorResponsesProgress.ts`, do you think?

CarlosNZ: Also -- just wanted to point out that even without this change, the "Start" link only appears for sections that have questions answered:
Screen Shot 2022-04-29 at 11 03 18 AM

CarlosNZ:
Okay, @nmadruga -- I think I have a fix for this! d0ed8186

It seems to work for the case we're considering ("Start" action shows up on Assignment page AND can submit consolidation with unanswered  questions), but I'm concerned that it may break something else.

See my inline comments also.

I'm concerned with all this logic, that there is just so much potential for things to not work right, and one little fix can easily mess something else up. We need to have a think about how this how chain of convoluted logic can be simplified a lot -- something for another day, but good to start thinking about!

src/utils/helpers/structure/generateConsolidatorResponsesProgress.ts
CarlosNZ: This fix is working great now. Have fixed the other problem it caused -- see below.

src/utils/helpers/structure/generateReviewSectionActions.ts
```diff
@@ -51,11 +51,11 @@ const actionDefinitions: ActionDefinition[] = [
   {
     // Show "start" when is new review submitted to consolidation (even for partial reviews)
     action: ReviewAction.canStartReview,
-    checkMethod: ({ reviewStatus, reviewLevel, isSecondReview, isPendingReview }) =>
-      reviewStatus === ReviewStatus.Pending &&
+    checkMethod: ({ reviewAssignmentStatus, reviewLevel, isSecondReview, reviewStatus }) =>
+      reviewAssignmentStatus === ReviewAssignmentStatus.Assigned &&
       reviewLevel > 1 &&
       !isSecondReview &&
-      isPendingReview,
```
CarlosNZ, line 58: One thing I noticed is that `isPendingReview` is inconsistent between first review and consolidation. (It's present before starting for review, but not consolidation). This is the kind of thing that is causing us headaches!

src/utils/helpers/structure/generateReviewSectionActions.ts
```diff
@@ -131,10 +131,13 @@ const generateReviewSectionActions: GenerateSectionActions = ({
...
 
+    const reviewStatus = thisReview?.current.reviewStatus
+
     const isReviewable =
-      (totalReviewable || 0) > 0 &&
       assignmentStatus === ReviewAssignmentStatus.Assigned &&
-      assignedSections.includes(section.details.code)
+      assignedSections.includes(section.details.code) &&
+      // Only check totalReviewable if review already exists
+      (reviewStatus ? (totalReviewable || 0) > 0 : true)
```
CarlosNZ, line 140: We might get away with removing the `(totalReviewable || 0) > 0` check completely, but I kept it in just in case -- but only if the review exists

nmadruga:
Carl, I think the problem is still there unfortunately. I tested the snapshot present in instance 50008 - which I believe included this PR, and even though I can start the consolidation, it **cannot** be submitted - still the same problem as before. I'll show this in images, since it is a very difficult Bug and communicate has been a challenge with this one also:
Screen Shot 2022-04-29 at 12 45 11 PM

During the trip has been very hard to focus on development, I can only get 1 hour of focused time so, very difficult to find a proper solution to this. I didn't like the use of `isPendingReview` or `isReviewable` from the start, is too difficult to get this working properly considering each case of Consolidation with partial reviews or reviews being submitted, Re-review, etc...

I would love to have a simpler solution to that, and agree it's not feasible during this trip. But since this needs a fix I'll spend some time today looking for a fix that doesn't require much code changes, but at least allow the Consolidation to work.

nmadruga:
See fix on new PR #1168

PR #1170, opened 2022-05-02
Small tweak -- more sensible default minimum for number input element

CarlosNZ:
It was annoying me how I could press the stepper and have negative values for the shelf life inputs in Project Reg. Was about to edit the template to set `minValue: 0`, but then realised it would just be better to have `0` as the default minimum for number input. Configuration can set a lower minimum in the rare cases we want to allow negative input.

Tiny change (and update to docs)


nmadruga (approved):
👍  Good idea. Annoyed me as well

Issue #1172, opened 2022-05-02
Bug: Review level problems -- summary


CarlosNZ:
Issue to list all the current review level / assignment UI and behaviour problems -- add to this as we encounter them

- [ ] "Make Decision" option not working -- when toggled on, the review page has no "Submit" button appear at all, so can't do anything
- [ ] It's possible to start and finish a consolidation when only 1 section of previous level reviews has been submitted. Which then triggers advancement to next stage etc, even though most of the sections haven't even been reviewed yet!

CarlosNZ:
> * "Make Decision" option not working -- when toggled on, the review page has no "Submit" button appear at all, so can't do anything

Weirdly, I got the "Submit" area to show up once, no idea what was different. But it still didn't work, because it was saying 
"Not all sections have been reviewed" when trying to submit, even though none of the questions were reviewable.

PR #1176, opened 2022-05-03
1175 Update assignment logic to take level into account

### LINKED ISSUE #1175: Assignment to respect selected level
When level + 1 is assigned and `all levels` filter is on, assigned person is also assigned level 1 assignments.

Sorry could be duplicate, this was raised by @nmadruga:

From telegram:

```
Assignment showing twice once the ProdRegistration is on Stage2 Level 2 (Chefe da secção). It should only be showing the last one
```
Screen Shot 2022-05-04 at 12 34 09 AM

andreievg:
closes #1175 

Created new type `AssignedSectionsByLevel` which is `SectionAssignee` grouped by levels, this type is used in assignment logic.

Noticed a couple of things: 

* Should assigned level 1 be able to assign level above one ?
* Looks like we can approve consolidation when approving just one section (if other sections are not submitted by another reviewer yet)
* Review assignment of level one, that are not assigned, but when level one is fully assigned, are still AVAILABLE ? Is that ok ?

### DIFF
--- a/src/containers/Review/assignment/AssignmentRows.tsx
+++ b/src/containers/Review/assignment/AssignmentRows.tsx
@@ -51,14 +51,13 @@ const AssignmentRows: React.FC<AssignmentRowsProps> = ({
                 sectionCode={code}
                 reviewLevel={Number(level)}
                 structure={fullStructure}
-                assignedSectionsState={[
-                  assignedSectionsByLevel[level] || defaultAssignedSections,
-                  (assignedSections) =>
-                    setAssignedSectionsByLevel({
-                      ...assignedSectionsByLevel,
-                      [level]: assignedSections,
-                    }),
-                ]}
+                assignedSections={assignedSectionsByLevel[level] || defaultAssignedSections}
+                setAssignedSections={(assignedSections) =>
+                  setAssignedSectionsByLevel({
+                    ...assignedSectionsByLevel,
+                    [level]: assignedSections,
+                  })
+                }
                 setEnableSubmit={setEnableSubmit}
                 setAssignmentError={setAssignmentError}
               />
--- a/src/containers/Review/assignment/AssignmentRows.tsx
+++ b/src/containers/Review/assignment/AssignmentRows.tsx
@@ -2,6 +2,7 @@ import React, { SetStateAction } from 'react'
 import { Header, Segment } from 'semantic-ui-react'
 import { useReviewStructureState } from '../../../contexts/ReviewStructuresState'
 import {
+  AssignedSectionsByLevel,
   AssignmentDetails,
   FullStructure,
   LevelAssignments,
@@ -14,8 +15,8 @@ interface AssignmentRowsProps {
   fullStructure: FullStructure
   assignmentInPreviousStage: AssignmentDetails
   assignmentGroupedLevel: LevelAssignments
-  assignedSections: SectionAssignee
-  setAssignedSections: React.Dispatch<SetStateAction<SectionAssignee>>
+  assignedSectionsByLevel: AssignedSectionsByLevel
+  setAssignedSectionsByLevel: React.Dispatch<SetStateAction<AssignedSectionsByLevel>>
   setEnableSubmit: React.Dispatch<SetStateAction<boolean>>
   setAssignmentError: React.Dispatch<SetStateAction<string | null>>
 }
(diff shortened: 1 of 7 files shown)


src/utils/types.ts
```diff
@@ -752,15 +757,15 @@ export type FilterTypeMethod = (filterKey: string, options?: FilterTypeOptions)
 
 export type FilterTypeDefinitions = {
   [filterType in
-    | 'number'
```
andreievg, line 755: Something is up with my formatter, sorry

src/utils/hooks/useUpdateAssignment.tsx
```diff
@@ -25,23 +26,25 @@ const useUpdateAssignment = ({ fullStructure }: { fullStructure: FullStructure }
...
-              variables: {
-                assignmentId: assignment.id,
-                assignmentPatch,
-              },
-            })
-          )
-        }
-      })
+      assignmentsFiltered
+        .filter(({ level: assignmentLevel }) => assignmentLevel === level)
```
andreievg, line 30: only added filter line

src/containers/Review/assignment/AssignmentTab.tsx
```diff
@@ -28,14 +29,8 @@ const AssignmentTab: React.FC<{
     userState: { currentUser },
   } = useUserState()
   const [enableSubmit, setEnableSubmit] = useState<boolean>(false)
-  const [assignedSections, setAssignedSections] = useState<SectionAssignee>(
```
andreievg, line 31: this is relocated to AssignmentRows

src/containers/Review/assignment/AssignmentRows.tsx
```diff
@@ -14,22 +15,30 @@ interface AssignmentRowsProps {
...
-  assignedSections,
-  setAssignedSections,
+  assignedSectionsByLevel,
+  setAssignedSectionsByLevel,
   setEnableSubmit,
   setAssignmentError,
 }) => {
   const { reviewStructuresState } = useReviewStructureState()
 
+  const defaultAssignedSections = Object.values(fullStructure.sections).reduce(
```
andreievg, line 34: Relocated from AssignmentTab

src/utils/types.ts
CarlosNZ: I re-auto-formatted that file for you: 0d703ed3

CarlosNZ:
> * Looks like we can approve consolidation when approving just one section (if other sections are not submitted by another reviewer yet)

Yeah, that's pretty bad -- I'd noted it here already: https://github.com/openmsupply/application-manager-web-app/issues/1172#issue-1223446072

> * Should assigned level 1 be able to assign level above one ?

Not sure exactly what you're asking. If you mean, should as assigner be able to assign a higher level than what is currently in progress, then yes -- that's something that Jess was wanting. So a supervisor could go through and do all the assignments at once for the whole stage. Not sure what's actually implemented now though -- if it's possible I think it's more by accident 😊

CarlosNZ (commented):
Great, thanks. Looks pretty good I think, but we should probably get @nmadruga to test it out before we merge it, since I'm not 100% clear the conditions it needs to be tested in.

src/utils/types.ts
andreievg: thank you

nmadruga:
> * Should assigned level 1 be able to assign level above one ?

Only if this person has `ASSIGN` permissions - which is valid for the whole stage.

nmadruga:
> > * Looks like we can approve consolidation when approving just one section (if other sections are not submitted by another reviewer yet)
> 
> Yeah, that's pretty bad -- I'd noted it here already: [#1172 (comment)](https://github.com/openmsupply/application-manager-web-app/issues/1172#issue-1223446072)

I agree. This shouldn't be happening. Also have a linked issue on our board: https://github.com/openmsupply/application-manager-web-app/issues/1180

nmadruga:
> * Review assignment of level one, that are not assigned, but when level one is fully assigned, are still AVAILABLE ? Is that ok ?

This was changed to simplify the code on re-assignment. Now the AVAILABLE doesn't mean it's possible to assign, but only means it is someone that is assignable for that section - in case no one else is already assigned. Which means, when the assigner clicks on re-assign the system will list all the people AVAILABLE.

nmadruga (approved):
Thanks! I can see the same code is kept, but fixed and moved the logic to the right places. Which is good - I have tested using the app, so good to me. 

One question: When submitting the reviewAssignments, if you are re-assigning for one level for example, it would still send the updated for for each level. This doesn't seem required, but I don't think it's a problem.

PR #1179, opened 2022-05-04
Subfolder parameter for fileUpload

CarlosNZ:
This is part of a request of Adam's. The "subfolder" can already be specified in the `/upload` endpoint -- this just makes it available to the front-end plugin.

It should rarely be needed, but it will be useful for configurations when we are trying to specify where certain files should go via the UI, such as carbone templates.


nmadruga (approved):
LGTM

PR #1186, opened 2022-05-05
#1185 delete template snapshots after loading

### LINKED ISSUE #1185: Importing template should delete snapshot files after loading
When we import a template using the "Templates" page "import" button, the uploaded template snapshot is kept in the "snapshots" folder, so whenever we go the "/admin/snapshots" page, it's cluttered up with a lot of useless template-only snapshots. 

The template builder "import" function should delete the uploaded template snapshot once it has successfully loaded it.

The "snapshots" folder (and therefore the /snapshots page) should only contain snapshots we've explicitly uploaded using the snapshots UI (not the Template builder)

CarlosNZ:
Closes #1185 

Ah, no more clutter (well, less anyway) on the snapshots page after importing loads of templates!

### DIFF
--- a/src/containers/TemplateBuilder/shared/OperationContextHelpers.tsx
+++ b/src/containers/TemplateBuilder/shared/OperationContextHelpers.tsx
@@ -26,6 +26,7 @@ const snapshotsBaseUrl = `${config.serverREST}/admin/snapshot`
 const takeSnapshotUrl = `${snapshotsBaseUrl}/take`
 export const snapshotFilesUrl = `${snapshotsBaseUrl}/files`
 const useSnapshotUrl = `${snapshotsBaseUrl}/use`
+const deleteSnapshotUrl = `${snapshotsBaseUrl}/delete`
 const templateExportOptionName = 'templateExport'
 const uploadSnapshotUrl = `${snapshotsBaseUrl}/upload`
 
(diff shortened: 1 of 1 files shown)


src/containers/TemplateBuilder/shared/OperationContextHelpers.tsx
```diff
@@ -238,11 +239,16 @@ export const importTemplate: ImportTemplateHelper =
...
 
-      return await safeFetch(
+      const snapshotResult = await safeFetch(
         `${useSnapshotUrl}?name=${snapshotName}&optionsName=${templateExportOptionName}`,
         '{}',
         setErrorAndLoadingState
       )
+
+      // Delete the snapshot cos we don't want snapshots page cluttered with individual templates
+      safeFetch(`${deleteSnapshotUrl}?name=${snapshotName}`, {}, () => {})
```
CarlosNZ, line 249: Don't need to await this, since the result doesn't affect anything on the current page.

nmadruga:
Carl - I'll also leave this one for a review tomorrow. Ran out of time today...

I just want to understand this first before reviewing... Seems a bit risky business

CarlosNZ:
> I just want to understand this first before reviewing... Seems a bit risky business

Cool -- check out the original issue for an explicit explanation.

Also -- I just realised after I merged it that #1188 was based from this. So you'll see the changes from there in this PR too now. Sorry about that.

nmadruga (approved):
LGTM

PR #1189, opened 2022-05-08
#1134 useToast improvements

### LINKED ISSUE #1134: Fix useToast

CarlosNZ:
Fix #1134 

See demo at http://localhost:3000/dev/toast-demo (also proper use-case implementation in #1190)

Basically works as it did previously (with hook to instantiate), but with these main changes:
- Global context rather than separate instances of hook, so there is only ever one set of notifications for whole app
- This allows us to keep the actual UI component persistent and outside the main UI container, so we don't have any of the positioning problems. Toast container is a fixed-posisition (so has no impact on  main app containers) full-height, full-width empty container which spawns absolutely-positioned children for individual toast blocks.
- Simplified offset. The fine-tune offset values were only to deal with the annoying positioning issues in some places. But now this is solved we can simplify positioning to just the main six general positions (top-left, top-middle, etc...)
- Added usage documentation

Further explanations in inline comments


src/contexts/Toast/ToastProvider.tsx
```diff
@@ -0,0 +1,115 @@
...
+            {toasts
+              .filter((toast) => toast.position === Position[key as keyof typeof Position])
+              .map((toast) => (
+                <Toast toast={toast} removeToast={() => removeToast(toast.uid)} key={toast.uid} />
+              ))}
+          </div>
+        ))}
+      </div>
+      {children}
+    </ToastProviderContext.Provider>
```
CarlosNZ, line 64: `toast-container` is the fixed-position full-size container
`block ${positionClass}` are the six blocks, one of each position (top-left, etc.), positioned absolutely relative to the toast container
`<Toast />` -- the individual toasts, positioned relative to their respective blocks
`{children}` -- the rest of the app

src/contexts/Toast/ToastProvider.tsx
```diff
@@ -0,0 +1,115 @@
...
+  return showToast
+}
+
+export const Toast = ({ toast, removeToast }: { toast: ToastProps; removeToast: () => void }) => {
+  const [visible, setVisible] = useState(false)
+  const timerId = useRef<number>(0)
+  const { timeout } = toast
+
+  useEffect(() => {
+    setTimeout(() => setVisible(true), 50)
```
nmadruga, line 83: I don't undertand what is this `setTimeout` this part seems very confusing and hard to follow what the code here is doing. I can undertand that's how the Toast will be removed after the transition_duration ends, but why do you need to call setTimeout here that many times?

nmadruga (changes requested):
Ok - I quickly run the test now (this is taking me some time to review, it is a new feature and very good to have Thanks, Carl! Although at the same time I fell that I can't take to long)

Anyway my only thing to add here is about positioning. Should we be showing a toast on top of the Top area? Or on top of the footer. Check out the results that I get when I'm using this options"

src/contexts/Toast/ToastProvider.tsx
CarlosNZ: The first timeout is for the `<Transition>` component - the timeout makes the Toast start invisible, then become visible, which is how the "Transition" component can apply the nice "fade-in" effect. If we start with it visible, you don't get the effect.

(And the 2nd timeout is obviously to close the Toast)

I can see how that's not obvious though, so have added a comment above the first one: 7ab8c456

CarlosNZ:
> Anyway my only thing to add here is about positioning. Should we be showing a toast on top of the Top area? Or on top of the footer. Check out the results that I get when I'm using this options:

Yeah, personally I think it's fine. The notifications are meant to temporarily sit "on top" of everything else, and the small drop-shadow makes it clear it's "above" the content. They're intended as a temporary alert, and they're dismissable, so there's no problem with obscuring important content in any significant way.

That seems to be a pretty standard convention, as seen here in open msupply, where it obscures the bottom menu and info-bar temporarily:
Screen Shot 2022-05-24 at 2 21 44 PM

(Admittedly, those ones are a bit prettier than these, but we don't get them for free with Semantic UI -- we're just repurposing the Message box lol)

nmadruga (approved):
Cool! Not problem to keep it at the top of other components then, and thanks for adding that comment. I think it will be helpful in the future.

PR #749, opened 2021-05-28
#704 fix application title o application and review homepages

nmadruga:
Fixes #704 

### Changes

- Show application title and link to the summary page from `ReviewHome`
- Update ApplicationHomeWrapper to display title and subtitle dependent on which page is wrapping, the create page will show the same title and subtitle as before and the update page (`ApplicationHome`) will show title and subtitle as defined in Invision:
Screen Shot 2021-05-28 at 10 01 13 PM

- Also fixes the large title that was showing on latest screencast to Debbie.

Issue #1075, opened 2021-12-10
Outcome configuration UI


CarlosNZ:
# Step 0 -- rename/consolidate outcomes and lookup tables

"Outcomes" is poorly named and should be changed ASAP. I think we agreed on "data" or "database", but can continue to discuss.

I suggest that lookup table and outcome table names should share a common prefix, as there's not really a reason to distinguish between them. So they can all become "data_<name>". Also, we should register new outcome "data" tables in the (currently) "lookup_table" table similar to lookup tables so the system can have a better way to keep track of its data.

So that would be:
- rename "lookup_table" to "data_table"
- update the "modifyRecord" action -> "createTable" method to also register the table in "data_table" on creation (and also: add a field to "data_table" called "is_lookup_table", so we have a way to keep track of which is which -- the only real difference would be that lookup tables can be exported/imported using the Lookup table UI, whereas "normal" data tables can't be)
- write migration code (back-end SQL?) to change existing snapshots when first launched
- add the prefix "data_" to all new tables created with modifyRecord or lookup table import

# Step 0.5 Fix lookup tables

They're currently quite broken. Also, they don't handle types other than strings. I actually think we can use the modifyRecord action as the data writing module, since it's already built to handle table creation, type detection and data entry. Presumably it will be fine to use it it iterate over hundreds of records when inserting. I think this will be easier and more useful ultimately than trying to fix the existing import routine.

# Step 1 -- link Data tables to Templates

- Add  "template_id"  field to "outcome_display" table (should be optional as outcomes shouldn't *have* to be linked to templates, and in many cases won't be.
- Update "outcomes" endpoints to filter outcomes to only return the ones that are connected to template versions that are "active", or have no template_id
- Export/import linked outcome records with template import/export

# Step 2 -- UI

It occurs to me that the Outcome configurations themselves can be just displayed as "Outcomes" themselves -- so the "display" side of this is already done! (~~I'll quickly whip up a demo and add it to core templates~~ Demo available in snapshot: `.../application-manager-templates/dev/snapshots/core_templates_outcomes_demo`)

As far as the editing side of things goes, the [proposal for editing lookup tables](https://github.com/openmsupply/application-manager-server/issues/629) pretty much applies here too. So if we do that work, then we've pretty much got outcome configuration for free. There'll be some specific requirements for how to edit some of the more fiddly columns, but in essence it should basically work the same.

[...]

PR #1191, opened 2022-05-09
1075 Part 1: Rename outcomes

CarlosNZ:
Front-end complement for [back-end PR #749](https://github.com/openmsupply/application-manager-server/pull/749)

Just renaming, no logic changes.


src/utils/defaultLanguageStrings.ts
```diff
@@ -257,7 +257,7 @@ export default {
   PAGE_TITLE_REVIEW: 'Review %1',
   PAGE_TITLE_TEMPLATES: 'Templates',
   PAGE_TITLE_LOOKUP_TABLES: 'Lookup Tables',
-  PAGE_TITLE_OUTCOMES: 'Outcomes',
+  PAGE_TITLE_DATA_DISPLAY: 'Data Display',
```
nmadruga, line 260: I have done the changes on `core_templates` Pt strings, in case we use that in the future to base another Pt speaking country. 🤭

nmadruga (approved):
The renaming is a good thing, it has been annoyance for a while now. 

I have mentioned on the back-end before (and also on issue #1075 - I think the table and endpoint should be `data-views`) but I can see now that the endpoint is actually `data` so that looks good to me. All approved...

Just a (hopefully) tiny thing to suggest:
Screen Shot 2022-05-24 at 11 58 31 AM

Also since you are on it. Can this layout be improved a little bit?

And I noticed that the headers (in `components/Admin.tsx` aren't localised).

CarlosNZ:
> If you are on it. Can this layout be improved a little bit? And the headers (in `components/Admin.tsx`) aren't localised.

Yikes, I didn't even know that page existed! It's not a big deal, as you can't get to it from the UI at all, but on the off-chance someone goes to it directly, I've done a quick tidy up and localisation.  ec7a990c

nmadruga:
Much better! 👏 

Screen Shot 2022-05-25 at 4 27 35 PM

PR #1192, opened 2022-05-09
1075 Part 2: Migrate lookup tables

CarlosNZ:
Requires [back-end PR #750](https://github.com/openmsupply/application-manager-server/pull/750)

Updates lookup table UI to comply with table naming standards. It also allows more flexibility in the entered name, as whatever is entered will be converted to a singular snake_case table name.

See back-end notes and inline comments for more details.


src/LookupTable/contexts/LookUpTableImportCsvContext.tsx
```diff
@@ -48,8 +49,7 @@ const LookUpTableImportCsvReducer = (
     case LookUpTableImportCsvActions.ImportCSV:
       return { ...state, file: action.payload }
     case LookUpTableImportCsvActions.SetTableName: {
-      const restrictCharacters = action.payload.replace(/[^A-z_]/gm, '')
-      return { ...state, tableName: restrictCharacters }
+      return { ...state, tableName: action.payload }
```
CarlosNZ, line 52: This restriction not necessary. The "name" can be anything and it will automatically be converted to a singular snake_case name for the table by the back-end.

src/utils/graphql/queries/getAllLookupTableStructures.query.ts
```diff
@@ -2,11 +2,11 @@ import { gql } from '@apollo/client'
 
 export default gql`
   query getAllLookupTableStructures {
-    lookupTables {
+    dataTables(condition: { isLookupTable: true }) {
```
CarlosNZ, line 5: This is how we distinguish "lookup tables" from other custom data tables. (since we don't want other data tables to be editable, at least not here)

nmadruga (commented):
Just confirming. Which this work after we run modifyRecord action we would be adding another `data_table_<new>` which isn't listed in the Lookup tables UI since it's set without the flag?

CarlosNZ:
> Which this work after we run modifyRecord action we would be adding another `data_table_<new>` which isn't listed in the Lookup tables UI since it's set without the flag?

When you run modifyRecord and create a new table, the `is_lookup_table` flag *won't* be set -- is that what you're asking?

I was toying with the idea of allowing `isLookupTable` as an optional parameter on modifyRecord, but i dont' think that's a good idea since lookup tables have such a restriction on how they're structured (must have id column etc.) and we don't want to have to enforce all that in modifyRecord (not yet anyway)

nmadruga (approved):

Issue #1204, opened 2022-05-15
Joined fields for uploading document (Checklist + Upload button)


nmadruga:

CarlosNZ:
This will be solved once we make fields configurable as reviewable or non-reviewable. #1199 The first field (with the checkboxes) can just be marked "non-reviewable".

nmadruga:
- The original idea is to use a wrapper to be used in order to combine 2 existing plugins. Cofnigration for a wrapper might be difficult, also considering that the final result should be a combination of both (something that is setup per Wrapper).
- Option to use the wrapper should show as another plugin, but the parameters would be added for both
- This has been a very arguable solution, since it could be achievable by setting the checklists as non-reviewable 
- To do this using only configuration would add another step to configuration, although when using the search would help to repeat the 2-fields for next questions...

nmadruga:
Why not to do a specific plugin instead?
- Because the list of plugins would be bigger
- It would be less work than doing the wrapper

nmadruga:
Any more complexity added to configuration should be reviewed. The point is to make this **simpler** to configure and review

nmadruga:
We considered that for just a plugin it would take 2 days. The 4 days solution is considering a more generic way to use wrappers

PR #1222, opened 2022-06-02
Small change on the upload file plugin

nmadruga:
No issue related...

Done as a minor change to improve what has been requested (internally) on #1204 

Since we are using 2 fields for one type of question in the latest template being build (no naming since this is public), basically the changes are to instead of showing both like this:
Screen Shot 2022-06-03 at 10 14 44 AM

Moving the description to the bottom. So it displays like this:
Screen Shot 2022-06-03 at 10 00 07 AM

### DIFF
--- a/src/formElementPlugins/fileUpload/src/ApplicationView.tsx
+++ b/src/formElementPlugins/fileUpload/src/ApplicationView.tsx
@@ -153,7 +153,6 @@ const ApplicationView: React.FC<ApplicationViewProps> = ({
           <Markdown text={label} semanticComponent="noParagraph" />
         </label>
       )}
-      <Markdown text={description} />
       <Segment.Group>
         {/* Dummy input button required, as Semantic Button can't
         handle file input. Link between this input and Semantic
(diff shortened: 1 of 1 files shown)


CarlosNZ:
But this makes it worse and inconsistent when its used on its own. Every plugin goes Label, Description, Content, so it's just confusing and messy if we have one plugin that doesn't follow the convention.

I really don't think it's a problem having it where it is, even in this use case. However, if you really think it needs to change we could do one of these options:

- Just put the description in the Checkbox plugin so it's at the top of the whole block (config only)
- Add an additional "TextInfo" element below with just the description in  it -- that should look exactly like your screenshot (also config only)
- I think either of those would be fine, but if you really need to change the plugin, please make it an optional parameter so it still displays normally the rest of the time by default.

CarlosNZ:
Closing -- this can now be replicated by just setting the "Checklist" element to NEVER reviewable, as per: https://github.com/openmsupply/conforma-server/pull/783

PR #1223, opened 2022-06-29
#766B Logo customisation

CarlosNZ:
For back-end PR https://github.com/openmsupply/conforma-server/pull/767

**Quick summary**:
- Moved `usePrefs` into a Context so we only use one instance of it (since it's global state)
- Changed user area to use the brand logo defined in "preferences", or fallback to our usual if not defined
- Small style tweak to make the logo sit better in the menu area

To see, load the snapshot `angola_2022_06_29` (from templates repo)

It should look like this:
Screen Shot 2022-06-29 at 4 31 21 PM

Screen Shot 2022-06-29 at 4 37 41 PM


src/contexts/SystemPrefs.tsx
```diff
@@ -0,0 +1,55 @@
...
+      })
+      .catch((err) => {
+        setPrefsState({ loading: false, error: err })
+      })
+  }
+
+  return (
+    <SystemPrefsContext.Provider value={{ ...prefsState, refetchPrefs: fetchPrefs }}>
+      {children}
+    </SystemPrefsContext.Provider>
```
CarlosNZ, line 51: Exactly the same functionality as previous hook, just moved into context so we don't keep querying the server for the logo url every time the user area renders.

andreievg: looks good

src/containers/User/UserArea.tsx
```diff
@@ -182,17 +183,23 @@ const MainMenuBar: React.FC<MainMenuBarProps> = ({ templates, dataViews }) => {
 
 const BrandArea: React.FC = () => {
   const { strings } = useLanguageProvider()
```
andreievg, line 185: this is now unused

src/containers/User/UserArea.tsx
```diff
@@ -182,17 +183,23 @@ const MainMenuBar: React.FC<MainMenuBarProps> = ({ templates, dataViews }) => {
...
+  const { preferences } = usePrefs()
+
+  const logoUrl = preferences?.brandLogoOnDarkFileId
+    ? `${config.serverREST}/public/file?uid=${preferences.brandLogoOnDarkFileId}`
```
andreievg, line 189: `nit:` it kind of feels like the server should be giving the full url (minus domain), or maybe a helper to get the file by id, this is constructed twice in f/e

src/containers/User/UserArea.tsx
```diff
@@ -182,17 +183,23 @@ const MainMenuBar: React.FC<MainMenuBarProps> = ({ templates, dataViews }) => {
 
 const BrandArea: React.FC = () => {
   const { strings } = useLanguageProvider()
+  const { preferences } = usePrefs()
```
andreievg, line 186: hmm, this potentially will cause a flicker between defaultBrandLogo and logo in prefs on slower networks.  

probably ok, but would typically wait for preferences in usePrefs before rendering Image.

src/index.tsx
```diff
@@ -89,4 +90,9 @@ const App: React.FC = () => {
   )
 }
 
-ReactDOM.render(<App />, document.getElementById('root'))
+ReactDOM.render(
+  <SystemPrefsProvider>
```
andreievg, line 94: Hmm, I think it should be in the App

andreievg (commented):
Want to approve, but just holding of for your reply to comments

CarlosNZ:
> Want to approve, but just holding of for your reply to comments

Great, thanks for getting to these. I'll reply/fix this afternoon. :)

src/containers/User/UserArea.tsx
CarlosNZ: Yeah okay. I'd left in cos it was used by the commented-out code below, but I've just removed them both now, we don't need it anymore.

src/containers/User/UserArea.tsx
CarlosNZ: The reason I didn't provide the whole url in the server is that the "preferences.json" file that provides it is currently manually created. The whole prefs file gets imported/exported with snapshots, but the contents is just hand-made for now. (Until we make a UI for editing prefs). And so I figured it was better to keep the prefs content to just the essentials (i.e. fileID) rather than the whole url (which could in theory change, or more likely get mistyped -- trailing slash? No slash? -- that kind of thing).

Also, we've generally used hard-coded values for endpoints in the front-end throughout, so just kept with that pattern. Ideally we should have a global endpoint map, but that seems like something for another day as I'd want to replace all hard-coded endpoint urls.

src/containers/User/UserArea.tsx
CarlosNZ: In "index.tsx", the entire app doesn't get rendered until prefs have loaded (it's always been this way, since we need the language strings before we can render anything). So by the time it gets to here, preferences will be available from the global "Prefs" context.

src/index.tsx
CarlosNZ: It needs to be outside it, since "App" calls "usePrefs", which it can only do within the Provider.

(I did try a few variations of this, and this structure was the best I could come up with -- open to suggestions though. But the main requirement is that `usePrefs` must be called outside the Language provider, since the default language code needs to be passed to the LanguageProvider)

Ideally it would go:
```
<PrefsProvider>
  <ApolloProvider>
    <ToastProvider>
      <LanguageProvider>
```
But we need to make sure that "Prefs" is loaded before rendering all the rest, which I couldn't see how to do without using the `usePrefs` hook and waiting for `loading` to complete, which really needs to happen in "App" as far as I can tell.

src/containers/User/UserArea.tsx
CarlosNZ: 5ec7d65d

src/containers/User/UserArea.tsx
andreievg: Cool, thanks

src/containers/User/UserArea.tsx
andreievg: ohh true, oops, i though prefs are re-fetched here, but that's not the case.

src/index.tsx
andreievg: Ok, oops should have checked where userPrefs is used. Sounds like a change for later, but yeah if we usePrefs within LanguageProvider, it would work

andreievg (approved):
Looks good thanks

PR #1224, opened 2022-07-03
#1209 Preview documents UI

### LINKED ISSUE #1209: Configurable per Stage/Level/Decision new option and Template Builder changes
- Schema change
- Template Builder UI
- Need to configure a message that is based on the decision: Approve or Reject.
- Permissions tab have a way to add specific messages depending on the decision. We should have default messages to be shown in case: is last level for conform, Changes requested (to reviewer) list of question (to applicant) or non-conform and also when is not last level to explain it is going to the next level review.

CarlosNZ:
Fixes #1209 

Uses the endpoint created in https://github.com/openmsupply/conforma-server/pull/775, so will require that branch.

Screen Shot 2022-07-04 at 11 39 37 AM

To test: use snapshot `PreviewDocs_UI` in the same folder as the previous ones.

You can go straight to an almost-completed review at: http://localhost:3000/application/S-LVH-0004/review/8?active-sections=S1%2CS2%2CS3%2CS4 -- need to be logged in as **BasicReviewer**

There you can test the Previews for all three possible decisions: "Conform", "NonConform", "LOQ".


semantic/src/site/elements/loader.overrides
```diff
@@ -1,3 +1,13 @@
...
+
+.ui.dimmer .ui.loader {
+    color: @loaderTextColor;
+  }
+.ui.dimmer .ui.loader:before {
+    border: @loaderLineWidth solid @loaderFillColor;
+}
+.ui.dimmer .ui.loader:after {
+    border-color: @shapeBorderColor;
+}
```
CarlosNZ, line 13: I've no idea why the default Semantic Loaders had different colour scheme when launched within a dimmer (Modal), but this just resets it back to the same as the standard Loader.

semantic/src/site/modules/modal.variables
```diff
@@ -1,6 +1,3 @@
 /*******************************
          Site Overrides
 *******************************/
-
-// Prevent rounded corners of full-screen modal
```
CarlosNZ, line 5: This should only be applied to full-screen modals, so have moved to "modal.overrides" and applied to correct class

src/components/Review/DecisionPreview/PreviewItems.tsx
```diff
@@ -0,0 +1,94 @@
...
+  )
+}
+
+const FallbackPreview = ({ item }: { item: GenericPreviewData }) => {
+  return (
+    <div className="item fallback-preview">
+      <p>
+        {item.displayString}
+        <br />
+        <pre>{item.text}</pre>
```
CarlosNZ, line 71: Just shows a stringified view of the output JSON for now -- we can make additional preview types as we need them.

src/components/Review/DecisionPreview/ReviewPreviewModal.tsx
```diff
@@ -0,0 +1,94 @@
...
+  const { strings } = useLanguageProvider()
+  const [data, setData] = useState<ActionResultPreviewData[] | null>(null)
+  const [loading, setLoading] = useState(true)
+  const [error, setError] = useState(false)
+
+  useEffect(() => {
+    if (!open) {
+      setData(null)
+      setError(false)
+      return
```
CarlosNZ, line 34: Reset state when closing modal, otherwise it's still there next time you open it.

src/components/Review/DecisionPreview/ReviewPreviewModal.tsx
```diff
@@ -0,0 +1,94 @@
...
+    setLoading(true)
+    fetchPreviews(reviewId, previewData).then((result) => {
+      if (result.error) setError(result.error)
+      else setData(result.displayData)
+      setLoading(false)
+    })
+  }, [open])
+
+  return (
+    <Modal id="preview-modal" open={open} closeOnDimmerClick={false}>
```
CarlosNZ, line 45: `closeOnDimmerClick={false}` -- thought I'd make it a bit harder to accidentally close the modal, cos it generates new previews every time it's opened.

src/containers/TemplateBuilder/shared/TextIO.tsx
```diff
@@ -6,8 +6,8 @@ const iconLink = 'https://react.semantic-ui.com/elements/icon/'
 const colourLink = 'https://htmlcolorcodes.com/'
 
 type TextIOprops = {
-  text?: string
-  title?: string
+  text?: string | null
+  title?: string | null
```
CarlosNZ, line 10: Was causing problems setting the "code" to empty string, as I've added database constraint to make the code unique per template, and this was making them all empty strings if not specified, which broke the constraint. 

So now they can remain null if they start as null

CarlosNZ:
Just going to add some screenshots here for @jagoje and @adamdewey to comment on.

Bottom of review page:
Screen Shot 2022-07-05 at 10 11 38 AM

When you click the "Preview" button:
Screen Shot 2022-07-05 at 10 12 47 AM
Then...
Screen Shot 2022-07-05 at 10 11 51 AM
Document links open PDF in new tab.
Email notifications are expandable:
Screen Shot 2022-07-05 at 10 12 01 AM

And if you click "Preview" with "Non-conform":
Screen Shot 2022-07-05 at 10 12 31 AM

And with "List of Questions:
Screen Shot 2022-07-05 at 10 12 37 AM

adamdewey:
Really cool! 👏

adamdewey:
That first image shows 'List of Questions' instead of 'Conform' - is that just a mashup of screenshots?

CarlosNZ:
> That first image shows 'List of Questions' instead of 'Conform' - is that just a mashup of screenshots?

Yes, capturing the "Fetching" in action was tricky, so I just used the first one I got ;)

src/containers/TemplateBuilder/shared/TextIO.tsx
CarlosNZ: Had to revisit this, as it caused another bug #1243. 

I've made changes now to allow the text values to be `null` throughout, and where it *can't* be, it falls back to an empty string. Not super happy with it, but best I could do without a significant re-factor.

a9802b10

src/components/Review/DecisionPreview/ReviewPreviewModal.tsx
```diff
@@ -0,0 +1,94 @@
...
+    </Modal>
+  )
+}
+
+export default ReviewPreviewModal
+
+const fetchPreviews = async (reviewId: number, applicationDataOverride: { [key: string]: any }) => {
+  const JWT = localStorage.getItem(config.localStorageJWTKey)
+  try {
+    const result = await postRequest({
```
andreievg, line 85: i think postRequest already sets authorization (and json type for body)

src/components/Review/DecisionPreview/ReviewPreviewModal.tsx
```diff
@@ -0,0 +1,94 @@
...
+        {error && (
+          <Message
+            error
+            icon="warning sign"
+            header={strings.REVIEW_DECISION_PREVIEW_ERROR_HEADER}
+            content={strings.REVIEW_DECISION_PREVIEW_ERROR_TEXT}
+          />
+        )}
+        <div id="preview-items">
+          {data &&
```
andreievg, line 63: stylistic preference again again `{data?.map(() => ) || (<Message info header= strings.REVIEW_DECISION_NO_PREVIEWS_AVAILABLE} />)`

src/components/Review/DecisionPreview/ReviewPreviewModal.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+  useEffect(() => {
+    if (!open) {
+      setData(null)
+      setError(false)
+      return
+    }
+    setLoading(true)
+    fetchPreviews(reviewId, applicationDataOverride).then((result) => {
+      if (result.error) setError(result.error)
```
andreievg, line 38: if (!open) return

otherwise a small glitch if you close before fetch is loaded and then open again

src/components/Review/DecisionPreview/ReviewPreviewModal.tsx
andreievg: Hmm, can you add a comment about it in code please

src/containers/Review/ReviewSubmit.tsx
```diff
@@ -23,19 +24,21 @@ const ReviewSubmit: React.FC<ReviewSubmitProps> = (props) => {
...
 
   return (
     <Form id="review-submit-area">
-      <ReviewDecision
-        decisionOptions={decisionOptions}
-        setDecision={setDecision}
-        isDecisionError={isDecisionError}
-        isEditable={thisReview?.current.reviewStatus === ReviewStatus.Draft}
-      />
+      <div className="flex-row-space-between" style={{ alignItems: 'flex-end' }}>
```
andreievg, line 33: nit: add `flex-row-space-between-flex-end`

src/containers/TemplateBuilder/template/Actions/ActionConfig.tsx
```diff
@@ -133,6 +135,16 @@ const ActionConfig: React.FC<ActionConfigProps> = ({ templateAction, onClose })
           <div className="spacer-10" />
           <div className="config-container-outline">
             <div className="flex-column-start-center">
+              <TextIO
```
andreievg, line 138: Would it have been ok to just:

`text={state.code || ''}`
`setState({ ..state: code: test || null}))` here ? 
Sorry I have no idea what the side effects are of chaging `TextIOprops` for text to string | null, it seems easier just to map null to '' for text and '' to null when saving back to state

andreievg: also saw this in console (I think it's related '`value` prop on `input` should not be null'

src/components/Review/DecisionPreview/ReviewPreviewModal.tsx
CarlosNZ: 👍🏼  f66b942b

semantic/src/site/globals/site.overrides
```diff
@@ -669,6 +732,7 @@ a:hover {
 
 #review-submit-area {
   margin-bottom: 30px;
+  max-width: 500px;
```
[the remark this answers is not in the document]
CarlosNZ: Right, yeah I see what you mean. Let the textarea fill its container and restrict the "review-submit-area" container so we don't have to keep them both in "sync".

f66b942b

src/components/Review/DecisionPreview/PreviewItems.tsx
```diff
@@ -0,0 +1,94 @@
+import React, { useState } from 'react'
+import { Icon, Accordion, Image } from 'semantic-ui-react'
+import { useLanguageProvider } from '../../../contexts/Localisation'
+import MarkdownBlock from '../../../utils/helpers/semanticReactMarkdown'
+import { ActionQueueStatus } from '../../../utils/generated/graphql'
+import config from '../../../config'
+
+const fileEndpoint = `${config.serverREST}/public/file?uid=`
```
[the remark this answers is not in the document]
CarlosNZ: Yup, I see you've made this point a few times now.

Made an issue: https://github.com/openmsupply/conforma-web-app/issues/1253

Should be fairly straightforward when we have a moment.

src/components/Review/DecisionPreview/PreviewItems.tsx
```diff
@@ -0,0 +1,94 @@
...
+const NotificationPreview = ({ item }: { item: NotificationPreviewData }) => {
+  const { strings } = useLanguageProvider()
+  const [open, setOpen] = useState(false)
+  return (
+    <Accordion className="item notification-preview" style={{}}>
+      <Accordion.Title active={open} onClick={() => setOpen(!open)}>
+        <Icon name="dropdown" />
+        {strings.REVIEW_PREVIEW_NOTIFICATION} <strong>{item.displayString}</strong>
+      </Accordion.Title>
+      <Accordion.Content active={open} className="content">
```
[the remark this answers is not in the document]
CarlosNZ: Ah yes, good point.

src/components/Review/DecisionPreview/ReviewPreviewModal.tsx
CarlosNZ: Ah yeah. Thanks. I think I've left it in a few other places where is isn't needed too. I'll do a tidy-up.

CarlosNZ: Yeah, removed it from quite a few places: 21a427ce

Thanks

src/components/Review/DecisionPreview/ReviewPreviewModal.tsx
CarlosNZ: That's not quite the same though, is it? We need to check that not only does `data` exist, but it has a length of more than 0. In your version it won't show the "NO PREVIEWS" message if data is an empty array, will it?

src/components/Review/DecisionPreview/ReviewPreviewModal.tsx
CarlosNZ: Sorry -- not quite following. I already have:
```
if (!open) {
      setData(null)
      setError(false)
      return
    }
```

Did you mean somewhere else as well?

src/containers/Review/ReviewSubmit.tsx
CarlosNZ: Sure. 85062fbc

src/containers/TemplateBuilder/template/Actions/ActionConfig.tsx
CarlosNZ: > Would it have been ok to just:
> 
> `text={state.code || ''}` `setState({ ..state: code: test || null}))` here ?

I think I tried something like that at first, but then it caused [another problem](https://github.com/openmsupply/conforma-web-app/pull/1224#discussion_r924119245) in that it wouldn't save empty values at all if changing them from existing text content.

I'll have another look at though, there should be a better way.

CarlosNZ: Yeah, that worked better. Thanks for the suggestion: 9c6e82a1

And I was able to have them fall back to either `null` or `""` depending on whether or not the database field is nullable.

src/components/Review/DecisionPreview/ReviewPreviewModal.tsx
andreievg: Ahh ok true could do data?.length > 0 ?

src/containers/TemplateBuilder/template/Actions/ActionConfig.tsx
andreievg: thanks

andreievg (approved):

PR #1227, opened 2022-07-05
File upload enhancements

CarlosNZ:
Requires back-end [PR #778](https://github.com/openmsupply/conforma-server/pull/778)

Will be required for the file management template outlined in https://github.com/openmsupply/conforma-web-app/issues/1177.

### Improvements: 

- Plugin can optionally show a "description" input field for each file, which we update on the server without re-uploading file (specify `showDescription: true` parameter)
- Split Display components into seperate files just for easier code management
- "Delete" icon now only shows when hovering over the file, which looks *way* better

Here's a template that shows the fileUpload plugin with the "showDescription" field on:
[fileUpload.zip](https://github.com/openmsupply/conforma-web-app/files/9042764/fileUpload.zip)

### DIFF
--- a/src/formElementPlugins/fileUpload/src/ApplicationView.tsx
+++ b/src/formElementPlugins/fileUpload/src/ApplicationView.tsx
@@ -1,14 +1,11 @@
 import React, { useEffect, useState, useRef } from 'react'
-import { Button, Icon, Grid, List, Image, Message, Segment, Loader, Input } from 'semantic-ui-react'
+import { Button, Icon, List, Segment } from 'semantic-ui-react'
 import { nanoid } from 'nanoid'
 import { ApplicationViewProps } from '../../types'
 import { useLanguageProvider } from '../../../contexts/Localisation'
 import { useUserState } from '../../../contexts/UserState'
 import { postRequest } from '../../../utils/helpers/fetchMethods'
-import prefs from '../config.json'
-import './styles.css'
-import { FileDisplay } from './components/FileDisplay'
-import { FileDisplayWithDescription } from './components/FileDisplayWIthDescription'
+import { FileDisplay, FileDisplayWithDescription } from './components'
 
 export interface FileResponseData {
   uniqueId: string
--- a/src/formElementPlugins/fileUpload/src/ApplicationView.tsx
+++ b/src/formElementPlugins/fileUpload/src/ApplicationView.tsx
@@ -1,18 +1,22 @@
 import React, { useEffect, useState, useRef } from 'react'
-import { Button, Icon, Grid, List, Image, Message, Segment, Loader } from 'semantic-ui-react'
+import { Button, Icon, Grid, List, Image, Message, Segment, Loader, Input } from 'semantic-ui-react'
 import { nanoid } from 'nanoid'
 import { ApplicationViewProps } from '../../types'
 import { useLanguageProvider } from '../../../contexts/Localisation'
 import { useUserState } from '../../../contexts/UserState'
 import { postRequest } from '../../../utils/helpers/fetchMethods'
 import prefs from '../config.json'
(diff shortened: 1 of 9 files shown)


src/formElementPlugins/fileUpload/src/ApplicationView.tsx
```diff
@@ -176,77 +201,20 @@ const ApplicationView: React.FC<ApplicationViewProps> = ({
...
-                      </Grid.Row>
-                    </>
-                  )}
-                </Grid>
-              </List.Item>
+        <List className="file-list" horizontal={!showDescription} verticalAlign="top">
+          {uploadedFiles.map((file) =>
+            showDescription ? (
+              <FileDisplayWithDescription
+                key={file.key}
```
andreievg, line 208: noticed that key is missing from normal FileDisplay, and that both just get the key from `file`. Would suggest to remove `key` from `FileDisplayWithDescription` and get `key` from file and make sure List.Item has key (for react)

src/formElementPlugins/fileUpload/src/ApplicationView.tsx
```diff
@@ -263,6 +231,12 @@ const ApplicationView: React.FC<ApplicationViewProps> = ({
       otherBody: fileData,
     })
   }
+  async function updateFileDescription(uniqueId: string, description: string | null) {
```
andreievg, line 234: can you move uplaodFile and updateFielDescription from component please

src/formElementPlugins/fileUpload/src/components/FileDisplayWIthDescription.tsx
```diff
@@ -0,0 +1,107 @@
...
+  downloadUrl,
+  updateDescription,
+}: FileDisplayDescriptionProps) => {
+  const { getPluginStrings } = useLanguageProvider()
+  const strings = getPluginStrings('fileUpload')
+  const { loading, error, errorMessage, filename, fileData, key } = file
+  const [description, setDescription] = useState<string>(fileData?.description ?? '')
+  return (
+    <List.Item>
+      {fileData && (
```
andreievg, line 24: should remove this check here `   {fileData && (` otherwise errors are not shown

src/formElementPlugins/fileUpload/src/ApplicationView.tsx
CarlosNZ: Doesn't the immediate child of in a component array need the key (for React)? So inside the `<List>` component, the immediate children are the `<FileDisplayWithDescription>` and `<FileDisplay>` components, so they must have a key for React. The `<List.Item>` components within each of those don't need a key as there's only one of them for each "FileDisplay" component. (The "FileDisplay" components are really just a wrapper for "List.Item"). Is that right? 

Anyway, have added key to "FileDisplay": efd92223

src/formElementPlugins/fileUpload/src/components/FileDisplayWIthDescription.tsx
CarlosNZ: Cool -- b66bf0e4

src/formElementPlugins/fileUpload/src/ApplicationView.tsx
andreievg: Ohh, i thought they meant to be outside and din't actually have any extra params, sorry, I found it weird having the methods after return, thus commented

src/formElementPlugins/fileUpload/src/ApplicationView.tsx
andreievg: ok, i wasn't sure about the key being on the immediate child. I just saw it was different between two similar components

andreievg (approved):
thanks

Issue #1234, opened 2022-07-11
Hide "Preview Decision" button when not appropriate


CarlosNZ:
Currently, it just shows "No previews available" in the modal when there's nothing to preview, but it actually seems wrong to see it at all in many situations. 

Simple solution is to also query template_actions during structure building, then we can check if there's any "ON_PREVIEW" actions for the current template.

This is only a simple on/off --i.e. it'll still show up for levels stages where it's not appropriate, but I think that's the simplest thing to do for now, and is a lot better than showing it on application types that never have preview options.

PR #1239, opened 2022-07-17
#1236 #1234 UI for Extend deadline

### LINKED ISSUE #1236: UI for "Extend deadline" button
Part of Epic https://github.com/openmsupply/conforma-web-app/issues/1225

See [agreed solution](https://github.com/openmsupply/conforma-web-app/issues/1225#issuecomment-1178466161)

CarlosNZ:
Fix #1236 and #1234 

Requires back-end branch from [PR #790](https://github.com/openmsupply/conforma-server/pull/790)

Use the snapshot suggested there for testing.

Submit an application (as jane_smith), review it with LOQ (as basic_reviewer), and note the new extension UI on the review home page:
Screen Shot 2022-07-17 at 12 40 35 PM

ADDITIONAL: Have created a generic "useConfirmationModal" hook as the existing Confirmation modal was super-annoying to use -- needed to add a bunch of additional logic into the current component rather than just calling it with a few props. I'll make a new PR to use this modal instead of the current one (it's only in a couple of places) and get rid of the existing ones.


src/containers/Review/overview/Overview.tsx
```diff
@@ -68,9 +87,57 @@ export const Overview: React.FC<{
                 <strong>{strings.REVIEW_OVERVIEW_SERIAL}: </strong>
                 {serial}
               </p>
+              {applicantDeadline && deadlineActive && (
+                <p className="right-item">
+                  <strong>{strings.REVIEW_OVERVIEW_DEADLINE}: </strong>
+                  {DateTime.fromJSDate(applicantDeadline).toLocaleString()}
+                </p>
+              )}
```
CarlosNZ, line 95: The logic for whether the deadline shows up is slightly different to whether or not to display the "Extend" button/selector. For the deadline, we only show it if it's active (i.e. not expired) and exists. But for the "Extend" button, we also need to show it when it's *not* active (i.e. expired), but also the application needs to be either "Expired" or have status "Changes requested".

src/utils/defaultLanguageStrings.ts
```diff
@@ -290,6 +291,13 @@ export default {
   REVIEW_OVERVIEW_STAGE: 'Stage',
   REVIEW_OVERVIEW_OUTCOME: 'Outcome',
   REVIEW_OVERVIEW_SERIAL: 'Serial',
+  REVIEW_OVERVIEW_DEADLINE: 'Applicant Deadline',
+  REVIEW_OVERVIEW_EXTEND_BY: 'Extend deadline by',
+  REVIEW_OVERVIEW_DAYS: 'days',
```
CarlosNZ, line 296: It's possibly problematic to split the sentence like this -- other languages might structure it differently. But it's not really feasible to put a substitution in when the substitution needs to be a Form element (number selector).

src/utils/hooks/useLoadApplication.ts
```diff
@@ -38,7 +37,7 @@ const useLoadApplication = ({ serialNumber, networkFetch }: UseGetApplicationPro
   const { strings } = useLanguageProvider()
   const [isLoading, setIsLoading] = useState(true)
   const [structureError, setStructureError] = useState('')
-  const [structure, setFullStructure] = useState<FullStructure>()
+  const [structure, setStructure] = useState<FullStructure>()
```
CarlosNZ, line 40: Oops -- how long has this been mis-named? 😳

src/utils/hooks/useLoadApplication.ts
```diff
@@ -163,6 +163,20 @@ const useLoadApplication = ({ serialNumber, networkFetch }: UseGetApplicationPro
 
     const canApplicantMakeChanges = application.template?.canApplicantMakeChanges ?? true
 
+    const scheduledEvents: ApplicationScheduledEvent[] = (
+      (application?.triggerSchedules?.nodes || []) as {
+        id: number
+        timeScheduled: string
+        eventCode: string
+        isActive: boolean
```
CarlosNZ, line 171: Can't just use `as ApplicationScheduledEvent[]` since the incoming data has the date as an ISO string, but I've converted it to a JS Date to put in the structure

src/utils/graphql/queries/getApplication.query.ts
```diff
@@ -21,13 +21,26 @@ export default gql`
             ...TemplateStage
           }
         }
+        # Count PREVIEW actions so we know whether to show the option to Preview
+        templateActions(condition: { trigger: ON_PREVIEW }) {
```
andreievg, line 25: can you alias this please to `previewActions: templateAction`, although I think access to this might be restricted to backend only (or at least it's easier that way), maybe we can do something with visibility of fields on templateActions for non admin user, i'll make a note of it

src/utils/graphql/queries/getApplication.query.ts
```diff
@@ -21,13 +21,26 @@ export default gql`
...
+        }
       }
       user {
         ...User
       }
       org {
         ...Organisation
       }
+      # Get scheduled events so we can extend deadlines
+      triggerSchedules {
```
andreievg, line 36: Here also I would make an alias: `applicationDeadlineSchedules: triggerSchedules (.. query for eventCode)`

src/containers/Review/overview/Overview.tsx
```diff
@@ -1,26 +1,45 @@
...
 }) => {
   const { strings } = useLanguageProvider()
   const { Outcome } = useLocalisedEnums()
+  const [deadlineDays, setDeadlineDays] = useState(5)
+  const { ConfirmModal, showModal } = useConfirmModal()
   const applicant = `${user?.firstName} ${user?.lastName}`
   const organisation = org?.name
   const { started, completed } = getDates(activityLog)
   const stage = current.stage.name
+  const applicantDeadlineEvent = scheduledEvents.filter(
```
andreievg, line 35: I feel like this should be extracted in structure, i.e. structure should have `applicationDeadlineEvent:? {isActive, deadline}`

src/containers/Review/overview/Overview.tsx
```diff
@@ -68,9 +87,57 @@ export const Overview: React.FC<{
...
+              {applicantDeadline && deadlineActive && (
+                <p className="right-item">
+                  <strong>{strings.REVIEW_OVERVIEW_DEADLINE}: </strong>
+                  {DateTime.fromJSDate(applicantDeadline).toLocaleString(DateTime.DATETIME_SHORT)}
+                </p>
+              )}
             </div>
+            {applicantDeadline &&
+              (outcome === ApplicationOutcome.Expired ||
+                current.status === ApplicationStatus.ChangesRequired) && (
```
andreievg, line 99: do we really need to be able to extend when `ChangesRequired` ? What would be the difference between it being `ChangesRequired` and `Draft` (i.e. when it's in LOQ and being edited by the user)

src/utils/hooks/useConfirmModal.tsx
```diff
@@ -0,0 +1,101 @@
...
+  type: 'warning' | 'confirmation'
+  title: string
+  message?: string
+  OKText: string
+  cancelText: string
+  onOK: () => void
+  onCancel: () => void
+}
+
+const useConfirmModal = ({
```
andreievg, line 15: Hmm, i'll comment about this in upstream PR (in omSupply we use global confirm modal and context)

CarlosNZ: I've made a generic useConfirmModal in the next PR, cos this was annoying me too.

CarlosNZ: Oh sorry, this is it. The next PR refactors existing to use it.

andreievg:
something for another issue, do we want to record deadline extension in the "log" (like the history in review overview)

andreievg:
noticed something, not sure if it's worth looking into, I can set deadline to 0, and confirmation modal shows that it will be extended by one day, but it doesn't actually extend

CarlosNZ:
> something for another issue, do we want to record deadline extension in the "log" (like the history in review overview)

That's done in subsequent PR #1244

src/utils/graphql/queries/getApplication.query.ts
CarlosNZ: I'm reluctant to alias it here, as this is actually fetching *all* trigger schedules for this application, not just the deadline ones, so calling it `applicationDeadlineSchedules` would be misleading. Wouldn't it be better to change the name when we insert a filtered version into the structure further down?

src/utils/graphql/queries/getApplication.query.ts
CarlosNZ: > can you alias this please to `previewActions: templateAction`

If I change this, when I try to refer to it in useLoadApplication, I get a type error on `hasPreviewActions: application.template.previewActions.totalCount > 0,` as its expecting auto-generated back-end "Application" type for this structure which has "templateActions" not "previewActions" on "application.template".

Any idea how to fix?

src/containers/Review/overview/Overview.tsx
CarlosNZ: Okay, I think this works (for both this and the previous comment): 95a4f2cd

src/containers/Review/overview/Overview.tsx
CarlosNZ: Just trying to think through the condition here. It needs to be active when:
- a deadline exists(active or other)
- application is expired
- application is not expired but in Draft or ChangesRequested

But NOT active when:
- application is APPROVED or REJECTED or WITHDRAWN (i.e. it's properly "Done")

So, okay -- I think this covers it: 8d022747

Any case I've missed?

CarlosNZ:
> noticed something, not sure if it's worth looking into, I can set deadline to 0, and confirmation modal shows that it will be extended by one day, but it doesn't actually extend

Okay, this was just a problem with my plural logic. Fixed now, and setting to 0 displays correctly but does nothing when submitted: 426ec571

src/utils/graphql/queries/getApplication.query.ts
andreievg: There is a cast further up `const application = data.applicationBySerial as Application`, which kind of negates query specific types that were auto generated, if I remove the `as` it seems to work, but get some compile errors in the ActivityLog (it has EXTENSION now, I think it's just some backend/frontend mismatch)

src/utils/graphql/queries/getApplication.query.ts
andreievg: oki, I don't think I communicated it well, was suggesting we also filter `triggerSchedules` by the eventType. But i guess it's ok to just filter it out in structure building

src/containers/Review/overview/Overview.tsx
andreievg: Hmm, I think it should just be whenever application is in the hand of applicant, it's a bit tricky actually:

 status `draft` or `changes_required` 
 or
 status 'completed' and outcome 'expired'
 would be great as a method `is_applicant_action` (ts method), although would also be great as gql resolver on application

CarlosNZ: It can't include the case when it's a first draft, as there's nothing to expire yet. Though there wouldn't be a deadline event created yet, so that's alright.

> status 'completed' and outcome 'expired'

The status is redundant -- it's always COMPLETED when outcome is anything other than "PENDING" (database enforced)

Okay, thinking about it a bit more clearly this morn, I think this should do it:  1565b252

Basically, must meet ALL of the following criteria:
- deadline exists
- status either DRAFT or CHANGES_REQUIRED
- outcome either PENDING or EXPIRED

Sound right?

src/utils/graphql/queries/getApplication.query.ts
CarlosNZ: Okay, got ya. I think my commit mentioned in the [comment below](https://github.com/openmsupply/conforma-web-app/pull/1239#discussion_r935540774) should take care of it. No filtering and renaming here in the query (I left it like that cos we may want to use other events at some point), but I've filtered, renamed and structured it as it goes into the full structure, so the "Overview" page doesn't need to do any additional logic.

src/containers/Review/overview/Overview.tsx
andreievg: I thought that when outcome goes to anything other then pending, status auto changes to `completed`  ?

I still think logic should be as per my last comment

src/utils/graphql/queries/getApplication.query.ts
CarlosNZ: Okay, thanks -- think this is good now: 672001f7

src/containers/Review/overview/Overview.tsx
CarlosNZ: Okay -- this time?: 6746cb7e 🤞

I think that's pretty much what I started with, except the extra "DRAFT" option.

andreievg (approved):
Thank for the changes

PR #1240, opened 2022-07-17
Refactor confirmation modals

CarlosNZ:
After creating a new generic "ConfirmationModal" hook in #1239, this PR just re-factors existing warning/confirm modals to use the new hook and deletes the two old ones.

Plus a few tweaks to the useConfirmationModal hook itself.


src/containers/Review/ReviewPage.tsx
```diff
@@ -66,19 +66,26 @@ const ReviewPage: React.FC<{
...
+  useEffect(() => {
+    if (!reviewStructure) return
+    if (thisReview?.current.reviewStatus === ReviewStatus.Pending) {
+      console.log('Should show modal')
+      showWarningModal({
+        onConfirm: () => push(`/application/${reviewStructure.info.serial}/review`),
+        onCancel: () => push(`/application/${reviewStructure.info.serial}/review`),
+      })
+    }
+  }, [reviewStructure])
```
CarlosNZ, line 84: Needed to put in a `useEffect` so it only runs this check once.

src/components/Admin/AdminLocalisations.tsx
```diff
@@ -16,7 +16,11 @@ export const AdminLocalisations: React.FC = () => {
   const [exportDisabled, setExportDisabled] = useState(true)
   const [importDisabled, setImportDisabled] = useState(true)
   const showToast = useToast({ position: topLeft })
-  const [showModalWarning, setShowModalWarning] = useState<ModalProps>({ open: false })
+  const { ConfirmModal: WarningModal, showModal: showWarningModal } = useConfirmationModal({
+    type: 'warning',
+    title: strings.LOCALISATION_DELETE_WARNING_TITLE,
+    confirmText: strings.BUTTON_CONFIRM,
+  })
```
CarlosNZ, line 23: All modal parameters can either be provided when the hook is first instantiated, or when the "showModal" method is called. As a general rule, I try to put them in the hook call except when either:
- the values might be dynamic (e.g. message dependent on context)
- they refer to items that haven't been defined yet (e.g. functions)

src/components/Admin/AdminLocalisations.tsx
```diff
@@ -16,7 +16,11 @@ export const AdminLocalisations: React.FC = () => {
   const [exportDisabled, setExportDisabled] = useState(true)
   const [importDisabled, setImportDisabled] = useState(true)
   const showToast = useToast({ position: topLeft })
-  const [showModalWarning, setShowModalWarning] = useState<ModalProps>({ open: false })
+  const { ConfirmModal: WarningModal, showModal: showWarningModal } = useConfirmationModal({
```
andreievg, line 19: This is very cool, returning component from the hook, although I don't know about react internals enough to figure out if there are any side effects. I was suggesting context, but I guess this should work pretty well, if we have any glitches for any reason would suggest changing to context

src/containers/Review/ReviewPage.tsx
```diff
@@ -66,19 +66,26 @@ const ReviewPage: React.FC<{
...
+    title: strings.REVIEW_STATUS_PENDING_TITLE,
+    message: strings.REVIEW_STATUS_PENDING_MESSAGE,
+  })
+
+  useEffect(() => {
+    if (!reviewStructure) return
+    if (thisReview?.current.reviewStatus === ReviewStatus.Pending) {
+      console.log('Should show modal')
+      showWarningModal({
+        onConfirm: () => push(`/application/${reviewStructure.info.serial}/review`),
```
andreievg, line 80: this seems like a one button modal

src/containers/Review/ReviewPage.tsx
```diff
@@ -66,19 +66,26 @@ const ReviewPage: React.FC<{
...
+  const { ConfirmModal: WarningModal, showModal: showWarningModal } = useConfirmationModal({
+    type: 'warning',
+    title: strings.REVIEW_STATUS_PENDING_TITLE,
+    message: strings.REVIEW_STATUS_PENDING_MESSAGE,
+  })
+
+  useEffect(() => {
+    if (!reviewStructure) return
+    if (thisReview?.current.reviewStatus === ReviewStatus.Pending) {
+      console.log('Should show modal')
```
andreievg, line 78: console.log

src/utils/hooks/useConfirmationModal.tsx
```diff
@@ -79,17 +89,17 @@ const useConfirmModal = ({
...
-  }: Partial<ConfirmModalState>) => {
+  }: Partial<ConfirmModalState> = {}) => {
     const newState: Partial<ConfirmModalState> = {}
     if (title) newState.title = title
     if (message) newState.message = message
-    if (OKText) newState.OKText = OKText
+    if (OKText) newState.confirmText = OKText
     if (cancelText) newState.cancelText = cancelText
-    if (onOK) newState.onOK = onOK
+    if (onOK) newState.onConfirm = onOK
```
andreievg, line 102: hmm, not sure why OKText and onOk

andreievg (approved):
This is very cool. I had a read through but haven't really digested all of the refactor. The modal hook looks sounds though

src/containers/Review/ReviewPage.tsx
CarlosNZ: Yeah good point. I just added an additional property to hide the cancel button and applied it here (and a couple of others) 

13317a0e

src/utils/hooks/useConfirmationModal.tsx
CarlosNZ: Oops - -yeah. I renamed this at one point and I guess this instance automatically aliased it rather than fully renamed it. Thanks. 799f4f2e

andreievg (approved):
Thanks for the changes, looks good

PR #1244, opened 2022-07-18
#1238 Timeline improvements to show Deadline extensions

### LINKED ISSUE #1238: Check/fix timeline display for "Extend" events
Part of Epic https://github.com/openmsupply/conforma-web-app/issues/1225

See [agreed solution](https://github.com/openmsupply/conforma-web-app/issues/1225#issuecomment-1178466161)

CarlosNZ:
Fix #1238 

Requires back-end from https://github.com/openmsupply/conforma-server/pull/792

Had to re-factor some of the existing code so that the special-case events (that only show as the *last* item in timeline) continue to work, but I think I've simplified it overall.

To test, load the same snapshot as the last PR (`/dev/snapshots/Deadline_Column_In_List_2022_07_18`) and submit and review an application (with LOQ), then notice the Timeline when extending the deadline.

(May need to manually run `yarn migrate` in the back-end)


src/utils/hooks/useTimeline/useTimeline.tsx
```diff
@@ -86,20 +94,31 @@ const buildTimeline = (
...
+          TimelineEventType.Ignore,
+          TimelineEventType.ApplicationChangesRequired,
+          TimelineEventType.ReviewChangesRequested,
+        ].includes(timelineEvent.eventType)
       )
         stages[stageIndex].events.push(timelineEvent)
     }
   })
-  // Put final outcome at the end of the event list
-  if (finalOutcome) stages[stageIndex].events.push(finalOutcome)
```
andreievg, line 102: Not sure about this, looks like we not showing final outcome ?

andreievg (commented):
This is the first time I am looking at useTimeline:

Have suggestion, not to use index: https://github.com/openmsupply/conforma-web-app/compare/1238-applicant-deadline-in-timeline...1238-suggestion

```
  STAGE: () =>
    // Already handled in caller
    ({ eventType: TimelineEventType.Ignore, displayString: '' }),
```
Not used ? I guess if it's removed there will be missing key in EventType.

`eventInterpretation` -> wow ..  
🏃‍♂️

andreievg:
Want to ask a question re outcome being removed and changes from .Ignore, easier to do in person i think

andreievg:
I just realised, that we could potentially have the same issue with timeline done on server (the trigger issue, where server hasn't quite processed log into event log, with existing triggers it would be instant), but it's slightly different i guess since you can't change timeline from overview tab

andreievg (approved):
Comments were clarified in our conversation, thanks

PR #1248, opened 2022-07-19
#1246 Let developers set templates as draft even when there are applications

### LINKED ISSUE #1246: Template builder -- allow setting "draft" with existing applications in Dev mode
While in dev it would be a lot nicer if we could set templates as drafts for editing even with existing applications. It's kind of annoying to have to create new versions for every little change when testing, or manually deleting the existing applications.

CarlosNZ:
Closes #1246 

This will make my life easier, for one. ;)


andreievg (approved):
nice

PR #1249, opened 2022-07-19
#1245 Add "evaluatorFunctions" to template builder action evaluations

### LINKED ISSUE #1245: Add "evaluatorFunctions" to template builder so we can preview them
Just needs to be an exact copy of the "evaluatorFunctions" file in the back-end (and we need to remember to keep them in sync).

Then add the "functions" object to evaluator params in the template builder evaluation preview, making sure it matches back-end structure.

~~Also -- to do at the same time -- update the existing "getYear" function to take an optional parameter for number of digits, i.e.:~~ (DONE -- squeezed it into #1244 )
```
getYear() => 2022
getYear("short") => 22
```

CarlosNZ:
Closes #1245 

Now, when selecting the "objectFunctions" operator in the Template Builder (Actions only), you can preview the output.

The default "functions.getYear" value in the template builder will work -- should return `"2022"`


andreievg (commented):
would assume we should add type to getYear on back end ?

andreievg:
What do you think of mono repo ?

andreievg (approved):

CarlosNZ:
> What do you think of mono repo ?

Yeah, I think it might be nice, but I don't think the advantages make it super urgent for us right now.

With this one, I think it would be better if the back-end was able to serve these functions rather than duplicating it in both repos, but we'd probably do that once we have the back-end serving all the plugins etc. as well.

PR #1252, opened 2022-07-21
commit from new branch which is branched from develop

fergie-nz:
New branch which pushes to right base (develop)

### DIFF
--- a/src/formElementPlugins/checkbox/src/ApplicationView.tsx
+++ b/src/formElementPlugins/checkbox/src/ApplicationView.tsx
@@ -23,6 +23,7 @@ const ApplicationView: React.FC<ApplicationViewProps> = ({
   onSave,
   Markdown,
   initialValue,
+  validationState
 }) => {
   const { getPluginStrings } = useLanguageProvider()
   const strings = getPluginStrings('checkbox')
@@ -30,6 +31,7 @@ const ApplicationView: React.FC<ApplicationViewProps> = ({
   const { label, description, checkboxes, type, layout, resetButton = false, keyMap } = parameters
 
   const [isFirstRender, setIsFirstRender] = useState(true)
+  const [toggled, setToggled] = useState(false)
 
   const [checkboxElements, setCheckboxElements] = useState<Checkbox[]>(
     getInitialState(initialValue, checkboxes, keyMap, isFirstRender)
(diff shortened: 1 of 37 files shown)


fergie-nz:
[CR2-4.zip](https://github.com/openmsupply/conforma-web-app/files/9155168/CR2-4.zip)

Template used to test.

Page 2 of form has conditional checkboxes at the end as a declaration (both need to be checked to remove validation message)

src/formElementPlugins/checkbox/src/ApplicationView.tsx
```diff
@@ -23,13 +23,15 @@ const ApplicationView: React.FC<ApplicationViewProps> = ({
...
   initialValue,
+  validationState
 }) => {
   const { getPluginStrings } = useLanguageProvider()
   const strings = getPluginStrings('checkbox')
   const { isEditable } = element
   const { label, description, checkboxes, type, layout, resetButton = false, keyMap } = parameters
 
   const [isFirstRender, setIsFirstRender] = useState(true)
+  const [toggled, setToggled] = useState(false)
```
CarlosNZ, line 34: I can see where you're going with this -- making it so it doesn't immediately show up as invalid. Good idea, the only problem is this value gets reset every reload, so if you change page and go back to it, the warning has disappeared.

Also -- when you press the "Submit" button, the error message should appear then too, which is doesn't do with this approach.

Normally we check if the current response is `null` and use that to consider whether the user has "interacted" with the element yet or not. The problem with the checkboxes is that we save a response immediately, as "no selection" needs to be a valid selection for checkboxes.

I think what we might need to do for this is create a new parameter called "allowNoSelection", which defaults to "true". But when it's "false", then we *don't* save the response on initial load, so not selecting anything will be considered "invalid". 

This is a bit trickier than the initial issue and will require a bit of thorough checking for all the possible situations. Let's discuss.

CarlosNZ (changes requested):
Hey, thanks @fergie-nz -- good start, does the job.

I've highlighted one problem regarding the validation of the checkboxes and their initial state. It's a bit complicated -- perhaps we can go through it together and I'll show you what I'm talking about.

If you want to make that little change to use the "Label" instead of the Form.Field, then let me know and we'll take it from there.

Cheers

fergie-nz:
New branch created from base develop

PR #1254, opened 2022-07-24
#1253 global url constructor for endpoints

### LINKED ISSUE #1253: Global url constructor for endpoints
To avoid having this all throughout the front-end:
```
const fileEndpoint = `${config.serverREST}/public/file?uid=` 
```

Create a global helper method, so we can just do something like this when we need it:
```
<Image src={getServerUrl('file', fileId, isThumbnail) style={{ maxHeight: 50 }} />
```

Something like that I guess?

CarlosNZ:
Fix #1253 

Ask and ye shall receive, @andreievg.

May have a gone a little overboard, but the syntax for using the endpoint is so much more straightforward now, e.g.

`getServerUrl('snapshot', 'use', name)`

### Notes

- Have abstracted all the url-building logic out of "config.ts" and now it's purely the reference list, with url-building logic all in `endpointUrlBuilder.ts`
- Used strict typing to ensure that the canonical list of endpoints in "config.ts" is exhaustively and restrictively handled by the module (i.e. if you add one without defining a case for it, it won't compile)
- Almost all the changed files here are implementation, the actual logic is pretty much all in `endpointUrlBuilder` and its associated `types` file.


andreievg (approved):
This is epic ! Thanks Carl. 

I would prefer to use objects:
* Reduce chance of error if there are consecutive string params
* Clarity when calling the method
* Clarity within the method implementation (can destructure with confidence0

Here is an [example](https://www.typescriptlang.org/play?#code/C4TwDgpgBAYglgGwgUQHYBMwHs6uFAXigG0INtdgAuKAcgDNEJaAaKRpASXRoGdgATrgDmbXnABeEAPx9BIgLoBYAFChIUADIBDVMICu24SnI48hEmUxnqdBLoNHmbe3sPGAwlnQQ5QvcoqAPRBUOrQACoAFnC8AOqICABCAhDaANbIAB5R2vr8cABu0ETEUFYUeDS0EDl5BcWsULxYALYQwDF6yAi8vs3yelCB4VAAShD8aNaUvBbwSNOV+AA+Wg7uJjPmIVBr0bEJCMmpGdm5+cBFEKqqAMZYqPzNEALFAmPIAMoRFrR9b1eAH0KjZaPdHs9UlNTLMLABvVRQdhMaocZxIqCuRzGapgfQAIwQcDuQWxm1YmNqFwa-VoqCC2nBKgAvrcVA8nvhWtpgHcogBBY4WAAUQJoqAg7wAlBKpa9CAA+KCIlTIzoCLAAdygkp1yAEmoEIpqdUu1ygPL5UVo0tUbJUEK5UGMwC+r3eE2eRBFADp-doBMJeDQvcAljZeLKBv5hEqVZjOc9QZQANIQEAWQPB4gABgUAG5E5D8Cm8AAFXlRCzQ8OwvC8Uj14DpkCBTG8LVwa1QEVllsZ6UJtXIqB3bR9OwbJy0KiY0djktY6eebwlKDZxsARkCC+RuyT+E6+laBKzQcbACZd3vUsB9AJUFAAAYAEnhAM93wiLPf-crnQskE77kk4Xg+Cyz7zmOE7QAwTCztByKHiiXDoOeOY7khi7OuIUgYVeN4LneD5Pm+H4eq8nw-L+8L-lWLLSPocDoAQ77otwtF4dA0gvvCABk3FsR+kgQJBUDVLQkHQT49B5Ag1DYVa-JCggfbNq2dojsRHSkXQzLIg6DpOi0SC+ggWDCCKrruoCHyTMAJrok0tCXrmYBZC5DxYAghQsRAWC2lpB6Qj5EDmZZ1kdLZnoOU5CFsLQx6nqg2iIF5WA+X5PiBYluCFNoxLoEF7K7Nk2itGASBQF2nRQAA8gSABWEB3MAvDsqMCxbMsl4IuUzZoglqEQNwfgiGIomyDGIhQA6ow6G4TgRpQfVEPCA3bLYtCgcYTS7RA4H9PwsZzaouyjAc8SJCkaSZGatL9f21TUvUVyNGIbQdF0wg9H041DPN4DQGGK0NmtsBMGDwB9Wsi04j1Nh9SE+wxNdxy3WcD3vTcjockuNmUfZ-AQyKuD4rYoPNrwl7Rids0EMqqrIUu9F1UQtbQ425P6MAvr9u2I6dt2-K9jzfP9kOzOjuOk47SuzBztpMtLhtB1HXNFji0Wyv7qEKEbbE0QngSqWIDQyVniyWuoBT2EkY+L7vp+VHfrRbNREBIEK0d0nK7LcHOUre44c8G0cegk34dbRDa-bumO+RLsfG7f7NgBntMSxwkR1xolQLxz4CUJzuieJkl+wusnyYpuuWryKnHGTtu81pIcO0+tAGWdrLskmYURVZhN2WGl4ihtz10M5bAR9UbkeU03HVN5vn+YFc3SsF+uhWZFlD9FROj+Pm3LENSBNLPdDz55Ud0iv2UBU0+WFSx1TP0VtAb1p7KqEAA) (i deliberately made size on file vs thumbnail, i would assume you used isThumbnail: 'thumbnail'  to avoid consecutive string params ? btw with the way you've done it, using boolean could work, although it's not explanatory when you call it, i see.

CarlosNZ:
> I would prefer to use objects:

I dunno, I think this:

```
getServerUrl('snapshot', 'use', name)
```
is way nicer than;
```
getServerUrl( {endpoint: 'snapshot', action: 'use', name: name} )
```

And in the simpler case: `getServerUrl('prefs')` beats `getServerUrl({endpoint: 'prefs'})`


I think this is a good case for using `args`. 

>     * Reduce chance of error if there are consecutive string params

I don't think there's any where this is the case. Most strings are basically enums, or a list of string literals (such as the list of available endpoint values).

>     * Clarity when calling the method

See examples above -- I think the less verbose one is way cleaner, especially when it's going to be used all over the place.

>     * Clarity within the method implementation (can destructure with confidence0

The named tuples defined in `types.ts` ensure that you can't get this wrong, and destructuring has type safety just like with objects. If you put the wrong parameter types in the wrong order anywhere, typescript will tell you immediately.

I kind of agree for cases where there are a lot of variable/optional parameters as it can get confusing quickly, but in this case, the args follow a common (and fairly obvious IMHO) pattern -> `( endpoint, action/type, value)` or `(endpoint, value)`, or just `(endpoint)`

Where there is more, or optional parameters, I've used a parameters object as the third argument, such as for file upload.

This seems to be a common pattern in a lot of packages I've seen/used and I quite like it --> multiple expected arguments, then a final "options" object as the last argument with all the optional parameters.

src/LookupTable/components/ImportCsvModal.tsx
```diff
@@ -56,9 +57,9 @@ const ImportCsvModal: React.FC<any> = ({
 
     await axios
       .post(
-        `${config.serverREST}/admin/lookup-table/import${
-          tableStructureID ? '/' + String(tableStructureID) : ''
-        }?name=${name}`,
+        tableStructureID
+          ? getServerUrl('lookupTable', 'update', tableStructureID)
+          : getServerUrl('lookupTable', 'import', name),
```
andreievg, line 62: this looks different

andreievg:
If you start extending endpoints that have multiple args as parameters (especially optional ones), you'll have to start using `undefined` or change parameters to be an object.

> See examples above -- I think the less verbose one is way cleaner, especially when it's going to be used all over the place.

Mostly yes, but for edge cases, to me the following are somewhat confusing:

```
getServerUrl('dataViews', tableName, recordId)
getServerUrl('dataViews', tableName, apiQueries)
getServerUrl('dataViews')

getServerUrl('lookupTable', 'update', tableStructureID)
getServerUrl('lookupTable', 'import', name)
```

Sorry I am very convinced that object are better when unions are involved, even though I am not requesting changes for it, I still need to drive my point (below might have been improved in latest typescript versions though, and maybe not an issue anymore):

## Type errors are more explanatory:

Screen Shot 2022-08-03 at 9 31 13 PM

Screen Shot 2022-08-03 at 10 08 06 PM

## Type hints

`in endpointUrlBuilder`

Screen Shot 2022-08-03 at 10 03 00 PM

Screen Shot 2022-08-03 at 10 02 45 PM

[in example branch](https://github.com/openmsupply/conforma-web-app/compare/1253-global-url-constructor-for-endpoints...1253-example)

First with args (this should be 'one' | 'two'):
Screen Shot 2022-08-03 at 10 13 08 PM

But fine with object

Screen Shot 2022-08-03 at 10 13 23 PM

## Type safety:

I noticed this when removing `isReview`, can remove from type but can still use ([see example branch](https://github.com/openmsupply/conforma-web-app/compare/1253-global-url-constructor-for-endpoints...1253-example) `'language' endpoint, not_used argument.


I think it's similar stylistic preference in which we differ that was used in expressionEvaluator (I think arrays should have been avoided where possible in place of objects, and generic `children` key when arrays are needed to be named more specific).

Anyways, I am not asking for any changes related to object shape, but still prefer objects in this case.

I want to have another look through the changes, I found something while doing the type example, will comment and request changes.

src/utils/hooks/useDataViews.tsx
```diff
@@ -58,14 +55,12 @@ export const useDataViewsTable = ({ tableName, apiQueries }: DataViewTableProps)
   } = useUserState()
 
   useEffect(() => {
-    const queryElements = []
-    if (first) queryElements.push(`first=${first}`)
-    if (offset) queryElements.push(`offset=${offset}`)
-    if (orderBy) queryElements.push(`orderBy=${orderBy}`)
-    if (ascending) queryElements.push(`ascending=${ascending}`)
```
andreievg, line 65: nice

src/utils/helpers/endpoints/endpointUrlBuilder.ts
```diff
@@ -0,0 +1,144 @@
...
+      return `${serverREST}${endpointPath}${buildQueryString(checkUniqueParams as BasicObject)}`
+
+    case 'dataViews':
+      const tableName = args[1]
+      const itemId = typeof args[2] === 'string' ? args[2] : undefined
+      const query = typeof args[2] === 'object' ? args[2] : undefined
+      // Detail view
+      if (itemId) return `${serverREST}${endpointPath}/table/${tableName}/item/${itemId}`
+      // Table view
+      if (tableName)
```
andreievg, line 75: I think this should be at the top
```
if (!tableName)
return `${serverREST}${endpointPath}`
```

andreievg (commented):
Oki, went through all of them, just some minors, only one I think in question is the snapshot optionsName

CarlosNZ:
> Sorry I am very convinced that object are better when unions are involved, even though I am not requesting changes for it, I still need to drive my point (below might have been improved in latest typescript versions though, and maybe not an issue anymore):

I'll have to get you to show me in person what you mean -- I think I'm seeing different things in my VS code when it comes to type hinting and safety. I thought it was giving exactly what I would want, but maybe I'm missing something. (I thought named tuples are essentially the same as objects when it comes to typescript features -- they're kind of supposed to be the best of both worlds)

I'd be open to making anything after the first parameter an object, but it seems crazy to me to make the first parameter part of a named key/value object when it's always the endpoint enum.

So it'd be `getServerUrl('endpoint', {options})` or just `getServerUrl('endpoint')` for the simple ones. (I've already done it like that for the more complicated ones like "fileUpload.) Sound reasonable?

Can discuss it in person tomorrow when I'm back in the office.

CarlosNZ:
> I think it's similar stylistic preference in which we differ that was used in expressionEvaluator (I think arrays should have been avoided where possible in place of objects, and generic `children` key when arrays are needed to be named more specific).

I just sent you an invite to my version 2 repo for the evaluator that I'm working on, which has the best of both approaches (objects first, but full backwards compatibility with existing approach).

src/formElementPlugins/fileUpload/src/ApplicationView.tsx
```diff
@@ -253,21 +240,20 @@ const uploadFile = async (
   const fileData = new FormData()
   await fileData.append('file', file)
   return await postRequest({
-    url: `${uploadUrl}?user_id=${userId}&application_serial=${serialNumber}&application_response_id=${applicationResponseId}${
-      subfolder ? '&subfolder=' + subfolder : ''
-    }`,
+    url: getServerUrl('upload', {
+      userId,
```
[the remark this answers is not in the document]
CarlosNZ: Yup, there's a general method `combineRequestParams` on (most?) server routes, that also standardises casing of keys (`objectKeysToSnakeCase`).

src/utils/hooks/useDataViews.tsx
CarlosNZ: Yeah, nice not to have to break it apart only to re-assemble it again! ;)

src/utils/helpers/endpoints/endpointUrlBuilder.ts
CarlosNZ: Cool. This plus above suggestions: b5887213

src/utils/helpers/endpoints/endpointUrlBuilder.ts
```diff
@@ -0,0 +1,144 @@
...
+      const options = args[3]
+
+      // "download" is direct download url
+      if (action === 'download') return `${serverREST}${endpointPath}/files/${name}.zip`
+
+      if (action === 'upload' || action === 'delete' || !options)
+        return `${serverREST}${endpointPath}/${action}?name=${name}`
+
+      // Options AND name present, so must be 'take' or 'use'
+      return `${serverREST}${endpointPath}/${action}?name=${name}&optionsName=${options}`
```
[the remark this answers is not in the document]
andreievg: I just saw this (in Snapshot.tsx):

Screen Shot 2022-08-04 at 11 38 25 AM

And saw that it would be different with endpoinUrlBuilder.ts

andreievg:
> Can discuss it in person tomorrow when I'm back in the office.

Sounds good, maybe my IDE is playing up

CarlosNZ:
Okay, going to merge this now with the more strictly typed "options" object in the tuple. We can revisit at a later stage.

As discussed in person, the excessive `as` clauses are a limitation of Typescript when re-assigning a variable whose type would otherwise be strictly understood within each `case`. Hopefully this will improve in future Typescript version.

PR #1256, opened 2022-07-25
add drop down add other option

fergie-nz:
This allows users to save new options to drop down lists.

I believe there is a more elegant solution, but this is the best I can come up with.

### DIFF
--- a/src/formElementPlugins/dropdownChoice/src/ApplicationView.tsx
+++ b/src/formElementPlugins/dropdownChoice/src/ApplicationView.tsx
@@ -1,6 +1,5 @@
 import React, { useEffect, useState } from 'react'
 import { Dropdown, Label } from 'semantic-ui-react'
-import consolidatorResponseFragment from '../../../utils/graphql/fragments/consolidatorResponse.fragment'
 import { ApplicationViewProps } from '../../types'
 
 const ApplicationView: React.FC<ApplicationViewProps> = ({
--- a/src/formElementPlugins/dropdownChoice/src/ApplicationView.tsx
+++ b/src/formElementPlugins/dropdownChoice/src/ApplicationView.tsx
@@ -26,86 +26,86 @@ const ApplicationView: React.FC<ApplicationViewProps> = ({
   } = parameters
 
   const [selectedIndex, setSelectedIndex] = useState<number>()
-  const [localOptions, setLocalOptions] = useState<any[]>([])
   //saving selected value in local state, so that when options are re rendered, we retain this value
-  const [addedOptions, setAddedOptions] = useState<any[]>([])
+  const [addedOptions, setAddedOptions] = useState<Array<{}>>([])
 
   const { isEditable } = element
-
-  useEffect (() => {
-    const optionsMapped = options.map((option: any, index: number) => {
-      return {
-        key: `${index}_${option}`,
-        text: optionsDisplayProperty ? option[optionsDisplayProperty] : option,
-        value: index,
-      }
-    })
-    if (addedOptions && addedOptions.length > 0) {
-      const newOptionList = [...optionsMapped, ...addedOptions]
-      setLocalOptions(newOptionList)
-    } else {
-      setLocalOptions(optionsMapped)
-    }
-  }, [options, addedOptions])
 
   useEffect(() => {
     // This ensures that, if a default is specified, it gets saved on first load
     if (!currentResponse?.text && defaultOption !== undefined) {
-      const optionValue = getDefaultIndex(defaultOption, options)
+      const optionIndex = getDefaultIndex(defaultOption, options)
       onSave({
         text: optionsDisplayProperty
-          ? options[optionValue][optionsDisplayProperty]
-          : options[optionValue],
-        selection: options[optionValue],
-        optionValue,
+          ? options[optionIndex][optionsDisplayProperty]
+          : options[optionIndex],
+        selection: options[optionIndex],
+        optionIndex,
       })
-      setSelectedIndex(optionValue)
+      setSelectedIndex(optionIndex)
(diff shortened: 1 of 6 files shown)


src/formElementPlugins/dropdownChoice/src/ApplicationView.tsx
```diff
@@ -21,53 +22,93 @@ const ApplicationView: React.FC<ApplicationViewProps> = ({
     search,
     options,
     optionsDisplayProperty,
+    hasOther,
     default: defaultOption,
   } = parameters
 
-  const [selectedIndex, setSelectedIndex] = useState<number>()
+  const [selectedIndex, setSelectedIndex] = useState<any>()
```
CarlosNZ, line 29: Change type back to `number`

src/formElementPlugins/dropdownChoice/src/ApplicationView.tsx
```diff
@@ -21,53 +22,93 @@ const ApplicationView: React.FC<ApplicationViewProps> = ({
...
+    } else {
+      setLocalOptions(optionsMapped)
+    }
+  }, [options, addedOptions])
+
   useEffect(() => {
     // This ensures that, if a default is specified, it gets saved on first load
     if (!currentResponse?.text && defaultOption !== undefined) {
-      const optionIndex = getDefaultIndex(defaultOption, options)
+      const optionValue = getDefaultIndex(defaultOption, options)
```
CarlosNZ, line 56: Why change this variable name? (Doesn't look like you've changed any functionality)

src/formElementPlugins/dropdownChoice/src/ApplicationView.tsx
```diff
@@ -1,5 +1,6 @@
 import React, { useEffect, useState } from 'react'
 import { Dropdown, Label } from 'semantic-ui-react'
+import consolidatorResponseFragment from '../../../utils/graphql/fragments/consolidatorResponse.fragment'
```
CarlosNZ, line 3: Remove this --looks like accidental import

src/formElementPlugins/dropdownChoice/src/ApplicationView.tsx
```diff
@@ -21,53 +22,93 @@ const ApplicationView: React.FC<ApplicationViewProps> = ({
...
     optionsDisplayProperty,
+    hasOther,
     default: defaultOption,
   } = parameters
 
-  const [selectedIndex, setSelectedIndex] = useState<number>()
+  const [selectedIndex, setSelectedIndex] = useState<any>()
+  const [localOptions, setLocalOptions] = useState<any[]>([])
+  //saving selected value in local state, so that when options are re rendered, we retain this value
+  const [addedOptions, setAddedOptions] = useState<any[]>([])
```
CarlosNZ, line 32: We should create a proper `type` for this and not use `any`.

src/formElementPlugins/dropdownChoice/src/ApplicationView.tsx
```diff
@@ -21,53 +22,93 @@ const ApplicationView: React.FC<ApplicationViewProps> = ({
...
-          ? options[optionIndex][optionsDisplayProperty]
-          : options[optionIndex],
-        selection: options[optionIndex],
-        optionIndex,
-      })
-    // Reset response if selection cleared
-    else onSave(null)
+  //addItemHandler is called when new data is entered. This in turn calls original hook to re-render displayed options
+  const addItemHandler = (e: any, data: any) => {
+    //console.log('data is', data)
```
CarlosNZ, line 70: Remove console.log

src/formElementPlugins/dropdownChoice/src/ApplicationView.tsx
```diff
@@ -21,53 +22,93 @@ const ApplicationView: React.FC<ApplicationViewProps> = ({
...
-        text: optionsDisplayProperty
-          ? options[optionIndex][optionsDisplayProperty]
-          : options[optionIndex],
-        selection: options[optionIndex],
-        optionIndex,
-      })
-    // Reset response if selection cleared
-    else onSave(null)
+  //addItemHandler is called when new data is entered. This in turn calls original hook to re-render displayed options
+  const addItemHandler = (e: any, data: any) => {
```
CarlosNZ, line 69: Type should be `string` based on what you've done below.

CarlosNZ (changes requested):
Thanks @fergie-nz , good idea to use a separate array for the "addedOptions". 

Having done that though -- I don't think you need to have the "localOptions" state value as well, as it's just a combination of "options" and "addedOptions".

You should just merge "options" and "addedOptions" together to create "dropdownOptions" like it was originally. In other words, use dropdownOptions like it was working previously, just add your "addedOptions" to it.

Then you won't need the new useEffect either.

Also, to simplify things a bit more, I wouldn't bother storing any index in the addedOptions array -- I'd just keep it as an array of strings. Only add the index and key when you merge it with "options", as the indexes are pretty meaningless until they're merged.

Let me know if you need any further clarification.

src/formElementPlugins/dropdownChoice/src/ApplicationView.tsx
CarlosNZ: Also, you'll need to populate the initial value. Currently when saving an additional item and reloading, the "added option" no longer shows up.

After "options" as loaded, if the "currentReponse" text is not included in "options", then you'll need to add it to "addedOptions" at that point.

CarlosNZ:
You'll need to update the reference docs in `Element_type_specs.md` as well. Just follow the pattern used for the others. (E.g. I think "radioButton" has an "hasOther" option)

PR #1259, opened 2022-07-26
1258 Fix navigation to password reset

### LINKED ISSUE #1258: Navigating to password reset after verification goes to login
could be related: https://github.com/openmsupply/conforma-web-app/issues/1154

I can replicate this pretty consistently, use verification link in incognito tab, when clicking to navigate to reset password it goes to login screen. Didn't want to setup email so just added `console.log(parametersEvaluated)` to https://github.com/openmsupply/conforma-server/blob/f1f94dc408a0c52cde723914cd02bd07e8851c05/src/components/actions/executeAction.ts#L82 (can see url in server console log).

The problem is with `location` not being fully available via `useLocation` in `useRouter` (i.e. pathname has `?sessionId=` but `search` doesn't), thus when AuthenicatedWrapper.tsx checks for `if(sessionId)` from useRouter, it's not immediately available.
I was logging `location` in useRouter and in AuthenicatedWrapper.tsx, and was seeing that location is "fully" available when state key is present (that's when `search` has correct value), so fast way to fix for now is to check that, and otherwise return loading. Not that happy with the suggested fix, since there could be other areas being affected (TODO add issue).

andreievg:
closes #1258 

Fix as per issue description

### DIFF
--- a/src/containers/Main/AuthenticatedWrapper.tsx
+++ b/src/containers/Main/AuthenticatedWrapper.tsx
@@ -11,15 +11,14 @@ const AuthenticatedContent: React.FC = () => {
     location,
     query: { sessionId },
   } = useRouter()
-  if (isLoggedIn()) return <SiteLayout />
-  // If there is a sessionId in the URL, then need to login as nonRegistered
-  // before continuing
-
   // Location is not fully loaded unless 'state' is present, TODO try upgrading react router
   // without this sessionId is missing when navigating via Click Here in forgot password verification
   if ('state' in location) {
     let { pathname, search } = location
+    // If there is a sessionId in the URL, then need to login as nonRegistered
+    // before continuing
     if (sessionId) return <NonRegisteredLogin option="redirect" redirect={pathname + search} />
+    if (isLoggedIn()) return <SiteLayout />
     return <Redirect to={{ pathname: '/login', state: { from: location.pathname } }} />
   }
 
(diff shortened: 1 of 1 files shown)


src/containers/Main/AuthenticatedWrapper.tsx
```diff
@@ -4,17 +4,26 @@ import { useRouter } from '../../utils/hooks/useRouter'
...
-  if (sessionId) return <NonRegisteredLogin option="redirect" redirect={pathname + search} />
-  return <Redirect to={{ pathname: '/login', state: { from: location.pathname } }} />
+
+  // Location is not fully loaded unless 'state' is present, TODO try upgrading react router
+  // without this sessionId is missing when navigating via Click Here in forgot password verification
+  if ('state' in location) {
+    let { pathname, search } = location
+    if (sessionId) return <NonRegisteredLogin option="redirect" redirect={pathname + search} />
+    return <Redirect to={{ pathname: '/login', state: { from: location.pathname } }} />
+  }
```
CarlosNZ, line 24: Cool, so this gets updated from "location" props as soon as "state" becomes available, and then the component re-renders?

Looks good, thanks. I can't really test this locally as it seems to work regardless, it's only on the server that it seems to be a problem. I'll make a new build and include this and see how it goes.

CarlosNZ: Hmm, didn't seem to fix it.

Will look into it more later

CarlosNZ (approved):
Okay -- let's merge this now and keep an eye on it going forward to see if there's any outstanding problems.

Issue #1261, opened 2022-07-28
Add "Loading" spinners/waiting for more items


CarlosNZ:
Noticed during the recent presentation that there's not UI feedback when certain items are loading. We don't notice it in dev because they appear super-fast, but on a slow connection, it's quite disconcerting, so could use some user feedback.

In particular:
- [x] Loading spinners PR #1262 
- [x] Application list - PR #1262 
- [x] Assignments (on Assignments tab) -- #1375
- [ ] Templates page
- probably some others...

CarlosNZ:
Maybe you could try, @fergie-nz ?

If you do, check in with me before you start and I'll point you in the right direction.

CarlosNZ:
Seems like we've done most of these...

PR #1262, opened 2022-07-29
Feature/loading spinners upon slow loading

fergie-nz:


fergie-nz:
This is working fine but the "filters" box jumps bigger than smaller upon render. I can't figure out why or how to fix this.

src/components/List/ApplicationsList.tsx
```diff
@@ -22,6 +24,7 @@ const ApplicationsList: React.FC<ApplicationsListProps> = ({
   loading,
   refetch,
 }) => {
+
```
CarlosNZ, line 27: These are just whitespace changes, so can you revert the file to what it was?

src/components/Dashboard.tsx
```diff
@@ -85,9 +94,10 @@ const TemplateComponent: React.FC<{ template: TemplateInList }> = ({ template })
   )
 }
 
-const FilterComponent: React.FC<{ template: TemplateDetails; filter: Filter }> = ({
+const FilterComponent: React.FC<{ template: TemplateDetails; filter: Filter; setLoading:any }> = ({
```
CarlosNZ, line 97: Try not to use `any` unless it's really necessary.

In this case the correct type for `setLoading` is `(loading: boolean) => void`

src/components/Dashboard.tsx
```diff
@@ -106,7 +120,7 @@ const FilterComponent: React.FC<{ template: TemplateDetails; filter: Filter }> =
       .map(([key, value]) => `${key}=${value}`)
       .join('&')}`
 
-  if (loading) return null
+  // if (loading) return <LoadingSmall/>
```
CarlosNZ, line 123: Can remove this comment now.

src/components/Dashboard.tsx
```diff
@@ -57,6 +58,11 @@ const TemplateComponent: React.FC<{ template: TemplateInList }> = ({ template })
       ? USER_ROLES.APPLICANT
       : USER_ROLES.REVIEWER
 
+  const [loadingFilters, setLoadingFilters] = useState(false)
+  const setLoading = (loadingState:boolean) => {
+    setLoadingFilters(loadingState)
+  }
+
```
CarlosNZ, line 65: Do you need this extra function? You should be able to just pass `setLoadingFilters` down to each component.

src/components/LoadingSmall.tsx
```diff
@@ -0,0 +1,16 @@
...
+import { Loader, Segment } from 'semantic-ui-react'
+import { useLanguageProvider } from '../contexts/Localisation'
+
+const LoadingSmall: React.FC = () => {
+  const { strings } = useLanguageProvider()
+  return (
+    <Segment basic style={{ height: '50%', margin: 0}}>
+      <Loader active size="mini">
+        {/* {strings.LABEL_LOADING} */}
+      </Loader>
```
CarlosNZ, line 11: Without the inner string, we can just make this a self-closing tag:
```
<Loader active size="small" />
```

Also -- I think "small" is better in this case, when compared with the block that appears. However, I can see that "mini" or "tiny" will probably be better when you get to the "Assignment" page lines, so why not make it a prop with a default.

So all up component would be something like:

```ts
const LoadingSmall: React.FC<{ size?: SemanticSIZES }> = ({ size = 'small' }) => {
  return (
    <Segment basic>
      <Loader active size={size} />
    </Segment>
  )
}
```

CarlosNZ (changes requested):
Hey, looks good thanks.

I've made a few small comments, but the main problem I can see is that `loadingFilters` filters will get set to `true` as soon as the *first* filter is finished loading, but we want it to wait for *all* of them before it disappears.

Have a think about how you might achieve that, and come and have a chat if you want to talk it through.

Cheers.

src/components/List/ApplicationsList.tsx
fergie-nz: Sounds good

src/components/Dashboard.tsx
```diff
@@ -10,6 +10,7 @@ import useListTemplates from '../utils/hooks/useListTemplates'
 import usePageTitle from '../utils/hooks/usePageTitle'
 import { TemplateDetails, TemplateInList } from '../utils/types'
 import LoadingSmall from './LoadingSmall'
+import { filter } from 'lodash'
```
CarlosNZ, line 13: I don't think you meant to import this.

src/components/Dashboard.tsx
```diff
@@ -58,7 +59,7 @@ const TemplateComponent: React.FC<{ template: TemplateInList }> = ({ template })
       ? USER_ROLES.APPLICANT
       : USER_ROLES.REVIEWER
 
-  const [loadingFilters, setLoadingFilters] = useState(false)
+  const [loadingFilters, setLoadingFilters] = useState<number>(0)
```
CarlosNZ, line 62: FYI -- when initialising a variable, if it's type can be inferred, you don't need to specify it.
So: `const [loadingFilters, setLoadingFilters] = useState(0)` is fine here. Typescript will enforce it as `number` elsewhere.

CarlosNZ: Also, maybe name it to something a little more specific -- `loadedFiltersCount` perhaps?

If you're not already familiar with it, the VSCode command "Rename symbol" (right-click menu) is super-useful for renaming variables.

src/components/Dashboard.tsx
```diff
@@ -91,12 +92,10 @@ const TemplateComponent: React.FC<{ template: TemplateInList }> = ({ template })
   )
 }
 
-const FilterComponent: React.FC<{ template: TemplateDetails; filter: Filter; setLoading:(loading:boolean) => void; filters: Array<{}>; i:number }> = ({
+const FilterComponent: React.FC<{ template: TemplateDetails; filter: Filter; setLoadingFilters:Function; filters: Array<{}>; loadingFilters: number}> = ({
```
CarlosNZ, line 95: You don't need to pass `loadingFilters` down to the FilterComponent. That component doesn't need to know the overall state of the filters, it just needs a mechanism to update it, which it does through `setLoadingFilters`.

src/components/Dashboard.tsx
```diff
@@ -106,9 +105,7 @@ const FilterComponent: React.FC<{ template: TemplateDetails; filter: Filter; set
   })
 
   useEffect(() => {
-    if (i+1 === filters.length) {
-    setLoading(loading)
-    }
+    if (loading === false) setLoadingFilters((loadingFilters: number) => loadingFilters + 1)
```
CarlosNZ, line 108: As I mentioned above, you can remove "loadingFilters" from the props. So the "loadingFilters" value here is just the parameter for the function you're sending to `setLoadingFilters`, so can be any name.

I'd suggest: `setLoadingFilters((currentCount) => currentCount + 1)`

src/components/Dashboard.tsx
```diff
@@ -69,12 +70,12 @@ const TemplateComponent: React.FC<{ template: TemplateInList }> = ({ template })
...
             </a>
             <Icon name="chevron right" />
-            {loadingFilters && 
+            {loadingFilters !== filters.length &&
               <LoadingSmall/>
             }
           </Label>
           {filters.map((filter, i:number, filters) => (
-            <FilterComponent key={filter.id} template={template} filter={filter} setLoading={setLoadingFilters} filters={filters} i={i}/>
+            <FilterComponent key={filter.id} template={template} filter={filter} setLoadingFilters={setLoadingFilters} filters={filters} loadingFilters={loadingFilters}/>
```
CarlosNZ, line 78: Also -- I don't think you've got your Prettier auto-formatting configured correctly. Cos once this auto-formats, it should look like:
```ts
         {filters.map((filter, i: number, filters) => (
            <FilterComponent
              key={filter.id}
              template={template}
              filter={filter}
              setLoadingFilters={setLoadingFilters}
              filters={filters}
            />
          ))}
```

Let me know if you can't get it working -- I suggest having "Format on Save" enabled then you never need to think about it.

src/components/Dashboard.tsx
CarlosNZ: Type for "setLoadingFilters" should be something like `setLoadingFilters: (setState: Function) => void`
And you can remove the "filters" as mentioned earlier.

CarlosNZ (changes requested):
Looking good, and working as expected.

Just a few little refinements I've suggested. Let me know if you need further explanation/clarification.

src/components/Dashboard.tsx
fergie-nz: That's great, thanks Carl. I very much appreciate the tips and tricks - I hadn't heard of that one before.

src/components/LoadingSmall.tsx
```diff
@@ -0,0 +1,14 @@
...
+import { useLanguageProvider } from '../contexts/Localisation'
+import { SemanticSIZES } from 'semantic-ui-react'
+
+const LoadingSmall: React.FC<{size?: SemanticSIZES}> = ({size ='small'}) => {
+  return (
+    <Segment basic style={{ height: '50%', margin: 0}}>
+      <Loader active size={size} />
+    </Segment>
+  )
+}
```
CarlosNZ, line 12: You've lost the changes you made here.

src/components/List/ApplicationsList.tsx
```diff
@@ -14,6 +14,8 @@ interface ApplicationsListProps {
   refetch: Function
 }
 
+
+
```
CarlosNZ, line 18: Revert

src/components/Dashboard.tsx
```diff
@@ -66,9 +69,16 @@ const TemplateComponent: React.FC<{ template: TemplateInList }> = ({ template })
...
-          {filters.map((filter) => (
-            <FilterComponent key={filter.id} template={template} filter={filter} />
+          {filters.map((filter, i: number, filters) => (
+            <FilterComponent
+              key={filter.id}
+              template={template}
+              filter={filter}
+              setFiltersCount={setLoadedFiltersCount}
+              filters={filters}
+            />
```
CarlosNZ, line 81:
```suggestion
          {filters.map((filter) => (
            <FilterComponent
              key={filter.id}
              template={template}
              filter={filter}
              setFiltersCount={setLoadedFiltersCount}
            />
```
Not using index or filters now.

src/components/Dashboard.tsx
```diff
@@ -85,17 +95,22 @@ const TemplateComponent: React.FC<{ template: TemplateInList }> = ({ template })
...
 }
 
-const FilterComponent: React.FC<{ template: TemplateDetails; filter: Filter }> = ({
-  template,
-  filter,
-}) => {
+const FilterComponent: React.FC<{
+  template: TemplateDetails
+  filter: Filter
+  setFiltersCount: (setState: Function) => void
```
CarlosNZ, line 101: Sorry, I put you wrong with the type here, this doesn't compile.  For now, just use `setFiltersCount: Function`

src/components/Dashboard.tsx
```diff
@@ -85,17 +95,22 @@ const TemplateComponent: React.FC<{ template: TemplateInList }> = ({ template })
...
+}> = ({ template, filter, setFiltersCount: setLoadingFilters }) => {
   const templateType = template.code
   const { loading, applicationCount } = useListApplications({
     type: templateType,
     perPage: 1,
     ...filter.query,
   })
 
+  useEffect(() => {
+    if (loading === false) setLoadingFilters((currentCount: number) => currentCount + 1)
```
CarlosNZ, line 111:
```suggestion
    if (!loading) setLoadingFilters((currentCount: number) => currentCount + 1)
```
Slightly more typical

CarlosNZ (commented):
Great, nearly there -- just a couple of minor comments.

src/components/Dashboard.tsx
```diff
@@ -66,9 +69,15 @@ const TemplateComponent: React.FC<{ template: TemplateInList }> = ({ template })
               {template?.namePlural || `${name} ${strings.LABEL_APPLICATIONS}`}
             </a>
             <Icon name="chevron right" />
+            {loadedFiltersCount !== filters.length && <LoadingSmall />}
           </Label>
-          {filters.map((filter) => (
-            <FilterComponent key={filter.id} template={template} filter={filter} />
+          {filters.map((filter: { id: any }) => (
```
CarlosNZ, line 74:
```suggestion
          {filters.map((filter) => (
```
This doesn't actually compile like this, as the destructuring makes the parent "filter" not available, but "filter" is required for the filter prop. (And you're not actually using this id anywhere)

src/components/LoadingSmall.tsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react'
+import { Loader, Segment } from 'semantic-ui-react'
+import { SemanticSIZES } from 'semantic-ui-react'
+
+const LoadingSmall: React.FC<{size?: SemanticSIZES}> = ({size ='small'}) => {
+  return (
+    <Segment basic style={{ height: '50%', margin: 0}}>
```
CarlosNZ, line 7:
```suggestion
    <Segment basic>
```
Remember we'd already changed this.

CarlosNZ (commented):
Awesome :) 
Just those two tiny little changes and we're good to go.

I don't love how the spinner takes up space and then the boxes get smaller once loaded. I know you've had some ideas about that, so it'd be good to have a chat about that when you're in next week.

src/components/LoadingSmall.tsx
fergie-nz: Thanks Carl. I managed to lose those changes, and was unable to find them again.

CarlosNZ (approved):
Great, thanks a lot. Can merge this now.

Can you also go over to the issue page (#1261) and tick off the first two items on the list (and write PR#1262 next to them) so we can keep track of progress. Thanks.

Issue #1264, opened 2022-08-01
Await trigger on application restart and allow view history when last answer is null


andreievg:
Noticed a couple of things when reviewing https://github.com/openmsupply/conforma-web-app/pull/1235
* When restarting application (from LOQ), we don't await trigger (brings you straight to summary)
* When viewing history for element where last answer is null (i.e. always reviewable element which went trough loq a few times and the last time value was remove, set to null), element history fails to load

PR #1265, opened 2022-08-01
1264 await trigger on application restart and allowe view history when last answer is null

andreievg:
closes: #1264


CarlosNZ (approved):
Great, thanks.

Issue #1242, opened 2022-07-18
#1256 Add "Other" option functionality to "Drop-down" element


CarlosNZ:
Key things:
- response shape needs to remain the same.
- use the existing "Allow additions" functionality of Semantic Drop-down: https://react.semantic-ui.com/modules/dropdown/#usage-allow-additions
- optional parameter (hasOther) needs to have default `false` so existing ones don't change
- see the "hasOther" functionality in "radioChoice" element and make the pattern consistent with that (as much as possible)

---

Also, check out but not if it's going to be too complicated (and do them in separate PRs):

- Allow multiple selections in Drop-down? 
- Add "other" option for Check-boxes

PR #1266, opened 2022-08-01
#1242 Drop-down custom option tweak

CarlosNZ:
@fergie-nz -- this is branched from your one so you can compare it against what you already had.

See inline comments for more info.

Was a bit trickier than I initially thought. Basic functionality not too bad, the hard part was getting it to recognise and display a custom option when loading, when the options list isn't immediately available -- e.g. from an API.

Here's an alt version of the feature showcase to test against: [Demo-FeatureShowcase_NEW.zip](https://github.com/openmsupply/conforma-web-app/files/9237502/Demo-FeatureShowcase_NEW.zip)
 
In particular -- check out the drop-down at the top of page 5, where it displays a list of names from public API. Now you can add your own custom option and it will still persist even when reloading the page and waiting for the options list.

@fergie-nz you can merge this into your branch when you're happy with it. Feel free to leave it till you're back in the office if you want to clarify anything with me.

### DIFF
--- a/docs/internal/Element-Type-Specs.md
+++ b/docs/internal/Element-Type-Specs.md
@@ -244,7 +244,7 @@ _Multi-choice question, with one allowed option, displayed as Drop-down list (Co
 - **default**: `string`/`number` -- if not provided, defaults to index 0.
 - **search**: `boolean` (default: `false`) -- if `true`, the list of options can be searched and filtered by user
 - **optionsDisplayProperty**: If `options` (above) consists of an array of objects, this parameter specifies the field of each object to be displayed in the options list. For example, if `options` was a list of organisation objects (i.e. `{orgId, name, licenceNumber}`), you'd probably specify `name` as the `optionsDisplayProperty`. Note that even though one field is displayed to the user in the Dropdown list, the _entire_ selected object is saved as the selection. And if `optionsDisplayProperty` refers to a field that doesn't exist on the supplied object, the plugin will fail and show in error in the application.
-- **hasOther**: `boolean` (default `false`) -- if `true`, displays an additional "Other" option with a free text field for inputting additional user-defined option.
+- **hasOther**: `boolean` (default `false`) -- if `true`, allows the user to enter a custom "free text" value instead of one of the pre-defined options.
 
 #### Response type
 
@@ -253,6 +253,7 @@ _Multi-choice question, with one allowed option, displayed as Drop-down list (Co
   optionIndex: <integer> (index from the options array)
   text: <string> (actual text from options array)
   selection: <string | object> (entire object or string from the supplied options list)
+  isCustomOption: <boolean> (if the selection is a custom "hasOther" option)
 }
 
 ```
--- a/src/formElementPlugins/dropdownChoice/localisation/default/strings.json
+++ b/src/formElementPlugins/dropdownChoice/localisation/default/strings.json
@@ -1,0 +1,3 @@
+{
+  "ADD_CUSTOM_OPTION_LABEL": "Add custom value: "
+}
--- a/src/formElementPlugins/dropdownChoice/src/ApplicationView.tsx
+++ b/src/formElementPlugins/dropdownChoice/src/ApplicationView.tsx
@@ -1,6 +1,12 @@
 import React, { useEffect, useState } from 'react'
 import { Dropdown, Label } from 'semantic-ui-react'
+import { ResponseFull } from '../../../utils/types'
 import { ApplicationViewProps } from '../../types'
+import { useLanguageProvider } from '../../../contexts/Localisation'
+
+type ObjectOption = { [key: string]: any }
+type ObjectOptions = { options: ObjectOption[]; optionsDisplayProperty: string }
+type StringOptions = { options: string[]; optionsDisplayProperty: never }
 
 const ApplicationView: React.FC<ApplicationViewProps> = ({
   element,
@@ -14,6 +20,8 @@ const ApplicationView: React.FC<ApplicationViewProps> = ({
   Markdown,
   getDefaultIndex,
 }) => {
+  const { getPluginStrings } = useLanguageProvider()
+  const strings = getPluginStrings('dropdownChoice')
   const {
     label,
     description,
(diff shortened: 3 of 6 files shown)


docs/internal/Element-Type-Specs.md
```diff
@@ -253,6 +253,7 @@ _Multi-choice question, with one allowed option, displayed as Drop-down list (Co
   optionIndex: <integer> (index from the options array)
   text: <string> (actual text from options array)
   selection: <string | object> (entire object or string from the supplied options list)
+  isCustomOption: <boolean> (if the selection is a custom "hasOther" option)
```
CarlosNZ, line 256: Not essential, but thought it might be useful to indicate in the response whether the selection is a custom value or not. An action might do something different depending on this.

src/formElementPlugins/dropdownChoice/localisation/default/strings.json
```diff
@@ -0,0 +1,3 @@
+{
+  "ADD_CUSTOM_OPTION_LABEL": "Add custom value: "
```
CarlosNZ, line 2: The "Add " text that comes up when adding a custom option needs to be localised. Follow the same pattern as localised strings in other form elements.

src/formElementPlugins/dropdownChoice/src/ApplicationView.tsx
```diff
@@ -1,6 +1,12 @@
 import React, { useEffect, useState } from 'react'
 import { Dropdown, Label } from 'semantic-ui-react'
+import { ResponseFull } from '../../../utils/types'
 import { ApplicationViewProps } from '../../types'
+import { useLanguageProvider } from '../../../contexts/Localisation'
+
+type ObjectOption = { [key: string]: any }
+type ObjectOptions = { options: ObjectOption[]; optionsDisplayProperty: string }
+type StringOptions = { options: string[]; optionsDisplayProperty: never }
```
CarlosNZ, line 9: The point of these types is that it allows us to tell typescript that an "optionsDisplayProperty" parameter should always be paired with an options list made of objects, and vice-versa.

src/formElementPlugins/dropdownChoice/src/ApplicationView.tsx
```diff
@@ -23,11 +31,17 @@ const ApplicationView: React.FC<ApplicationViewProps> = ({
...
+    placeholder?: string
+    search?: boolean
+    hasOther?: boolean
+    default: any
+  } & (ObjectOptions | StringOptions)
 
   const [selectedIndex, setSelectedIndex] = useState<number>()
-  //saving selected value in local state, so that when options are re rendered, we retain this value
-  const [addedOptions, setAddedOptions] = useState<Array<{}>>([])
+  const [addedOption, setAddedOption] = useState<string | null>(null)
```
CarlosNZ, line 44: Ended up being easier to keep the custom option as a single value rather than an array. If a new custom value is added, this just gets replaced.

src/formElementPlugins/dropdownChoice/src/ApplicationView.tsx
```diff
@@ -37,75 +51,60 @@ const ApplicationView: React.FC<ApplicationViewProps> = ({
...
-              ? dropDownOptions[optionIndex][optionsDisplayProperty]
-              : dropDownOptions[optionIndex]['text'],
-            selection: dropDownOptions[optionIndex],
-            optionIndex,
-          })
-        }
-        // Reset response if selection cleared
-        else onSave(null)
-  }
+  const handleChange = (value: number | string, addedText?: string) => {
```
CarlosNZ, line 75: The reason why I've added the second (optional) parameter here is because we want to run `handleChange` from `addItemHandler`. And because it runs as part of the same component render, the value of "addedOption" is not reliably available just after it's set. So we just pass it in instead.

In fact, we can almost do without having `addedOption` as a state value at all, since it can be generated from the response as part of "getDropdownOptions". The only time it's really needed is so its value can persist while another option is selected (I think).

src/formElementPlugins/dropdownChoice/src/ApplicationView.tsx
```diff
@@ -116,16 +115,22 @@ const ApplicationView: React.FC<ApplicationViewProps> = ({
...
         clearable
         search={search || hasOther}
         allowAdditions
-        onAddItem={(e, data)=>addItemHandler(e, data.value as string)}
+        additionLabel={strings.ADD_CUSTOM_OPTION_LABEL}
+        onAddItem={(e, data) => addItemHandler(data.value as string)}
         placeholder={placeholder}
-        options={dropDownOptions}
-        onChange={handleChange}
+        options={getDropdownOptions(
```
CarlosNZ, line 126: Making `getDropdownOptions` a separate function ensures it'll always reflect the most up-to-date state. I was having problems with it being updated when options was slow to load.

src/formElementPlugins/dropdownChoice/src/ApplicationView.tsx
```diff
@@ -137,3 +142,36 @@ const ApplicationView: React.FC<ApplicationViewProps> = ({
...
+) => {
+  // Figure out if the currentResponse contains an added custom response. This
+  // is just for initial loading if we've had to wait for the list of options to
+  // populate
+  const initialCustomResponse =
+    !addedOption &&
+    currentResponse?.optionIndex === options.length &&
+    typeof currentResponse.selection === 'string'
+      ? currentResponse.selection
+      : null
```
CarlosNZ, line 161: This is what allows the custom option to persist after a slow options load. When the component first loads, the saved response won't correspond to the options list. But after options loads, if the saved index is one higher than the options list size, then we assume it's meant to be a custom option for the current list.

fergie-nz: Nice. I'm wondering if the options dynamically changed to a list that was 1 shorter than the previous and the last option on the list was selected, the selected option could be preserved and brought over to the new list. However, that seems like a very unlikely case and no big deal if it does occur.

CarlosNZ: Yeah, exactly. I figured if and when the options list changes from under you, you're going to want to make a new selection anyway, so the current response becomes irrelevant. It's possible it could "auto-select" the extra option, but that problem's always been there -- if the list changes the index of your current selection will be selected in the new list.

PR #1269, opened 2022-08-03
#1268 Option to connect to remote server

### LINKED ISSUE #1268: Config option to connect to online server
I think it would be useful in development to be able to use a remote back-end when developing front-end. Some things are just easier to test with a live server, as you can pick up on bugs that might only occur with a realistic network speed as opposed to the almost-instant response we get from local server.

We can either use `.env` or another local (non-committed) file to put the server url in, and the existing "config.ts" should check this for a custom value, or default to its current value.

CarlosNZ:
Closes #1268 

Added Dotenv to parse .env file.

By default, there will be no .env file, so config will use the normal local server. You need to create your own as we don't commit it to the repo.

Example `.env` file:

```
REMOTE_SERVER=angolaTesting
SERVERS={"angolaTesting":"https://conforma-demo.msupply.org:50008","demoServer":"https://conforma-demo.msupply.org:50000"}
```

This will use the server defined in SERVERS, i.e. "angolaTesting" -> `https://conforma-demo.msupply.org:50008`

To revert back to local server, just change "REMOTE_SERVER" to an invalid value, e.g. `REMOTE_SERVER=NONE`


src/config.ts
```diff
@@ -2,9 +2,20 @@
 const isProductionBuild = process.env.NODE_ENV === 'production'
 const { version } = require('../package.json')
 
+// To connect to a remote server, store them in .env file
+// REMOTE_SERVER => name of server in SERVERS
+// SERVERS => JSON.stringified object {serverName: url, ...}
```
andreievg, line 7: I think it's worth adding to README.md.

andreievg (approved):
Looks good, a bit unconventional, typically `#` comment is used (maybe even quicker 🤷 ), and would include local dev server, i.e.
```env
#SERVER_URL="https://conforma.demo.msupply.org:50008" # angola testing
SERVER_URL="http://localhost:8080/api" # local dev
#SERVER_URL="https://conforma-demo.msupply.org:50000" # demo server
```

CarlosNZ:
> Looks good, a bit unconventional, typically `#` comment is used (maybe even quicker 🤷 ), and would include local dev server, i.e.

Yeah, that is better. Sorry, didn't realise you could put comments in .env file. (The standard "add comment" shortcut key didn't do anything).

I'll change it.

CarlosNZ:
@andreievg changed it to reflect your suggestion and added README note.  Thanks.
 
>  and would include local dev server

Didn't include localhost server cos there's two (REST/GraphQL), and the url construction is different for localhost than the live server. (live server has common base url, localhost has seperate ports.

You think it's okay to leave the demo server urls in the public README?

andreievg:
> You think it's okay to leave the demo server urls in the public README?

Thanks yeah, sounds great.

andreievg (approved):

PR #1273, opened 2022-08-08
Quick fix for faulty applicationData.config

CarlosNZ:
Have manually re-constructed the "config" object in applicationData, so it provides "serverREST" prop and "serverGraphQL" prop as before.


nmadruga (approved):
Is that what was causing the API problem on latest version?

CarlosNZ:
> Is that what was causing the API problem on latest version?

Yes.

Issue #1193, opened 2022-05-12
Bug: Filter by Assigner doesn't properly load the list of applications


nmadruga:

nmadruga:
Going to close this issue for now, since Assigner filter isn't in use. I added a comment about this on the file `applicationFilters`

PR #1275, opened 2022-08-09
Add default property to filters be displayed by default

### LINKED ISSUE #1274: Setting the filters to be static

nmadruga:
Fixes #1274 

Small change to allow these selected filters to display by default once the Applications list page is loaded

Screen Shot 2022-08-09 at 1 29 59 PM


CarlosNZ:
Looks good -- I'll test it shortly. 

Just one question -- how are you deciding which filters should be "default: true" ? Is there a list somewhere?

CarlosNZ:
> Just one question -- how are you deciding which filters should be "default: true" ? Is there a list somewhere?

Do we want to make it a preference that is saved in snapshot, so each installation can specify their preferred filters?

nmadruga:
> Looks good -- I'll test it shortly.
> 
> Just one question -- how are you deciding which filters should be "default: true" ? Is there a list somewhere?

Yep, I got from this doc: https://docs.google.com/document/d/1vsuJ7HGeQ-Ch8Hr1PhRoqiQayRkibdrpgYfZkhTBD2w

nmadruga:
> > Just one question -- how are you deciding which filters should be "default: true" ? Is there a list somewhere?
> 
> Do we want to make it a preference that is saved in snapshot, so each installation can specify their preferred filters?

That would require a bigger change, since we would need to move the `applicationFilters` to a config file and that would be on the back-end to get saved with the snapshot. Or do you though just as a config file in the front-end?

CarlosNZ:
> That would require a bigger change, since we would need to move the `applicationFilters` to a config file and that would be on the back-end to get saved with the snapshot. Or do you though just as a config file in the front-end?

I was thinking it would just be the same as how we're doing the logo. Stored in back-end `preferences.json`. Front-end checks the pref from the prefs context, or reverts to defaults (the ones you've specified) if not defined.

It's quite a simple change, I reckon it'd be worth it. But no worries if you want to leave it for now.

nmadruga:
> I was thinking it would just be the same as how we're doing the logo. Stored in back-end `preferences.json`. Front-end checks the pref from the prefs context, or reverts to defaults (the ones you've specified) if not defined.
> 
> It's quite a simple change, I reckon it'd be worth it. But no worries if you want to leave it for now.

I had a quick look and I think it's not worth for now. There are these properties called `options` to build each array of values that depends on a few functions in the Front-end, I will add an issue to have that done when we are working on a second project.

src/containers/List/ListFilters/ListFilters.tsx
```diff
@@ -23,6 +23,11 @@ const getDisplayableFilters = (filterDefinitions: FilterDefinitions) => {
   return result
 }
 
+const getDefaultDisplayFilters = (filterDefinitions: FilterDefinitions) =>
+  Object.entries(filterDefinitions)
+    .filter(([filterName, filterValue]) => filterValue.default)
+    .map(([filterName, filterValue]) => filterName)
```
CarlosNZ, line 29:
```suggestion
    .filter(([_, filterValue]) => filterValue.default)
    .map(([filterName, _]) => filterName)
```
For the unused fields on each entry

CarlosNZ (commented):
Looks good. Thanks.

One problem this highlights -- and it's not a problem with this PR specifically as it was already there, but we should also have filters restricted by user role. Like, if `user_role=applicant`, then they shouldn't be able to filter by reviewer/assigner etc. Basically, they shouldn't see a filter for a column they don't see, so perhaps we can tie the column definitions to the filter visibility?

You want to do that now, or make a new issue?

src/containers/List/ListFilters/ListFilters.tsx
```diff
@@ -31,9 +36,10 @@ const ListFilters: React.FC<{
   const { query, updateQuery } = useRouter()
 
   const displayableFilters = getDisplayableFilters(filterDefinitions)
+  const defaultDisplayFilters = getDefaultDisplayFilters(filterDefinitions)
 
   const filterNames = Object.keys(displayableFilters)
-  const [activeFilters, setActiveFilters] = useState<string[]>([])
+  const [activeFilters, setActiveFilters] = useState<string[]>(defaultDisplayFilters)
```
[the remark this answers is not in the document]
nmadruga: I rather keep this way. I think nested things a little harder to read

nmadruga:
> One problem this highlights -- and it's not a problem with this PR specifically as it was already there, but we should also have filters restricted by user role. Like, if `user_role=applicant`, then they shouldn't be able to filter by reviewer/assigner etc. Basically, they shouldn't see a filter for a column they don't see, so perhaps we can tie the column definitions to the filter visibility?
> 
> You want to do that now, or make a new issue?

That's a good idea. I was going to quickly translate the other template - GrantUserPermission and come back to this. I'll see if this can be done as part of this PR indeed. I also want to keep looking at that problem in issue #1193 so will keep looking on the same place

src/containers/List/ListFilters/ListFilters.tsx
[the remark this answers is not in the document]
CarlosNZ: Okay, but you can still lose the type definition now.

nmadruga:
For testing @CarlosNZ, just add this to your `preferences.json` on the backend please:

src/containers/List/ListWrapper.tsx
```diff
@@ -8,18 +9,20 @@ import { useLanguageProvider } from '../../contexts/Localisation'
...
 import ListFilters from './ListFilters/ListFilters'
 import { useApplicationFilters } from '../../utils/data/applicationFilters'
 import Loading from '../../components/Loading'
+import { usePrefs } from '../../contexts/SystemPrefs'
 
 const ListWrapper: React.FC = () => {
   const { strings } = useLanguageProvider()
-  const APPLICATION_FILTERS = useApplicationFilters()
+  const { preferences } = usePrefs()
+  const APPLICATION_FILTERS = useApplicationFilters(preferences?.defaultListFilters || [])
```
CarlosNZ, line 25: I think the fallback should be a couple of the more common filters, not empty array. Cos we want some defaults even on older snapshots and on snapshots that don't have anything defined in prefs.

I'd imagine that if the hard-coded defaults are fine for most cases, adding to the preferences won't be necessary, it's only if they want something different.

src/contexts/SystemPrefs.tsx
```diff
@@ -9,6 +9,7 @@ interface PrefsState {
     defaultLanguageCode: string
     brandLogoFileId?: string
     brandLogoOnDarkFileId?: string
+    defaultListFilters: string[]
```
CarlosNZ, line 12:
```suggestion
    defaultListFilters?: string[]
```
Should be optional

src/utils/helpers/list/mapColumnsByRole.ts
```diff
@@ -89,14 +96,19 @@ export const useMapColumnsByRole = () => {
...
     REVIEWER_ACTION: {
       headerName: strings.COLUMN_REVIEWER_ACTION,
-      sortName: 'outcome',
+      sortName: 'reviewer-action',
       ColumnComponent: ReviewerActionCell,
     },
     APPLICANT_ACTION: {
       headerName: strings.COLUMN_APPLICANT_ACTION,
-      sortName: 'outcome',
+      sortName: 'applicant-action',
```
CarlosNZ, line 111: I think this might be the problem. We don't want to change the sorting. The "Reviewer action" column (etc.). shouldn't be sortable.

src/containers/List/ListWrapper.tsx
```diff
@@ -105,7 +108,17 @@ const ListWrapper: React.FC = () => {
     }
   }
 
-  if (loading ) return <Loading />
+  if (loading) return <Loading />
+
+  let filterDefinitions: FilterDefinitions = {}   
+  
+  columns.forEach((column) => {
+    const filter = Object.entries(APPLICATION_FILTERS).find(([filterName, _]) => camelCase(column.sortName) === filterName)
```
CarlosNZ, line 116: Yeah, I think the problem is that you've changed the "sortName" for some of the columns which has messed up how the list looks/works.

And it would be better if this logic was hidden away in the hook it relates to rather than here. You might be able to run the function to check whether filter should be visible in the `useApplicationFilters` hook itself.

So each filterDefinition could be something like:
```
  applicantDeadline: {
      type: 'date',
      default: defaultFilters.some((filter) => filter === 'applicantDeadline'),
      visible: checkFilterVisibility("DEADLINE_DATE")
      title: strings.FILTER_APPLICANT_DEADLINE,
      options: { namedDates: NAMED_DATE_RANGES },
    },
```

And then create a function `checkFilterVisibility` to do the logic you've done here, ie. compare against column visibility. And then you don't need to match by "sortName" (which causes the problems I described elsewhere), you can just directly pass the column to compare against to "checkFilterVisibility" for each filter.

CarlosNZ (changes requested):
Couple of immediate things I noticed:

- we don't want to add more columns to the table. Now the "consolidator" column is showing up, which looks weird and unnecessary.
- when in "applicant" mode, the "Stage" filter is not available at all.
- there's this weird arrow that has appeared over the last columns which looks odd:
Screen Shot 2022-08-10 at 10 20 20 AM
- basically the look and functionality of the list itself should not be changing

I think the problem is caused by chaning the "sortName" on some of the columns. See my comments for possible solution.

src/containers/List/ListWrapper.tsx
nmadruga: Thanks Carl, I didn't think this was a very good solution either. I was trying to make it without many changes... I still don't get one thing though... I'll check with you before starting on this

nmadruga: Ok! Quite a lot of changes... sorry! But the result of what you and I talked is there. Hopefully much more like what was envisioned at the start : )

src/containers/List/ListWrapper.tsx
```diff
@@ -1,5 +1,6 @@
 import React, { useEffect, useState } from 'react'
 import { Label, Button, Search, Header, Icon } from 'semantic-ui-react'
+import { camelCase } from 'lodash'
```
CarlosNZ, line 4: Can remove now

src/containers/List/ListWrapper.tsx
```diff
@@ -8,18 +9,20 @@ import { useLanguageProvider } from '../../contexts/Localisation'
 import { findUserRole, checkExistingUserRole } from '../../utils/helpers/list/findUserRole'
 import { useUserState } from '../../contexts/UserState'
 import { useMapColumnsByRole } from '../../utils/helpers/list/mapColumnsByRole'
-import { ApplicationListRow, ColumnDetails, SortQuery } from '../../utils/types'
+import { ApplicationListRow, ColumnDetails, FilterDefinitions, SortQuery } from '../../utils/types'
```
CarlosNZ, line 12: `FilterDefinitions` no longer used

src/containers/List/ListWrapper.tsx
```diff
@@ -8,18 +9,20 @@ import { useLanguageProvider } from '../../contexts/Localisation'
...
-import { useApplicationFilters } from '../../utils/data/applicationFilters'
+import { useGetFilterDefinitions } from '../../utils/helpers/list/useGetFilterDefinitions'
 import Loading from '../../components/Loading'
+import { usePrefs } from '../../contexts/SystemPrefs'
 
 const ListWrapper: React.FC = () => {
   const { strings } = useLanguageProvider()
-  const APPLICATION_FILTERS = useApplicationFilters()
+  const { preferences } = usePrefs()
+  const FILTER_DEFINITIONS = useGetFilterDefinitions(preferences?.defaultListFilters || [])
```
CarlosNZ, line 25: Better I think to have ` const { preferences } = usePrefs()` in `useGetFilterDefinitions` rather than passing it in here. Hide all that stuff in the hook rather than cluttering this component, I'd say.

src/utils/data/columnsPerUserRole.ts
```diff
@@ -1,28 +1,28 @@
...
-    APPLICATION_COLUMNS.STATUS,
-    APPLICATION_COLUMNS.OUTCOME,
-    APPLICATION_COLUMNS.APPLICANT_ACTION,
+    LIST_COLUMNS.SERIAL_NUMBER,
+    LIST_COLUMNS.LAST_ACTIVE_DATE,
+    LIST_COLUMNS.DEADLINE_DATE,
+    LIST_COLUMNS.APPLICATION_NAME,
+    LIST_COLUMNS.STATUS,
+    LIST_COLUMNS.OUTCOME,
+    LIST_COLUMNS.APPLICANT_ACTION,
```
CarlosNZ, line 16: Good idea to rename this!

src/utils/defaultLanguageStrings.ts
```diff
@@ -76,13 +76,15 @@ export default {
...
-  COLUMN_APPLICANT: 'Applicant',
+  COLUMN_APPLICANT: '',
   COLUMN_ORGANISATION: 'Organisation',
+  COLUMN_REVIEWER: '',
   COLUMN_CONSOLIDATOR: 'Consolidator',
   COLUMN_STAGE: 'Stage',
   COLUMN_STATUS: 'Status',
   COLUMN_OUTCOME: 'Outcome',
   COLUMN_REVIEWER_ACTION: '',
+  COLUMN_ASSIGNER_ACTION: '',
```
CarlosNZ, line 87: Just wondering why the empty strings?

If we're not using them, why not remove them completely, or if you want to keep them around for completion sake, then we may as well have proper values in them, no?

src/utils/helpers/list/useGetFilterDefinitions.ts
```diff
@@ -1,21 +1,23 @@
...
+import getOrganisationFilterList from '../../graphql/queries/applicationListFilters/getOrganisationFilterList'
+import getReviewersFilterList from '../../graphql/queries/applicationListFilters/getReviewersFilterList'
+import getStageFilterList from '../../graphql/queries/applicationListFilters/getStageFilterList'
+import { FilterDefinitions, GetFilterListQuery, NamedDates } from '../../types'
+import { useLanguageProvider } from '../../../contexts/Localisation'
+import { USER_ROLES } from '../../data'
+import LIST_FILTERS from '../../data/listFilters'
 
-export const useApplicationFilters = () => {
+export const useGetFilterDefinitions = (defaultFilters: string[]) => {
```
CarlosNZ, line 20: As suggested above, I think it would be better to read the prefs inside here rather than pass in a parameter.

src/utils/helpers/list/useGetFilterDefinitions.ts
```diff
@@ -109,19 +110,25 @@ export const useApplicationFilters = () => {
...
+      visibleTo: checkFilterVisibility("STAGE"),
       title: strings.FILTER_STAGE,
       options: {
         getListQuery: constructFilterListQuery(getStageFilterList, 'applicationListFilterStage'),
       },
     },
     search: {
       type: 'search',
+      default: defaultFilters.some((filter) => filter === 'search'),
+      visibleTo: ['applicant' as USER_ROLES, 'reviewer' as USER_ROLES], // Always Visible
```
CarlosNZ, line 123:
```suggestion
      visibleTo: [USER_ROLES.APPLICANT, USER_ROLES.REVIEWER], // Always Visible
```

CarlosNZ: Also, personally I like this definition way more. I think it would be better if they were all like this, I don't see why we need a whole extra map and function. This is super readable and clear, why complicate it?

src/utils/helpers/list/useGetFilterDefinitions.ts
```diff
@@ -197,3 +225,32 @@ const getDateRangeForUnit: GetDateRangeForUnit = (unit, value = 0) => {
...
+    LIST_FILTERS.OUTCOME,
+    LIST_FILTERS.STAGE,
+    LIST_FILTERS.ORGANISATION,
+    LIST_FILTERS.APPLICANT,
+    LIST_FILTERS.ASSIGNERS,
+    LIST_FILTERS.REVIEWERS,
+    LIST_FILTERS.REVIEWER_ACTION,
+  ]
+  if (visibleToReviewer.find((visibleFilter) => visibleFilter === filterKey)) filtersAreVisibleTo.push(USER_ROLES.REVIEWER)
+  return filtersAreVisibleTo
```
CarlosNZ, line 255: This seems like a lot of unnecessary bloat when they could all be as simple as the "search" filter above:
```
visibleTo: [USER_ROLES.APPLICANT, USER_ROLES.REVIEWER],
```

Why do you think we need all this?

src/utils/hooks/useListApplications.tsx
```diff
@@ -14,7 +15,8 @@ const useListApplications = ({
   type,
   ...queryFilters
 }: BasicStringObject) => {
-  const APPLICATION_FILTERS = useApplicationFilters()
+  const { preferences } = usePrefs()
+  const FILTER_DEFINITIONS = useGetFilterDefinitions(preferences?.defaultListFilters || [])
```
CarlosNZ, line 19: As per my above comment, you don't need these pref imports here if you do it all in the hook.

CarlosNZ (changes requested):
Cool, it's working well now. See inline comments for some minor tweaks, but basically:

- I really don't see the need for the whole "checkFilterVisibility" function. They way you did it for the "search" filter is just fine, and way easier to understand.
- When there is no defaults defined by the back-end, we should have some hard-coded defaults, not an empty array.
- Why is Stage not showing up for the applicant? I guess that's something that Angola requested, but it seems strange to hard-code that into the app. Perhaps column and filter visibility should also be defined in preferences -- though that can wait for another day :)

CarlosNZ:
Also -- one other little bug I just noticed. Not a big deal, so don't go too far trying to fix it, just if it's simple.

The behaviour of "Clear all filters" is changing depending on whether the default ones are there or not.

To see what I mean:
- reload the page and have your default filters showing up.
- if you click "Clear all filters" now, they disappear
- reload the page, get your defaults back
- make some changes (add new filter), or select a value in one of your filters
- if you click "Clear all filters" now, the reset back to your defaults, not all gone.

I actually don't know which one I think should be the "expected" behaviour, but they should probably be consistent if possible. I'm leaning towards always going back to your defaults, but I'm open to persuasion. Although, if that's the case, I'd probably want to make the label say "Reset filters" rather than "Clear all filters".

src/utils/helpers/list/useGetFilterDefinitions.ts
```diff
@@ -59,46 +61,45 @@ export const useApplicationFilters = () => {
...
     },
     outcome: {
       type: 'enumList',
+      default: defaultFilters.some((filter) => filter === 'outcome'),
+      visibleTo: checkFilterVisibility("OUTCOME"),
       title: strings.FILTER_OUTCOME,
       options: { enumList: Object.values(ApplicationOutcome) },
     },
-    orgName: {
+    org: {
```
CarlosNZ, line 99: Oh -also this breaks the list -- the query to the back-end no longer works, so we'll need to change this back to "orgName"

nmadruga:
Before I have a look in each point, here is my initial thoughts on these points:

> * I really don't see the need for the whole "checkFilterVisibility" function. The way you did it for the "search" filter is just fine, and way easier to understand.

Ok - I'll have a look and try to change that

> * When there is no defaults defined by the back-end, we should have some hard-coded defaults, not an empty array.

I think if there is no one defined we can leave as is, since it's argue-able

> * Why is Stage not showing up for the applicant? I guess that's something that Angola requested, but it seems strange to hard-code that into the app. Perhaps column and filter visibility should also be defined in preferences -- though that can wait for another day :)

Yes, that was what I thought you had suggested originally, and have answered about that... Anyway I think we can leave this one for now as well. Ok to improve in another delivery

src/containers/List/ListWrapper.tsx
nmadruga: Hmmm Ok - I'll have a go

src/utils/defaultLanguageStrings.ts
nmadruga: Oh, yea. Just forgot to remove these. There are some like that though, but it's because we want to have the column, and a title is expected, but it doesn't display any text currently...

src/containers/List/ListWrapper.tsx
nmadruga: Nah, not simple to do. I'll keep as is

src/utils/helpers/list/useGetFilterDefinitions.ts
nmadruga: Yes, it could be better... But there is a thing with getting the userRole.

src/utils/helpers/list/useGetFilterDefinitions.ts
nmadruga: Oh, totally agreed!

src/utils/helpers/list/useGetFilterDefinitions.ts
nmadruga: Sorry, not sure what is your suggestion on this one : (

src/containers/List/ListWrapper.tsx
CarlosNZ: You must have misunderstood what I was suggesting -- it's literally 2 minutes work and is much tidier. Please see comparison: https://github.com/openmsupply/conforma-web-app/compare/%231274-show-default-display-filters-on-top...%231274-show-default-display-filters-on-top-change-hook-input

nmadruga:
> The behaviour of "Clear all filters" is changing depending on whether the default ones are there or not.

Ok - I think it's fixed now. Thanks

src/utils/defaultLanguageStrings.ts
CarlosNZ: Okay cool, that makes sense.

src/utils/helpers/list/useGetFilterDefinitions.ts
CarlosNZ: It's the same suggestion as above -- you don't need "checkFilterVisiblity" at all.

src/utils/helpers/list/useGetFilterDefinitions.ts
CarlosNZ: > But there is a thing with getting the userRole

What thing? 😕

src/containers/List/ListWrapper.tsx
nmadruga: Very good. Thanks!

src/utils/helpers/list/useGetFilterDefinitions.ts
```diff
@@ -1,23 +1,29 @@
...
+import { USER_ROLES } from '../../data'
+import { usePrefs } from '../../../contexts/SystemPrefs'
 
-export const useApplicationFilters = () => {
+export const useGetFilterDefinitions = () => {
   const { strings } = useLanguageProvider()
 
+  const { preferences } = usePrefs()
+
+  const defaultFilters = preferences?.defaultListFilters || []
```
CarlosNZ, line 25: You can leave this for now, but I just wanted to point out that having `[]` as the fallback instead of some hard-coded defaults means we're making more work for ourselves. Whenever we upgrade a server, Jess is going to ask "why are the default filters not showing up?", and then we're going to have to go through every instance and manually add "defaultFilters" to their prefs file.

If you're happy to do that, that's fine, I was just trying to avoid having to re-edit existing prefs files unless they needed to be different to the agreed "standard".

src/containers/List/ListFilters/ListFilters.tsx
```diff
@@ -65,7 +71,7 @@ const ListFilters: React.FC<{
 
   const resetFilters = () => {
     updateQuery(Object.fromEntries(activeFilters.map((filterName) => [filterName, undefined])))
-    setActiveFilters([])
+    setActiveFilters(defaultDisplayFilters)
```
CarlosNZ, line 74: Thanks for fixing this. 

Just one tiny nit -- can you change the string for the label from "Clear all filters" to "Reset all filters", as "clear" doesn't quite make sense now if we're leaving the default ones up.

src/containers/List/ListWrapper.tsx
```diff
@@ -14,12 +14,12 @@ import { Link } from 'react-router-dom'
 import ApplicationsList from '../../components/List/ApplicationsList'
 import PaginationBar from '../../components/List/Pagination'
 import ListFilters from './ListFilters/ListFilters'
-import { useApplicationFilters } from '../../utils/data/applicationFilters'
+import { useGetFilterDefinitions } from '../../utils/helpers/list/useGetFilterDefinitions'
```
CarlosNZ, line 17: Tiny -- but there's a blank link in the middle of the imports (line 3). Might want to remove.

nmadruga:
POk - after our chat this morning @CarlosNZ I'll be merging this feature in

CarlosNZ (approved):
Great, thanks for this. All looking good now.

I made a couple of minor comments you might want to consider, but nothing important.

PR #1280, opened 2022-08-10
#1278 Password reset fix (again!)

### LINKED ISSUE #1278: Bug: Password reset still not working properly
Even after attempted fix #1259 by @andreievg, the same problem persists.

Need to investigate and fix please!

CarlosNZ:
Fix #1278

Yikes, this was a bit gnarly. @andreievg I'd be interested to see what you think and if there's a better way to go about it.

Basically, the problem with the solution you made a couple of weeks ago is that, if `state` is not present at the beginning, then even though you accounted for the case, your solution relied on "location" changing and forcing a re-render of the `AuthenticatedWrapper` component. Unfortunately, this doesn't happen, and so it just sits there perpetually re-loading.

Seems like whatever bug/issue is causing the "location" object to be missing "state", and "query" on load is also not notifying React of a change when it *is* available.

So I went with a blunt solution -- ignore React Router completely and DIY it with my own function to manually parse `window.location`. It works, but it sucks that we need to resort to this -- surely there's a better way? 

Couple of tiny offroad fixes too. See inline comments for more.

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
     "react-json-view": "^1.21.3",
     "react-markdown": "^5.0.3",
     "react-progress-stepper": "^0.2.2",
-    "react-router-dom": "^5.2.0",
+    "react-router-dom": "^5.3.3",
     "react-semantic-ui-datepickers": "^2.13.1",
     "react-string-replace": "^0.4.4",
     "serve": "^11.3.2"
--- a/src/components/Review/DecisionPreview/ReviewPreviewModal.tsx
+++ b/src/components/Review/DecisionPreview/ReviewPreviewModal.tsx
@@ -44,7 +44,7 @@ const ReviewPreviewModal: React.FC<PreviewProps> = ({
     // generates new Previews every time it's opened
     <Modal id="preview-modal" open={open} closeOnDimmerClick={false}>
       <Modal.Header>{strings.REVIEW_DECISION_PREVIEW_HEADER}</Modal.Header>
-      <Modal.Content>
+      <Modal.Content scrolling>
         {strings.REVIEW_DECISION_PREVIEW_TEXT} <strong>{decision}</strong>
         {loading && (
           <Loader active size="huge">
--- a/src/containers/Main/AuthenticatedWrapper.tsx
+++ b/src/containers/Main/AuthenticatedWrapper.tsx
@@ -1,21 +1,18 @@
-import React, { useEffect } from 'react'
+import React from 'react'
 import SiteLayout from './SiteLayout'
 import { useRouter } from '../../utils/hooks/useRouter'
 import isLoggedIn from '../../utils/helpers/loginCheck'
 import NonRegisteredLogin from '../User/NonRegisteredLogin'
 import { Redirect } from 'react-router'
-import { Loading } from '../../components'
 
 const AuthenticatedContent: React.FC = () => {
-  const {
-    location,
-    query: { sessionId },
-  } = useRouter()
-  // Location is not fully loaded unless 'state' is present, TODO try upgrading react router
-  // without this sessionId is missing when navigating via Click Here in forgot password verification
-  // if ('state' in location) {
+  const { location } = useRouter()
+
+  const sessionId = getSessionIdFromUrl()
+
   console.log('Location state', location)
   console.log('sessionId', sessionId)
+
   let { pathname, search } = location
   // If there is a sessionId in the URL, then need to login as nonRegistered
   // before continuing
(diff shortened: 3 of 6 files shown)


package.json
```diff
@@ -39,7 +39,7 @@
     "react-json-view": "^1.21.3",
     "react-markdown": "^5.0.3",
     "react-progress-stepper": "^0.2.2",
-    "react-router-dom": "^5.2.0",
+    "react-router-dom": "^5.3.3",
```
CarlosNZ, line 42: I hoped updating React Router would help. Alas it doesn't, but may as well leave this here. This is as high as we can go before v6, which would require a major re-factor as it's very different.

src/containers/Main/AuthenticatedWrapper.tsx
```diff
@@ -4,25 +4,35 @@ import { useRouter } from '../../utils/hooks/useRouter'
...
-import { Loading } from '../../components'
 
 const AuthenticatedContent: React.FC = () => {
-  const {
-    location,
-    query: { sessionId },
-  } = useRouter()
-  // Location is not fully loaded unless 'state' is present, TODO try upgrading react router
-  // without this sessionId is missing when navigating via Click Here in forgot password verification
-  if ('state' in location) {
```
CarlosNZ, line 16: This was how @andreievg tried to fix the problem initially. However, it doesn't work fully, as the "query" (and "state") props never refresh, so they never actually get a value.

src/components/Review/DecisionPreview/ReviewPreviewModal.tsx
```diff
@@ -44,7 +44,7 @@ const ReviewPreviewModal: React.FC<PreviewProps> = ({
     // generates new Previews every time it's opened
     <Modal id="preview-modal" open={open} closeOnDimmerClick={false}>
       <Modal.Header>{strings.REVIEW_DECISION_PREVIEW_HEADER}</Modal.Header>
-      <Modal.Content>
+      <Modal.Content scrolling>
```
CarlosNZ, line 47: Fixes the overflow problem noticed by @nmadruga when content was too long. This makes it scroll within the modal.

src/utils/helpers/endpoints/types.ts
```diff
@@ -40,7 +40,7 @@ export type CheckUniqueEndpoint = [
         table: string
         field: string
       }
-  ) & { value: string }
+  ) & { value: string; caseInsensitive?: boolean }
```
CarlosNZ, line 43: Makes type consistent with change done by @fergie-nz in https://github.com/openmsupply/conforma-server/pull/810

src/containers/Main/AuthenticatedWrapper.tsx
```diff
@@ -4,25 +4,35 @@ import { useRouter } from '../../utils/hooks/useRouter'
...
+// (i.e. when clicking "Click here" in password reset verification). So we fetch
+// and parse it ourselves directly from window.location.search instead. We
+// should upgrade to ReactRouter 6 at some point and hopefully this will be
+// sorted.
+export const getSessionIdFromUrl = () => {
+  const sessionIdRegex = /^\?(sessionId)=(.+)$/
+  const matches = window.location.search.match(sessionIdRegex)
+  if (!matches) return null
+  if (matches[1] === 'sessionId') return matches[2]
+}
```
nmadruga, line 38: Works fine for me : ) tested with our latest angola snapshot

nmadruga (commented):
Have tested and looked briefly seems ok

nmadruga:
Ok to merge this one in @CarlosNZ ? Just because we will need this feature on the next release. But like you said a better solution considering n update on packages is required for the early future.

nmadruga (approved):
👍 

Required to be merged in develop for next release

src/containers/Main/AuthenticatedWrapper.tsx
andreievg: I was contemplating manually parsing location too

src/containers/Main/AuthenticatedWrapper.tsx
```diff
@@ -4,25 +4,35 @@ import { useRouter } from '../../utils/hooks/useRouter'
...
+
+// This is a nasty hack that we need due to the fact that ReactRouter doesn't
+// have "query" values available when this component is loaded from another
+// (i.e. when clicking "Click here" in password reset verification). So we fetch
+// and parse it ourselves directly from window.location.search instead. We
+// should upgrade to ReactRouter 6 at some point and hopefully this will be
+// sorted.
+export const getSessionIdFromUrl = () => {
+  const sessionIdRegex = /^\?(sessionId)=(.+)$/
+  const matches = window.location.search.match(sessionIdRegex)
```
andreievg, line 35: I would have used query-string (it's already in our dependencies)

CarlosNZ: Oh okay. Yeah, I'll change it.

CarlosNZ: Okay, I think that this whole "missing sessionId" thing is actually a bit of a red herring.

Explanation and new PR here: #1285 

Hopefully that'll be the last of it for a while. 🙄

PR #1281, opened 2022-08-11
Change ordering of templates returned to Dashboard to orderBy code

nmadruga:
Quick fix to try sort out the problem found in https://github.com/openmsupply/conforma-templates/issues/19

### DIFF
--- a/src/utils/generated/graphql.tsx
+++ b/src/utils/generated/graphql.tsx
@@ -61,6 +61,36 @@ export type Query = Node & {
   counters?: Maybe<CountersConnection>;
   /** Reads and enables pagination through a set of `DataTable`. */
   dataTables?: Maybe<DataTablesConnection>;
+  /** Reads and enables pagination through a set of `DataTableActiveIngredient`. */
+  dataTableActiveIngredients?: Maybe<DataTableActiveIngredientsConnection>;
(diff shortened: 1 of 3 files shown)


CarlosNZ (approved):
Awesome -- quick and easy solution 👍

Just one note -- the order of the categories is being affected too. Which ever template comes first, its category is first displayed. Is this okay, or do we want to have categories display in order by their code, too?

src/utils/generated/graphql.tsx
```diff
@@ -61,6 +61,36 @@ export type Query = Node & {
   counters?: Maybe<CountersConnection>;
   /** Reads and enables pagination through a set of `DataTable`. */
   dataTables?: Maybe<DataTablesConnection>;
+  /** Reads and enables pagination through a set of `DataTableActiveIngredient`. */
```
CarlosNZ, line 64: Don't want to commit the Angola types to repo.

nmadruga: Ok - cool I'll revert this change by using core_templates loaded in

nmadruga:
> Just one note -- the order of the categories is being affected too. Which ever template comes first, its category is first displayed. Is this okay, or do we want to have categories display in order by their code, too?

Probably good to have categories sorted by code as well for consistency. Will work on it now

nmadruga:
Done!

PR #1282, opened 2022-08-11
#1160 show decision in tooltip improvement

### LINKED ISSUE #1160: Improve message displayed to reviewer to help with decision
1. Add warning next to button to **Reject** (when selected) only for reviewer 1 - should check the level. To say something about this will go to Applicant in case it's send as LOQ.

2. Change overall comment as markup text - if possible....

After changed the Portuguese to this:
`"LABEL_REVIEW_OVERALL_COMMENT": "Comentário geral:\n Se for selecionado \"Em confomidade\": Favor inserir um comentário para anotações internas da decisão interna.\n Se for selecionado \"Observações\": Favor inserir um comentário com o resumo de anotações da decisão interna.\nSe for selecionado \"Não-conformidade\": Favor inserir os detalhes com os motivos da rejeição deste formulário.\n"`

It doesn't show as expected:
Screen Shot 2022-04-27 at 9 16 51 AM

[Updated 12/08/2022 by Nicole] Bringing this from another issue - with another version of decision's descriptions
---

From issue https://github.com/openmsupply/conforma-server/issues/724:

To improve the user experience with the overall comment box.
-----------
Update 1:
Can the following please be added:
The text above overall comment be changed to:
_‘Overall comment
If application outcome is 'Conform': Please use this field to include a summary of the internal decision notes.
If application outcome is 'Request information from the applicant': Please use this field to include a summary of the internal decision notes.
If application 'Non-conform': Please use this field to detail the reasons for rejection to the applicant.'_

image

------------
Update 2:

Can the decision radio button for list of questions be amended as follows to help internal staff:
List of questions to _'Request information from applicant. Note: Any non-conforming comments will be displayed to the applicant for actioning."_

nmadruga:
fixes #1160

~~Adding description for each type of decision inline with the decision. Currently just showing when the decision is selected. Opinions @jagoje @CarlosNZ @fergie-nz ?~~

Ok - after talking to Carl a new solution was made, basically moved the overall comment long explanation to a Tooltip. 
[Pt-an]
Screen Shot 2022-08-12 at 12 17 31 PM

[En-nz]
Screen Shot 2022-08-12 at 12 18 01 PM

OBS: Tooltip only display after user clicks on it.

@CarlosNZ we agreed on a more reusable solution for Tooltip but for now I don't see further options to be added, or the need to use it as a hook - as it was done for `useConfirmationModal`. Let me know if you have ideas for improvement anyway


CarlosNZ:
Is the purpose of this just to show an explanation of what each option does?

If so, couple of ideas:
- the most obvious thing seems to change the actual text to something more appropriate. Like "Conform" should just be "Approve", no need for both. 
- if we still need more info, this seems like a good use for a tool-tip or popover: https://react.semantic-ui.com/modules/popup/ rather than cluttering the UI. 
- we're not going to show the actual list of questions are we? Cos that's already in the the Preview.

CarlosNZ (changes requested):
That's great, works really well and does what it needs to without being intrusive.

We should think about other places this component would be useful.

General comments:
- Might be time to move "Tooltip", and the "Loading"/"NoMatch" components into a subfolder called "Common"?
- I think the icon could do with some styling. Not sure exactly what, but I think this makes it look a bit more like a "typical" tooltip icon:
  -  icon to "help circle" (I'd be inclined to make the icon an (optional) prop, but can do that when we need something different)
  - color: grey (more subtle)
  - margin (on both sides) so it's not so close to the text

See inline comments for more.

src/components/Review/ReviewComment.tsx
```diff
@@ -44,6 +45,7 @@ const ReviewComment: React.FC<ReviewCommentProps> = ({ reviewDecisionId, isEdita
     <div>
       <p>
         <strong>{strings.LABEL_REVIEW_OVERALL_COMMENT}:</strong>
+        <Tooltip message={strings.LABEL_REVIEW_OVERALL_COMMENT_TOOLTIP} />
```
CarlosNZ, line 48: I think setting a max-width of around 500 looks better for this one in particular:
Screen Shot 2022-08-12 at 2 59 43 PM

Probably should add a few basic style props to this component cos different cases would want different widths.

You probably want at least:
- min-width
- max-width

Make sure they have defaults though (probably 500px for max-width, and just let the Popup component use its own min-width)

CarlosNZ:
> * I think the icon could do with some styling. Not sure exactly what, but I think this makes it look a bit more like a "typical" tooltip icon:

Speaking of "typical" tooltip, I just noticed the Github one right here on this page!
Screen Shot 2022-08-12 at 3 15 26 PM

I think the question mark is fine, but overall this one looks pretty good and is kind of what I'm thinking of in my comments above.

CarlosNZ:
Here's my suggested changes all in one comparison: https://github.com/openmsupply/conforma-web-app/compare/%231160-show-decision-description-improvement...%231160-show-decision-description-style-tweaks

Please merge into here if you're happy with it. Basically just makes layout a little tidier and made "Tooltip" component more generic (but with defaults) for future re-use.

nmadruga:
All good for me, Carl. Can this be merged?

CarlosNZ (approved):
👍

Looking good :)

PR #1283, opened 2022-08-12
#1270 Admin differentiation

### LINKED ISSUE #1270: Differentiate between "Admin" and "SysAdmin"
This will be needed for Angola so it needs to be part of Phase 2 work.

The issue is, we need to be able to give senior staff "administrative" rights, which would allow them to access "Admin" category templates such as "Grant permissions", "Manage docs", etc, but NOT have access to our actual "isAdmin" permission stuff such as the template builder.

### Suggested solution (needs further discussion and agreement):

- Terminology: to avoid confusion, "Admin" refers to staff-admin stuff, that a senior staff member would have access to. And what we're currently calling "Admin" (i.e. template builder, etc), will be called "SysAdmin". Senior staff will be "Admin" , but only developers (to start with) will have the actual "isSysAdmin" permissions.
- Therefore first thing will be to go through and change the name of the "isAdmin" permission to "isSysAdmin", otherwise it's going to be really confusing. This will require changing on all templates and the hard-coded "isAdmin" permission in the codebase
- Make the current "Admin" menu just for "staff admin" stuff. Currently is displays system admin stuff and "Admin"-category templates, but it should just be "Admin"-category templates
- The proper "sysAdmin" stuff should be accessed from our special little "Dev" menu in the top-right that currently only shows in development mode. We'll need to do a bit of work making this a bit nicer and useable, and also find a way to still do our "dev" stuff (i.e. quickly change user) when in dev mode.
- Need to update templates that use the "isAdmin" apply permission to use whatever the senior administrator's permission is going to be.

CarlosNZ:
Fix #1270

Requires back-end branch from PR: https://github.com/openmsupply/conforma-server/pull/815

I've added a snapshot to the templates repo for testing: `/dev/snapshots/admin_differentiation_demo_2022_08_12.zip`

(You'll probably have to `yarn generate` in the front-end after loading the snapshot to recognise the new enum value)

There's 3 Admin/Manager users (the usual password):
- "admin" -- the usual one with the standard "admin" permission
- "manager" -- new user with just "systemManager" permission
- "admin_manager" -- new user with BOTH the above permissions

Having "systemManager" permission should let you see the "Manage" menu, with all the utility templates such as "Create internal user" , "Change permissions", etc.

The pre-existing "admin" permission allows you to see the "Configuration" menu, with Template builder, snapshots, etc.

Small note: If changing users with the dev tool "user swap", you'll need to refresh the page to get the menu bar items to refresh.

The good thing about this is that the "Manage" menu is completely optional for a setup. If we wanted to create demo systems where everything is under the "Config" menu as before, then we'd just keep all the templates with "admin" permission and with the UI location "Admin" as it is currently.

I haven't done anything to the main "Angola" snapshot yet. I'll wait until this is approved, then I'll update that and merge this at the same time so it all stays in sync.


src/containers/User/UserArea.tsx
```diff
@@ -127,14 +129,14 @@ const MainMenuBar: React.FC<MainMenuBarProps> = ({
...
+    configOptions.splice(1, 0, {
       key: 'snapshots',
       text: 'Snapshots',
       value: '/admin/snapshots',
     })
 
-  // Add Admin templates to Admin menu
-  adminOptions.push(
+  // Add Config/Admin templates to Config menu (requires Admin permission)
+  configOptions.push(
```
CarlosNZ, line 139: Templates can still be added to the "Config" menu if they are linked to the UI-location "Admin", if we want.

CarlosNZ, line 139: Templates can still be added to the "Config" menu if they are linked to the UI-location "Admin", if we want.

src/utils/defaultLanguageStrings.ts
```diff
@@ -142,7 +142,7 @@ export default {
   FILTER_ASSIGNER_ACTION: 'Assigner Action',
   FILTER_DATE_BEFORE: 'Before',
   FILTER_DATE_AFTER: 'After',
-  FILTER_CLEAR_ALL: 'Clear all filters',
+  FILTER_RESET_ALL: 'Reset all filters',
```
CarlosNZ, line 145: Not part of this issue, but #1275 got merged before this suggestion was implemented.

src/utils/helpers/fetchUserInfo.ts
```diff
@@ -3,14 +3,19 @@ import config from '../../config'
...
+import { usePrefs } from '../../contexts/SystemPrefs'
 
 interface SetUserInfoProps {
   dispatch: Dispatch<UserActions>
 }
 
 const fetchUserInfo = ({ dispatch }: SetUserInfoProps, logout: Function) => {
+  const { preferences } = usePrefs()
+
+  const managementPrefName = preferences?.systemManagerPermissionName || 'systemManager'
```
CarlosNZ, line 15: By default, the permission name to access "Management" templates is `systemManager`, but this can be overridden by preferences.json if they want it to be called something different. (Or, say, they just wanted to make one of the existing permissions the "management" one as well)

src/containers/Dev/UserSelection.tsx
```diff
@@ -40,9 +40,9 @@ const UserSelection: React.FC = () => {
     } else localStorage.setItem(config.localStorageJWTKey, loginResult.JWT)
 
     // Organisation login (auto-select first in list)
-    const { JWT, user, templatePermissions, orgList, isAdmin } = loginResult
+    const { JWT, user, templatePermissions, permissionNames, orgList, isAdmin } = loginResult
```
CarlosNZ, line 43: We need a simple list of "permissionNames" in order to determine if the current user has "management" access. Plus, it'll probably useful in other places too, I'd imagine.

CarlosNZ: I could have just returned `isManager` from back-end like `isAdmin`, but thought a list of currently active permission names is probably useful for other things. (e.g. It'll automatically be available to evaluator expressions as it's in the "currentUser" state, so we can make form elements conditional on having a given permission, say)

src/containers/User/UserArea.tsx
```diff
@@ -159,6 +169,12 @@ const MainMenuBar: React.FC<MainMenuBarProps> = ({
     push(value)
   }
 
+  const handleManagementChange = (_: SyntheticEvent, { value }: any) => {
+    console.log('Value', value)
```
nmadruga, line 173: Pleas remove

src/utils/helpers/fetchUserInfo.ts
nmadruga: Ok - I see that's determinate on the back-end so no risk for the "managerPermissionName" to be changed to some other users permissions on the Front-end

nmadruga (commented):
Ok - looks good to me. I'm going to test this now : )

nmadruga (approved):
Ok - it seems to be working fine. Thanks for doing that asap!

I haven't tested the `resetPassword` with this users... But would imagine there wouldn't be a problem with that?

Also I noticed a difference between `manager` and `admin-manager` users on the Manage menu:
Screen Shot 2022-08-15 at 11 18 28 AM
Screen Shot 2022-08-15 at 11 17 54 AM

Also just curious: if we would like to only grant `manager` permission to IT staff, wouldn't we want to include the `Lookup tables` for the manager to control? I know it does data-edition, but I would say it's very controlled and management task.

CarlosNZ:
> Also just curious: if we would like to only grant `manager` permission to IT staff, wouldn't we want to include the `Lookup tables` for the manager to control? I know it does data-edition, but I would say it's very controlled and management task.

Yeah, that's true. We probably need to have a chat about how to achieve that properly. Currently the lookup tables (and all the other admin tasks) are behind "admin" routes in the back-end, so must have the special "admin" permission to access.

Not sure how we'd split that up further without restructuring the back-end routing, so we should probably talk it through.

PR #1285, opened 2022-08-12
#1278 Password reset fix again again!

CarlosNZ:
Hmmm, prompted by @andreievg 's comment [here](https://github.com/openmsupply/conforma-web-app/pull/1280/files#r944026127), I looked into this a little bit more. Our custom `useRouter` hook is already using the suggested "queryString" to fetch the url queries, and it seems that "query" is actually available when clicking through, even when "state" is missing.

I now think that the thing that was wrong was not actually the missing "sessionId", but was due to the fact that there was an `isLoggedIn()` check on load of the "NonRegisteredLogin" component, and if you were already logged in as "nonRegistered" (which you would be after filling in a password reset request) then you'd get re-directed away back to the login screen.

I think the missing "state"  / "sesssionId" was actually a bit of a red herring

I already made a fix in that last PR to allow a logged-in user to go to the "password_reset" page, and I now think that this is enough to prevent the re-direction.

So I've reverted the `AuthenticatedWrapper` to use "query" from useRouter, and I've left this other little change:
```
  if (query.sessionId && !isLoggedIn())
    return <NonRegisteredLogin option="redirect" redirect={pathname + search} />
```
So now it only re-logs you in if you're not already logged in (as nonRegistered). If you are, it just loads `<SiteLayout/>` with the password reset application correctly.

I can't find any situations where this doesn't work now. Have tried:
- Forgot password, then logging out (clear local storage) BEFORE verification and PW reset
- Forgot password, then verifying and reset without logging out
- Create internal user, verifying and resetting password as separate user WITHOUT logging staff member out
- Create internal user, verifying and resetting password AFTER logging out

Please let me know if you find any.

### DIFF
--- a/src/containers/Main/AuthenticatedWrapper.tsx
+++ b/src/containers/Main/AuthenticatedWrapper.tsx
@@ -6,15 +6,13 @@ import NonRegisteredLogin from '../User/NonRegisteredLogin'
 import { Redirect } from 'react-router'
 
 const AuthenticatedContent: React.FC = () => {
-  const { location } = useRouter()
-
-  const sessionId = getSessionIdFromUrl()
+  const { location, query } = useRouter()
 
   const { pathname, search } = location
 
   // If there is a sessionId in the URL, then need to login as nonRegistered
   // before continuing
-  if (sessionId && !isLoggedIn())
+  if (query.sessionId && !isLoggedIn())
     return <NonRegisteredLogin option="redirect" redirect={pathname + search} />
 
   if (isLoggedIn()) return <SiteLayout />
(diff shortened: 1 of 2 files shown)


nmadruga (approved):
Great finding, Andrei and fix Carl : )

andreievg:
The testing that I was doing where sessionId was missing from query was:
* Forgot password
* Follow verification link in a new incognito window, on clicking link would go to login screen.

If above tests pass then nice going !

andreievg:
> The testing that I was doing where sessionId was missing from query was:
> * Forgot password
> * Follow verification link in a new incognito window, on clicking link would go to login screen.

> If above tests pass then nice going !

Hmm above test failed for me

andreievg:
> Forgot password, then logging out (clear local storage) BEFORE verification and PW reset

This also failed for me:

* Forgot password
* Close all tabs 
* chrome://settings/content/all -> then clear data for localhost:3000
* Paste link in browser
* Click on link in verification

CarlosNZ:
Merging -- but let's continue to keep an eye on this.

Issue #1294, opened 2022-08-22
Cancelamentos do AIM list fails to load. 


fergie-nz:
Also the listbuilder is failing for cancellation of products. However when filtered for drafts, this works. Not sure what is going on here.
This may be a template issue, or a front end issue, or a backend issue.

fergie-nz:
This may be a config issue (result of not copying from coreTemplates) and not a required change.

fergie-nz:
Closing - core actions not implemented correctly in template caused the issue.

PR #1295, opened 2022-08-22
#1294 fixing failing to load

fergie-nz:
outcome is initially returning as null in stead of PENDING. I am unsure exactly why this is. 

This catches this error and delays render attempt until outcome is returning a value

### DIFF
--- a/src/components/List/Cells/OutcomeCell.tsx
+++ b/src/components/List/Cells/OutcomeCell.tsx
@@ -21,8 +21,9 @@ const useOutcomeDisplayMap = () => {
 const OutcomeCell: React.FC<CellProps> = ({ application }) => {
   const outcomeDisplayMap = useOutcomeDisplayMap()
   const { outcome } = application
+  console.log(outcome)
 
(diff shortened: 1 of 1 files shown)


CarlosNZ (commented):
Hmm, I'm not sure exactly the problem this fixes, but I'd like to get to the bottom of it a bit more. The "outcome" field in the application structure should never be "null", so we should probably address that rather than band-aid over the effects.

Could you add a bit more detail to the [issue](#1294) to explain more precisely what the problem is? I'm a bit confused about what "list" you're referring to. It seems like you mean the "application list" based on this PR, but then you talk about the "listBuilder" which is a completely different thing.

 I'll check in with you tomorrow to clarify what's going on. Thanks.

fergie-nz:
Hi Carl,

I'm not certain of the terminology, but the list was the list of applications to cancel a product. 
However, Andrei showed me that this was in fact a template issue where core actions hadn't been implemented properly. 

I understand that there is no issue here once core actions are fixed, so I will close it.

CarlosNZ:
Great, thanks @fergie-nz

Issue #1289, opened 2022-08-17
Making sections not visible


jagoje:
Making sections in an application field different based on selections with the form.

nmadruga:
Hi @andreievg any feedback on this one? I was going to have a look on it now.

andreievg:
I haven't had a chance to look at this yet, probably a good idea for you to check it out.

PR #1296, opened 2022-08-23
#1289 - Making sections hidden according to elements conditions

nmadruga:
Step 1: Remove the visibility of sections when there are no elements to editor are visible ✔️ 
Step 2: Set reviewable section only for `active` - Nothing needed to be done  ✔️


CarlosNZ:
Cool -- look forward to trying this out :)

nmadruga:
Ok - you were right @CarlosNZ the review Sections (which are hidden) can be assigned but don't show on the Review page since there is nothing to review:
Screen Shot 2022-08-25 at 12 33 52 PM

I haven't tested how it shows when there is something to review, but the workflow seems all good when they are not visible - I was able to move from Screening to Assessment just assigning and approving the visible sections.

Let me know when you can test this - I'll pass the template that I have been using when you are ready.

nmadruga:
I've added a template for testing that on the folder Snapshots where we are keeping the latest versions of it

src/utils/helpers/application/getSectionsDetails.ts
```diff
@@ -26,6 +26,7 @@ const getSectionDetailsArray = (sectionsList: TemplateSection[]): SectionDetails
     const totalPages = pageBreaks.length + 1
 
     return {
+      active: true,
```
CarlosNZ, line 29: I assume this is set to `true` initially, because we haven't yet been able to evaluate the visibility conditions of the elements?

src/utils/helpers/structure/updateSectionsAndPages.ts
```diff
@@ -0,0 +1,25 @@
...
+    .map((section) =>
+      Object.values(section.pages).sort((pageOne, pageTwo) => pageOne.number - pageTwo.number)
+    )
+    .flat()
+
+  // Set if section is or not active (in case no element is visible & active - should be inactive/hidden)
+  sortedSections.forEach((section) => {
+    section.details.active = Object.values(section.pages).some((page) =>
+      Object.values(page.state).some(
+        ({ element: { isVisible, isEditable } }) => isVisible && isEditable
```
CarlosNZ, line 17: I think it should be just `isVisible`, not `isEditable`. Cos if there's uneditable elements, we would still want to see them.

CarlosNZ (changes requested):
Nice, working really well.

Just that one comment about "isEditable" -- I think we should still see sections that have uneditable elements. 

We don't hide uneditable elements in normal sections (that's what "isVisible" is for), so why should we hide them in these sections?

src/utils/helpers/application/getSectionsDetails.ts
nmadruga: Sure. I can change that

src/utils/helpers/structure/updateSectionsAndPages.ts
nmadruga: No problem - we still can't review it, so I think it's ok.

I'll also remember to remove the visibility from static fields in this case...

src/utils/helpers/application/getSectionsDetails.ts
CarlosNZ: I didn't mean to change it, was just a question.

src/utils/helpers/application/getSectionsDetails.ts
```diff
@@ -26,7 +26,7 @@ const getSectionDetailsArray = (sectionsList: TemplateSection[]): SectionDetails
     const totalPages = pageBreaks.length + 1
 
     return {
-      active: true,
+      active: false,
```
CarlosNZ, line 29: Sorry, I wasn't actually suggesting you change this, just asking a question. Not sure what the implications are, just go with whichever looks best as it loads, I guess.

CarlosNZ (approved):
👍

src/utils/helpers/application/getSectionsDetails.ts
nmadruga: I think this is probably best indeed. It's just the default, before it checks the elements in the page to set an `active` flag for the sections, so in case there isn't any element - which shouldn't happen, it would be active = false. And that seems more accurate to me.

PR #1309, opened 2022-08-24
#1303 - Part 2 - Fix assignment available for applications Reviewed or Assigned by others

### LINKED ISSUE #1195: Bug: Reviewer can self-assign after applicant has re-submitted to LOQ
To reproduce this problem:
- Application is put into 'request clarification (List of Questions)' state by User 1
- Appears to User 2 as self-assignable and can self-assign it

See Adam for more details (posted on Subforma on 10th May 2022)

### LINKED ISSUE #1303: Bugs: Reviewer's action, ability to assign and viewing applications not yet submitted
1. One application is currently awaiting amendment by the applicant - is there a status that should be showing here to indicate that to an internal user?

2. Reviewer can see application that has not been submitted 😨 

4. One application has already been reviewed by ‘Jess’ and now user Nicole is prompted to assign the application to themselves - even though the application is awaiting amendment by the applicant.

nmadruga:
Fixes #1303 & fixes #1195

Not sure why this was missing it. But it does fix the problems 2 and 3 described in the bug.... Problem 1 is still wip, it will require a little more work than I was imagined.

Fix for 2 problems listed related to trigger not being called when self-assignment, unassignment or re-assignment happened

### DIFF
--- a/src/utils/hooks/useUpdateAssignment.tsx
+++ b/src/utils/hooks/useUpdateAssignment.tsx
@@ -96,11 +96,7 @@ const createAssignmentPatch = (
   addedSections.forEach((code) => assignedSectionsCodes.add(code))
 
   const trigger =
-    assignedSectionsCodes.size === 0
-      ? Trigger.OnReviewUnassign
-      : assignment.isSelfAssignable
-      ? Trigger.OnReviewSelfAssign
-      : Trigger.OnReviewAssign
+    assignedSectionsCodes.size === 0 ? Trigger.OnReviewUnassign : Trigger.OnReviewAssign
 
   const patch: AssignmentPatch = {
     status:
--- a/src/utils/hooks/useUpdateAssignment.tsx
+++ b/src/utils/hooks/useUpdateAssignment.tsx
@@ -1,6 +1,7 @@
 import { AssignmentDetails, FullStructure, SectionAssignee } from '../../utils/types'
 import {
   ReviewAssignmentStatus,
+  Trigger,
   UpdateReviewAssignmentMutation,
   useUpdateReviewAssignmentMutation,
 } from '../../utils/generated/graphql'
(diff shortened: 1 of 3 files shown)


src/utils/hooks/useUpdateAssignment.tsx
```diff
@@ -86,12 +95,20 @@ const createAssignmentPatch = (
   removedSections.forEach((code) => assignedSectionsCodes.delete(code))
   addedSections.forEach((code) => assignedSectionsCodes.add(code))
 
+  const trigger =
+    assignedSectionsCodes.size === 0
+      ? Trigger.OnReviewUnassign
+      : assignment.isSelfAssignable
+      ? Trigger.OnReviewSelfAssign
+      : Trigger.OnReviewAssign
```
CarlosNZ, line 103: In our configs, are we actually distinguishing between "OnReviewSelfAssign" and "OnReviewAssign"?

Seems like there'd be quite a bit of duplication between them?

What is the use case when a Self-Assign trigger would cause different actions to an "Assign" trigger?

CarlosNZ (approved):
Cool, LGTM!

But check my comment about multiple triggers -- just wondering if we actually need both?

src/utils/hooks/useUpdateAssignment.tsx
nmadruga: Hmmm I had a look on that, and you are totally right! I think we can get rid of that event! I'll have to do some changes to the back end - and also to remove the old Trigger `Trigger.OnReviewSelfAssign` from core_templates but I'll do it in the Part 1 Back-end (so I don't need to add a new one)

CarlosNZ: In that case, can we get rid of the "OnReviewSelfAssign" trigger altogether (i.e. remove it from possible Trigger enum values)?

nmadruga: I tried, but it would mean it fails on snapshots that are still using it. Should it be safe to do?

PR #1314, opened 2022-08-25
#1199 Optionally reviewable elements

### LINKED ISSUE #1199: To allow for internal staff to review elements in the form that have not been filled by applicants.
### Reference

Final Requirements Matrix - Customisations - Row 1

* 4 days of development + tests
* 1 day review/documentation

### Requirement discussions
As discussed at recent meeting, we need a way to allow certain questions to be optionally:

- reviewable, even when there is no applicant response
- NON reviewable even when there IS a response (eg the checkbox array tied to the file upload)

My suggestion would be to add a new field to template_element schema called is_reviewable (along with is_visible, etc.) with three possible states:
- NULL (default), behaves same as it does currently
- true -- always reviewable regardless of applicant response
- false -- never reviewable, regardless of applicant response

For the "true" case, the way to make it work would be to generate a blank default response when the application is created, which will ensure that application responses are not pruned and associated review_responses are created.

For the "false" case, we check during creation of review responses, and don't make one if this field is "false"

Slight problem: this seems like it would be mainly back-end logic, so this field wouldn't be able to be dynamic like all the others. I would think that this is probably fine, but it bothers me slightly that this is inconsistent with other similar fields (is_visisble, is_required, is_editable, etc). Will have a bit more of a think about it.

--- 

### Issues (as per [suggested solution](https://github.com/openmsupply/conforma-web-app/issues/1199#issuecomment-1174445626) below):

- [x] https://github.com/openmsupply/conforma-server/issues/782
- [x] #1229
- [x] #1230
- [x] #1231

CarlosNZ:
Fix #1199 

Requires back-end branch from https://github.com/openmsupply/conforma-server/pull/821

To test, please use snapshot from templates repo: `/conforma-templates/dev/snapshots/Optional_review_testing.zip`

You probably need to re-run `yarn generate` on both ends after loading the snapshot.

There is an application by "jane_smith" which you can start reviewing as "basic_reviewer". Note the reviewability of various questions, and confirm that they behave as expected.


src/components/PageElements/Elements/ReviewApplicantResponse.tsx
```diff
@@ -156,14 +173,16 @@ const ReviewApplicantResponse: React.FC<ReviewApplicantResponseProps> = ({
...
               >
-                {!decisionExists && (
-                  <ReviewElementTrigger
-                    title={triggerTitle} // Review or Re-review
-                    onClick={() => setIsActiveEdit(true)}
-                  />
-                )}
+                {!decisionExists &&
+                  (reviewType === 'OptionallyReviewable' ? (
+                    <AddIcon onClick={() => setIsActiveEdit(true)} />
```
CarlosNZ, line 179: I thought this "+" icon would be a suitable alternative for the "optionally" reviewable items. It needs to be significantly different, as when you click "Approve All", these ones won't be approved, so we want to avoid confusion as to why not.

Open to other suggestions though?

src/components/PageElements/Elements/ReviewInlineInput.tsx
```diff
@@ -101,7 +101,7 @@ const ReviewInlineInput: React.FC<ReviewInlineInputProps> = ({
           <div className="review-button-container">
             <Button
               primary
-              disabled={isInvalidComment}
+              disabled={isInvalidComment || !reviewResponse.decision}
```
CarlosNZ, line 104: Not technically part of this PR, but seemed like it should have been like this all along. Otherwise you can click the "Approve" button before selecting a decision, but nothing happens, which is confusing.

src/components/PageElements/Elements/ReviewInlineNewResponse.tsx
```diff
@@ -0,0 +1,132 @@
...
+  stageNumber: number
+}
+
+/*
+This component is visually the same as "ReviewInlineInput", but is used when reviewing elements that don't have an applicantResponse. Instead of updating a review response, it needs to create a new review response along with its associated applicationResponse.
+*/
+const ReviewInlineNewResponse: React.FC<ReviewInlineNewResponseProps> = ({
+  setIsActiveEdit,
+  summaryViewProps,
+  reviewInfo,
```
CarlosNZ, line 22: We need reviewInfo so we can get the reviewId.

src/components/PageElements/Elements/ReviewInlineNewResponse.tsx
```diff
@@ -0,0 +1,132 @@
...
+        templateElementId,
+        applicationId,
+        reviewId,
+        stageNumber,
+        decision: reviewResponse.decision as ReviewResponseDecision,
+        comment: reviewResponse.comment,
+        timeSubmitted,
+      },
+      // The order of these refetch queries appears to be important!
+      refetchQueries: ['getAllResponses', 'getReviewResponses'],
```
CarlosNZ, line 66: @nmadruga This is what I needed to fix the "Decision" not updating, and the order seems to be important!!

src/components/PageElements/Elements/ReviewInlineNewResponse.tsx
```diff
@@ -0,0 +1,132 @@
+import React, { useState } from 'react'
+import { Button, Form, Radio, TextArea } from 'semantic-ui-react'
+import { useLanguageProvider } from '../../../contexts/Localisation'
+import { SummaryViewWrapperProps } from '../../../formElementPlugins/types'
+import { ReviewResponseDecision } from '../../../utils/generated/graphql'
+import { useCreateReviewResponseMutation } from '../../../utils/generated/graphql'
+import { ReviewDetails } from '../../../utils/types'
+
+interface ReviewInlineNewResponseProps {
```
CarlosNZ, line 9: This component is mostly the same as `ReviewInlineInput`, just with a few modifications to handle creating a new response rather than updating an existing one.

src/components/PageElements/PageElements.tsx
```diff
@@ -44,12 +50,11 @@ const PageElements: React.FC<PageElementProps> = ({
   stages,
   canEdit = false,
   isConsolidation = false,
-  isReview = false,
+  reviewInfo,
```
CarlosNZ, line 53: Rather than passing _both_ `isReview` and `reviewInfo` down through multiple levels, I just passed `reviewInfo` and deduce `isReview` when required (it's only in one place)

src/components/PageElements/PageElements.tsx
```diff
@@ -60,6 +65,7 @@ const PageElements: React.FC<PageElementProps> = ({
   } = useRouter()
 
   const visibleElements = elements.filter(({ element }) => element.isVisible)
+  const isReview = !!reviewInfo
```
CarlosNZ, line 68: As mentioned -- deduce `isReview` rather than passing it down.

nmadruga (commented):
Found a problem to reply to Request further information (LOQ) as applicant : (

This is the application after reviewer send one Rejected (Optional element) back to `jane_smith`. And then after I click to "Update" it doesn't show the problem and I can't start:
Screen Shot 2022-08-26 at 10 19 13 AM
Screen Shot 2022-08-26 at 10 19 08 AM

CarlosNZ:
> Found a problem to reply to Request further information (LOQ) as applicant : (

Good catch. I suspect this might be caused by unintended effects of the application_response being `null`. I'll look into it.

src/components/PageElements/Elements/ReviewInlineNewResponse.tsx
```diff
@@ -0,0 +1,136 @@
...
+  summaryViewProps: SummaryViewWrapperProps
+  reviewInfo: ReviewDetails
+  stageNumber: number
+}
+
+/*
+This component is visually the same as "ReviewInlineInput", but is used when
+reviewing elements that don't have an applicantResponse. Instead of updating a
+review response, it needs to create a new review response along with its
+associated applicationResponse.
```
nmadruga, line 20:
```suggestion
This component is visually the same as "ReviewInlineInput", but is used when
reviewing elements that don't have an applicantResponse. Instead of updating a
reviewResponse, it needs to create a new one and its associated applicationResponse.
```

src/components/PageElements/Elements/ReviewInlineNewResponse.tsx
```diff
@@ -0,0 +1,136 @@
...
+  const reviewOptions = [
+    {
+      label: strings.LABEL_REVIEW_APPROVE,
+      decision: ReviewResponseDecision.Approve,
+    },
+    {
+      label: strings.LABEL_REVIEW_RESSUBMIT,
+      decision: ReviewResponseDecision.Decline,
+    },
+  ]
```
nmadruga, line 56: Maybe add a comment on top to explain this is currently only allowed for first_level therefor only review decisions that would be associated with a simple review

src/components/PageElements/Elements/ReviewInlineNewResponse.tsx
nmadruga: Oh, cool! Yes, I think is because that's the same order `ReviewPageWrapper` does it

src/components/PageElements/PageElements.tsx
```diff
@@ -27,6 +32,7 @@ interface PageElementProps {
   stages: StageDetails[]
   canEdit: boolean
   isConsolidation?: boolean
+  reviewInfo?: ReviewDetails
   isReview?: boolean
```
nmadruga, line 36: `isReview` can be removed here also

src/components/PageElements/PageElements.tsx
nmadruga: Ok - that's better : )

src/containers/TemplateBuilder/template/Form/ElementConfig.tsx
```diff
@@ -253,6 +253,10 @@ const ElementConfig: React.FC<ElementConfigProps> = ({ element, onClose }) => {
                 options={[
                   { value: IsReviewableStatus.Always, text: 'Always' },
                   { value: IsReviewableStatus.Never, text: 'Never' },
+                  {
+                    value: IsReviewableStatus.OptionalIfNoResponse,
+                    text: 'Optional (if no application response)',
+                  },
```
nmadruga, line 259: Aff everything here is not localised. But that's ok - work for the future

nmadruga (commented):
Looks great. I'm so happy to see this added, and the code is very separated which is awesome! I'll wait on you to check that problem found and review this again today - hopefully so we can release that beta-version including this work

src/containers/TemplateBuilder/template/Form/ElementConfig.tsx
CarlosNZ: Ya, not much in the template builder is localised!

CarlosNZ:
> This is the application after reviewer send one Rejected (Optional element) back to `jane_smith`. And then after I click to "Update" it doesn't show the problem and I can't start:

Cool, fixed this now: 1fed7d6a

I'd just forgotten to set `recommendedApplicantVisibility: ORIGINAL_RESPONSE_VISIBLE_TO_APPLICANT` in the new response, I didn't realise it needed be this value.

Works for me now. 👍

src/components/PageElements/Elements/ReviewInlineNewResponse.tsx
CarlosNZ: 43c20a76

src/components/PageElements/PageElements.tsx
CarlosNZ: 43c20a76

nmadruga:
Thanks, Carl. It does fix something, although when trying again I made it slightly different and found still another thing. I have created a snapshot where the problem is present an uploaded in Templates repo called `Optional_review_elements_LOQ`. 

The problem is that: I have Rejected 2 Optional elements (one in Section 1 and another in Section 3). The one in Section 3 is showing correctly, but not the one in Section 1. See screenshots for more details:

### Review done by `basic_reveiwer`:
Screen Shot 2022-08-29 at 9 57 17 AM
Screen Shot 2022-08-29 at 9 57 27 AM
Screen Shot 2022-08-29 at 9 57 09 AM

#### Application on Draft for `jane_smith` after changes required:
Screen Shot 2022-08-29 at 9 56 37 AM
Screen Shot 2022-08-29 at 9 56 43 AM

nmadruga (changes requested):
Please check the problem found above. 

I also tried to fix the conflicts problem with no luck

CarlosNZ:
> Thanks, Carl. It does fix something, although when trying again I made it slightly different and found still another thing. I have created a snapshot where the problem is present an uploaded in Templates repo called `Optional_review_elements_LOQ`.

That's interesting. When I went through a similar process, the application showed the review_responses correctly to "jane_smith" as expected. 

When I look at your snapshot, I can see that the review_response in question still has "is_visible_to_applicant" set to false, which is what is causing the problem. This is because the "recommended_applicant_visibility" is set to "ORIGINAL_RESPONSE_NOT_VISIBLE_TO_APPLICANT", so it's not getting set to visible to applicant. If you manually change the value of this to "true", it'll display as expected to applicant.

This is very odd, though as the last change I made ensured that this would be set. If you look at the "createReviewResponse" mutation, you'll see that the value "ORIGINAL_RESPONSE_VISIBLE_TO_APPLICANT" is hard-coded into the mutation, and it's working okay for me.

Is there any chance you started your review before my latest commit (after your first review on Friday)? If so, that'll be the problem. You didn't perhaps start a review before pulling the latest, and then pull my changes before submitting but AFTER creating that review response?

Would you be able to run through it again with latest changes and see if you can reproduce this? Cos I don't see how that field could be set incorrectly for newly created review_responses when the correct value is hard-coded. 🤔

nmadruga (approved):

PR #1315, opened 2022-08-25
#1291 Fix "Manage" menu bar item

### LINKED ISSUE #1291: Full horizontal list doesn't render upon log in.
List is failing to render fully for Admin - unsure yet if this is affecting other users such as internal or external client users.

Toolbar won't show all options on log in. After refreshing, all options are rendered. It seems that the missing item can vary. Possibly a back-end issue.

On initial log in:

incompleteAdminList

After refreshing:

refreshedAdminList

CarlosNZ:
Fix #1291 

It was getting the correct `isManage` pref when already logged in, but when logging in, it would run the `else` part of the `onLogin` method (in UserState), which was setting `isManager` to `false`. I think I assumed it would get updated by a subsequent `fetchUserInfo` call, but it doesn't.

So this just fixes that.

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -51,6 +51,7 @@ const config = {
   localStorageJWTKey: 'persistJWT',
   applicantDeadlineCode: 'applicantDeadline',
   isProductionBuild,
+  defaultSystemManagerPermissionName: 'systemManager',
 }
 
 export default config
--- a/src/contexts/UserState.tsx
+++ b/src/contexts/UserState.tsx
@@ -3,6 +3,7 @@ import { useApolloClient } from '@apollo/client'
 import fetchUserInfo from '../utils/helpers/fetchUserInfo'
 import { OrganisationSimple, TemplatePermissions, User } from '../utils/types'
 import config from '../config'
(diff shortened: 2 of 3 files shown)


src/config.ts
```diff
@@ -51,6 +51,7 @@ const config = {
   localStorageJWTKey: 'persistJWT',
   applicantDeadlineCode: 'applicantDeadline',
   isProductionBuild,
+  defaultSystemManagerPermissionName: 'systemManager',
```
CarlosNZ, line 54: Putting this in here so don't have to hard-code it in two separate places.

src/contexts/UserState.tsx
```diff
@@ -135,7 +140,7 @@ export function UserProvider({ children }: UserProviderProps) {
         newPermissionNames: permissionNames || [],
         newOrgList: orgList || [],
         newIsAdmin: !!isAdmin,
-        newIsManager: false,
+        newIsManager: permissionNames.includes(managementPrefName),
```
CarlosNZ, line 143: This is the actual fix

nmadruga (approved):
Looks good - will be nice to have on the release

PR #1317, opened 2022-08-26
#1316 Number input improvements

### LINKED ISSUE #1316: Improvement for Number plugin
The "suffix" or "prefix" should be outside the number input box. For two reasons:
- We can have the prefix/suffix appear before anything is entered (which is more informative for user)
- We can keep the input field as a "number" type so it keeps the stepper

CarlosNZ:
Fix #1316 

Thanks for the suggestions @fergie-nz and @nmadruga, I think this is a much better use of the units suffix now.

To see in action, use latest snapshot: "angola_2022_08_26_1459" -- the 3 shelf-life fields (Mod1, pg3) have been updated to use this to display the "month" units more effectively:
Screen Shot 2022-08-26 at 2 47 43 PM

Have updated the docs with correct parameters. See inline comments for further explanation.

### DIFF
--- a/docs/internal/Element-Type-Specs.md
+++ b/docs/internal/Element-Type-Specs.md
@@ -623,11 +623,13 @@ _Input for numeric fields_
 - **minValue** -- minimum allowed value (default: 0)
 - **maxValue** -- maximum allowed value (default: no limit)
 - **step** -- `number` (default: `1`) If `simple == true` (above), the `step` value specifies the amount the number will be incremented or decremented by when using the stepper.
-  **NOTE**: The parameters below are only relevant is `simple == false` (above)
+- **prefix** / **suffix** -- `string` If specified, the number will display these values either side of the number input. They'll also be pre/appended to the "text" representation of the number in the saved response. Useful if you want to define units with the input number (e.g. `12 km`)
+- **suffixPlural** -- `string` (only relevant if `suffix` is specified above) Changes the displayed/stored suffix depending on the value of the number based on pluralisation rules.  
+(e.g. if `suffix = "month"` and `suffixPlural = "months"`, then when number is 1: "1 month", when number is 2:  "2 months")  
+**NOTE**: The parameters below are only relevant if `simple == false` (above)
 - **locale** -- `string` specifies the international "locale" code (e.g `'ja-JP'`) for displaying the calendar in local format. Default is the local setting.
 - **currency** -- `string` If specified, number will be formatted as a currency value (e.g. $4.95). Should be specified in ISO4217 country code format (e.g. "USD", "JPY") See: [https://www.iban.com/currency-codes](https://www.iban.com/currency-codes)
 - **maxSignificantDigits** -- `number` If specified, number will be rounded to the specified number of significant figures
-- **prefix** / **suffix** -- `string` If specified, formatted number will include these values in the string. Useful if you want to define units with the input number (e.g. `12 km`)
 
 #### Response type
 
@@ -636,6 +638,11 @@ _Input for numeric fields_
   text: <Formatted version of number (as specifed in parameters)>
   number: <number>
   type: <integer | float>
+  currency: <string> (from parameters, only stored if defined)
+  locale: <string> (from parameters, only stored if defined)
+  prefix: <string> (from parameters, only stored if defined)
+  suffix: <string> (from parameters, only stored if defined)
+  suffixPlural: <string> (from parameters, only stored if defined)
 }
 ```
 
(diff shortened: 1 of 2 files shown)


src/formElementPlugins/number/src/ApplicationView.tsx
```diff
@@ -29,16 +29,17 @@ const ApplicationView: React.FC<ApplicationViewProps> = ({
     description,
     default: defaultValue,
     maxWidth,
-    type = NumberType.FLOAT,
+    type = NumberType.INTEGER,
```
CarlosNZ, line 32: The docs said "Integer", the plugin was set to "Float". Made them consistent (default Integer)

src/formElementPlugins/number/src/ApplicationView.tsx
```diff
@@ -74,26 +77,45 @@ const ApplicationView: React.FC<ApplicationViewProps> = ({
...
+    if (validation.isValid)
+      onSave({
+        text: fullText,
+        number,
+        type,
+        currency,
+        locale,
+        prefix,
+        suffix,
+        suffixPlural,
```
CarlosNZ, line 107: May as well store these values with the response, so the strings can be rebuilt from parts if required.

src/formElementPlugins/number/src/ApplicationView.tsx
```diff
@@ -74,26 +77,45 @@ const ApplicationView: React.FC<ApplicationViewProps> = ({
...
-    setTextValue(text)
+    setTextValue(formattedNumber)
   }
 
   const styles = maxWidth
     ? {
         maxWidth,
       }
+    : prefix || suffix
+    ? { maxWidth: 100 }
```
CarlosNZ, line 118: If there is a prefix or a suffix, display the number input small by default. (Normally it'll be full-width)

src/formElementPlugins/number/src/ApplicationView.tsx
```diff
@@ -162,18 +196,32 @@ const parseInput = (
...
-  if (!textInput) return [null, null]
+  prefix: string | undefined,
+  suffix: string | undefined,
+  suffixPlural: string | undefined
+): {
+  number: number | null
+  formattedNumber: string | null
+  currentSuffix: string | undefined
+  fullText: string | null
+} => {
```
CarlosNZ, line 207: Was starting to get unwieldy, so returning named object fields easier than array.

Quick explanation:
- `number`: the actual number (as JS number type)
- `formattedNumber`: stringified version of number, formatted according to locale/currency/simple settings (e.g. 1,000, $100.00)
- `currentSuffix`: which suffix to use (plural or singular) depending on the number value
- `fullText`: "formattedNumber" plus any prefix or suffix values (e.g. "12 months") -- this is the value that gets stored in the "text" field of the saved response

src/formElementPlugins/number/src/ApplicationView.tsx
```diff
@@ -162,18 +196,32 @@ const parseInput = (
...
-  if (rawDigits === '') return [null, textInput]
+  if (rawDigits === '')
+    return { number: null, formattedNumber: textInput, currentSuffix: suffixPlural, fullText: null }
   const number: number = Number(rawDigits)
-  const text: string = simple
-    ? String(number)
-    : ((prefix + numberFormatter.format(number) + suffix) as string)
-  return [number, text]
+  const prefixString = prefix ? prefix + ' ' : ''
+  const suffixString = suffix ? ' ' + (number === 1 ? suffix : suffixPlural) : ''
```
CarlosNZ, line 216: Pad prefix and suffix with a space for concatenating with formatted number value.

docs/internal/Element-Type-Specs.md
```diff
@@ -623,11 +623,13 @@ _Input for numeric fields_
 - **minValue** -- minimum allowed value (default: 0)
 - **maxValue** -- maximum allowed value (default: no limit)
 - **step** -- `number` (default: `1`) If `simple == true` (above), the `step` value specifies the amount the number will be incremented or decremented by when using the stepper.
-  **NOTE**: The parameters below are only relevant is `simple == false` (above)
+- **prefix** / **suffix** -- `string` If specified, the number will display these values either side of the number input. They'll also be pre/appended to the "text" representation of the number in the saved response. Useful if you want to define units with the input number (e.g. `12 km`)
+- **suffixPlural** -- `string` (only relevant if `suffix` is specified above) Changes the displayed/stored suffix depending on the value of the number based on pluralisation rules.  
+(e.g. if `suffix = "month"` and `suffixPlural = "months"`, then when number is 1: "1 month", when number is 2:  "2 months")
```
nmadruga, line 628: Yey! That's awesome : )

nmadruga (approved):
Also looking pretty good, I do like the idea of incrementing the general number plugin like that

PR #1319, opened 2022-08-29
#1299 Assign reviewer to all levels at once

### LINKED ISSUE #1299: Option to auto-assign all sections to reviewer (even when it's not last level)
Currently the auto-assignment to all sections only happens (after the reviewer is selected in one of the sections) when the  stage only has one level or when this is the last level review. We would like to add a new option to have the auto-assignment switched on for a specific stage/level.

### Implement in the front-end
Don't have any option for the template. just add helper to assign one reviewer to all the available sections (and if this person doesn't have permissions to one section it will not get assigned to that section...) 
Add a dropdown at the top and list all possible reviewers (from each section's dropdown)

Let's see if this simpler option is valid.

### LINKED ISSUE #1313: UX issue with dropdown tool to assign reviewer - some options hidden behind banner
Very small thing but causes a lot of UX pain. Can't see all possible reviewers in certain cases when selecting who should review an application.

I propose small change to make the dropdown have a scroll. However I seem to remember discussion on this and we decided against it. I think this particular use case wasn't considered, so worth taking another look.

Alternative option could be force the dropdown list to sit on top of a banner. I think this would be the best solution because then we wouldn't need the scroll which is a bit clunky at times.

Critical UX! Cannot select all assignments due to dropdown being blocked

CarlosNZ:
Fix #1299 

Screen Shot 2022-08-30 at 9 16 44 AM

Adds dropdown to select a reviewer to "assign all" to. It only assigns sections that are:
- allowed to be assigned to that reviewer
- not already assigned

There is a Tooltip info icon which explains this to the user.

Only shows when it's *not* last level, as last level is already assigned to all.

Placement: I tried putting it above the Sections, but I couldn't get it to look decent with the other elements (Filter, Stage). I tried right-align, left-align, and even putting it in between these two, but none looked very good. Open to other suggestions regarding placement.

I've tried to account for as many review/assignment states as I can, but please see if there's any that I've missed.

I was testing with the same "Optional Review testing" snapshot (in templates repo) that I used for my last PR, so use that one if you want to quickly go through an application/review process.

**Update**: there's a small unrelated fix for scrolling in the Preview window. See inline comment for details.

**Update**: another small CSS tweak to fix #1313 4663a17f


src/components/Tooltip.tsx
```diff
@@ -20,20 +22,22 @@ const Tooltip: React.FC<TooltipProps> = ({
   minWidth = 300,
   maxWidth = 550,
   style,
+  popupStyle = style,
+  iconStyle,
```
CarlosNZ, line 26: I realised that the current `style` prop only applied to the pop itself, not the icon. Now we can discriminate between them.

src/containers/Review/assignment/AssignAll.tsx
```diff
@@ -0,0 +1,66 @@
...
+        onChange={(_, { value }) => {
+          const reviewerId = value !== '' ? options[value as number].reviewerId : 0
+          setReviewerForAll(reviewerId)
+          setSelected(value as number | string)
+        }}
+      />
+      <Tooltip
+        message={strings.ASSIGN_ALL_TOOLTIP}
+        maxWidth={400}
+        iconStyle={{ height: '1.4em' }}
```
CarlosNZ, line 41: Not sure why this was necessary, but it doesn't center-align vertically without this.

semantic/src/site/globals/site.overrides
```diff
@@ -548,6 +548,7 @@ a:hover {
     gap: 10px;
     max-width: 650px;
     margin-top: 20px;
+    margin-bottom: 60px; // to offset the Actions area that covers it.
```
CarlosNZ, line 551: This is to make the scrolling work properly for Preview modal. I thought I'd already fixed this, but I hadn't allowed for the height of the "Actions" area at the bottom (with the OK button), which covers the main content.

The actions area is absolutely positioned, so the main content sits behind it and doesn't show a scroll bar until it's taller than the whole preview area, including the area covered by the actions area. This extra margin is the height of the actions area, so content won't get stuck behind it without a scroll bar.

nmadruga:
> **Update**: there's a small unrelated fix for scrolling in the Preview window. See inline comment for details.

Cool. Should this be a good moment to also add this improvement? #1313

nmadruga (changes requested):
The code looks good - I haven't seen anything that requires changes. Although while testing (with our latest snapshot) I found an issue. I think we do allow the `ReviewAssignment.allowSections` to be Null in case all sections are allowed. And in that case the function `getReviewerList` is returning empty array. So we don't have the **AssignAll** component displaying.

CarlosNZ:
> Cool. Should this be a good moment to also add this improvement? #1313

Sure -- this should fix #1313 4663a17f

CarlosNZ:
> The code looks good - I haven't seen anything that requires changes. Although while testing (with our latest snapshot) I found an issue. I think we do allow the `ReviewAssignment.allowSections` to be Null in case all sections are allowed. And in that case the function `getReviewerList` is returning empty array. So we don't have the **AssignAll** component displaying.

Okay, have removed the additional filter, as we discussed in person, and added a comment to the commented-out line. 4663a17f

nmadruga (approved):

PR #1320, opened 2022-08-29
Add correct config info to applicationData

CarlosNZ:
Quick fix for the missing logo issue in the Company Data View.

The applicationData that was being passed to the "SummaryView" components in Data Views was incorrect due to a recent change to how the server urls are generated. I'd updated all the other applicationData's, but somehow missed this one.

Have also updated the comment to make it clearer what's going on here.

### DIFF
--- a/src/formElementPlugins/fileUpload/src/SummaryView.tsx
+++ b/src/formElementPlugins/fileUpload/src/SummaryView.tsx
@@ -5,6 +5,7 @@ import getServerUrl from '../../../utils/helpers/endpoints/endpointUrlBuilder'
 import { SummaryViewProps } from '../../types'
 
 const SummaryView: React.FC<SummaryViewProps> = ({ parameters, Markdown, response }) => {
+  console.log('files', response?.files)
   return (
     <Form.Field required={parameters.isRequired}>
       {parameters?.label && (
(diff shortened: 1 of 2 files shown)


nmadruga (changes requested):
I'm not sure if it's associated but the file that has been previously uploaded in the example "Create organisation" serial ORG-0046-2022 on the latest snapshot doesn't open correctly.
I can see that the file is present in the data base with unique_id `3dLFqxr27Q2a52F3IMoFc`, but this is displayed when I hover the link to this document (during Review):
`http://localhost:8080/api/public/file?uid=undefined` 

Well, I just realised that this is the same URL showing for any of the images/documents in the form.

CarlosNZ:
> I'm not sure if it's associated but the file that has been previously uploaded in the example "Create organisation" serial ORG-0046-2022 on the latest snapshot doesn't open correctly. I can see that the file is present in the data base with unique_id `3dLFqxr27Q2a52F3IMoFc`, but this is displayed when I hover the link to this document (during Review): `http://localhost:8080/api/public/file?uid=undefined`
> 
> Well, I just realised that this is the same URL showing for any of the images/documents in the form.

Okay, good catch -- but not related to this issue (it's on develop like that). I'll investigate.

CarlosNZ:
> > I'm not sure if it's associated but the file that has been previously uploaded in the example "Create organisation" serial ORG-0046-2022 on the latest snapshot doesn't open correctly. I can see that the file is present in the data base with unique_id `3dLFqxr27Q2a52F3IMoFc`, but this is displayed when I hover the link to this document (during Review): `http://localhost:8080/api/public/file?uid=undefined`
> > Well, I just realised that this is the same URL showing for any of the images/documents in the form.
> 
> Okay, good catch -- but not related to this issue (it's on develop like that). I'll investigate.

It's a quick fix, fortunately, so I've just squeezed it in here: 49dd92bf

It's annoying though, I thought I had the typing really strict in the "getServerUrl" method, so I didn't think incorrect parameters like this could have slipped through. Hmmmm....

src/formElementPlugins/fileUpload/src/SummaryView.tsx
```diff
@@ -5,6 +5,7 @@ import getServerUrl from '../../../utils/helpers/endpoints/endpointUrlBuilder'
 import { SummaryViewProps } from '../../types'
 
 const SummaryView: React.FC<SummaryViewProps> = ({ parameters, Markdown, response }) => {
+  console.log('files', response?.files)
```
nmadruga: Needs removing!

nmadruga (commented):
I wasn't able to properly test... But let's get it in for the server upgrade

nmadruga (approved):

PR #1323, opened 2022-08-30
Quick fix: Display checkbox summary even when no response

CarlosNZ:
@nmadruga this fixes the problem you were having with checkbox display. Made the SummaryView display the CheckboxDisplay component regardless, and just pass in a "dummy" empty response if no response exists.


nmadruga (approved):
Looking great. Thanks for the effort!

PR #1325, opened 2022-08-31
add commands to reset to top of window when navigating to any new pag…

fergie-nz:
…e in form

### DIFF
--- a/src/components/Application/ProgressBar.tsx
+++ b/src/components/Application/ProgressBar.tsx
@@ -62,6 +62,7 @@ const ProgressArea: React.FC<ProgressAreaProps> = ({
   const handleChangeToPage = (sectionCode: string, pageNumber: number) => {
     if (!isLinear) {
       push(`/application/${structure.info.serial}/${sectionCode}/Page${pageNumber}`)
+      window.scrollTo({ top: 0 })
       return
     }
 


nmadruga (changes requested):
That's great Fergus! 

Could you also add that on Page change using the Navigation Bar please?

CarlosNZ (changes requested):
Nice, thanks for this. Can we also add it to the "Previous" and "Next" buttons within the application? Cheers.

src/formElementPlugins/checkbox/src/SummaryView.tsx
```diff
@@ -35,15 +35,25 @@ const SummaryView: React.FC<SummaryViewProps> = ({ parameters, Markdown, respons
...
+            keyMap,
+            true
+          )}
           disabled
           type={type}
           layout={layout}
           onChange={() => {}}
         />
-      ) : null
+      )
```
CarlosNZ, line 56: This is not from your PR. Can you check that you're in sync with develop?

CarlosNZ: > This is not from your PR. Can you check that you're in sync with develop?

Just merge `develop` into yours, should be good to go.

CarlosNZ (approved):
Works for me! 

Please merge develop into this before merging (see my comment)

nmadruga (approved):

PR #1326, opened 2022-08-31
#1324 Assign all fix

### LINKED ISSUE #1324: Problem with "Assign All Sections" when "allowed sections" is null
If the "allowed_sections" field in review_assignment is NULL, we interpret this as meaning "can do ALL sections". However the new "Assign all" functionality hasn't allowed for this case.

CarlosNZ:
Fix #1324 

This handles when the "allowedSections" array is empty -- as that actually means "allow all".

I actually don't really like this, I would much prefer if the "generateReviewAssignements" action put a full list of sections in when we want to allow all, but I think the ramifications of changing that right now could get messy and be hard to test thoroughly.

So this will have to do for now.

### DIFF
--- a/src/containers/Review/assignment/AssignAll.tsx
+++ b/src/containers/Review/assignment/AssignAll.tsx
@@ -2,17 +2,22 @@ import React, { useState } from 'react'
 import { Dropdown, Label } from 'semantic-ui-react'
 import Tooltip from '../../../components/Tooltip'
 import { useLanguageProvider } from '../../../contexts/Localisation'
-import { AssignmentDetails } from '../../../utils/types'
+import { AssignmentDetails, User } from '../../../utils/types'
 
 interface AssignAllProps {
   assignments: AssignmentDetails[]
+  currentUser: User | null
   setReviewerForAll: (value: number) => void
 }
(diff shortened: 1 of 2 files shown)


src/containers/Review/assignment/AssignAll.tsx
```diff
@@ -56,11 +65,12 @@ const getReviewerList = (assignments: AssignmentDetails[]) => {
       const {
         reviewer: { id, firstName, lastName },
       } = assignment
+      const displayName = id === currentUserId ? selfString : `${firstName} ${lastName}`
```
CarlosNZ, line 68: Added so it displays "Yourself" for the current user, for consistency with the section selectors.

nmadruga (approved):
Looks good. But would be easy to not display this when all sections are currently assigned?

Screen Shot 2022-08-31 at 3 54 48 PM

CarlosNZ:
> Looks good. But would be easy to not display this when all sections are currently assigned?
> 
> 
> 
> Screen Shot 2022-08-31 at 3 54 48 PM
> 
> 

Yeah good call. Will update 👍

PR #1328, opened 2022-08-31
#946 Don't show invalid review responses in history panel

### LINKED ISSUE #946: Bug: Review history showing reviewResponse without value
Another small issue, when Re-review the history shows something that hasn't been reviewed yet

Before changing - shouldn't show this reviewResponse
photo_2021-09-02_10-02-26

After changing (this one is ok)
photo_2021-09-02_10-02-30

CarlosNZ:
Fix #946 

I think this is all it needs. There was existing code that caught responses with no stageNumber defined, and set `stageNumber = 0`. The problem was, the *current* review response (i.e. the one just created for this new review) doesn't have a stage number yet, so we can just exclude these ones and get the correct history display.

I don't actually like how the new review responses don't get a stageNumber set until the review is responded to -- it should really be included in the "createReview" mutation. But it provides a convenient way for us to detect which review responses to exclude from the history list, so I guess we can live with it.

@nmadruga -- do you know why we don't set a stageNumber initially? Is it just an oversight, or was there a good reason for it?

### DIFF
--- a/src/utils/hooks/useGetQuestionReviewHistory.tsx
+++ b/src/utils/hooks/useGetQuestionReviewHistory.tsx
@@ -50,51 +50,44 @@ const useGetQuestionReviewHistory = ({ isApplicant, ...variables }: UseGetQuesti
 
     applicationResponses.nodes.forEach((applicantResponse) => {
       if (!applicantResponse) return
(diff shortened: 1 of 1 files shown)


nmadruga (approved):
Looking great! Thanks

PR #1331, opened 2022-09-01
#820B improvement review responses part of latest review submission

nmadruga:
Back-end PR: https://github.com/openmsupply/conforma-server/pull/831

Testing instructions on Back-end!

This will just set the `review_response` on the newly submitted Review to have field `is_latest_review_submission` as TRUE

### DIFF
--- a/docs/internal/Level-1-Review-And-LOQ.md
+++ b/docs/internal/Level-1-Review-And-LOQ.md
@@ -45,7 +45,8 @@ This schema structure allows new responses to be created (previous responses to 
 
 **Front End**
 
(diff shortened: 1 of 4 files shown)


CarlosNZ (commented):
As per my comment on the back-end -- this can all happen directly in the database, front-end doesn't even need to know about it.

https://github.com/openmsupply/conforma-server/pull/831#pullrequestreview-1092952158

There's already a bunch of other review_response triggers that run on the back-end that the front-end doesn't get involved with, so this would be consistent with that pattern.

PR #1335, opened 2022-09-05
#1333 Search plugin improvements (and list builder)

### LINKED ISSUE #1333: Search plugin -- not great when only one field in result objects
The "displayFormat" and "resultFormat" paramaters are not behaving very nicely when there isn't at least 2 fields in the search results -- 1 for "title" and one for "description".

Also -- try and make it look a little nicer for simple results (card is a bit much)

CarlosNZ:
Fix #1333 

- can now handle when there is only one of either `title` or `description` provided in "displayFormat"
- additional parameter "displayType" with the option to display results as 'list', which looks much nicer when there is just a single field of information:  
Screen Shot 2022-09-05 at 1 06 03 PM
- applied same "list" option to listbuilder, which already has a "displayType" parameter.
- updated docs (listBuilder was missing documentation for existing "inline" view)

Basically, all these changes make the search and listBuilder plugins look nicer with simple results (i.e. only a single field)

See them in action in Product registration (Angola):
- Active ingredient search (in listbuilder) (Module 1, pg 6)
- Dosage forms listBuilder (Module 1, pg 6)


src/formElementPlugins/listBuilder/src/styles.less
```diff
@@ -18,3 +18,14 @@
...
+/* This allows the delete (x) icon to become visible when hovering over the
+whole list item */
+
+.list-list-item-icon {
+  visibility: hidden;
+}
+
+.list-list-item:hover .list-list-item-icon {
+  visibility: visible;
+}
```
CarlosNZ, line 31: Same pattern used elsewhere to create hover-delete icons

src/formElementPlugins/search/src/ApplicationView.tsx
```diff
@@ -126,27 +136,10 @@ const ApplicationView: React.FC<ApplicationViewProps> = ({
...
-      case 0:
-        return fieldType === 'title' ? strings.MESSAGE_NO_RESULTS : ''
-      case 1:
-        return fieldType === 'title' ? fields[0] : result[fields[0]]
-      default:
-        return fieldType === 'title'
-          ? `${fields[0]}: ${result[fields[0]]}`
-          : `${fields[1]}: ${result[fields[1]]}`
-    }
-  }
```
CarlosNZ, line 145: Just moved this outside component so it can be called from the Display component as well

src/formElementPlugins/listBuilder/src/displayComponents/index.tsx
```diff
@@ -1,5 +1,6 @@
 import ListCardLayout from './ListCardLayout'
 import ListTableLayout from './ListTableLayout'
 import ListInlineLayout from './ListInlineLayout'
+import ListListLayout from './ListListLayout'
 
-export { ListCardLayout, ListTableLayout, ListInlineLayout }
+export { ListCardLayout, ListTableLayout, ListInlineLayout, ListListLayout }
```
nmadruga, line 6: They would be better called `ListBuilder...` but up to you if you want to change that

src/formElementPlugins/search/src/ApplicationView.tsx
```diff
@@ -243,6 +272,7 @@ export const DisplaySelection: React.FC<DisplayProps> = ({
             {subtitle && (
               <Card.Meta>
                 <Markdown text={substituteValues(subtitle, item)} semanticComponent="noParagraph" />
+                SUBTITLE
```
nmadruga, line 275: Hmm do we already have a localisation file for this plugin?

nmadruga (approved):
Looks cool. I was wondering why to use the default as "cards" if that's the one that requires specific formatting (for `displayFormat`) and not just use list?

CarlosNZ:
> I was wondering why to use the default as "cards" if that's the one that requires specific formatting (for `displayFormat`) and not just use list?

Based on how it is now, yes I agree "list" would be better. However, that would break a lot of existing templates that are using the default, so I thought best not to change now.

src/formElementPlugins/search/src/ApplicationView.tsx
CarlosNZ: This is a mistake -- sorry, I'll put it back to how it was.

CarlosNZ: But yes, there is a localisation file for this one. 

Removed string: a4449590

src/formElementPlugins/listBuilder/src/displayComponents/index.tsx
CarlosNZ: Yeah, it's not great, will rename them all at some point.

PR #1338, opened 2022-09-06
#1214 - Warning on Discontinued reviews (Unassigned)

### LINKED ISSUE #1214: Show message to user unassigned to review on the ReviewPage
IMPORTANT: PSM wrote: Approved and recommended to be in initial MVP (02/06/2022)

### Reference 

**Final Requirements Matrix 2** - Customisations - Row 12 

### Requirement discussion
- [x] Fix https://github.com/openmsupply/conforma-web-app/issues/1216 (Not requested - but should helps to remove confusion)
It's an edge case, since usually the re-assignment is done because the original reviewer is not working on the review, but it happened during a workshop. and it's confusing to the person reviewing when try to submit and nothing happens.

nmadruga:
Fixes #1214 
Back-end PR: [837](https://github.com/openmsupply/conforma-server/pull/837)

Tests: Please use snapshot `dev/snapshots/Unassignment_warning_test` for testing this. It won't work currently in out angola snapshot since the actions on event ON_REVIEW_ASSIGN and ON_REVIEW_UNASSIGN are not correctly set/missing.

Once you load the `Unassignment_warning_test` snapshot please run `yarn migrate` to get changes on the back-end working.

Then test: 

1. load as `jess` and see the Review of http://localhost:3000/application/S-OSW-0002/review?tab=assignment
2. Open the review page, make changes if desired - but don't submit
3. Open an incognito browser and log in as `nicole` then in http://localhost:3000/application/S-OSW-0002/review?tab=assignment do Un-assignment of jess. 
4. In the other tab looking at the Review page, if you try to add a new review_element, click "Approve all", submit the review or reload it should show the Warning:
Screen Shot 2022-09-06 at 1 12 50 PM

No changes can be done - but you see the previous added elements:
Screen Shot 2022-09-06 at 1 13 08 PM

If this the other user (`nicole`) now Re-assigning back to user `jess`. Then the review should be able to load fine and continued.

### Changes
- Back-end: Fix activity_log functions that caused a problem when running action `ChangeStatus` to set the Review to DISCONTINUED on Un-assignment.
- Updated elements on page to not-editable when the Review status is not DRAFT. Strange that this was allowed before!

### DIFF
--- a/src/containers/Review/ReviewPage.tsx
+++ b/src/containers/Review/ReviewPage.tsx
@@ -76,19 +76,16 @@ const ReviewPage: React.FC<{
   })
 
   useEffect(() => {
-    if (reviewStatus === ReviewStatus.Pending) {
-      console.log('reviewStatus: PENDING')
+    if (reviewStatus === ReviewStatus.Pending)
       showWarningModal({
         title: strings.REVIEW_STATUS_PENDING_TITLE,
         message: strings.REVIEW_STATUS_PENDING_MESSAGE,
       })
-    } else if (reviewStatus === ReviewStatus.Discontinued) {
-      console.log('reviewStatus: DISCONTINUED')
+    else if (reviewStatus === ReviewStatus.Discontinued)
       showWarningModal({
         title: strings.REVIEW_STATUS_DISCONTINUED_TITLE,
         message: strings.REVIEW_STATUS_DISCONTINUED_MESSAGE,
       })
-    }
   }, [reviewStructure])
 
   if (error) return <Message error title={strings.ERROR_GENERIC} list={[error]} />
@@ -100,10 +97,7 @@ const ReviewPage: React.FC<{
     reviewStructure?.thisReview?.current.reviewStatus !== ReviewStatus.Discontinued
   ) {
     const {
-      info: {
-        name,
-        current: { stage },
-      },
+      info: { name },
     } = reviewStructure
 
     return (
--- a/src/containers/Review/ReviewPage.tsx
+++ b/src/containers/Review/ReviewPage.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useState } from 'react'
+import React, { useEffect } from 'react'
 import { Button, Header, Icon, Label, Message, Segment } from 'semantic-ui-react'
 import {
   Loading,
@@ -52,7 +52,6 @@ const ReviewPage: React.FC<{
   const {
     userState: { currentUser },
   } = useUserState()
-  const [reviewStatus, setReviewStatus] = useState(ReviewStatus.Submitted)
 
   const { push } = useRouter()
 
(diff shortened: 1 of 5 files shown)


src/containers/Review/ReviewPage.tsx
```diff
@@ -48,9 +48,11 @@ const ReviewPage: React.FC<{
   fullApplicationStructure: FullStructure
 }> = ({ reviewAssignment, previousAssignment, fullApplicationStructure }) => {
   const { strings } = useLanguageProvider()
+  const { serial } = fullApplicationStructure.info
   const {
     userState: { currentUser },
   } = useUserState()
+  const [reviewStatus, setReviewStatus] = useState(ReviewStatus.Submitted)
```
CarlosNZ, line 55: Not entirely sure why this needs to be a state variable. Can't it just be derived from `reviewStructure` at any time?

I think it works to just declare it after the `useGetReviewStructureForSections` hook:

```ts
const { reviewStructure, error } = useGetReviewStructureForSections({
    reviewAssignment,
    reviewStructure: fullApplicationStructure,
  })

const reviewStatus = reviewStructure?.thisReview?.current.reviewStatus ?? ReviewStatus.Submitted
```

It'll still get updated whenever the value in `reviewStatus` changes (as reviewStatus will be updated from the hook).

src/containers/Review/ReviewPage.tsx
```diff
@@ -68,19 +70,30 @@ const ReviewPage: React.FC<{
...
+    showCancel: false,
   })
 
   useEffect(() => {
-    if (!reviewStructure) return
-    if (thisReview?.current.reviewStatus === ReviewStatus.Pending) {
+    if (reviewStructure?.thisReview?.current.reviewStatus) {
+      console.log(reviewStructure.thisReview.current.reviewStatus)
+      setReviewStatus(reviewStructure.thisReview.current.reviewStatus)
+    }
```
CarlosNZ, line 81: If you remove the `useState` variable (previous comment), then you won't need this either.

src/containers/Review/ReviewPage.tsx
```diff
@@ -68,19 +70,30 @@ const ReviewPage: React.FC<{
...
+      console.log('reviewStatus: DISCONTINUED')
       showWarningModal({
-        onConfirm: () => push(`/application/${reviewStructure.info.serial}/review`),
-        showCancel: false,
+        title: strings.REVIEW_STATUS_DISCONTINUED_TITLE,
+        message: strings.REVIEW_STATUS_DISCONTINUED_MESSAGE,
       })
     }
-  }, [reviewStructure])
+  }, [reviewStructure, reviewStatus])
```
CarlosNZ, line 96: And you won't need it here either, as it gets updated whenever `reviewStructure` changes anyway.

CarlosNZ: Fyi -- have tested this without requiring `useState`, and it seems to behave the same.

CarlosNZ (changes requested):
Cool, this works nicely. Good job tracking down all (hopefully) the places this is affected by.

Just one main suggestion -- see comments.

src/containers/Review/ReviewPage.tsx
nmadruga: Yes, good call. I made this change earlier when I didn't realise there was a problem about the changeStatus action not running correctly. So I haven't change this after to check that. I'll change this back as you suggested. 👍

CarlosNZ (approved):
Looks good, thanks 👍🏼

PR #1342, opened 2022-09-09
Permission of type view in template builder

fergie-nz:


nmadruga (approved):
👍 all good

Thanks Fergus. I have tested with a user that had the permission it it did allow to view someone else review after done the following steps:
- Add new permissionPolicy with rules as described in [comment](https://github.com/openmsupply/conforma-web-app/issues/524#issuecomment-1241434538)
- Add new associated permissionName and join to a user & organisation
- (Using this PR) Added the permission to VIEW ALL on the Template that should be allowed for the user associated to the permission to view all reviews - I didn't realised before that this was required, so we need to remember that when doing it on the main snapshot.

PR #1344, opened 2022-09-14
#849F Delete template "snapshot" after duplicate

CarlosNZ:
Fix [back-end #848](https://github.com/openmsupply/conforma-server/issues/849) Part 2:

> 2\. when _duplicating_ a template, delete the "snapshot" after import so it doesn't appear in the list (we currently do this from the front-end when _importing_ a template, but needs to happen for duplication too)

Part 1 PR [here](https://github.com/openmsupply/conforma-server/pull/850)

### DIFF
--- a/src/containers/TemplateBuilder/shared/OperationContextHelpers.tsx
+++ b/src/containers/TemplateBuilder/shared/OperationContextHelpers.tsx
@@ -216,7 +216,7 @@ export const duplicateTemplate: TemplateOperationHelper = async (
 
   if (!result) return false
 
-  return await safeFetch(
+  const snapshotResult = await safeFetch(
     getServerUrl('snapshot', {
       action: 'use',
       name: snapshotName,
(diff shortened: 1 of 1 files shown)


src/containers/TemplateBuilder/shared/OperationContextHelpers.tsx
```diff
@@ -225,6 +225,11 @@ export const duplicateTemplate: TemplateOperationHelper = async (
     body,
     setErrorAndLoadingState
   )
+
+  // Delete the snapshot cos we don't want snapshots page cluttered with individual templates
+  safeFetch(getServerUrl('snapshot', { action: 'delete', name: snapshotName }), {}, () => {})
+
+  return snapshotResult
```
CarlosNZ, line 232: This is exactly the same as the change I made to the "importTemplate" method a while back (just below this one)

I should have done it at the same time, but alas I forgot about duplication as well as import.

nmadruga (approved):
All good

PR #1345, opened 2022-09-15
Small fix: Text fallback if optionsDisplayProperty is invalid

CarlosNZ:
This will prevent the problem that occurred due to misconfigured dropDown.

There was an "optionsDisplayProperty" specified, but it wasn't needed as the options query was already returning an array of strings, not an array of objects (as is required for optionsDisplayProperty to work). So it was trying to index a property of a string for its saved text, which resulted in `undefined` (therefore saving `null` to response value).

This change just checks the type of the options value and only tries to find the `optionsDisplayProperty` if it's an object. Otherwise the `optionsDisplayProperty` is quietly ignored.


nmadruga (approved):
Great! Thanks for fixing that Carl. It's was bugging me why it wouldn't fail when running the Graphql query but would when using it in the Dropdown

PR #1347, opened 2022-09-15
Postman collection

CarlosNZ:
I did a big tidy up of this the other day, so thought it would probably be useful to keep it in the repo.

### DIFF
--- a/utils/Conforma.postman_collection.json
+++ b/utils/Conforma.postman_collection.json
@@ -1,0 +1,1104 @@
+{
+  "info": {
+    "_postman_id": "675a8958-a451-49dd-941f-5c6a5c6d9edc",
+    "name": "Conforma",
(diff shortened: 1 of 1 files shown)


nmadruga (approved):
Yes, that's indeed helpful. : )

And we should remember to add in a section for Admins when using Template Builder functionalities to have that would be very helpful.

CarlosNZ:
> And we should remember to add in a section for Admins when using Template Builder functionalities to have that would be very helpful.

Do you mean in the collection? The snapshot endpoints are already included. Or were you meaning something else?

nmadruga:
> Do you mean in the collection? The snapshot endpoints are already included. Or were you meaning something else?

I meant in the documentation TODO for Admin and Template Builder stuff : )

PR #1348, opened 2022-09-15
#1346 Update substituteValues in plugins

### LINKED ISSUE #1346: "Search" and "ListBuilder" substituteValues function should handle nested object paths
The "substituteValues" function in "utilityFunctions" does this correctly.

We should swap out the element-specific "substituteValues" functions from "Search" and "ListBuilder" plugins to use this generic one, which is more powerful. 

(But check there's no regressions either)

CarlosNZ:
Fix #1346 

Removes the plugin-specific "substituteValues" functions and replaces them with the generic one from utils. It does exactly the same thing, except the generic one is recursive, so it can handle nested object paths.

Noticed this when I wanted the `displayFormat` value to include `genericNames.text`, but it wouldn't work cos it didn't drill down through the object path.

Generic one does the trick (and should have been done a while ago).

Required for the "displayFormat" of the Product card in "Cancel Product" to display correctly.

### DIFF
--- a/src/formElementPlugins/listBuilder/src/displayComponents/ListCardLayout.tsx
+++ b/src/formElementPlugins/listBuilder/src/displayComponents/ListCardLayout.tsx
@@ -1,7 +1,7 @@
 import React from 'react'
 import { Card, Icon, Label } from 'semantic-ui-react'
 import { ListLayoutProps } from '../types'
-import { substituteValues } from '../helpers'
+import { substituteValues } from '../../../../utils/helpers/utilityFunctions'
 
 const ListCardLayout: React.FC<ListLayoutProps> = ({
   listItems,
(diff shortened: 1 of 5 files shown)


CarlosNZ:
FYI, the effect of this change can be seen on the "Product Cancellation" template, where the card format references a nested property `genericNames.text`:
 
Screen Shot 2022-09-16 at 11 25 08 AM

Without this branch active, you'll notice the "Nome comum" doesn't show up.

nmadruga (approved):
I trust this is doing its job, although I don't really understand why this helps to display the other property from the displayFormat object 🤣

CarlosNZ:
> I don't really understand why this helps to display the other property from the displayFormat object

Because it's recursive, whereas the internal one was just one-level. (It's basically the same as what the "objectProperties" operator in evaluator uses)

PR #1353, opened 2022-09-19
#1351 - Hide option to edit/add reviews on elements in re-assigned sections

### LINKED ISSUE #1351: Bug: Reviewable elements on section un-assigned from review
Just brief explanation - when All sections were assigned and review started, if one or more sections got re-assigned to a different user, even though the review completeness would only consider review of element in the assigned sections, all the elements were still "reviewable". 

The expected behaviour would be that elements on un-assigned sections don't show the option to review anymore.  
Also check that - if for some reason - these sections are re-assigned back to this reviewer, these elements are set back to be reviewable in this review : )

nmadruga:
Fixes #1351 

We already have the logic in place to only consider elements reviewed that are in assigned sections to determinate the overall decision. So, the only change required was to hide the option to edit elements that aren't part of the assigned sections...

### Example
1. The reviewer `Nicole Madruga` had all sections assigned, and started to review Section 1:
#### Overview of sections assigned
Screen Shot 2022-09-19 at 12 59 11 PM

#### Detailed Section 1 in progress
Screen Shot 2022-09-19 at 12 59 24 PM

2. Now review `Nicole Madruga` has been un-assigned from Section 1:
#### Overview of sections assigned
Screen Shot 2022-09-19 at 1 00 12 PM

#### Detailed Section 1 - Un-assigned
PS: The user can still see their previous reviews, not sure if I should hide this too...
Screen Shot 2022-09-19 at 12 59 59 PM

### Overview of sections after Section 2 & 3 are completed
Reviewer can submit after finishing the assigned work (previous logic)
Screen Shot 2022-09-19 at 1 00 39 PM

### DIFF
--- a/src/components/PageElements/Elements/ReviewApplicantResponse.tsx
+++ b/src/components/PageElements/Elements/ReviewApplicantResponse.tsx
@@ -190,17 +190,24 @@ const ReviewApplicantResponse: React.FC<ReviewApplicantResponseProps> = ({
                   ))}
               </ApplicantResponseElement>
               {/* Current review response */}
-              {decisionExists && (
-                <ReviewResponseElement
-                  isCurrentReview={true}
-                  isConsolidation={false}
-                  reviewResponse={reviewResponse}
-                >
-                  {canEdit && isActiveReviewResponse && (
-                    <UpdateIcon onClick={() => setIsActiveEdit(true)} />
+              {isActiveReviewResponse
+                ? canEdit &&
+                  decisionExists && (
+                    <ReviewResponseElement
+                      isCurrentReview={true}
+                      isConsolidation={false}
+                      reviewResponse={reviewResponse}
+                    >
+                      {canEdit && <UpdateIcon onClick={() => setIsActiveEdit(true)} />}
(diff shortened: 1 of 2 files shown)


CarlosNZ:
> PS: The user can still see their previous reviews, not sure if I should hide this too...

It would probably be better if these were hidden too. Can a different user see the review responses of another reviewer's unsubmitted reviews? I don't think so, right? In which case it makes sense not to show unsubmittted responses to _any_ reviewer not assigned to the section.

Would this be hard to do, do you think?

nmadruga:
> > PS: The user can still see their previous reviews, not sure if I should hide this too...
> 
> It would probably be better if these were hidden too. Can a different user see the review responses of another reviewer's unsubmitted reviews? I don't think so, right? In which case it makes sense not to show unsubmittted responses to _any_ reviewer not assigned to the section.
> 
> Would this be hard to do, do you think?

No, it was easy. Changed now, I understood your point and agree.

CarlosNZ (approved):
That's great thanks.

I have a question about how it works with Consolidation, but I'll ask you in the morning.

PR #1354, opened 2022-09-19
#1353 Fix search box

### LINKED ISSUE #1352: Bug: Search box on List page is broken
As soon as you type a single character, the search re-runs, and we lose focus of the search box. So you have to click back in to keep typing. And this keeps happening after every character.

- Fix focus problem (don't lose focus on re-render)
- add some debouncing (use existing `useDebounce` hook)

~~NOTE: This probably broke as part of implementing the "default filters" on the list page, as the problem is not present on older versions that don't have this, such as: https://conforma-demo.msupply.org:50002~~ (Nope, it was earlier than that)

CarlosNZ:
Fix #1352 

See inline comment for what the problem was.

Have also added basic debouncing, but this is unrelated to the main bug

### DIFF
--- a/src/containers/List/ListWrapper.tsx
+++ b/src/containers/List/ListWrapper.tsx
@@ -15,7 +15,7 @@ import ApplicationsList from '../../components/List/ApplicationsList'
 import PaginationBar from '../../components/List/Pagination'
 import ListFilters from './ListFilters/ListFilters'
 import { useGetFilterDefinitions } from '../../utils/helpers/list/useGetFilterDefinitions'
-import Loading from '../../components/Loading'
+import useDebounce from '../../formElementPlugins/search/src/useDebounce'
 
 const ListWrapper: React.FC = () => {
   const { strings } = useLanguageProvider()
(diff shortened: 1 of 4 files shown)


src/containers/List/ListWrapper.tsx
```diff
@@ -105,8 +108,6 @@ const ListWrapper: React.FC = () => {
     }
   }
 
-  if (loading) return <Loading />
-
```
CarlosNZ, line 109: This was the problem! Was showing the "Loading" component in place of the whole ListWrapper which was wiping out the Search (and filters). There is already a "Loading" component within the list, so we don't need one here as well.

nmadruga (approved):
Great! I guess this makes the "Loading" show less and if there is an error with the Source - i.e GraphQL error - it would display the error after the debouncing time is over?

Also, just flagging this which was noticed now - not translated error. 
Screen Shot 2022-09-21 at 11 16 11 AM

If you could add the Pt version of "Nenhum resultado encontrado" that would be great. Thanks!

CarlosNZ:
> if there is an error with the Source - i.e GraphQL error - it would display the error after the debouncing time is over?

The GraphQL request only gets sent after the debouncing time, so yes.

CarlosNZ:
> If you could add the Pt version of "Nenhum resultado encontrado" that would be great. Thanks!

Cool, done.

PR #1355, opened 2022-09-20
Tweak for dates in correct format (using Locale from localisation file)

nmadruga:
Just add locale to our localisation file and use in Notes tab (Review) and Applicant Deadline

Notes tab example (using locale: 'pt'):
Screen Shot 2022-09-20 at 1 08 24 PM

Considering... do we need to change this as well?
Screen Shot 2022-09-20 at 1 07 04 PM

### DIFF
--- a/src/containers/Review/notes/NotesTab.tsx
+++ b/src/containers/Review/notes/NotesTab.tsx
@@ -44,7 +44,10 @@ const NotesTab: React.FC<{
   state: NotesState
   setState: (state: NotesState) => void
 }> = ({ structure: fullStructure, state, setState }) => {
-  const { strings } = useLanguageProvider()
+  const {
+    strings,
+    selectedLanguage: { locale },
+  } = useLanguageProvider()
   const {
     userState: { currentUser },
   } = useUserState()
@@ -132,7 +135,7 @@ const NotesTab: React.FC<{
                   </p>
                   <p className="slightly-smaller-text dark-grey">
                     {DateTime.fromISO(note.timestamp)
-                      .setLocale(strings.LOCALE)
+                      .setLocale(locale)
                       .toLocaleString(DateTime.DATE_FULL)}
                   </p>
                 </div>
--- a/src/containers/Review/notes/NotesTab.tsx
+++ b/src/containers/Review/notes/NotesTab.tsx
@@ -131,7 +131,9 @@ const NotesTab: React.FC<{
                     <strong>{note.user?.fullName}</strong>
                   </p>
                   <p className="slightly-smaller-text dark-grey">
-                    {DateTime.fromISO(note.timestamp).toLocaleString(DateTime.DATETIME_MED)}
+                    {DateTime.fromISO(note.timestamp)
+                      .setLocale(strings.LOCALE)
+                      .toLocaleString(DateTime.DATE_FULL)}
                   </p>
                 </div>
               </div>
--- a/src/formElementPlugins/datePicker/src/ApplicationView.tsx
+++ b/src/formElementPlugins/datePicker/src/ApplicationView.tsx
@@ -3,6 +3,8 @@ import { ApplicationViewProps } from '../../types'
 import SemanticDatepicker from 'react-semantic-ui-datepickers'
 import { DateTime } from 'luxon'
 import 'react-semantic-ui-datepickers/dist/react-semantic-ui-datepickers.css'
+import { useLanguageProvider } from '../../../contexts/Localisation'
+import { LocaleOptions } from 'react-semantic-ui-datepickers/dist/types'
 
 // Stored response date format
 interface DateSaved {
@@ -35,13 +37,15 @@ const ApplicationView: React.FC<ApplicationViewProps> = ({
   const [selectedDate, setSelectedDate] = useState<SelectedDateRange>(
     fromDateSaved(currentResponse?.date)
   )
+  const {
+    selectedLanguage: { locale },
+  } = useLanguageProvider()
   const { isEditable } = element
   const {
     label,
     description,
     default: defaultDate,
     allowRange = Array.isArray(defaultDate),
-    locale = 'en-US',
     displayFormat = 'short',
     entryFormat = 'YYYY-MM-DD',
     minDate,
@@ -87,7 +91,7 @@ const ApplicationView: React.FC<ApplicationViewProps> = ({
       )}
       <Markdown text={description} />
       <SemanticDatepicker
-        locale={locale}
+        locale={(locale as LocaleOptions) || 'en-US'}
         onChange={handleSelect}
         type={allowRange ? 'range' : 'basic'}
         value={selectedDate}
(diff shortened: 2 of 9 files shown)


src/utils/defaultLanguageStrings.ts
```diff
@@ -208,6 +208,7 @@ export default {
   LABEL_VERIFICATION_FAILED: 'Verificaton Failed',
   LINK_NEW_ACCOUNT: 'Create new account',
   LINK_RESET_PASSWORD: 'Forgot my password',
+  LOCALE: 'en-GB',
```
CarlosNZ, line 211: This shouldn't be a string. It should be in the top-level "languages.json" file, e.g.:
```
  {
    "languageName": "Portugese",
    "description": "Portugese translation for Angola users",
    "code": "pt_an",
    "flag": "🇦🇴",
    "locale": "pt_PT"
    "enabled": true
  },
```

Fetch in the Language Provider and make it part of the global Language state.

CarlosNZ:
> Considering... do we need to change this as well?

If you follow my last comment, then you'll be able to fetch this from global language state (but keep this as fallback)

src/utils/defaultLanguageStrings.ts
nmadruga: Yes, that's cool. I think we can actually use the existing `code` field, no?

nmadruga: Oh, no I see we use that to differentiate 2 types of the same language too...

CarlosNZ: > Yes, that's cool. I think we can actually use the existing `code` field, no?

Yeah, originally I thought the "code" would be equivalent to the locale, but we've kind of diverged from that.

nmadruga:
Ok @CarlosNZ have done the changes I imagine needs to be in place to get the `locale` on the languages file. Please use the snapshot `dev/snasphots/Test_dates_using_locale` to test. I have also added a one-line PR on Front-end with change related to this: LINK

src/formElementPlugins/datePicker/src/ApplicationView.tsx
```diff
@@ -35,13 +37,15 @@ const ApplicationView: React.FC<ApplicationViewProps> = ({
...
+  const {
+    selectedLanguage: { locale },
+  } = useLanguageProvider()
   const { isEditable } = element
   const {
     label,
     description,
     default: defaultDate,
     allowRange = Array.isArray(defaultDate),
-    locale = 'en-US',
```
CarlosNZ, line 44: We don't want to remove this -- the locale should still be over-ride-able from the element configuration.

Just change the default value instead:
```
locale = selectedLanguage.locale
```
(And remove the nested destructure above to avoid name duplication)

src/formElementPlugins/datePicker/src/ApplicationView.tsx
```diff
@@ -87,7 +91,7 @@ const ApplicationView: React.FC<ApplicationViewProps> = ({
       )}
       <Markdown text={description} />
       <SemanticDatepicker
-        locale={locale}
+        locale={(locale as LocaleOptions) || 'en-US'}
```
CarlosNZ, line 94: Undo, as per above comment.

src/utils/defaultLanguageStrings.ts
CarlosNZ, line 211: Just need to remove this now.

src/containers/Review/notes/NotesTab.tsx
```diff
@@ -131,7 +134,9 @@ const NotesTab: React.FC<{
                     <strong>{note.user?.fullName}</strong>
                   </p>
                   <p className="slightly-smaller-text dark-grey">
-                    {DateTime.fromISO(note.timestamp).toLocaleString(DateTime.DATETIME_MED)}
+                    {DateTime.fromISO(note.timestamp)
+                      .setLocale(locale)
+                      .toLocaleString(DateTime.DATE_FULL)}
```
CarlosNZ, line 139: Shouldn't it still be `DateTime.DATETIME_MED` ?

src/utils/hooks/useTimeline/useTimeline.tsx
```diff
@@ -157,15 +163,17 @@ const generateTimelineEvent: {
...
+    locale: string
   ) => EventOutput
 } = {
   STAGE: () =>
     // Already handled in caller
     ({ eventType: TimelineEventType.Ignore, displayString: '' }),
   STATUS: (event, fullLog, _, __, strings) => getStatusEvent(event, fullLog, strings),
   OUTCOME: (event, _, __, ___, strings) => getOutcomeEvent(event, strings),
-  EXTENSION: (event, _, __, ___, strings) => getExtensionEvent(event, strings),
+  EXTENSION: (event, _, __, ___, strings, _____, locale) =>
```
CarlosNZ, line 175:
```suggestion
  EXTENSION: (event, _, __, ___, strings, ____, locale) =>
```
Getting a little carried away with your underscores there! 😜

src/utils/localisation/exportLanguages.ts
```diff
@@ -53,6 +53,8 @@ export const exportLanguages = async (
     const row3code = ['Code:', '', ...languageOptions.map((opt) => opt.code)]
     const row4Flag = ['Flag:', '', ...languageOptions.map((opt) => opt.flag)]
     const row5Enabled = ['Enabled?', '', ...languageOptions.map((opt) => opt.enabled)]
+    const row6Locale = ['Locale:', '', ...languageOptions.map((opt) => opt.locale)]
```
CarlosNZ, line 56: I think the "Enabled" row should always be last.

CarlosNZ (changes requested):
Cool, looking good. 

There's also a couple of cells in the Application List you might want to update too:
- DeadlineCell
- LastActiveDateCell

src/containers/Review/notes/NotesTab.tsx
nmadruga: No because it shows dates as Sept. or Oct. instead of the full "Setembro" as we normally use in other languages. And it also looks better this way - in my opinion.

src/utils/hooks/useTimeline/useTimeline.tsx
nmadruga: : P

src/utils/localisation/exportLanguages.ts
nmadruga: Hmmmmm ok

src/containers/Review/notes/NotesTab.tsx
CarlosNZ: Okay, that's fine -- just make sure it looks fine in NZ/English as well. 👍

nmadruga:
All done - just didn't change on `LastActiveDateCell` because that one isn't a date, but a period like "2 days ago..."

CarlosNZ (changes requested):
Sorry -- just one more thing!

Need to update the localisation import script to match the output. Currently, it just crashes (and breaks the back-end language files) when importing.

I think all you need to do is update the fields in here:
Screen Shot 2022-09-21 at 10 57 06 AM

And then try doing an export, then re-import the same file and make sure it works.

CarlosNZ (approved):
Great, looking good thanks 👍

PR #1357, opened 2022-09-23
#32T (Front-end) Check for fully approved review

### LINKED ISSUE #1180: Bug: consolidation should require all sections to submitted before APPROVAL
- Improve the actions to also communicate if the consolidation can be partially done or if all sections are received and can be completed. 
- They would like to do iterative consolidations
- Needs to check if is correctly blocking other reviews to continue if it went back to Applicant with LOQ
- Make sure the Consolidation can only be APPROVED once all sections have been received - check that if one section is empty it isn't required.

nmadruga:
Part 2 - Fixes https://github.com/openmsupply/conforma-templates/issues/32 & fixes #1180 

~~[Back-end Base PR](880) - Part 1~~

### Testing
~~Run `yarn migrate` on latest snapshot~~ It will require to first do some config changes in it!
Use snapshot in `conforma-templates` repo in: `dev/snapshots/AA_Testing_Fully_reviewed_submission_3` to test with one partial review already in place. Log in as this fake user `reviewer`/`ARMED1234`. There are 2 applications for ProdRegistration in to be reviewed. I have already assigned so 1 can be completed (as fully assigned) and the other can't. 
The setup is that applications:
 - "Test Medicamento" only has Section 1 elements submitted => If Reviewer approves it should go to Stage 2
 - "Test Medicamento 2" has one extra element in Section 2 submitted => Reviewer has been assigned to Section 1 only. When submitting the review, it should show a warning and allow to submit but **keep the application in Stage 1**.
 - You can try after that assign the other Section and try to submit... I haven't tried this step.

### Changes
- Warning when user submits Review/Consolidation that doesn't have all assigned and submitted questions (reviewable)
- The warning shows, but user can still submit - It should allow submitting the partial reviewsin order to allow for different users to review different sections. The last one submitting will trigger the action to Increment Stage correctly.
- I have also to update the **CORE-ACTION** _IncrementStage_ to be checking for fully approved elements... That's what prevent it moving to next stage when not fully reviewed. 
- **This hasn't been added to core_actions snapshot - I'll do before this PR is merged**


src/containers/Review/ReviewSubmit.tsx
```diff
@@ -21,12 +25,29 @@ type ReviewSubmitProps = {
...
+
+  const [fullyAssigned, setFullyAssigned] = useState(false)
+
+  const { data, loading, error } = useGetReviewableQuestionsQuery({
+    variables: {
+      applicationId: info.id,
+      stageId: thisReview?.current.stage.id as number,
+      levelNumber: assignment?.assigneeLevel as number,
+    },
+  })
```
CarlosNZ, line 39: I thought we were going to include this info in the full structure so we don't have to query for it here?

src/containers/Review/ReviewSubmit.tsx
```diff
@@ -21,12 +25,29 @@ type ReviewSubmitProps = {
...
   const reviewDecision = thisReview?.reviewDecision
   const { decisionOptions, getDecision, setDecision, getAndSetDecisionError, isDecisionError } =
     useGetDecisionOptions(canApplicantMakeChanges, assignment, thisReview)
 
+  useEffect(() => {
+    if (!data) return
+    const { assignedQuestions, reviewableQuestions } = data
+    const fullyAssigned = assignedQuestions?.totalCount === reviewableQuestions?.totalCount
+    setFullyAssigned(fullyAssigned)
+  }, [data])
```
CarlosNZ, line 49: In which case, this won't be necessary.

src/containers/Review/ReviewSubmit.tsx
nmadruga: Yes, so I looked at this option but it would be tricky since we don't have the assignment (and which is the level) to pass on when the static part of the structure is queried, and I thought it was simpler to do this here

CarlosNZ: What about in the "getReviewStructure" part? (You probably have a good reason, I'm just wondering)

nmadruga: The logic on the getReviewStructure is way convoluted now to get this added to it... I just thought it would be simpler to do here. But it could be done there, I think

nmadruga:
Well - I'm still not 100% happy, but I have fixed one issue that was causing all review_responses to get removed. I hope that's an ok change, it's in file `useSubmitReview`

src/containers/Review/ReviewSubmit.tsx
```diff
@@ -21,12 +25,29 @@ type ReviewSubmitProps = {
 
 const ReviewSubmit: React.FC<ReviewSubmitProps> = (props) => {
   const {
-    structure: { thisReview, assignment, canApplicantMakeChanges },
+    structure: { info, thisReview, assignment, canApplicantMakeChanges },
   } = props
+
+  const [fullyAssigned, setFullyAssigned] = useState(false)
```
CarlosNZ, line 31: This is that same "derived state" issue I mentioned on something else a while back -- you don't need to create a React state variable for values that are derived purely from existing state variables. 

In this case `fullyAssigned` is just `assignedQuestions?.totalCount === reviewableQuestions?.totalCount`, so you there's no can just state the value directly and no need to wrap it in a `useEffect` either.

src/containers/Review/ReviewSubmit.tsx
CarlosNZ, line 49: I think something like this will work (assuming you remove the "fullyAssigned" state variable above:
```ts
if (!data) return null

const { assignedQuestions, reviewableQuestions } = data
const fullyAssigned = assignedQuestions?.totalCount === reviewableQuestions?.totalCount
```

Without the `useEffect`. There'll be a lot less unnecessary re-renders.

src/containers/Review/ReviewSubmit.tsx
```diff
@@ -21,12 +25,29 @@ type ReviewSubmitProps = {
...
+
+  const { data, loading, error } = useGetReviewableQuestionsQuery({
```
CarlosNZ, line 33: `loading` and `error` not used, so should probably remove.

src/containers/Review/ReviewSubmit.tsx
```diff
@@ -98,7 +127,8 @@ const ReviewSubmitButton: React.FC<ReviewSubmitProps & ReviewSubmitButtonProps>
   })
 
   const { ConfirmModal, showModal: showConfirmModal } = useConfirmationModal({
-    ...messages.REVIEW_SUBMISSION_CONFIRM,
+    onConfirm: () => submission(),
+    showCancel: false,
```
CarlosNZ, line 131: Can we keep the cancel button please? It's very strange not to have an obvious "Cancel" option.

src/utils/defaultLanguageStrings.ts
```diff
@@ -316,6 +316,8 @@ export default {
   REVIEW_SUBMISSION_CONFIRM_OPTION: 'Submitting Review',
   REVIEW_SUBMISSION_FAIL: 'Not all sections have been reviewed',
   REVIEW_SUBMISSION_ERROR: 'Something wrong with server - review mutation failed',
+  REVIEW_SUBMISSION_INCOMPLETE_MESSAGE:
+    'You are about to submit your review. Some sections have not be completed by others, which means your decision is not changing the outcome - if aplicable. Are you sure you wish to proceed?',
```
CarlosNZ, line 320: Typo on `if applicable`. 

But I'm not actually sure what "if applicable" means in this case.

> `Some sections have not be completed by others,`

Do you mean lower level reviewers? If so, we should probably be specific that we mean that. 

And what happens if they go ahead and submit anyway? I thought we wanted to prevent that altogether.

src/utils/hooks/useSubmitReview.ts
```diff
@@ -33,8 +33,6 @@ const useSubmitReview: UseSubmitReview = (reviewId, reload) => {
 
     const reviewResponsesPatches = reviewResponses.map(({ thisReviewLatestResponse }) => ({
       patch: {
-        decision: thisReviewLatestResponse?.decision,
-        comment: thisReviewLatestResponse?.comment,
```
CarlosNZ, line 37: Is this the fix for the incorrectly trimmed review responses? If so, why are we *removing* these? (I'm sure there's a good reason, I'm just a bit confused)

CarlosNZ (changes requested):
> Although I'm still seeing one issue related to the other Reviewer can't start the review on the Assignment tab.

I'm wondering if this is a consequence of changing the "is_reviewable" value from NULL to something else. I know when I added the stuff to do with optional reviewability, I was using the reviewability status to determine various things, like can review be submitted, stuff like that, so maybe we need to check this.

I suggest that you add that "NOT NULL" constraint on the back-end, then re-run the typescript codegen in both repos and I *think* Typescript should give an error wherever the front-end is referring to the "null" state on this field.

src/utils/defaultLanguageStrings.ts
nmadruga: > Do you mean lower level reviewers? If so, we should probably be specific that we mean that.

Well - it could be also other reviewers in the same level - for other sections

nmadruga: > And what happens if they go ahead and submit anyway? I thought we wanted to prevent that altogether.

I think preventing would add a little more headache and the need to make a better process for this. For example, if you have 2 reviewers: One for section 1 and another for section 2. Even if the one for section 2 haven't finished yet, you are submitting your sections 1 as APPROVED. The message is only to say - This is a partial submission. It will have to wait until user approves section 2 to move on to the next stage.

src/containers/Review/ReviewSubmit.tsx
nmadruga: Sure! I'll do that now

src/utils/defaultLanguageStrings.ts
CarlosNZ: > It will have to wait until user approves section 2 to move on to the next stage.

> Well - it could be also other reviewers in the same level - for other sections

Okay, I guess I'm just a little confused as to where and how we're checking for this. It seems like two different scenarios:

1. Consolidator is submitting before all lower-level reviews are done -- in which case we should probably prevent this submission completely (if possible?)
2. First reviewer submitting their section with other sections not completed -- warning seems good, but how are we actually checking for this when running the `incrementStage` action in the back-end?

nmadruga:
> I suggest that you add that "NOT NULL" constraint on the back-end, then re-run the typescript codegen in both repos and I _think_ Typescript should give an error wherever the front-end is referring to the "null" state on this field.

Thanks for the suggestion. I have added the constraint now and saved a new snapshot (I'll update to templates repo shortly). On the front-end interestingly I couldn't found any usage of isReviewableStatus = NULL. I'll keep looking, but seems no big changes to this current logic

src/utils/defaultLanguageStrings.ts
nmadruga: > 1. Consolidator is submitting before all lower-level reviews are done -- in which case we should probably prevent this submission completely (if possible?)

No. Users actually liked this option where we allow submitting the consolidation prior to all lower level users have completed. Also because if there are problems if one section it can be fixed before other sections are reviewed.

We also have already implemented logic to block other lower level reviews to be submitted while waiting for applicant to respond. This also prevents many threads of the same review. It should have only one Consolidation done per application per time.

nmadruga: > 2\. First reviewer submitting their section with other sections not completed -- warning seems good, but how are we actually checking for this when running the `incrementStage` action in the back-end?

Once this is done - we also need to check on IncrementStage condition if `revieweable_questions_count` = `submitted_questions_count`. The reason why I wanted these 2 to be functions. We might even have a one function in the end to check for that

nmadruga:
Over a nightmare conflict resolution I have chosen to Re-PR this one. In #1385 
@CarlosNZ I have replied to your coments here and you can check for changes there please

PR #1358, opened 2022-09-28
Tweaks in application container

nmadruga:
Just 2 tweaks on the Application Home page.

1. Changed the string, which wasn't easy to translate in Portuguese
2. Re-ordered the Title (with Company name) and back button
* It has annoyed me for a while. I tried a few things, and re-ordering so the company shows first and the button to see Applications list after seems a little bit better. 

#### Previously
Screen Shot 2022-09-28 at 12 34 42 PM
Screen Shot 2022-09-28 at 12 35 29 PM

#### With this PR
Screen Shot 2022-09-28 at 2 57 51 PM

### DIFF
--- a/semantic/src/site/globals/site.overrides
+++ b/semantic/src/site/globals/site.overrides
@@ -455,8 +455,6 @@ a:hover {
 
 #content-area .top-container {
   margin-top: 20px;
-  display: flex;
-  flex-direction: column;
   flex-wrap: wrap;
   justify-content: space-between;
 }
--- a/semantic/src/site/globals/site.overrides
+++ b/semantic/src/site/globals/site.overrides
@@ -454,18 +454,20 @@ a:hover {
 }
 
 #content-area .top-container {
+  margin-top: 20px;
+  display: flex;
+  flex-direction: column-reverse;
   flex-wrap: wrap;
   justify-content: space-between;
-  padding-bottom: 20px;
 }
 
 // Back to application list label
 #content-area .back-label {
   min-width: 132px;
-  max-width: 350px;
   background-color: transparent;
   font-size: @fontSmall;
   font-weight: @regularWeight;
+  padding-left: 10%;
 }
 
 // Fix for company header on mobile to not clash with 'back to applications'
--- a/semantic/src/site/globals/site.overrides
+++ b/semantic/src/site/globals/site.overrides
@@ -456,7 +456,7 @@ a:hover {
 #content-area .top-container {
   margin-top: 20px;
   display: flex;
-  flex-direction: column-reverse;
+  flex-direction: column;
   flex-wrap: wrap;
   justify-content: space-between;
 }
--- a/src/utils/defaultLanguageStrings.ts
+++ b/src/utils/defaultLanguageStrings.ts
@@ -148,7 +148,7 @@ export default {
   FILTER_DATE_BEFORE: 'Before',
   FILTER_DATE_AFTER: 'After',
   FILTER_RESET_ALL: 'Reset all filters',
-  LABEL_APPLICATIONS: 'applications',
+  LABEL_APPLICATIONS: '%1 applications',
   LABEL_APPLICATION_SUBMITTED: 'Application Submitted',
   LABEL_APPLICATION_BACK: 'Back to form',
   LABEL_ASSIGNED_TO_YOU: 'Assigned to you',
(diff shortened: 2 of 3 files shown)


fergie-nz:
Looks great from my testing :)

semantic/src/site/globals/site.overrides
```diff
@@ -454,18 +454,20 @@ a:hover {
 }
 
 #content-area .top-container {
+  margin-top: 20px;
+  display: flex;
+  flex-direction: column-reverse;
```
CarlosNZ, line 459: Any particular reason why you used `column-reverse` as opposed to just reversing the order of the elements in the code?

nmadruga: No reason, but it's easier. I think they are not in the same component - so it would be tricky to change otherwise

CarlosNZ: >  I think they are not in the same component

Yes they are:
```ts
<div className={`top-container ${isNonRegistered ? 'hidden-element' : ''}`}>
        <Label
          className="back-label clickable"
          onClick={() => replace(`/applications?type=${code}`)}
        >
          <Icon name="chevron left" className="dark-grey" />
          {strings.LABEL_APPLICATIONS.replace('%1', name)}
        </Label>
        {currentUser?.organisation?.orgName && (
          <Header
            as="h2"
            className="heading-alt"
            textAlign="center"
            content={currentUser?.organisation?.orgName}
          />
        )}
</div>
```

Just becomes:
```ts
      <div className={`top-container ${isNonRegistered ? 'hidden-element' : ''}`}>
        {currentUser?.organisation?.orgName && (
          <Header
            as="h2"
            className="heading-alt"
            textAlign="center"
            content={currentUser?.organisation?.orgName}
          />
        )}
        <Label
          className="back-label clickable"
          onClick={() => replace(`/applications?type=${code}`)}
        >
          <Icon name="chevron left" className="dark-grey" />
          {strings.LABEL_APPLICATIONS.replace('%1', name)}
        </Label>
      </div>
```

And then it works fine with just `flex-direction: column` (with some slight adjustment to the spacing)

It's just very odd to list something backwards, then instruct it to display in reverse.

src/utils/defaultLanguageStrings.ts
```diff
@@ -148,7 +148,7 @@ export default {
   FILTER_DATE_BEFORE: 'Before',
   FILTER_DATE_AFTER: 'After',
   FILTER_RESET_ALL: 'Reset all filters',
-  LABEL_APPLICATIONS: 'applications',
+  LABEL_APPLICATIONS: '%1 applications',
   LABEL_APPLICATION_SUBMITTED: 'Application Submitted',
```
CarlosNZ, line 152: I'm so glad to see this for pluralisation, makes so much more sense and is easier to read/more consistent. Does that mean we can finally ditch the unnecessary (and largely unused) "name_plural" column?

CarlosNZ (commented):
Cool, looks much nicer thanks.

I still think you can remove the "column-reverse", see my comment.

src/utils/defaultLanguageStrings.ts
nmadruga: No. It's still used in the Dashboard. As mentioned previously Plural in other languages aren't as simple as English

nmadruga:
https://github.com/openmsupply/conforma-web-app/pull/1358#discussion_r986300127 Thanks for the suggestion Carl. I'll make this change snow so you can proceed with release. Is that all that was missing?

CarlosNZ:
> Thanks for the suggestion Carl. I'll make this change snow so you can proceed with release. Is that all that was missing?

Yup, looking good. 👍

src/utils/defaultLanguageStrings.ts
CarlosNZ: Okay, I'll have to get you to explain it to me again sometime, cos I don't quite get it. In English, it's just nicer to say "Org Registration applications" rather than "Org Registrations" (so you don't need a specific plural), but I accept that might not translate so well.

semantic/src/site/globals/site.overrides
```diff
@@ -454,18 +454,20 @@ a:hover {
 }
 
 #content-area .top-container {
+  margin-top: 20px;
+  display: flex;
+  flex-direction: column;
```
CarlosNZ, line 459: You actually don't need these two lines, as the parent class `#content-area` already has "display:flex" and "flex-direction: column:", so this is redundant here.

CarlosNZ (approved):
Great looking good.

I just pointed out that you don't need the additional flex definitions as the parent class already has them, so may as well save us a couple of lines of code before you merge. 👍

semantic/src/site/globals/site.overrides
CarlosNZ: Removed before merging :)

Issue #1359, opened 2022-09-29
Bug: Not changing pages on DataView


nmadruga:
Go to Database menu and click on one of the lists, for example "External users". You should check one that has more than one page, then click on the next page and see it doesn't change to the next. Even though it shows the page on the URL the data is the same as page 1. Example:

Check that page 1 and 2 are showing the same - I have removed the URL, Logo and organisation name for privacy:
![Screen Shot 2022-09-29 at 3 06 26 PM](https://user-images.githubusercontent.com/16461988/192922542-6b050cbf-24a9-40ce-b6ca-ed05473b148b.png)
![Screen Shot 2022-09-29 at 3 07 10 PM](https://user-images.githubusercontent.com/16461988/192922548-c6fc4118-d923-4880-bd2f-1b7668431518.png)

CarlosNZ:
Fixed in the branch I'm working on right now. ;)

PR #1360, opened 2022-09-29
866F Data view filtering

CarlosNZ:
Front-end part of https://github.com/openmsupply/conforma-server/pull/873

(Also includes a fix for #1359)

Test/demo with snapshot: `/conforma-templates/dev/snapshots/AA_DataViewImprovements.zip`

Screen Shot 2022-09-30 at 12 32 37 AM

This is pretty massive, sorry, and a lot of the code will probably seem kind of obscure. That's because I've mostly been able to adapt the existing Application List filters to work with the data view tables. So it's mostly a matter of wrangling the generic data view material into the shapes required by the existing filtering infrastructure. And I've had to extend and tweak the existing filter mechanisms to handle the larger range of use cases needed here.

Have a look at the different data views available (in Database menu), and check out the range of filters for each one.

There's so many possible scenarios for these filters, and I'm sure I haven't accounted for every single one, so please try and find additional cases which aren't handled correctly.

There's still a few more things I'd like to add, but this is getting pretty massive already, so I'll do them in a separate PR at some point. In particular, what I've got in mind:

**EDIT**: Most of these now addressed in: https://github.com/openmsupply/conforma-web-app/pull/1364

- [x] A filter for "Number" fields, with a max/min selector to specify a filter range
- [x] Check out how it handles other data types, in particular arrays (it might actually work fine, I just haven't really tried yet)
- [x] A mechanism to "auto-generate" the simplified "text" variants of complex data fields using evaluator expressions, which will save a lot of hassle in template configuration, and keep filter definitions where they belong ([Issue](https://github.com/openmsupply/conforma-server/issues/872) / [PR](https://github.com/openmsupply/conforma-server/pull/874))
- [x] In the "free text" search filters, I'd like the cursor focus to go to the input box when you open the filter pop-up, but this is proving really hard to do, due to the text input field being inside a Drop-down/modal
- [ ] Documentation
- [x] Stronger Typescript definitions (we were pretty loose with this in the early days)

See inline comments for further explanation.


src/utils/helpers/utilityFunctions.ts
```diff
@@ -35,3 +35,16 @@ export const substituteValues = (
...
+*/
+
+// Use anything obscure than doesn't have special meaning in urls or regex
+const MAGIC_STRING = '~~~'
+
+export const removeCommas = (value: string) => value.replace(/,/g, MAGIC_STRING)
+export const replaceCommas = (value: string) => value.replace(new RegExp(MAGIC_STRING, 'g'), ',')
+
+export const removeCommasArray = (values: string[]) => values.map((value) => removeCommas(value))
+export const replaceCommasArray = (values: string[]) => values.map((value) => replaceCommas(value))
```
CarlosNZ, line 50: The whole "commas within filter option values" thing was a major pain in my ass:


Because multiple values are separated in the url using a comma, the query parser has no way to tell which commas are separators and which are actual commas!

So I've made these little helpers, and we replace the commas from strings at the last possible moment (just before being inserted into the url (with `updateQuery`) and replace them again immediate after parsing the query.

Open to suggestions for the MAGIC_STRING value. Needs to be something not too ugly that draws attention to itself in the url, but obscure enough that it's very unlikely to be used in an actual string. Oh, and not use any characters that have semantic significance in urls (e.g. `&`, `?`) or Regex (e.g. `\`, `.`,`^` etc)

src/utils/hooks/useDataViews.tsx
```diff
@@ -42,35 +44,58 @@ export const useDataViewsList = () => {
...
   } = useUserState()
 
+  const updateDataViewTable = (newDataViewTable: DataViewsTableResponse) => {
+    // We only want to load filterDefinitions once (per dataView code),
+    // otherwise it gets into an infinite loop. Similarly, we don't want to set
+    // the dataViewTable data until we've loaded filterDefinitions, or the Table
+    // page will display the full list briefly before showing the filtered list.
+
+    // TO-DO: Fetch filter definitions in a separate query *before* requesting
+    // dataViewTable
```
CarlosNZ, line 75: The filter definitions are received in the same query/response as the actual table data.

This is not ideal, as when loading a page with filters already applied (in url), we have to fetch the full (unfiltered) data just to get the filter definitions (i.e. how to display the current url filters), and then fetch the filtered data again after that.

I've got this little workaround here to "hide" the re-fetch (i.e. don't render table until filter definitions are loaded), but it'd be better to make a separate endpoint just to query filter definitions *before* querying the table data (with filters applied). 

The only reason I didn't do that already is because it's slightly more efficient in the back-end to build both table columns and filter definitions in one go, since they share a lot of the same database queries and depend on each other to some extent. 

Something to think about for later...

src/utils/hooks/useDataViews.tsx
```diff
@@ -83,59 +108,37 @@ export const useDataViewsDetail = ({ dataViewCode, recordId }: DataViewDetailsPr
...
 }
 
-const processGetRequest = (
-  url: string,
-  setErrorMethod: (_: ErrorResponse | null) => void,
-  setLoadingMethod: (_: boolean) => void,
+interface RequestProps {
+  url: string
+  setError: (_: ErrorResponse | null) => void
+  setLoading: (_: boolean) => void
```
CarlosNZ, line 125: It was a dumb idea to split these into two separate methods for "get" and "post", I just did it cos I was in a hurry before. This is much nicer :)

src/utils/hooks/useDataViews.tsx
```diff
@@ -168,34 +174,33 @@ const buildFilterDefinitions = (tableData?: DataViewsTableResponse): FilterDefin
...
+
+  return returnFilterDefinitions
+}
 
-  return filterDefinitions
+const filterTypeMap: { [key: string]: FilterTypes } = {
+  string: 'dataViewString',
+  number: 'dataViewNumber',
+  boolean: 'dataViewBoolean',
+  Date: 'date',
```
CarlosNZ, line 205: In `<ListFilters/>`, the values "string", "boolean" etc. are already mapped to ApplicationList filters. So I'm remapping the data types to new keys here so we can map them to different ones. (Note that "date" is still using the ApplicationList date filter unmodified)

src/utils/types.ts
```diff
@@ -826,8 +843,14 @@ export type FilterTypeOptions = {
...
-  // For Data View filter lists
-  getFilterList?: (table: string, column: string) => Promise<string[]>
+  // For Data View filters
+  column?: string
+  code?: string
+  dataType?: string
+  showFilterList?: boolean
+  searchFields?: string[]
+  delimiter?: string
+  nullString?: string
```
CarlosNZ, line 853: This is horrible -- and really should be updated to union types for better type checking within each filter type.

src/containers/DataDisplay/DataViewTable.tsx
```diff
@@ -108,7 +120,8 @@ const DataViewTableContent: React.FC<DataViewTableContentProps> = ({
   const showToast = useToast({ style: 'negative' })
 
   const { headerRow, tableRows, totalCount } = dataViewTable
-  const showDetailsForRow = (id: number) => push(`/data/${dataViewCode}/${id}`)
+  const showDetailsForRow = (id: number) =>
+    push(`/data/${dataViewCode}/${id}`, { dataTableFilterQuery: location.search })
```
CarlosNZ, line 124: This passes the current filter (url query) state to the detail page, so when we click the "back to Table" link, it restores the correct filter state.

nmadruga:
> There's still a few more things I'd like to add, but this is getting pretty massive already, so I'll do them in a separate PR at some point. In particular, what I've got in mind:
> 
> * [ ]  A filter for "Number" fields, with a max/min selector to specify a filter range
> * [ ]  Check out how it handles other data types, in particular arrays (it might actually work fine, I just haven't really tried yet)
> * [ ]  A mechanism to "auto-generate" the simplified "text" variants of complex data fields using evaluator expressions, which will save a lot of hassle in template configuration, and keep filter definitions where they belong ([Issue](https://github.com/openmsupply/conforma-server/issues/872) -- I've started doing this and I think it's working great)
> * [ ]  In the "free text" search filters, I'd like the cursor focus to go to the input box when you open the filter pop-up, but this is proving really hard to do, due to the text input field being inside a Drop-down/modal
> * [ ]  Documentation
> * [ ]  Stronger Typescript definitions (we were pretty loose with this in the early days)

I suggest creating another issue with this listed. So we keep it on the record for next releases 👼

src/containers/DataDisplay/DataViewFilters.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+import React, { useState, useEffect } from 'react'
+import { Dropdown, Input } from 'semantic-ui-react'
+import { FilterContainer, FilterOptions } from '../List/ListFilters/common'
+import { useLanguageProvider } from '../../contexts/Localisation'
+import getServerUrl from '../../utils/helpers/endpoints/endpointUrlBuilder'
+import { postRequest } from '../../utils/helpers/fetchMethods'
+import useDebounce from '../../formElementPlugins/search/src/useDebounce'
+
+export const DataViewSearchableList: React.FC<any> = ({
```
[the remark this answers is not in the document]
nmadruga: Oh, I see. So it ended up not being a reuse, but fork? were there common components used for both, yes?

src/utils/helpers/list/buildQueryFilters.ts
```diff
@@ -67,8 +105,13 @@ const constructFilter = (
   filterValue: string
 ) => {
   const filter = filterTypeDefinitions[type](filterValue, options)
+
   // If all of the criteria is undefined skip filter
-  if (!Object.values(filter).find((value) => value !== undefined)) return {}
+  // if (!Object.values(filter).find((value) => value !== undefined)) return {}
```
nmadruga, line 110: Maybe remove

src/utils/helpers/list/buildQueryFilters.ts
```diff
@@ -78,10 +121,23 @@ const constructFilter = (
...
+    return { inInsensitive: replaceCommasArray(valuesArray) }
+
+  const nonBlankValues = valuesArray.filter((val) => val !== nullString)
+  return {
+    or: [
+      { [searchField]: { inInsensitive: replaceCommasArray(nonBlankValues) } },
+      { [searchField]: { isNull: true } },
+    ],
+  }
+}
```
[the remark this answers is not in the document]
nmadruga: Yep. Looks like it was fun indeed. But this part if very difficult to read that's for sure. I wonder it would be worth just adding those new fields `searchField` and `nullString` as optionals and set a default value. Wouldn't it be simple?

src/utils/helpers/utilityFunctions.ts
```diff
@@ -35,3 +35,16 @@ export const substituteValues = (
   // Match ${...} using regex and replace ... with property from object
   return parameterisedString.replace(/(\${)(.*?)(})/gm, getObjectProperty)
 }
+
+/*
+ We are using ',' as a delimiter for url queries with multiple values. This means that values with *actual* commas in them get messed up. These functions can be used to remove commas before updating url query, then re-inserting them after extraction from url query
+*/
```
nmadruga, line 41:
```suggestion
/*
 We are using ',' as a delimiter for url queries with multiple values. This means 
 that values with *actual* commas in them get messed up. These functions 
 can be used to remove commas before updating url query, then re-inserting 
 them after extraction from url query
*/
```

src/utils/helpers/utilityFunctions.ts
nmadruga: I do like the pipe `|` for that. And I think that was a smart solution.

nmadruga (commented):
Ok - I tried to look at all the code here. It's a lot to take indeed. I'll have to continue on Monday, since I still have things to Review in Portuguese translations... I haven't looked at the Backend code yet. And tested... But here is a initial feedback

src/containers/DataDisplay/DataViewFilters.tsx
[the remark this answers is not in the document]
CarlosNZ: All the render elements (FilterContainer, Input, Dropdown.Divider, FilterOptions) are the same, just slightly different props for slightly different data structures, but to the user it should look identical.

src/utils/helpers/list/buildQueryFilters.ts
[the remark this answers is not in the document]
CarlosNZ: I would have, but you can only have one optional argument. :( 

I would kind of like to re-write some more of this, but I was very conscious not to break existing ApplicationList functionality.

src/utils/helpers/utilityFunctions.ts
CarlosNZ: Yeah, that's not bad, but it would need to be at least a double one `||` as it's not unreasonable to expect a pipe in an actual data value.

You think it would be worth just putting this in preferences? And then if a country does find themselves using this character, then they can change it?

(I really would like to change it so that `+` is our global url delimiter -- that's more conventional, but that's a significant re-factor)

src/utils/helpers/list/buildQueryFilters.ts
CarlosNZ: Yeah, I was going to, but slightly concerned it might be needed for something in the ApplicationList. I haven't been able to see *what*, but I'm wary to remove it completely just in case. 

Maybe I'll remove it when we've had the filters feature branch functioning and well tested (along with the applications list). I'll do another overall tidy-up before merging the whole `feature/data-view-filters` into `develop`.

src/utils/helpers/utilityFunctions.ts
CarlosNZ: 34069ef9

src/utils/helpers/list/buildQueryFilters.ts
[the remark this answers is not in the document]
CarlosNZ: Ah -- there it is, now I can see the comment again. Odd 🤔

Anyway, changed the input parameters to an object, which is much better for handling multiple option arguments.
b322d6e8

src/utils/helpers/utilityFunctions.ts
nmadruga: Yes, I think it would be nice having that as a preferences - but probably wouldn't be used, so not to spent too long only if you think it's a quicker change. I wouldn't go with the significant refactor option for now LOL... But ideally the day we do this refactor is to also have the `+` substitution on the preferences (worth to put the idea on a issue though)

src/utils/helpers/list/buildQueryFilters.ts
```diff
@@ -78,10 +126,32 @@ const constructFilter = (
...
+
+  // Null values (passed in as nullString) require special handling
+  const nonBlankValues = valuesArray.filter((val) => val !== nullString)
+  return {
+    or: [
+      { [searchField]: { inInsensitive: replaceCommasArray(nonBlankValues) } },
+      { [searchField]: { isNull: true } },
+    ],
+  }
+}
```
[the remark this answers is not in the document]
nmadruga: I can still see my comment - but yes. This PR is getting massive! Going to have a second look on it now and hopefully we can merge this in to this feature branch and only will have the next PR with improvements to check and do further improvements from there YEY

src/utils/types.ts
CarlosNZ: Actually, I didn't really change anything here -- I just made the input props for each filter stricter (removed the `any`s)

nmadruga (changes requested):
Ok - on this second look AND tests I liked a lot how the filters are working - combined with sorting is very nice. I have just a few minor requests on a few filters using English terms... On the ApplicationList we have done something to fix that - which should be used similarly here:
The `Address` which is already a translated column on DataViewColumnDetails
Screen Shot 2022-10-05 at 4 55 29 PM
The values for `Estatus` needs to be using proper localisation
Screen Shot 2022-10-05 at 4 55 15 PM

CarlosNZ:
> The `Address` which is already a translated column on DataViewColumnDetails

> The values for `Estatus` needs to be using proper localisation

These are all just config settings and the ones here are for demo purposes only. Having said that, I've mapped the status to the correct Portuguese values in the next PR's snapshot anyway (and made a bunch of other changes, but again just for demo purposes)

src/containers/DataDisplay/DataViewFilters.tsx
```diff
@@ -0,0 +1,122 @@
...
+  const [filterList, setFilterList] = useState<string[]>()
+  const [error, setError] = useState<string>()
+  const [moreResults, setMoreResults] = useState(false)
+  const [debounceOutput, setDebounceInput] = useDebounce(searchText)
+
+  const { column, code, searchFields, delimiter } = options
+
+  useEffect(() => {
+    postRequest({
+      url: getServerUrl('dataViews', { dataViewCode: code, column }),
```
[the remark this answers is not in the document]
CarlosNZ: Which part, specifically? The code and column are dynamic values, and the `dataViews` is a strictly-typed string-literal so basically servers the same purpose as a CONSTANT., Or did you mean something else?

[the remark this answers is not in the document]
nmadruga: Yes, I meant `dataViews` I noticed we have been using constant straight on calls in other places too. That's recommended to be done with a constant to make it easier to refactor. But that's very minor

nmadruga (approved):

src/containers/DataDisplay/DataViewFilters.tsx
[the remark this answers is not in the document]
CarlosNZ: We did it with strict typing instead. So if you put a non-allowed value in there, it'll immediately show an error. So it's effectively an enum, defined the the "endpoint url builder".

We refactored this url builder quite substantially while you were away, so if you think it's missing anything, please make an issue (or bring it up in person).

PR #1362, opened 2022-09-29
#1361 Application list -- default filters fix

### LINKED ISSUE #1361: Bug: Default Filters not showing up on Application List
When you refresh the page or change lists from the list page they do, but when navigating from the Dashboard they don't.

CarlosNZ:
Fix #1361 

The problem was, when going to the page from the dashboard, the "defaultDisplayFilters" was empty, because the `userRole` had not yet been defined from the query, but we initialised `activeFilters` with this anyway. 

And then when the query updated, the active filters got over-written with `[...activeFilters, ...newFilters]`, which was also empty. And even though`defaultDisplayFilters` was updated, the state of `activeFilters` didn't change, since it was only the initialised value that depended on it.

This change just moves consideration of `defaultDisplayFilters` into the `useEffect` that happens when the query changes -- and if there are no filters already active or defined in the query, it'll use the default ones.

This also means that if you load a page with a filter already defined in the url, you'll only get those filters activated, not the default ones. I assume this is okay? I think it makes sense, but feel free to disagree.

---

I've also added defaults for when the "defaultFilters" is not defined in prefs. We should always have defaults for things like this, preferences should just be for when we want it to be different to the default.

### DIFF
--- a/src/containers/List/ListFilters/ListFilters.tsx
+++ b/src/containers/List/ListFilters/ListFilters.tsx
@@ -39,13 +39,15 @@ const ListFilters: React.FC<{
   const defaultDisplayFilters = getDefaultDisplayFilters(filterDefinitions)
 
   const filterNames = Object.keys(displayableFilters)
-  const [activeFilters, setActiveFilters] = useState<string[]>(defaultDisplayFilters)
+  const [activeFilters, setActiveFilters] = useState<string[]>([])
 
   // Add filters from URL to activeFilters, filters may not have criteria/options yet thus we have to use activeFilter state variable
   useEffect(() => {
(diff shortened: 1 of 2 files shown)


nmadruga (approved):
I think default should be no default filters (if not defined in the config file) but we already had this discussion previously. And changed here, I still prefer without this, but we want to have the fix added.

CarlosNZ:
> I think default should be no default filters (if not defined in the config file) but we already had this discussion previously. And changed here, I still prefer without this, but we want to have the fix added.

I was just avoiding the inevitable question from Jess: "Why have the default filters disappeared from Fiji?" 😜

PR #1364, opened 2022-10-01
#1363 Data view filtering minor issues and tidy-up

### LINKED ISSUE #1363: Data view filtering -- remaining issues
As noted in this PR: https://github.com/openmsupply/conforma-web-app/pull/1360

- [x] The "free text" filter is not preserving its value from the url when the page is (re)loaded (Fixed in #1360 )
- [x] A filter for "Number" fields, with a max/min selector to specify a filter range
- [x] Check out how it handles other data types, in particular arrays (it might actually work fine, I just haven't really tried yet)
- [x] In the "free text" search filters, I'd like the cursor focus to go to the input box when you open the filter pop-up, but this is proving really hard to do, due to the text input field being inside a Drop-down/modal
- [x] Documentation
- [x] ~~Stronger Typescript definitions (we were pretty loose with this in the early days)~~ (Low priority)

Couple of other little things I'll try and squeeze in here too:

- [x] Fix sorting -- change columns to two-state sort, and add a "Sort" filter display as per [this comment discussion](https://github.com/openmsupply/conforma-web-app/pull/1356#issuecomment-1267746272)
- [x] Umm.. there was something else, but it's slipped my mind. Will update if I remember.

CarlosNZ:
Fix #1363 

Can be demo-ed with snapshot: `/conforma-templates/dev/snapshots/AA_DataViewFilterMiscTidyUp.zip`

**Edit**: Use this updated snapshot in the same folder that also includes a [searchable ingredients list](https://github.com/openmsupply/conforma-server/issues/890): **AA_DataViewIngredients**

(Should use with back-end https://github.com/openmsupply/conforma-server/pull/874)

This PR seems pretty massive, but a lot of it is just tidying up and re-organising, along with a bunch of minor fixes and improvements. Explanations in comments.

Addressed the following from the issue:

> - [x] The "free text" filter is not preserving its value from the url when the page is (re)loaded

Already fixed in #1360

> - [x] A filter for "Number" fields, with a max/min selector to specify a filter range

Done here. Check out the "Amount" field on "ALL Products" view

> - [x] Check out how it handles other data types, in particular arrays

Arrays are a no-go here. It's way too complicated to parse and it looks like it might not even be possible to search partial strings within individual array items using GraphQL. However, with [this feature](https://github.com/openmsupply/conforma-server/pull/874), it's now pretty simple to write a definition that will auto-generate a delimited text string field from an array field, if we ever need it. So no point trying to make this even more complicated at this end.

> - [x] In the "free text" search filters, I'd like the cursor focus to go to the input box when you open the filter pop-up, but this is proving really hard to do, due to the text input field being inside a Drop-down/modal

Oh my god, this was a nightmare! But it's working well finally. I'll explain in inline comments in more detail, but basically it was very difficult to both auto-focus the input field *and* make it possible to elegantly close the pop-over filter list. Had to install a small package to handle the messy stuff, but it works really well now, and well worth it I think.

> - [ ] Documentation

Coming soon...

> - [x] Stronger Typescript definitions (we were pretty loose with this in the early days)

I got rid of most of the `any`s that were hanging around. Union types for the Filter options would be nice, but not hugely important now.

> - [x] Fix sorting -- change columns to two-state sort, and add a "Sort" filter display as per [this comment discussion](https://github.com/openmsupply/conforma-web-app/pull/1356#issuecomment-1267746272)

Done ✅

I think apart from Documentation, this should take care of all the front-end requirements for filtering, for now.


src/containers/List/ListFilters/common.tsx
```diff
@@ -10,39 +11,52 @@ import {
...
+        <Dropdown
+          multiple
+          text={!trigger ? title : undefined}
+          trigger={trigger && <div onClick={() => setIsOpen(!isOpen)}>{trigger}</div>}
+          icon={trigger ? null : undefined}
+          open={isOpen}
+          onOpen={() => {
+            setIsOpen(true)
+            setFocus()
+          }}
```
CarlosNZ, line 49: This (along with the `useOnClickOutside` hook) is what allows the auto-focus text input on open to work.

If you try and auto-focus the text input, it just closes the dropdown immediately, as by focusing the text input you're "onBlur"-ing the parent.

So I switched it to a controlled component with `isOpen/setIsOpen` management. But the problem with that is that we can't use `onBlur` to detect when to close the Dropdown, since (as mentioned above), the input element steals focus and forces this to close.

Solution: use the "useOnClickOutside" hook to detect when clicking outside the whole component, then `setIsOpen(false)`

Is seems so simple, but it took me ages to figure out how to get this to work properly -- I tried several different approaches. 😡

src/containers/List/ListFilters/common.tsx
```diff
@@ -90,5 +104,16 @@ const FilterOptions: React.FC<FilterOptionsProps> = ({
...
+  return (
+    <div className="filter-title">
+      {icon && <Icon name={icon} />}
+      <div className={title && criteria ? 'title-and-criteria' : ''}>
+        {criteria && <div className="filter-criteria">{criteria}</div>}
+        {title && <div>{title}</div>}
+      </div>
+    </div>
+  )
+}
```
CarlosNZ, line 118: Just copied and pasted from "Date" filter since it's now a common element.

src/utils/helpers/list/buildQueryFilters.ts
```diff
@@ -36,9 +36,8 @@ export default function buildQueryFilters(
 const filterTypeDefinitions: FilterTypeDefinitions = {
   number: (filterValue) => {
     const [fromNumber, toNumber] = filterValue.split(':')
-    const greaterThanOrEqualTo = fromNumber ? fromNumber : undefined
-    const lessThanOrEqualTo = toNumber ? toNumber : undefined
-
+    const greaterThanOrEqualTo = fromNumber !== '' ? parseInt(fromNumber) : undefined
+    const lessThanOrEqualTo = toNumber !== '' ? parseInt(toNumber) : undefined
```
CarlosNZ, line 40: We have no way to know in the front-end whether the database field is an `Int` or a `Float`, so I'm just making sure we're only sending `Ints` in the filter object.

CarlosNZ: It's actually enforced in the input components, this is just a last line of defence.

src/utils/hooks/useDataViews.tsx
```diff
@@ -55,49 +56,54 @@ export const useDataViewsList = () => {
...
 
-  const updateDataViewTable = (newDataViewTable: DataViewsTableResponse) => {
-    // We only want to load filterDefinitions once (per dataView code),
-    // otherwise it gets into an infinite loop. Similarly, we don't want to set
-    // the dataViewTable data until we've loaded filterDefinitions, or the Table
-    // page will display the full list briefly before showing the filtered list.
-
-    // TO-DO: Fetch filter definitions in a separate query *before* requesting
-    // dataViewTable
-
```
CarlosNZ, line 76: As mentioned earlier, all the changes in this file solve the "To-Do" mentioned here.

src/utils/hooks/useDataViews.tsx
```diff
@@ -55,49 +56,54 @@ export const useDataViewsList = () => {
...
+export const useDataViewFilterDefinitions = (dataViewCode: string) => {
   const { strings } = useLanguageProvider()
+  const [filterDefinitions, setFilterDefinitions] = useState<FilterDefinitions>()
   const [error, setError] = useState<ErrorResponse | null>(null)
-  const [loading, setLoading] = useState(false)
+  const [loading, setLoading] = useState(true)
+
+  useEffect(() => {
+    processRequest({
+      url: getServerUrl('dataViews', { dataViewCode, query: { first: '0' } }),
```
CarlosNZ, line 67: `{ first: '0' } ` means we're not querying the database for data records at all, we'll just get the filter (and column) definitions in this first request.

src/utils/hooks/useDataViews.tsx
CarlosNZ: Separating into two hooks makes it a lot simpler, and runs much nicer.

src/utils/hooks/useRouter.tsx
```diff
@@ -64,6 +65,10 @@ export function useRouter(): RouterResult {
           delete newQueryObject[key]
         } else newQueryObject[key] = value
       })
+
+      // Prevents un-necessary re-renders when query hasn't actually changed
+      if (isEqual(newQueryObject, queryFilters)) return
+
```
CarlosNZ, line 71: This was a big score! I suspect this has been the cause of many unnecessary re-renders throughout the app.

I could figure out why the `useDataViews` hooks were making so many repeated requests even when the `useEffect` dependencies weren't changing.

Turns out, whenever we run `updateQuery`, it creates a new instance of the `query` object. And React detects changes by *reference*, not *value*, so it was causing re-renders even when the query values didn't change (which happens a lot in the loading phase of many components).

We now do a deep-equality check on the updated query object, and if it hasn't actually changed values, we exit without updating the history or query state in the Router.

Sooo much smoother!

semantic/src/site/globals/site.overrides
```diff
@@ -1838,7 +1838,7 @@ a:hover {
 }
 
 // Layouts
-#list-filter {
+.list-filter-container {
```
CarlosNZ, line 1841: Changed this from id to class so it can be re-used within the same page

semantic/src/site/globals/site.overrides
```diff
@@ -1861,9 +1861,13 @@ a:hover {
...
 
-  // multiple is used to make sure filter menu stays open, but it adds extra margin on text
+  // multiple is used to make sure filter menu stays open, but it adds extra
+  // margin on text
   .ui.multiple.dropdown > .text {
-    margin: 0px;
+    margin: 0px ;    
+  }
+  .ui.multiple.dropdown {
+    padding: 5px 10px 5px 10px;
```
CarlosNZ, line 1870: As a result of the above change (id => class), the CSS precedence changes slightly (ids have higher priority than classes). So small adjustments required to keep it the same visually.

src/containers/DataDisplay/DataViewTable.tsx
```diff
@@ -134,14 +157,16 @@ const DataViewTableContent: React.FC<DataViewTableContentProps> = ({
...
+
+    if (ascending === 'false' || !apiQueries.orderBy) {
+      updateQuery({ sortBy: sortColumn })
+      setSortColumn({ title: columnTitle, ascending: true })
+    } else {
       updateQuery({
         sortBy: `${sortColumn}:desc`,
       })
+      setSortColumn({ title: columnTitle, ascending: false })
+    }
```
CarlosNZ, line 169: Changed to two-state sorting

src/containers/DataDisplay/Filters/SortReset.tsx
```diff
@@ -0,0 +1,45 @@
...
+              onClick={() => updateQuery({ sortBy: null })}
+            >
+              <Icon name="remove circle" />
+              {strings.FILTER_RESET_SORT}
+            </Dropdown.Item>
+          </Dropdown.Menu>
+        </Dropdown>
+      </div>
+    </div>
+  )
```
CarlosNZ, line 44: This is a very stripped-down version of the "FilterContainer" component in `common.tsx`

nmadruga (commented):
I couldn't look into this. Changed focus a bit to have the other investigation in progress. But on a quick test, and think I found something not expected behaviour:
1. Opened dataview of External users
2. Then selected filter "LastName", searched for "Man" then selected Manager
3. Reloaded the page (http://localhost:3000/data/user-public?last-name=Manager). The filter was persisted. But that's when I saw this problem:
4. Tried to select now "Username" filter. And nothing happens
It doesn't show on the right neither on the list of filters to select...

CarlosNZ:
> But on a quick test, and think I found something not expected behaviour:

@nmadruga which snapshot were you using? The one I suggested to use with this PR has no "user-public" data view.

So it's probably that, hopefully.

nmadruga (commented):
I moved this UI problem in here for fixing please: The loading after you click one filter option OR when sorting a column is showing weirdly. The list is visible on the bottom of the page and the loading takes most of the space

CarlosNZ:
> I moved this UI problem in here for fixing please: The loading after you click one filter option OR when sorting a column is showing weirdly. The list is visible on the bottom of the page and the loading takes most of the space

Okay, will sort that.

Part of the problem is that our "Loading" component has a set height -- which is good in some circumstances, but not others. Not sure what the solution is there. Maybe make the height override-able?

But the other part is that the list shouldn't remain while loading, so I'll deal with that too.

CarlosNZ:
> But the other part is that the list shouldn't remain while loading, so I'll deal with that too.

Good, that was simple: aa7be327 :)

Loading height seems fine now, so won't worry about updating that component at the moment.

src/containers/DataDisplay/DataViewTable.tsx
```diff
@@ -134,14 +157,16 @@ const DataViewTableContent: React.FC<DataViewTableContentProps> = ({
...
-    // Cycle through ascending/descending/not-sorted (default: id)
-    if (ascending === 'false') {
-      updateQuery({ sortBy: null })
-    } else if (!apiQueries.orderBy) updateQuery({ sortBy: sortColumn })
-    else
+
+    if (ascending === 'false' || !apiQueries.orderBy) {
+      updateQuery({ sortBy: sortColumn })
+      setSortColumn({ title: columnTitle, ascending: true })
+    } else {
```
nmadruga, line 164: I bit confused with the if statement `if (ascending === 'false'` and the setSortColumn using `ascending: true`

src/containers/DataDisplay/Filters/DataViewNumberFilter.tsx
```diff
@@ -0,0 +1,142 @@
+/*
+Filter uses existing filter-handling and building logic from Application List
+filters
+*/
+
+import React, { useState, useEffect, useRef, MutableRefObject } from 'react'
```
nmadruga, line 6:
```suggestion
import React, { useState, useEffect, useRef } from 'react'
```

Variable not used

nmadruga (commented):
Very small silly improvement, but why not use the same pattern of "Selection" as a pop-up circle at the top for DataViewNumberFilter? I was just curious - is it s lot of work? Just does't fit?
Screen Shot 2022-10-12 at 10 15 21 AM

CarlosNZ:
> I'm curious about this, where would be 'Another' and why this is the result when I select it?

It's just some random nonsense I put in the field for variety/testing as the current products all had the same values. The reason why it doesn't match up with the display value is that I manually entered it into the "genericIngredientsTextData", so it doesn't match its source. This would never happen in reality, since the text fields will be [auto-generated](https://github.com/openmsupply/conforma-server/pull/874).

Sorry for the confusion, you'll probably find a few things like that.

CarlosNZ:
> Very small silly improvement, but why not use the same pattern of "Selection" as a pop-up circle at the top for DataViewNumberFilter? I was just curious - is it s lot of work? Just does't fit?

I was just following Andrei's pattern for the Date filter. I know it's not particularly consistent, but as general rule, the "circle" label shows the *number* of selections, whereas the small text shows the actual values. If anything, the "Status" boolean is the one that breaks the pattern.

Happy to change to make consistent, let's bring it up at the next meeting cos I'm sure everyone has a different opinion on what it should show.

src/containers/DataDisplay/DataViewTable.tsx
CarlosNZ: "false" refers to the *current* value. So if it's false, change it to true (update query and sortColumn).

src/containers/DataDisplay/Filters/DataViewNumberFilter.tsx
CarlosNZ: 9c1516ab

nmadruga (approved):
All seems to be working fine. Besides just basically a few confusions and improvements suggestions. I'm also not clear on how the DataViewSearchableList works.
Overall comment is that I think the debouncing time to display result table while typing the filters is pretty cool. And how the preview display on each filter is great!

I still have to take a quick look on `useDataViews` to understand exactly the workflow, that's how the documentation would be helpful. But that's all good, please remember to add this when you think is a good time.

PR #1374, opened 2022-10-05
#1232 Fix Applicant summary page re-direct loop

### LINKED ISSUE #1232: Bug: Applicant going to their own Summary page after submission results in infinite re-direct
Keeps bouncing back and forth rapidly between the application home url and the /summary page

CarlosNZ:
Fix #1232 

Finally got to the bottom of this one. 

If you haven't seen the problem, I'm pretty sure you can re-create it by looking at a completed application as the applicant.

The problem was the application progress was being recalculated, even for completed applications, which was causing the validity check to fail on the summary page, and redirect back to an application page, etc.

I've kind of done a "double-fix" here. Either one on its own should fix the immediate problem, but setting the "shouldCalculateProgress" default based on the current status is more "correct", I think. 

The other fix is purely defensive -- we don't really need it if the first fix is working properly, but I thought it can't hurt to leave it in just in case (and since the result is so unpleasant if it fails)

### DIFF
--- a/src/utils/hooks/useGetApplicationStructure.tsx
+++ b/src/utils/hooks/useGetApplicationStructure.tsx
@@ -32,7 +32,7 @@ const useGetApplicationStructure = ({
   shouldRevalidate = false,
   minRefetchTimestampForRevalidation = 0,
   firstRunValidation = true,
-  shouldCalculateProgress = true,
+  shouldCalculateProgress = structure.info.current.status !== ApplicationStatus.Completed,
   shouldGetDraftResponses = true,
   forceRun = false,
 }: UseGetApplicationStructureProps) => {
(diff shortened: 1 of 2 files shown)


src/utils/hooks/useGetApplicationStructure.tsx
```diff
@@ -32,7 +32,7 @@ const useGetApplicationStructure = ({
   shouldRevalidate = false,
   minRefetchTimestampForRevalidation = 0,
   firstRunValidation = true,
-  shouldCalculateProgress = true,
+  shouldCalculateProgress = structure.info.current.status !== ApplicationStatus.Completed,
```
CarlosNZ, line 35: This was the main cause of the problem -- was defaulting to `true` even for completed applications.

(I'm not sure *why* it calculated incomplete progress for completed applications, but I didn't dig into that logic).

nmadruga (approved):
Seems LGTM

PR #1380, opened 2022-10-07
#892B Part1 improve filters in Dashboard with single query for each panel

nmadruga:
Part 1 of fixes https://github.com/openmsupply/conforma-server/issues/892

### Changes
1. This will attempt to fix the delay to load Dashboard page. As noticed by Carl, each single filter would be calling `getApplicationList` each time in order to retrieve each on those account for the applications related to actions. This PR changes that to be one call to getApplicationList per Panel (template). 
2. Also noticed we use pagination for this, so there is a limit of application to be checked, therefor I changed the order first application with latest `latestActiveDate` and the number of aplication checked is 1000 by default, but can be defined in the preferences (on the snapshot) with this new `web` property:
Screen Shot 2022-10-07 at 1 41 33 PM


src/components/Dashboard.tsx
```diff
@@ -94,38 +85,81 @@ const TemplateComponent: React.FC<{ template: TemplateInList }> = ({ template })
...
+  const { preferences } = usePrefs()
   const templateType = template.code
-  const { loading, applicationCount } = useListApplications({
+  const [loadedFiltersCount, setLoadedFiltersCount] = useState(0)
+  const [totalMatchFilter, setTotalMatchFilter] = useState<{ [key: number]: number }>(
+    filters.reduce((totalPerFilter, filter) => ({ ...totalPerFilter, [filter.id]: 0 }), {})
+  )
+
+  // Convert to string - Default is 100 is nothing is preset on preferences
+  const perPage = preferences?.dashboardFetchedApplications || '100'
```
CarlosNZ, line 100: I'm not sure why we'd want to paginate this. Isn't the dashboard query to determine the total number of each type (filter type) of application? So what would be the purpose of not fetching the full set?

nmadruga: Just that the `getApplicationsList` uses the pagination. I don't think it's simple to remove/separate a new query to not use it

nmadruga: It was already using it. I didn't change, and tried to keep with the minimal changes possible! Although I agree that it would be nicer if we could do for all applications - Just one things that I might consider - we also do filters for Approved/Rejected. So it would basically query for all existing in all history of the App

CarlosNZ: Ah okay, in which case we should just set a default page size in `getApplicationList` that is really big (something we'll never reach). It shouldn't be a preference, cos we basically want to "fetch all". 

Actually, it's probably not that hard to make the "numberToFetch" parameter optional and only specify a `first` value if defined. I dunno the easiest way, but basically the dashboard query needs to get ALL matching applications somehow.

CarlosNZ: I might be misunderstanding slightly, but one problem with this approach is that we're filtering on the front-end, which means the returned list of applications could be enormous.

If we have 10,000 applications in the database, one of these filters (e.g. "can start review") is likely to reduce that list down to a handful (the current reviewers' available work). But the way it is here, you'd be pulling down all 10,000 and then filtering on the front-end.

Is that correct, or am I following it wrong?

I was thinking that we'd still filter on the back-end, but we'd combine the filters into a single gql filter object. So we'd still be retrieving a "reasonable" size list of applications, then just *sorting* them by filter-type in the front-end. 

Does that make sense?

nmadruga: Thanks. I think that's much better. : )

And I noticed that after Andrei's change of indexes the App is faster, but the Dashboard should also benefit from this change. Also good call you made purposing this solution.

src/utils/helpers/list/buildQueryVariables.ts
```diff
@@ -30,10 +30,13 @@ const mapSortFields: any = {
...
+  paginationOffset?: number
 }
 
-export function getPaginationVariables(page: number, perPage = 20): PaginationValues {
-  return { numberToFetch: perPage, paginationOffset: (page - 1) * perPage }
+export function getPaginationVariables(
+  page: number,
+  perPage: number | undefined
+): PaginationValues {
+  return { numberToFetch: perPage, paginationOffset: perPage ? (page - 1) * perPage : undefined }
```
CarlosNZ, line 41: This will cause a conflict with the fix I've made in my filtering PRs to fix the pagination bug you found the other week. 

I guess it's probably easiest to just leave it like this, and I'll have to make a slight modification to mine before merging it.

CarlosNZ (commented):
Cool, looks like the pagination is taken care of.

But I'm still not seeing any filter object passed to the back-end in the gql queries, so you'll still be pulling down all the applications (per template), I think -- as per [this comment](https://github.com/openmsupply/conforma-web-app/pull/1380#discussion_r989673553)

Have I missed something? (I'm just inspecting the network console here, haven't look too closely at the code again)

I think you'll to take the filter objects that were being generated previously ("can start review", etc), and combine them into a single `{ or: {... }}` filter) to pass in the query. The back-end should still be doing the filtering, this change should just be to make it a single filter in a single query as opposed to multiple queries with multiple filters. The result set shouldn't be any larger than it was.

src/utils/helpers/list/buildQueryVariables.ts
nmadruga: Yep. I think it's probably nice to have the "Get ALL" solution as well

nmadruga:
> Cool, looks like the pagination is taken care of.
> 
> But I'm still not seeing any filter object passed to the back-end in the gql queries, so you'll still be pulling down all the applications (per template), I think -- as per [this comment](https://github.com/openmsupply/conforma-web-app/pull/1380#discussion_r989673553)
> 
> Have I missed something? (I'm just inspecting the network console here, haven't look too closely at the code again)
> 
> I think you'll to take the filter objects that were being generated previously ("can start review", etc), and combine them into a single `{ or: {... }}` filter) to pass in the query. The back-end should still be doing the filtering, this change should just be to make it a single filter in a single query as opposed to multiple queries with multiple filters. The result set shouldn't be any larger than it was.

Done. The code is not very straight forward, since I need to create an array of objects from each filter in Dashboard and used a new function to convert it to something with this format:
```
{ or: [
        { reviewerAction: { equalTo: START_REVIEW } }
        { reviewerAction: { equalTo: RESTART_REVIEW } }
      ]
}
```

But happy to announce this is NOT retrieving all applications anymore! 👍

nmadruga:
Closed this after having conflicts problem to rebase to branch #1363 

New PR in 1385

PR #1382, opened 2022-10-09
#1381 Fix when changing application list page

### LINKED ISSUE #1381: Page goes blank (requires refresh) when changing from one application list to another from the dropdown menu
Logged in as admin.

Error "the message port was closed before a response was received"

CarlosNZ:
Fix #1381 

Smallest PR ever?

Basically just handles the case when the `filter` in `displayableFilters` is not found, which is the case on the first render after changing list page.


fergie-nz:
Nice! Easy peasy

fergie-nz (commented):
Nice! Easy peasy

fergie-nz (approved):
Nice! Easy peasy

PR #1385, opened 2022-10-10
892B Part1 improve filters in dashboard with single query for each panel

nmadruga:
Based from #1363 

### Changes
1. This will attempt to help with the delay to load Dashboard page. As noticed by Carl, each single filter would be calling `getApplicationList` each time in order to retrieve each on those account for the applications related to actions. This PR changes that to be one call to getApplicationList per Panel (template). 
2. Also noticed we use pagination was not correctly set, and had the list of applications up to 20. Now changed to be checking the full list (by not passing any pagination parameter)
3. It was required to add a new function to parse into GraphQL query considering an OR statement for each type of filter discovered in the Filters configured for this Dashboard, which was different from what is currently done by `buildFilters` in useApplicationsList. So, the best approach - as discussed with Carl was to parse this property `queryMultiFilter` separetelly and pass on to be called straight to one GraphQL call by the `useApplicationList` hook.

### DIFF
--- a/semantic/src/site/globals/site.overrides
+++ b/semantic/src/site/globals/site.overrides
@@ -1648,12 +1648,10 @@ a:hover {
       }
 
       .filters {
-        .flex-column();
         flex-grow: 1;
       }
 
       .filter {
-        display: flex;
         align-items: center;
       }
     }
--- a/src/components/Dashboard.tsx
+++ b/src/components/Dashboard.tsx
@@ -152,8 +152,8 @@ const PanelComponent: React.FC<{
       .map(([key, value]) => `${key}=${value}`)
       .join('&')}`
 
-  if (!loading && applications.length === 0) return null
-  if (loadedFiltersCount < filters.length) return <LoadingSmall />
+  if (loading || loadedFiltersCount < filters.length) return <LoadingSmall />
+  if (applications.length === 0) return null
 
   return (
     <>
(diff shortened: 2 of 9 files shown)


nmadruga:
Not sure if I should upload the re-generated `src/utils/generated/graphql.tsx` compiled when using core_templates... It currently doesn't have the changed required for the useApplicationList to have optional parameters

CarlosNZ (commented):
Okay, I think this is generally working well, however we've made a bit of a mess of the base branches. There's files in there that are very different to the base `1363-data-view-filters----extra-issues`, yet they're not showing up in the diff (e.g. "buildQueryFilters").

Can I suggest:

- start from a fresh `1363-data-view-filters----extra-issues` base and just paste in the changes in a new branch
- I don't think we need the pagination changes any more. I think the filter construction you've added is enough to do what we want, and making the pagination variables optional is just adding unnecessary complexity that we don't actually need. Was a bit of a red herring I think, apologies if I sent you in the wrong direction.
- I don't think there's any rush to get this in to the stable branch. It's definitely useful, but I don't think it's mission-critical, as the indexing is helping a lot with loading times now. So we can probably park this until early next week if you like and then we can merge a simplified version once the filtering stuff is all in place. I'm just thinking of time -- if you're concerned about getting a stable release for this week, then you might not want to spend more time trying to figure this out right now, and we can live without it till next week. But currently it's breaking too many other things. Hopefully reverting that pagination change helps, but I can't know for sure without a closer inspection.

nmadruga:
New PR in #1387

PR #1386, opened 2022-10-11
#1384 Access documents in question history panel

### LINKED ISSUE #1384: Being able to view historical applicant documents in the UI

CarlosNZ:
Fix openmsupply/conforma-web-app#1384 

~~(Requires back-end branch from: https://github.com/openmsupply/conforma-server/pull/933)~~ (In develop now)

For quick access, use snapshot: `/conforma-templates/dev/snapshots/AA_History_Updates_2.zip`

and go to URL: http://localhost:3000/application/S-TSC-0003/review/2?active-sections=S1%2CS2%2CS3&show-history=Q_upload2

You should see several questions in the application that have a "View history" button available where you can see the element previews.

I'm not super happy with the look of this. It's basically a bare-bones implementation with the "text" value replaced with a whole SummaryView component.

If we want to refine it further, we should probably discuss it in next meeting to know exactly what it should show.


src/components/PageElements/Elements/HistoryResponseElement.tsx
```diff
@@ -1,23 +1,19 @@
...
+import { ElementState, HistoryElement } from '../../../utils/types'
 
-interface HistoryResponseElementProps {
-  author?: string
-  title: string
-  message: string
-  timeUpdated: Date
-  // attentionRequired?: boolean
-  reviewerComment?: string
-}
```
CarlosNZ, line 12: This interface not required, it's the same as the "HistoryElement" type defined in "types.ts"

src/components/PageElements/Elements/HistoryResponseElement.tsx
```diff
@@ -1,23 +1,19 @@
...
+const HistoryResponseElement: React.FC<HistoryElement> = ({
   author = '',
   title,
   message,
   timeUpdated,
   // attentionRequired = false,
   reviewerComment,
+  response,
+  elementTypePluginCode,
+  parameters,
```
CarlosNZ, line 16: These three items are required to be able to display a SummaryView component.

src/components/PageElements/Elements/HistoryResponseElement.tsx
```diff
@@ -31,7 +27,22 @@ const HistoryResponseElement: React.FC<HistoryResponseElementProps> = ({
...
+            element={
+              {
+                parameters,
+                pluginCode: elementTypePluginCode,
+                isRequired: true,
+                isVisible: true,
+              } as ElementState
+            }
+            response={response}
+            allResponses={{}}
```
nmadruga, line 41: Wondering if this could be changed in `SummaryViewWrapper` so it's optional - if not passed we won't use to validate the element

src/utils/hooks/useGetQuestionReviewHistory.tsx
```diff
@@ -62,6 +62,9 @@ const useGetQuestionReviewHistory = ({ isApplicant, ...variables }: UseGetQuesti
           title: strings.TITLE_HISTORY_SUBMITTED_BY_APPLICANT,
           // TODO translated message, that nothing is entered
           message: value?.text || '',
```
nmadruga, line 64: It seems this is causing a small bug here. I noticed that there is a different between what is displayed in the SummaryView and what is displayed inside the ReviewHistory window:
Screen Shot 2022-10-13 at 11 26 30 AM
Screen Shot 2022-10-13 at 11 26 24 AM

nmadruga (changes requested):
Wow this is very succinct which is great.

I wasn't able to move to the next Stage - due to a misconfig on this version of `core_templates` - I wanted to check with different stages... but just on the existing application the **View Review** looks good. I just noticed something small in UX, when you have a checkboxes element it seems as you could change it (even though you can't) - notice the blue border around the checkbox that I clicked. If that can be disabled I think it's best:
  
Screen Shot 2022-10-13 at 11 31 43 AM

src/utils/hooks/useGetQuestionReviewHistory.tsx
CarlosNZ: > It seems this is causing a small bug here. I noticed that there is a different between what is displayed in the SummaryView and what is displayed inside the ReviewHistory window:

Thanks for thorough testing. I suspected that not passing in a full "allResponses" object might cause problems. And I'm also slightly concerned it won't get the full data object it needs for certain evaluations.

So I think we need to be a bit careful with this.

I'll have another look at it soon, but I'm kinda worried that it might be difficult to get exactly right. One thing we might have to do -- and it's something I've thought would be useful for a while -- is add an additional "evaluated_parameters" field to application_response. Because as the application changes, dynamic parameters will have different values at different times, and therefore it's entirely possible to have incorrect values showing up in a Summary View component, as they get evaluated a new every time.

So what we would do is save the results of the *current* parameter evaluation along with each response, then every time that response is rendered going forward, it uses these pre-evaluated parameters rather than re-evaluating them.

But that's a bit more involved, so we should probably have a discussion about it before going there.

nmadruga: Ok - I'm not fully understanding this. I suggest we have a meeting to discuss this later on today? Then I think we can create another issue to make an improvement on allResponses.

For now what do. you think about the possibility of having this object as optional and in this case nothing is passed and the response would only show things related to itself - not sure if this is possible though without having a look.

CarlosNZ: > I suggest we have a meeting to discuss this later on today?

Sure, I'm good anytime before 4.

CarlosNZ: Made a new issue to address this: #1394 

Will do that one first, then come back to this.

(I'll set this PR to Draft in the meantime)

nmadruga: Sounds good - I noticed this PR is now based on the other. So I'll have a look on that one today - if time allows

CarlosNZ:
> I just noticed something small in UX, when you have a checkboxes element it seems as you could change it (even though you can't) - notice the blue border around the checkbox that I clicked. If that can be disabled I think it's best:

Cool, just a matter of wrapping it in a `<Form>` component so that the styles work the same as the main form: 4247c270

Screen Shot 2022-10-20 at 10 25 04 AM

(The "disabled" effect is pretty subtle, just a minor opacity reduction, but that's the same as the main form)

src/components/PageElements/Elements/HistoryResponseElement.tsx
CarlosNZ: It's already not optional -- which is why I'm passing an empty object.

It's better this way, as SummaryViewWrapper can still do evaluations -- it's just that the "pre-evaluated parameters" now take priority.

CarlosNZ:
Okay @nmadruga, have now update this based on #1395 and it's working correctly -- showing the parameter values as they were at the time the questions were answered:
Screen Shot 2022-10-20 at 10 53 03 AM

To test, use the new snapshot `AA_History_Updates_2.zip` and check out the URL:
http://localhost:3000/application/S-ELD-0004/review/3?active-sections=S1%2CS2

There's several elements on that page that have histories available, including the file uploads, and the checkboxes you mentioned earlier.

nmadruga:
> For quick access, use snapshot: `/conforma-templates/dev/snapshots/AA_History_Updates.zip`

Hmmm trying to load this one didn't work. I got this message:
Screen Shot 2022-10-27 at 4 14 55 PM

nmadruga:
Yea, the `AA_History_Updates` is not there. The only one I found was this, which has that problem to upload. : (

Screen Shot 2022-10-27 at 4 18 30 PM

CarlosNZ:
> Yea, the `AA_History_Updates` is not there. The only one I found was this, which has that problem to upload. : (

Check your Telegram!

CarlosNZ:
Okay @nmadruga it looks like the problem you're having with this snapshot is caused by the same thing as #1405 !

Looks like one of the `UNIQUE` constraints got lost when refactoring the database functions etc. Hopefully will be a quick fix.

CarlosNZ:
Righto -- fixed that now @nmadruga: https://github.com/openmsupply/conforma-server/pull/933

So if you re-load that `AA_History_Updates_2` snapshot with that branch active on the back-end, all the schema fixes should run and get everything up to date and working again.

nmadruga (commented):
Oh, I think there is something that didn't work so well. I have back-end on https://github.com/openmsupply/conforma-server/pull/933 to add the fix on top of the already merged PR https://github.com/openmsupply/conforma-server/pull/919

But now the button doesn't display. And it seems that the new ReviewResponses which should have been adding a button to Re-review doesn't seem to be working. I think that's an issue with back-end, so would be better to have a look prior to pre-release today.

CarlosNZ:
> Oh, I think there is something that didn't work so well. I have back-end on [openmsupply/conforma-server#933](https://github.com/openmsupply/conforma-server/pull/933) to add the fix on top of the already merged PR [openmsupply/conforma-server#919](https://github.com/openmsupply/conforma-server/pull/919)
> 
> But now the button doesn't display. And it seems that the new ReviewResponses which should have been adding a button to Re-review doesn't seem to be working. I think that's an issue with back-end, so would be better to have a look prior to pre-release today. Check out the current state of the Review history (not showing) on URL: http://localhost:3000/application/S-TSC-0003/review/2?active-sections=S1%2CS2

Okay -- this is fixed now with the [back-end fix for missing generated column data](https://github.com/openmsupply/conforma-server/pull/934). Can confirm that when you load this snapshot with that branch on the back-end, the page you showed above is appearing correctly. 👍

nmadruga (approved):

PR #1387, opened 2022-10-11
892B improve filters in dashboard with single query for each panel

nmadruga:
Take 3 🎬 
Based from #1363 

### Changes
1. This will attempt to help with the delay to load Dashboard page. As noticed by Carl, each single filter would be calling `getApplicationList` each time in order to retrieve each on those account for the applications related to actions. This PR changes that to be one call to getApplicationList per Panel (template). 
2. Also noticed we use pagination was not correctly set, and had the list of applications up to 20. Now changed to be checking the full list (by not passing any pagination parameter)
3. It was required to add a new function to parse into GraphQL query considering an OR statement for each type of filter discovered in the Filters configured for this Dashboard, which was different from what is currently done by `buildFilters` in useApplicationsList. So, the best approach - as discussed with Carl was to parse this property `queryMultiFilter` separetelly and pass on to be called straight to one GraphQL call by the `useApplicationList` hook.

### Explanation about 3rd PR on a roll

I couldn't see the problem - When testing locally with latest build (0.4.6-5) + snapshot it seems fine and I can use filters... So I'm not sure what is the problem. But in case I missed something - And when I compared there are some conflict resolutions on buildQueryFilters so maybe that's where you saw a problem. That's the reason for creating this 3rd version of the orginal.

Done

> I don't think we need the pagination changes any more. I think the filter construction you've added is enough to do what we want, and making the pagination variables optional is just adding unnecessary complexity that we don't actually need. Was a bit of a red herring I think, apologies if I sent you in the wrong direction.

It is required otherwise it would be using the default 20 - Checkout the change from:
https://github.com/openmsupply/conforma-web-app/blob/49ef9e54243ae1594b86b8864d73d352b841aec4/src/utils/helpers/list/buildQueryVariables.ts#L37-L39

> I don't think there's any rush to get this in to the stable branch. It's definitely useful, but I don't think it's mission-critical, as the indexing is helping a lot with loading times now. So we can probably park this until early next week if you like and then we can merge a simplified version once the filtering stuff is all in place. I'm just thinking of time -- if you're concerned about getting a stable release for this week, then you might not want to spend more time trying to figure this out right now, and we can live without it till next week. But currently it's breaking too many other things. Hopefully reverting that pagination change helps, but I can't know for sure without a closer inspection.

I would prefer to get this in. There is no point of keeping things out from develop if this is going to be tested on our test instances. It has been reviewed, so the problem that I can see here is only related to trying to get other branches in queue. I'm terrible sorry about a bad conflict resolution - this is all my fault. But I keep saying we need to use more of `develop` to test and keep moving forward. Otherwise it will give problems like this. I agree we should have kept this in a separate feature branch, but I have been pushing to get this on develop because it will help when deploying the newly reviewed branches. Now that we have added this to the angola instance, it would definitely be time to merge in - so we can release things more safely after all combined.

### DIFF
--- a/src/components/Dashboard.tsx
+++ b/src/components/Dashboard.tsx
@@ -10,6 +10,7 @@ import useListTemplates from '../utils/hooks/useListTemplates'
 import usePageTitle from '../utils/hooks/usePageTitle'
 import { TemplateDetails, TemplateInList } from '../utils/types'
 import LoadingSmall from './LoadingSmall'
+import { constructOrObjectFilters } from '../utils/helpers/utilityFunctions'
 
 const Dashboard: React.FC = () => {
   const { strings } = useLanguageProvider()
@@ -58,8 +59,6 @@ const TemplateComponent: React.FC<{ template: TemplateInList }> = ({ template })
       ? USER_ROLES.APPLICANT
       : USER_ROLES.REVIEWER
 
-  const [loadedFiltersCount, setLoadedFiltersCount] = useState(0)
-
   return (
     <div className="template">
       <div className="content">
@@ -69,16 +68,8 @@ const TemplateComponent: React.FC<{ template: TemplateInList }> = ({ template })
               {template?.namePlural || `${name} ${strings.LABEL_APPLICATIONS}`}
             </a>
             <Icon name="chevron right" />
-            {loadedFiltersCount !== filters.length && <LoadingSmall />}
           </Label>
-          {filters.map((filter) => (
-            <FilterComponent
-              key={filter.id}
-              template={template}
-              filter={filter}
-              setFiltersCount={setLoadedFiltersCount}
-            />
-          ))}
+          <PanelComponent key={`${template?.id}_filters`} template={template} filters={filters} />
         </div>
         {totalApplications === 0 && hasApplyPermission && <StartNewTemplate template={template} />}
       </div>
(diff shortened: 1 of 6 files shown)


CarlosNZ:
> I would prefer to get this in. There is no point of keeping things out from develop if this is going to be tested on our test instances.

I just meant for last week -- as in it wasn't essential for the soft-launch, but it'd be good to have now. :) 

It looks like all the compilation errors are sorted and it's working, so it's probably mostly all good now.

I'll take another look at it shortly, but it seems like it's mostly all there. Thanks for your perseverance.

src/components/Dashboard.tsx
```diff
@@ -94,38 +85,89 @@ const TemplateComponent: React.FC<{ template: TemplateInList }> = ({ template })
...
+
+  const arrayFilters = filters.reduce((arrayFilters: { [key: string]: string }[], element) => {
+    if (typeof element.query === 'object') {
+      const queryObj = element.query as { [key: string]: string }
+      arrayFilters.push(queryObj)
+    }
+    return arrayFilters
+  }, [])
+
+  const queryMultiFilters = {
```
CarlosNZ, line 106: Similarly, this name doesn't really capture what it is. We've already been using `graphQLFilterObject` for similar objects elsewhere, so I'd suggest using something similar, (cos that's actually what it is).

src/components/Dashboard.tsx
```diff
@@ -94,38 +85,89 @@ const TemplateComponent: React.FC<{ template: TemplateInList }> = ({ template })
...
+    templateCode: { equalToInsensitive: templateType },
+    ...constructOrObjectFilters(arrayFilters),
+  }
+
+  const { loading, applications } = useListApplications(
+    {}, // Passing empty to use already construct GraphQL query on queryMultiFilters
```
CarlosNZ, line 112: Suggested comment re-wording: `// Passing empty as graphQLFilterObject is already constructed` (using suggested variable name, so don't have to write variable name _and_ an explanation of what it is)

src/components/Dashboard.tsx
```diff
@@ -94,38 +85,89 @@ const TemplateComponent: React.FC<{ template: TemplateInList }> = ({ template })
...
+          const queryObj = query as { [key: string]: string }
+          const filteredApplications = Object.entries(applications).filter(([_, application]) => {
+            // Each filter is currently delimited to a single check!
+            const key = Object.keys(queryObj)[0]
+            const value = Object.values(queryObj)[0]
 
-  const applicationListUserRole =
+            switch (key) {
+              case 'outcome':
+                return application.outcome === value
```
CarlosNZ, line 128: Is this the same as the more common:
```
 case 'outcome':
    application.outcome === value
    break
```

Personally I prefer this syntax (cos you're not really returning anything, and we generally expect `switch` blocks to have `breaks` when not `return`ing), but I guess if it works, it's probably fine (it's at least concise). Up to you.

If it's *not* the same, can you explain the difference?

src/components/Dashboard.tsx
```diff
@@ -94,38 +85,89 @@ const TemplateComponent: React.FC<{ template: TemplateInList }> = ({ template })
...
-}> = ({ template, filter, setFiltersCount: setLoadingFilters }) => {
+  filters: Filter[]
+}> = ({ template, filters }) => {
   const templateType = template.code
-  const { loading, applicationCount } = useListApplications({
-    type: templateType,
-    perPage: 1,
-    ...filter.query,
-  })
+  const [loadedFiltersCount, setLoadedFiltersCount] = useState(0)
```
CarlosNZ, line 93: Hmmm, I think we *should* be able to get rid of "loadedFiltersCount". That's something I asked Fergus to add when he added the Loading spinner to these components, as it was the easiest way to keep track of when all the filter queries are finished.

But now we're just doing one query per panel, so I don't see why we'd need to keep a count anymore -- the `loading` value from `useListApplications` should be all you need.

src/components/Dashboard.tsx
```diff
@@ -94,38 +85,89 @@ const TemplateComponent: React.FC<{ template: TemplateInList }> = ({ template })
...
+    {}, // Passing empty to use already construct GraphQL query on queryMultiFilters
+    queryMultiFilters
+  )
 
   useEffect(() => {
-    if (!loading) setLoadingFilters((currentCount: number) => currentCount + 1)
-  }, [loading])
+    if (applications) {
+      filters.forEach(({ id, query }) => {
+        if (typeof query === 'object') {
```
CarlosNZ, line 119: Similar to above -- do we need this check?

And if you really *do* need it, why not just filter the "filters" array once at the top and use the result in both places?

src/components/Dashboard.tsx
```diff
@@ -94,38 +85,89 @@ const TemplateComponent: React.FC<{ template: TemplateInList }> = ({ template })
...
-    `/applications?type=${templateType}&user-role=${applicationListUserRole}&${Object.entries(
-      filter.query
-    )
+  const constructLink = (filter: Filter) =>
+    `/applications?type=${templateType}&user-role=${userRole(filter)}&${Object.entries(filter.query)
       .map(([key, value]) => `${key}=${value}`)
       .join('&')}`
 
-  if (applicationCount === 0) return null
+  if (loading || loadedFiltersCount < filters.length) return <LoadingSmall />
```
CarlosNZ, line 155: Yeah, as above -- if you get rid of "loadedFiltersCount", then `loading` is all you need to check.

src/utils/helpers/utilityFunctions.ts
```diff
@@ -51,3 +51,15 @@ export const replaceCommas = (value: string) => value.replace(new RegExp(MAGIC_S
 
 export const removeCommasArray = (values: string[]) => values.map((value) => removeCommas(value))
 export const replaceCommasArray = (values: string[]) => values.map((value) => replaceCommas(value))
+
+// Constructs OR filter for an objects-array i.e. [ {fieldName1: value1}, {fieldName1: value2} ]
+// Returns the GraphQL filter i.e: { or: {fieldName1: { equalsTo: value1 }}, {fieldName1: { equalsTo: value2} }}}
+// This is useful to filter same key with many values using OR statement
+export const constructOrObjectFilters = (filters: { [key: string]: string }[]) => ({
+  or: Object.values(filters).map((filter) => {
+    // Each filter is currently delimited to a single check!
```
CarlosNZ, line 60: "delimited"? Did you just mean "limited"?

src/utils/helpers/utilityFunctions.ts
```diff
@@ -51,3 +51,15 @@ export const replaceCommas = (value: string) => value.replace(new RegExp(MAGIC_S
 
 export const removeCommasArray = (values: string[]) => values.map((value) => removeCommas(value))
 export const replaceCommasArray = (values: string[]) => values.map((value) => replaceCommas(value))
+
+// Constructs OR filter for an objects-array i.e. [ {fieldName1: value1}, {fieldName1: value2} ]
+// Returns the GraphQL filter i.e: { or: {fieldName1: { equalsTo: value1 }}, {fieldName1: { equalsTo: value2} }}}
+// This is useful to filter same key with many values using OR statement
+export const constructOrObjectFilters = (filters: { [key: string]: string }[]) => ({
```
CarlosNZ, line 58: There's already a `constructOrFilter` function in `buildQueryFilters` that does pretty much exactly the same thing (just with slightly different input parmeters). Any chance you can just export and use that (and change what you pass to it slightly)?

src/utils/hooks/useListApplications.tsx
```diff
@@ -24,7 +21,12 @@ const useListApplications = ({
     userState: { currentUser },
   } = useUserState()
 
-  const filters = buildFilter({ type, ...queryFilters }, FILTER_DEFINITIONS)
+  // If queryMultiFilters is passed it will be considering only that
+  // for filtering the list of applications - otherwise will have to
+  // buildFilters parsing to GraphQL what is passed on the URL
+  const filters = queryMultiFilters
+    ? queryMultiFilters
+    : buildFilter({ type, ...queryFilters }, FILTER_DEFINITIONS)
```
CarlosNZ, line 29: Simpler comment suggestion (using preferred variable name):
```
The "filters" object is either passed in already constructed (graphQLFilterObject), OR we'll need to "buildFilters" from url query key-values.
```

CarlosNZ (commented):
Great, functionality seems good now, so no more major re-factoring! 👏👏

I just have a bunch of small suggestions and terminology nit-picks for you to consider.

Thanks.

nmadruga:
> I just have a bunch of small suggestions and terminology nit-picks for you to consider.
> 
> Thanks.

As mentioned in person - this PR has been merged unintentionally so I'll make another PR with this suggestion based on develop to be merged hopefully today as well. I'm thinking we should build another version today to test the latest merged in instance 50008

nmadruga:
@CarlosNZ Please have a look on the newly created PR with basically the changes from this PR + fixes as per suggestions

PR #1395, opened 2022-10-19
#1394 SummaryView components show pre-evaluated parameters

### LINKED ISSUE #1394: Create "evaluated parameters" field for each application response
As per [this comment](https://github.com/openmsupply/conforma-web-app/pull/1386#discussion_r993961537)

Implementation would be:
- Add column "evaluated_parameters" to "application_response" table
- When saving application responses, also save the current evaluated parameters for this element.
- No changes to "ApplicationView" components (they should always show "live" re-evaluated parameter values)
- Change "SummaryViewWrapper" to use the "evaluated_parameters" field rather than the template_element "parameters" field. (We can still leave the evaluation of the parameters in the Summary View though, cos passing it pre-evaluated parameters is fine and will just return the same values back)

Once this is done, I'll come back to #1386 and incorporate this change.

CarlosNZ:
Fix #1394 

Fairly simple, hope I haven't introduced any regressions (i.e. places where the old behaviour is preferred) -- let me know if you think of any.

Requires back-end PR: https://github.com/openmsupply/conforma-server/pull/919

To demonstrate, use snapshot `conforma-templates/dev/snapshots/DEV_SummaryViewEvaluatedParametersNEW.zip` and visit URL: http://localhost:3000/application/S-CKB-0006/review/3?activeSections=S1

Look at the last question on the first page. When in develop, we see the problem: 
Screen Shot 2022-10-19 at 1 43 19 PM
The label shows the name of the reviewer rather than the applicant, as it's being re-evaluated using current data.

But when you switch to this branch, it shows the correct value for the applicant, as it's using the saved parameter values from when the question was answered:

Screen Shot 2022-10-19 at 1 42 49 PM


src/formElementPlugins/ApplicationViewWrapper.tsx
```diff
@@ -124,6 +124,7 @@ const ApplicationViewWrapper: React.FC<ApplicationViewWrapperProps> = (props) =>
             value: response,
             isValid: validationResult.isValid,
             stageNumber,
+            evaluatedParameters: { ...simpleParameters, ...evaluatedParameters },
```
CarlosNZ, line 127: Saving the combined parameters object -- simple and dynamic

src/formElementPlugins/SummaryViewWrapper.tsx
```diff
@@ -58,7 +58,11 @@ const SummaryViewWrapper: React.FC<SummaryViewWrapperProps> = ({
   if (!pluginCode || !isVisible) return null
 
   const DefaultSummaryView: React.FC = () => {
-    const combinedParams = { ...simpleParameters, ...evaluatedParameters }
+    const combinedParams = {
+      ...simpleParameters,
+      ...evaluatedParameters,
+      ...response?.evaluatedParameters,
+    }
```
CarlosNZ, line 65: We still evaluate parameters (in `evaluatedParameters`), cos why not? It's possible we may want to have dynamically evaluated parameters in a DataView Summary element, in which case it's still possible.

src/utils/helpers/evaluateElements.ts
```diff
@@ -76,7 +76,7 @@ const evaluateSingleElement: EvaluateElement = async (
         ? await evaluateExpression(expressionOrValue, evaluationParameters)
         : expressionOrValue
     } catch (e) {
-      console.log(e, expressionOrValue)
+      // console.log(e, expressionOrValue)
```
CarlosNZ, line 79: This has been annoying me for while in the console, so just suppressed it for now. Would like to do a bit more debugging in here at some point, but for now it's easier to ignore in the console.

nmadruga:
> To demonstrate, use snapshot `conforma-templates/dev/snapshots/DEV_SummaryViewEvaluatedParameters.zip` and visit URL: http://localhost:3000/application/S-CKB-0006/review/3?activeSections=S1

Apperantly something went wrong with the migration code, when I tried to load snapshot `DEV_SummaryViewEvaluatedParameters`
Screen Shot 2022-10-27 at 3 29 16 PM

And then the front-end can't get types regenerated: 
Screen Shot 2022-10-27 at 3 29 51 PM

So, nothing showing on the page you linked : (

nmadruga:
Ok - I think I'll try the other Front-end PR to see if that's sorted in there and do the reviews of code separately, but test in one go

CarlosNZ:
> Ok - I think I'll try the other Front-end PR to see if that's sorted in there and do the reviews of code separately, but test in one go

Sorry, there was a glitch in the back-end due to a faulty conflict resolution in the migration code. I've fixed that in [#919](https://github.com/openmsupply/conforma-server/pull/919) now, so should be good to go again.

nmadruga:
> http://localhost:3000/application/S-CKB-0006/review/3?activeSections=S1

I'm still not seeing the Review in this URL : (

And front-end still shows the same error on the Page.... Not sure what else might be different

CarlosNZ:
Okay, @nmadruga, have fixed the problems with this snapshot (there was actually a missing migration in "views_functions_triggers" too which I've fixed), so it should work okay now, including that testing URL I posted above.

New snapshot is: `conforma-templates/dev/snapshots/DEV_SummaryViewEvaluatedParametersNEW.zip`

(P.S. If you get an error when "importing" the snapshot zip, you have two options:
1. Switch back-end to [#912 branch](https://github.com/openmsupply/conforma-server/pull/922), import it, then switch it back.
2. Manually unzip the snapshot and put in your "_snapshots" folder

nmadruga:
Hmmm a bit confused on why after loaded this snapshot I can't add any review to elements. Even though the review is in propgress. It may be because of manual changes, but I just wanted to flag that the Review buttons are missing...

nmadruga (approved):
I'm going to continue the test for this on the next PR, so far the code seems ok to me, thanks for adding that! It would be good to have that included so we can check if the problem listed in issue https://github.com/openmsupply/conforma-server/issues/932 is fixed

CarlosNZ:
> Hmmm a bit confused on why after loaded this snapshot I can't add any review to elements. Even though the review is in progress. It may be because of manual changes, but I just wanted to flag that the Review buttons are missing...

Okay, it seems this problem is gone once [this fix](https://github.com/openmsupply/conforma-server/pull/934) for the missing generated column data is in place, so I'll be safe to merge this one now.

PR #1397, opened 2022-10-21
#1389 Filter tweaks

### LINKED ISSUE #1389: Data view filters -- minor tweaks
To be done *after* main feature branch is merged.

Just making a list here and I'll do these together at some point:

- [x] Don't show the "Sort by" filter label, make it an option in the filter menu: "Reset sort order"
- [x] Set a limit on the length of search text in filters and truncate text if it exceeds
- [x] Allow a "default sort column" to be defined (currently just used ID, which should remain the default)
- [x] Changing dataview should reset filters (properly)
- [x] Filter definitions column defaults tweak: https://github.com/openmsupply/conforma-server/issues/902 (Also update table view colums to NOT return filterData by default)
- [x] Also add a "Reset pagination" option in Filter menu
- [x] Update Postman collection
- [x] "Can be blank" option for filter lists (filter parameters)
- [x] Documentation (back-end)

CarlosNZ:
Fix #1389

Bunch of small bug fixes and tweaks as listed in #1389. Requires back-end branch from PR: https://github.com/openmsupply/conforma-server/pull/924

Items addressed by front-end here:

- [x] Don't show the "Sort by" filter label, make it an option in the filter menu: "Reset sort order"
- [x] Set a limit on the length of search text in filters and truncate text if it exceeds
- [x] Changing dataview should reset filters (properly)
- [x] Also add a "Reset pagination" option in Filter menu
- [x] Update Postman collection

See inline comments for more details.


src/containers/DataDisplay/DataViewTable.tsx
```diff
@@ -52,30 +51,13 @@ const DataViewTable: React.FC = () => {
...
 
-  // Reset filters if navigating here from Menu bar
-  useEffect(() => {
-    if (location?.state?.resetFilters) {
-      setGqlFilter({})
-      setApiQueries({})
-      setSearchText('')
-      setDebounceInput('')
-    }
-  }, [dataViewCode])
```
CarlosNZ, line 68: This resetting wasn't quite capturing all cases (it still left "activeFilters", even though they had nothing in them), so moved to "ListFilters" instead. Works better there anyway, as just resetting "activeFilters" resets all of the above as a result.

src/containers/DataDisplay/DataViewTable.tsx
```diff
@@ -52,30 +51,13 @@ const DataViewTable: React.FC = () => {
     filtersReady: !filtersLoading,
   })
 
-  const [currentSortColumn, setCurrentSortColumn] = useState<SortColumn>(null)
```
CarlosNZ, line 55: Can get rid of quite a lot as a result of not showing the currently sorted column in the filter bar 👍🏼  (Most of the removed content)

src/containers/DataDisplay/Filters/DataViewSearchableList.tsx
```diff
@@ -32,13 +32,13 @@ export const DataViewSearchableList: React.FC<SearchableListProps> = ({
   const [moreResults, setMoreResults] = useState(false)
   const [debounceOutput, setDebounceInput] = useDebounce(searchText)
 
-  const { column, code, searchFields, delimiter } = options
+  const { column, code, ...otherOptions } = options
```
CarlosNZ, line 35: These properties (plus the `includeNull` prop) are not actually used in the front-end, but passed back unchanged to server with search queries, so no need to unpack them. And it means if we add more "filter_parameters" we won't need to deal with them here (unless they're relevant to front-end)

src/containers/Dev/Snapshots.tsx
```diff
@@ -64,7 +64,11 @@ const Snapshots: React.FC = () => {
         url: getServerUrl('snapshot', { action: 'use', name }),
       })
 
-      if (resultJson.success) return setIsLoading(false)
+      if (resultJson.success) {
+        setIsLoading(false)
+        location.reload(true)
+        return
+      }
```
CarlosNZ, line 71: Not related to this PR. This just makes the whole app reload after a snapshot is loaded, which is much nicer.

src/containers/List/ListFilters/ListFilters.tsx
```diff
@@ -49,6 +49,12 @@ const ListFilters: React.FC<{
 
   // Add filters from URL to activeFilters, filters may not have criteria/options yet thus we have to use activeFilter state variable
   useEffect(() => {
+    // Reset active filters if coming from Database menu selection
+    if (location?.state?.resetFilters) {
+      setActiveFilters([])
+      return
+    }
```
CarlosNZ, line 56: As mentioned above, moved from "DataViewTable"

src/containers/List/ListFilters/ListFilters.tsx
```diff
@@ -110,10 +116,27 @@ const ListFilters: React.FC<{
...
+          )}
+          {(query.page || query.perPage) && (
+            <Dropdown.Item
+              key="clear-pagination"
+              onClick={() => updateQuery({ page: null, perPage: null })}
+            >
+              <Icon size="small" color="grey" name="delete" />
+              {strings.FILTER_RESET_PAGINATION}
+            </Dropdown.Item>
+          )}
```
CarlosNZ, line 139: This is the "Reset Filters", "Reset Sort" and "Reset pagination" controls. I've made them all only appear if there's actually something to reset. Does that seem okay, or do you reckon it's better to have them in the menu all the time regardless?

src/utils/hooks/useDataViews.tsx
```diff
@@ -182,26 +182,22 @@ const buildFilterDefinitions = (
...
-          code,
-          dataType,
-          showFilterList,
-          searchFields,
-          delimiter,
-          booleanMapping,
-          nullString,
-        },
-      }
+  filterDefinitions.forEach(({ column, title, dataType, showFilterList, ...otherOptions }) => {
```
CarlosNZ, line 185: As above, just passing through `...otherOptions` without destructuring them

src/utils/defaultLanguageStrings.ts
```diff
@@ -133,7 +133,8 @@ export default {
   FILTER_START_TYPING: 'Start typing...',
   FILTER_SEARCH_LIST: 'Search list...',
   FILTER_SORT_BY: 'Sort by',
-  FILTER_RESET_SORT: 'Reset sort',
+  FILTER_RESET_SORT: 'Reset column sort',
+  FILTER_RESET_PAGINATION: 'Reset pagination',
```
CarlosNZ, line 137: @nmadruga You might want to pre-translate these into the Angola localisations file so they're already there when this PR is merged.

src/containers/Dev/Snapshots.tsx
nmadruga: Oh! Cool, I'll have a look on how this is looking : )

nmadruga (approved):
👍 I liked the improvements a lot

PR #1398, opened 2022-10-25
#892B part2 suggestion of improvements to filters in dashboard

nmadruga:
Replacement for closed PR https://github.com/openmsupply/conforma-web-app/pull/1387

@CarlosNZ please have a look on this one and reply to suggestions will be on original PR.

I'm having a problem with styles to display the filters now And would like some help with that if possible. Thanks!

### DIFF
--- a/src/components/Dashboard.tsx
+++ b/src/components/Dashboard.tsx
@@ -118,7 +118,7 @@ const PanelComponent: React.FC<{
         if (query !== null) {
           const queryObj = query
           const filteredApplications = Object.entries(applications).filter(([_, application]) => {
-            // Each filter is currently delimited to a single check!
+            // Each filter is currently limited to a single check!
             const key = Object.keys(queryObj)[0]
             const value = Object.values(queryObj)[0]
             switch (key) {
--- a/src/components/Dashboard.tsx
+++ b/src/components/Dashboard.tsx
@@ -114,30 +114,30 @@ const PanelComponent: React.FC<{
 
   useEffect(() => {
     if (applications) {
-      filters.forEach(({ id, query }) => {
-        if (query !== null) {
-          const queryObj = query
-          const filteredApplications = Object.entries(applications).filter(([_, application]) => {
-            // Each filter is currently limited to a single check!
-            const key = Object.keys(queryObj)[0]
-            const value = Object.values(queryObj)[0]
-            switch (key) {
-              case 'outcome':
-                return application.outcome === value
-              case 'status':
-                return application.status === value
-              case 'reviewerAction':
-                return application.reviewerAction === value
-              case 'assignerAction':
-                return application.assignerAction === value
-              default:
-                return false
-            }
-          })
-          totalMatchFilter[id] = filteredApplications.length
-        }
-        setTotalMatchFilter(totalMatchFilter)
-      })
+      const resultMatchingFilters = filters.reduce((machingFilters, { id, query }) => {
+        if (query === null) return machingFilters
+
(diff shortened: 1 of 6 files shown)


src/components/Dashboard.tsx
```diff
@@ -94,38 +85,92 @@ const TemplateComponent: React.FC<{ template: TemplateInList }> = ({ template })
...
 
   useEffect(() => {
-    if (!loading) setLoadingFilters((currentCount: number) => currentCount + 1)
-  }, [loading])
+    if (applications) {
+      filters.forEach(({ id, query }) => {
+        if (query !== null) {
+          const queryObj = query
+          const filteredApplications = Object.entries(applications).filter(([_, application]) => {
+            // Each filter is currently delimited to a single check!
```
CarlosNZ, line 121:
```suggestion
            // Each filter is currently limited to a single check!
```

src/components/Dashboard.tsx
```diff
@@ -94,38 +85,92 @@ const TemplateComponent: React.FC<{ template: TemplateInList }> = ({ template })
...
+              case 'outcome':
+                return application.outcome === value
+              case 'status':
+                return application.status === value
+              case 'reviewerAction':
+                return application.reviewerAction === value
+              case 'assignerAction':
+                return application.assignerAction === value
+              default:
+                return false
```
CarlosNZ, line 134: > It's not the same. I thought it could be but actually no, the thing is I need to return a boolean and just doing this isn't returning the proper result:
> 
> ```
> case 'outcome':
>     application.outcome === value
>     break
> ```
> 
> No changes done. I'm happy to try something else if you think it's worth. I actually don't. like this code so much, but haven't spent much time trying to improve this. I was thinking this would have been quicker, although very long review has been applied.

(Continuing from other PR...)

Ah I see now, and I see what you mean by "limited to a single check" -- is this likely to be a problem? Are you assuming that these pre-defined filters only have one criteria in them? I guess they do currently, but it's probably not safe to assume that they always would.

I do think this should probably not be "limited to a single check", but this is going on for a while now and it's probably fine for now. Could you make a new issue for this, something along the lines of "Dashboard filters should handle multiple filter criteria" so we don't forget about it? (Unless I'm misunderstanding it?)

CarlosNZ (approved):
Cool, I think this is all good. I haven't actually been able to view any filters in the Dashboard, as I can't find a snapshot that has any data in it now 😳.

So:

> I'm having a problem with styles to display the filters now And would like some help with that if possible. Thanks!

I'm not actually sure what the problem is here. Looks like you've reverted to what it was -- is that fine? 

Anyway, feel free to merge if you're happy with how it's looking, otherwise hit me up tomorrow and I'll take a look with you.

nmadruga:
> > I'm having a problem with styles to display the filters now And would like some help with that if possible. Thanks!
> 
> I'm not actually sure what the problem is here. Looks like you've reverted to what it was -- is that fine?

No... It doesn't load correct after I refresh the page...

This is what it should be displaying:
Screen Shot 2022-10-25 at 8 10 38 PM

But this is what happens after a refresh:
Screen Shot 2022-10-25 at 8 10 45 PM

nmadruga:
If you try with any snapshot that you can load as an Applicant and submit one application and have another one draft you should be able to see it isn't showing. I reverted because I couldn't get those changes to work now.

src/components/Dashboard.tsx
nmadruga: Yes! I'll make an issue now

PR #1409, opened 2022-10-28
#1408 improving assignee dropdown

### LINKED ISSUE #1408: Improve assignee selection dropdown
As mentioned on the meeting with the client, it would be desirable to have the option to assign yourself at the top of the list. Plus a few other things to improve:
- [x] Allow the dropdown to show scroll for long lists
- [x] Show list in alphabetical order
- [x] Add the option to assign yourself at the top

Note - in order to change to alphabetical order and adding the "Yourself" user to the top proved would be needed to do while improving the complex mechanism in `useGetAssignmentOptions` hook. So that will be pending for when we have more time to do.

nmadruga:
Fixes #1408

~~Almost there - just missing the `onChange` function to be working. After that's ready I'll move this to "Not draft".~~
Thanks Carl! Seems like this one is also finished. Review can start now : )

Screen Shot 2022-10-28 at 3 41 59 PM

### DIFF
--- a/src/containers/Review/assignment/AssignAll.tsx
+++ b/src/containers/Review/assignment/AssignAll.tsx
@@ -33,10 +33,9 @@ const AssignAll: React.FC<AssignAllProps> = ({ assignments, setReviewerForAll, c
         options={options}
         placeholder={strings.ASSIGN_ALL_PLACEHOLDER}
         value={selected}
-        clearable
+        scrolling
         onChange={(_, { value }) => {
-          const reviewerId = value !== '' ? options[value as number].reviewerId : 0
-          setReviewerForAll(reviewerId)
+          setReviewerForAll(value as number)
           setSelected(value as number | string)
         }}
       />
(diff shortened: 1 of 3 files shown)


CarlosNZ:
What about [this](https://github.com/openmsupply/conforma-web-app/issues/1408#issuecomment-1294271541)?

CarlosNZ:
> What about [this](https://github.com/openmsupply/conforma-web-app/issues/1408#issuecomment-1294271541)?

Actually, it's probably slightly overkill to use that one, but a searchable list would still be good -- I think just enabling the `search` prop in `<Dropdown>` should give you a searchable (pre-loaded) list.

nmadruga:
> Actually, it's probably slightly overkill to use that one, but a searchable list would still be good -- I think just enabling the `search` prop in `<Dropdown>` should give you a searchable (pre-loaded) list.

Ok - Cool I'll have a look on this soon

CarlosNZ:
Okay, have had a quick look at the problems you're having as discussed, and I think the problem is that the logic in `useGetAssignmentOptions` is becoming overly complicated (and therefore hard to keep track of what's actually happening).

I think we can do it in a simpler way -- the original logic is actually already fine, so all we need to do is add a little bit extra to put the "assigneeOptions.options" in a different order with "Yourself" at the top.

Something like this does the trick (*after* assigneeOptions is generated):

```
   // Get the option for the current user
    const currentUserOption = assigneeOptions.options.find((option) => option.text === 'Yourself')

   // Make a new list with current user at the top, and the rest (with current user filtered out) after
    const orderedOptions = [
      currentUserOption,
      ...assigneeOptions.options.filter((option) => option.text !== 'Yourself'),
    ]

    assigneeOptions.options = orderedOptions as any

    console.log('assigneeOptions', assigneeOptions)
```

This is super-crude -- just a proof of concept, but it's working for me now. Obviously you'll want to "find" by matching `id` (not `option.text === 'Yourself' like I did), and there's a bit of typescript tidying up that needs to happen.

But my main main point is: I wouldn't touch the existing logic -- it's too fragile; just add a little extra to handle the options ordering (similar for AssignAll) and the rest of the issue is a just a small cosmetic one (add `search` and `scrolling` to the Dropdown).

nmadruga:
Thanks that sounds good and makes sense, I don't want to change the logic indeed. It might be something we could improve when doing the bug fix for https://github.com/openmsupply/conforma-templates/issues/54 later.

Just flagging that I saw your comment and will work on it after I have manage to add required config changes to fix Fiji project

nmadruga:
Closing for now - to be implemented later after issue https://github.com/openmsupply/conforma-templates/issues/54

Both are going to touch same area, so better to do this - minor tweak - on top of changes for Front-end PR - coming soon.

PR #1415, opened 2022-11-01
Fix to get assigned and Reviewed elements in current level/stage

nmadruga:
Required to get proper check for fully approved, otherwise it's basically not checking since `assignedQuestions` may be the same as `reviewedQuestions`, but what we also need to check is if they are fully APPROVED, which so far works by using:
`filter: { or: [{ decision: { equalTo: APPROVE } }, { decision: { equalTo: AGREE } }] }`

### DIFF
--- a/src/containers/Review/ReviewSubmit.tsx
+++ b/src/containers/Review/ReviewSubmit.tsx
@@ -6,7 +6,7 @@ import ReviewDecision from '../../components/Review/ReviewDecision'
 import {
   Decision,
   ReviewStatus,
-  useGetReviewableQuestionsQuery,
+  useGetReviewableQuestionCountsQuery,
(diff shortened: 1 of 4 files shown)


CarlosNZ (approved):
Okay, looks good.

Just one minor suggestion -- could we maybe rename the query to `getReviewableQuestionCounts` to make it clear that it's just getting a count and not the actual "reviewable questions"? 

Otherwise we're (well "I'm") inclined to wonder where else this is being used and wonder if adding this filter might be incorrect for other situations. But if it's clear that we're just interested in the "counts", then it makes more sense.

Up to you, not a big deal.

nmadruga:
Cool. Have renamed and updated the comparison for consider more approved/assigned then reviewable elements

Issue #76, opened 2020-10-29
Epic #1 - Schema documentation


nmadruga:
## Description
Continue break-down of schema documents (export groups of tables using DBever if that's possible).
Break-down .sql insertion to one table per file. And script to import all tables in order. Separate files for importing each table and create one single GraphQL file with pre-configured data to be present in the system at the start. 
Add schema description of groups in Drive and schema groups exported images in the wiki (Backend).

### Sprint 1-2:
- [x] [Back-end#14] (https://github.com/openmsupply/application-manager-server/issues/14) Schema Documentation - User/Organisation - Effort 1
- [x] [Back-end#15] (https://github.com/openmsupply/application-manager-server/issues/15) Schema Documentation - Application - Effort 1
- [x] [Back-end#16](https://github.com/openmsupply/application-manager-server/issues/16) Schema Documentation - Template - Effort 2
- [x] [Back-end#17](https://github.com/openmsupply/application-manager-server/issues/17) Schema Documentation - Review Process - Effort 1 ~**closed with unmerged PR**
- [x] [Back-end#18](https://github.com/openmsupply/application-manager-server/issues/18) Schema Documentation - Permissions - Effort 1
- [x] [Back-end#22](https://github.com/openmsupply/application-manager-server/issues/22) Bug? Inserted application data has no template_id
- [x] [Back-end#35](https://github.com/openmsupply/application-manager-server/issues/35) Schema -- update Template/Version table structure as per discussion - Effort 1

PR #1419, opened 2022-11-03
#76 Warning still showing for user that has approved all sections

nmadruga:
- Refetch query to get assignedQuestions when question is reviewed


CarlosNZ (approved):
LGTM! :)

PR #1421, opened 2022-11-04
#1408 improving assignment

### LINKED ISSUE #1408: Improve assignee selection dropdown

### LINKED ISSUE #1423: Final Decision - Problem to make review
As pointed out by @fergie-nz on Friday, the Final Decision on 50004 is broken.

When logged in as "basic_reviewer", we get this error in the UI:

Screenshot 2022-11-06 at 4 22 37 PM

When running this locally, the console is reporting an error about calling a hook from somewhere other than the body of a functional component -- and indeed this is the case at line 149 of `ReviewSectionRowAction.tsx`:
```
        const remakeReview = useRemakePreviousReview({
          reviewStructure,
          reviewAssignment,
          previousAssignment,
        })
```

Which would suggest this bit of code has *never* worked, so must not actually be encountered very often. When I change it to circumvent this "if" block and just run `createReview()`, it sometimes works and sometimes complains about row-level permissions -- not sure what's going on there.

Still, it all comes back to the mess that is caused by allowing multiple assignments -- if we get rid of this, then we'll be able to remove a lot of this convoluted logic in the front-end. 

Not sure what the best approach is, we'll need to have a bit of a chat first. I would suggest that we immediately start enforcing proper back-end integrity around review assignments (i.e. not allow multiple assignments) via [this PR](https://github.com/openmsupply/conforma-server/pull/944), then go from there -- fix the problem that then arise whenever something (either an action or the front-end) tries to break this enforcement. This will probably mean a slight regression initially, but IMO will be worth it once we simplify this whole mess. (We'll just have to make sure we don't put any problematic build up online)

Just posting this here now so we have something to start the week with -- but we need to sort this out ASAP or we're going to be plagued by these annoying bugs over and over again.

nmadruga:
Fixes #1408 and fixes #1423 - new bug introduced in v0.4.6-13 

### Briefing about extra commits

This branch was based on `build-2022-11-03A`, but doesn't seem to be very much different from `develop`.

### Changes

- Added `scrollable` in Dropdowns - UX very desired feature
- Improving to remove option "NOT ASSIGNED" from dropdown 
- Keep a constant shared for NOT_ASSIGNED -1 for when nothing is selected
- Fix how is showing the selected Assigned, something was wrong with the hook `getAssignmentOptions` that I didn't want to dive in so much, but adding the `isSubmitted` helped to show correctly the selected assignment before clicking on the Apply
- Fix problem where Re-assignment wasn't showing all available options
- Fix nor reloading after unassigned reviewer

I tried not to change much the code, but be simplifying the structure with the removal of 1 Component and combining more variables to new hook `useLoadAssignment`.


src/components/Review/ReviewSectionRowAction.tsx
```diff
@@ -151,9 +145,14 @@ const GenerateActionButton: React.FC<ReviewSectionComponentProps> = ({
       )
 
     try {
-      if (isFinalDecision)
+      if (isFinalDecision && previousAssignment) {
+        const remakeReview = useRemakePreviousReview({
+          reviewStructure,
+          reviewAssignment,
+          previousAssignment,
+        })
```
nmadruga, line 153: Moved this to here since it's only used for stage of FinalDecision...

src/containers/Review/assignment/AssignAll.tsx
```diff
@@ -30,16 +33,25 @@ const AssignAll: React.FC<AssignAllProps> = ({ assignments, setReviewerForAll, c
...
-        clearable
-        onChange={(_, { value }) => {
-          const reviewerId = value !== '' ? options[value as number].reviewerId : 0
-          setReviewerForAll(reviewerId)
-          setSelected(value as number | string)
-        }}
-      />
+        text={
+          options.find(({ value }) => value === selected)?.text || strings.ASSIGNMENT_NOT_ASSIGNED
+        }
```
nmadruga, line 38: The previous `value={selected}` wasn't working after changing to use the header ASSIGNMENT_NOT_ASSIGNED

src/containers/Review/assignment/AssignAll.tsx
```diff
@@ -30,16 +33,25 @@ const AssignAll: React.FC<AssignAllProps> = ({ assignments, setReviewerForAll, c
...
+        <Dropdown.Menu>
+          <Dropdown.Header icon="tags" content={strings.ASSIGNMENT_NOT_ASSIGNED} />
+          <Dropdown.Divider />
+          {options.map((assignee) => (
+            <Dropdown.Item
+              {...assignee}
+              onClick={(_, { value }) => {
+                setReviewerForAll(value as number)
+                setSelected(value as number | string)
+              }}
```
nmadruga, line 50: This is the same as `onChange` was previously doing, but didn't work with this new shape of Dropdown

src/containers/Review/assignment/AssigneeDropdown.tsx
```diff
@@ -1,30 +1,38 @@
...
-  assignmentOptions: {
-    isCompleted: boolean
-    selected: number
-    options: {
-      key: number
-      value: number
-      text: string
-    }[]
-  }
+  assignmentOptions: AssignmentOptions
```
nmadruga, line 6: Just using existing type

src/containers/Review/assignment/AssigneeDropdown.tsx
```diff
@@ -1,30 +1,38 @@
...
+        <Dropdown.Header icon="tags" content={strings.ASSIGNMENT_NOT_ASSIGNED} />
+        <Dropdown.Divider />
+        {options.map((assignee) => (
+          <Dropdown.Item
+            {...assignee}
+            onClick={(_: any, { value }: any) => onChangeMethod(value as number)}
+          />
+        ))}
+      </Dropdown.Menu>
+    </Dropdown>
```
nmadruga, line 35: Very similar to what you see above for AssignAll, we should be able to combine this two components, but not right now

src/containers/Review/assignment/AssignmentSectionRow.tsx
```diff
@@ -128,7 +126,7 @@ const AssignmentSectionRow: React.FC<AssignmentSectionRowProps> = ({
           </Label>
         </Grid.Column>
         <Grid.Column className="centered-flex-box-row" width={9}>
-          {originalAssignee ? (
+          {originalAssignee && assignmentOptions.isSubmitted ? (
```
nmadruga, line 129: This was a **new** problem, happens after the selection of the assigned reviewer, it was interpreting the same way as an existing Assignment on the database, so I added this flag to determine when this is submitted assignment or un-submitted

src/containers/Review/assignment/AssignmentSectionRow.tsx
```diff
@@ -154,12 +152,11 @@ const AssignmentSectionRow: React.FC<AssignmentSectionRowProps> = ({
...
         <Grid.Row>
           <Reassignment
-            assignments={assignments}
             sectionCode={sectionCode}
-            isLastLevel={isLastLevel}
-            previousAssignee={assignmentOptions.selected}
+            isLastLevel={isLastLevel()}
             assignedSections={assignedSections}
             setAssignedSections={setAssignedSections}
+            assignmentOptions={assignmentOptions}
```
nmadruga, line 159: There was some extra variables not needed for Re-assignment

I was investigating why it used to show only the same assignment on the dropdown, I found out the problem and ended up removing some extra props that could be already used with `assignmentOptions`

src/containers/Review/assignment/AssignmentTab.tsx
```diff
@@ -1,82 +1,55 @@
...
+    loading,
+    assignmentsFiltered,
+    assignmentInPreviousStage,
+    assignmentGroupedLevel,
+    isFullyAssigned,
+  } = useLoadAssignments({
+    info: fullStructure.info,
+    sectionCodes,
+    filters,
   })
```
nmadruga, line 49: New hook, where most of the code removed from here is now

src/containers/Review/assignment/AssignmentTab.tsx
```diff
@@ -103,21 +74,59 @@ const AssignmentTab: React.FC<{
...
+                  reviewAssignment={assignment}
+                  previousAssignment={assignmentInPreviousStage}
+                />
+              ) : null
+            )}
+          </div>
+        ))}
+      </Segment>
+    )
+  }
```
nmadruga, line 129: This used to be another component `AssignmentRows` and it was confusing so I thought it would be better to move in here

CarlosNZ:
> * Fix how is showing the selected Assigned, something was wrong with the hook `getAssignmentOptions` that I didn't want to dive in so much, but adding the `isSubmitted` helped to show correctly the selected assignment before clicking on the Apply
> 
> * Fix problem where Re-assignment wasn't showing all available options
> 
> * Fix nor reloading after unassigned reviewer

Looks good -- would you be able to provide a snapshot where these bugs are apparent, so we can compare the difference this branch makes? Thanks.

nmadruga:
> > * Fix how is showing the selected Assigned, something was wrong with the hook `getAssignmentOptions` that I didn't want to dive in so much, but adding the `isSubmitted` helped to show correctly the selected assignment before clicking on the Apply
> > * Fix problem where Re-assignment wasn't showing all available options
> > * Fix nor reloading after unassigned reviewer
> 
> Looks good -- would you be able to provide a snapshot where these bugs are apparent, so we can compare the difference this branch makes? Thanks.

The later 2 problems can be checked in any snapshot running on branch prior to v0.4.6-13.

nmadruga:
The one about adding `isSubmitted` I think might not be required... But was required as explained [here](https://github.com/openmsupply/conforma-web-app/pull/1421#discussion_r1013535902)

"This was a new problem, happens after the selection of the assigned reviewer, it was interpreting the same way as an existing Assignment on the database, so I added this flag to determine when this is submitted assignment or un-submitted"

src/utils/hooks/useCreateFinalDecisionReview.ts
```diff
@@ -54,9 +54,12 @@ const useRemakePreviousReview: UseRemakePreviousReview = ({
...
     const reviewResponseCreate = reviewableElements.map(
       ({ isPendingReview, thisReviewLatestResponse, response, lowerLevelReviewLatestResponse }) => {
-        const applicationResponseId = previousAssignment.level > 1 ? undefined : response?.id
+        const applicationResponseId =
+          !previousAssignment || previousAssignment.level > 1 ? undefined : response?.id
         const reviewResponseLinkId =
-          previousAssignment.level === 1 ? undefined : lowerLevelReviewLatestResponse?.id
+          !previousAssignment || previousAssignment.level === 1
+            ? undefined
+            : lowerLevelReviewLatestResponse?.id
```
nmadruga, line 62: Part of fix #1423

src/utils/hooks/useCreateFinalDecisionReview.ts
```diff
@@ -12,16 +12,16 @@ import { useGetFullReviewStructureAsync } from './useGetReviewStructureForSectio
 type UseRemakePreviousReviewMutationReturnType = ReturnType<typeof useCreateReviewMutation>
 type PromiseReturnType = ReturnType<UseRemakePreviousReviewMutationReturnType[0]>
 // hook used to restart a review, , as per type definition below (returns promise that resolve with mutation result data)
-type UseRemakePreviousReview = (props: {
+type UseCreateFinalDecisionReview = (props: {
```
nmadruga, line 15: Renaming the hook to be clearer what's doing

src/utils/hooks/useCreateFinalDecisionReview.ts
```diff
@@ -12,16 +12,16 @@ import { useGetFullReviewStructureAsync } from './useGetReviewStructureForSectio
...
   reviewStructure: FullStructure
   reviewAssignment: AssignmentDetails
-  previousAssignment: AssignmentDetails
+  previousAssignment?: AssignmentDetails
```
nmadruga, line 18: Part of fix #1423

src/containers/Review/assignment/useGetAssignmentOptions.ts
```diff
@@ -57,15 +63,22 @@ const useGetAssignmentOptions = () => {
...
+    const currentUserOption = assigneeOptions.options.find(
+      (option) => option.value === currentUser?.userId
+    )
+
+    // Remove currentUser from the list and add at the top
+    const sortedOptions = assigneeOptions.options.filter(
+      (option) => option.key !== currentUser?.userId
+    )
+    if (currentUserOption) sortedOptions.unshift(currentUserOption)
+    assigneeOptions.options = sortedOptions
```
nmadruga, line 81: Suggestion you gave @CarlosNZ 

The missing part (to remove the duplicate user) was this:
```
const sortedOptions = assigneeOptions.options.filter(
      (option) => option.key !== currentUser?.userId
    )
```

nmadruga:
Ready for another review - before we can build version to install in 50004 today

src/containers/Review/assignment/AssignAll.tsx
```diff
@@ -2,20 +2,23 @@ import React, { useState } from 'react'
...
+const AssignAll: React.FC<AssignAllProps> = ({ assignments, setReviewerForAll }) => {
+  const {
+    userState: { currentUser },
+  } = useUserState()
   const { strings } = useLanguageProvider()
   const [selected, setSelected] = useState<number | string>('')
   const options = getReviewerList(
     assignments,
-    currentUser?.userId ?? 0,
+    currentUser?.userId || 0,
```
CarlosNZ, line 21: Just interested -- why `||` rather than `??` ? (They would both behave exactly the same in this situation)

src/containers/Review/assignment/AssigneeDropdown.tsx
```diff
@@ -1,28 +1,27 @@
...
   const { isCompleted, options, selected } = assignmentOptions
 
   return (
     <Dropdown
       className="reviewer-dropdown"
+      placeholder={strings.ASSIGNMENT_NOT_ASSIGNED}
       options={options}
       value={selected}
       disabled={isCompleted}
+      labeled
```
CarlosNZ, line 22: What exactly is `labeled` doing here? I don't see any difference.

src/containers/Review/assignment/AssignmentSectionRow.tsx
```diff
@@ -138,7 +136,7 @@ const AssignmentSectionRow: React.FC<AssignmentSectionRowProps> = ({
               setIsUnassignment={() => showModal({ onConfirm: () => unassignAssignee() })}
             />
           ) : (
-            assignmentOptions.options.length > 1 && (
+            assignmentOptions.options.length > 0 && (
```
CarlosNZ, line 139: Not sure I understand this change. Shouldn't it still show if there's only 1 option?

src/utils/hooks/useCreateFinalDecisionReview.ts
CarlosNZ: So much clearer! :)

CarlosNZ (approved):
Okay, looks good thanks. It'll be good to get this on the server and see if any other problems arise. But seems pretty solid I think.

Couple of very minor question -- the only one that's potentially important is the one about the options.length > 1

CarlosNZ:
Sorry, one more little thing. Maybe not part of this PR, but I noticed this message:
Screenshot 2022-11-07 at 3 07 40 PM

The message doesn't really make sense -- it shouldn't include the name when its "self-asssignable" -- you only "self-assign" yourself, so should just say "Can be self-assigned". 

Ignore it for now if it's a hassle, we can deal with it elsewhere.

CarlosNZ:
Sorry, one more little thing. Maybe not part of this PR, but I noticed this message:
Screenshot 2022-11-07 at 3 07 40 PM

The message doesn't really make sense -- it shouldn't include the name when its "self-asssignable" -- you only "self-assign" yourself, so should just say "Can be self-assigned". 

Ignore it for now if it's a hassle, we can deal with it elsewhere.

src/containers/Review/assignment/AssignmentSectionRow.tsx
nmadruga: Yes, so it would show if there is at least one option (instead of at least 2 options as before - because it was considering the NOT ASSIGNED not selectable option)

src/containers/Review/assignment/AssigneeDropdown.tsx
nmadruga: Oh, Cool! I think it was used to show the selected option before. I'll remove

src/containers/Review/assignment/AssignAll.tsx
nmadruga: Yes, I think that was required. I'll revert...

src/containers/Review/assignment/AssignmentSectionRow.tsx
CarlosNZ: Ah right -- got it. 👍

nmadruga:
Ok - Carl I have replied to your changes suggested now. I have also added the `AssignAll` similar changes and added sorting as well - I think that seems like all #1408 would be completed by this PR now. I'll edit the description.


The only sad part is that Re-assignment isn't showing on the same order... Even though it uses the same component (As assigning) and the same options... So very odd. But I thinks it's not worth investigating more than the 5 minutes I have.

Issue #1413, opened 2022-10-31
Tweak: Missing one Review_element decision in ViewHistory modal


nmadruga:
https://conforma-demo.msupply.org:50008/application/AIM-2022-0089/review/30?active-sections=M2%2CM4

User: tecnico1 (check password with me)

Not sure how did this happened, but basically the first decision of **Tecnico1** user to REJECT is not displayed

PS: It was a Optional Reviewable element

![Image](https://user-images.githubusercontent.com/16461988/199116311-961e41df-d6c6-4d05-903f-49468e6661f5.png)
![Image](https://user-images.githubusercontent.com/16461988/199116312-7fa0d696-03b6-4435-a409-6c98f31dc289.png)

PR #1424, opened 2022-11-07
#1413 Fix missing response in history

CarlosNZ:
See back-end for more detail: https://github.com/openmsupply/conforma-server/pull/954

This just updates the front-end mutation so it doesn't try to directly write the stage_number field.

Please let me know if you find any other places this needs updating.


CarlosNZ:
@nmadruga have made fixes here as per your [comment on the missing application response](https://github.com/openmsupply/conforma-server/pull/954#pullrequestreview-1175009874).

5ea4ead4

src/utils/defaultLanguageStrings.ts
```diff
@@ -417,7 +418,7 @@ export default {
   TEMPLATE_GEN_BUTTON_AVAILABLE_DISABLED:
     'At least one template with the same code is already available, or this template already available',
   TEMPLATE_GEN_BUTTON_DRAFT: 'Make Draft',
-  TEMPLATE_GEN_BUTTON_DRAFT_DISABLED: 'Already has appications or is draft',
+  TEMPLATE_GEN_BUTTON_DRAFT_DISABLED: 'Already has applications or is draft',
```
CarlosNZ, line 421: Typo!

src/utils/hooks/useGetQuestionReviewHistory.tsx
```diff
@@ -55,18 +53,18 @@ const useGetQuestionReviewHistory = ({ isApplicant, ...variables }: UseGetQuesti
       const { firstName, lastName } = application?.user as User
 
       if (stageNumber) {
-        if (!allResponsesByStage[stageNumber]) allResponsesByStage[stageNumber] = {}
-        // Set each entry using HistoryElement values
-        allResponsesByStage[stageNumber][timeUpdated] = {
+        if (!allResponsesByStage[stageNumber]) allResponsesByStage[stageNumber] = []
+
+        allResponsesByStage[stageNumber].push({
```
CarlosNZ, line 58: Easier to make this an array of responses rather than object indexed by timestamp (which can be identical!)

src/components/PageElements/Elements/HistoryResponseElement.tsx
```diff
@@ -15,41 +16,50 @@ const HistoryResponseElement: React.FC<HistoryElement> = ({
...
-                isVisible: true,
-              } as ElementState
-            }
-            response={response}
-            // Not passing any other reponses, as these elements will be using
-            // pre-evaluated paramaters, not doing new evaluations
-            allResponses={{}}
-          />
+        {message ? (
+          <>
```
nmadruga, line 25: I think this is causing a weird problem : (

Parts of the History gets hidden by mysterious blank area
Screen Shot 2022-11-11 at 9 49 13 AM

Screen Shot 2022-11-11 at 9 49 19 AM

nmadruga: Ok - after running another `migrate` and reopening the weird blank area stopped showing and instead I've noticed this blinking button on hover... Not sure what's going on.


https://user-images.githubusercontent.com/16461988/201203991-7338db44-a4ba-4f01-bf5f-2518ac350bc4.mov

nmadruga (changes requested):
Hey! 

The improvement looks great! But I've found another annoying thing. We could live with it, but I honestly think this is new, so might be worth tracking the cause.

nmadruga (approved):

PR #1425, opened 2022-11-07
#1420 Hide reassign from list actions

### LINKED ISSUE #1420: Hide "Reassign" action from list home page.
Image

The "Reassign" action shouldn't be visible on the List page as it makes it to easy to do - re-assigning should be an occasional activity.

Just replace the "Reassign" link with the standard "Go to review homepage" link (View or Arrow?)

Also, seems a bit silly having both Self-Assign and Assign, so make it only one of them shows:

Screenshot 2022-11-04 at 5 34 33 PM

(Logic should probably be: If *only* self-assignable, show "Self-Assign", otherwise show "Assign" (even if self-assignable too))

---

Also... (and this is very much only if it's quick to do):

Add a loading spinner back to the application list (but don't break the "Search" input like we did last time we tried to add this)

CarlosNZ:
Fix #1420 

Simplified the ReviewerActionCell so it can only ever show a single action -- reviewer actions take priority (which includes "Self-assign") over Assigner actions.

"Re-assign" is hidden, will just display the default "View" (chevron) in that case.

Have also tidied up the Loading a little bit -- see inline comments for details.


src/components/List/ApplicationsList.tsx
```diff
@@ -45,26 +45,20 @@ const ApplicationsList: React.FC<ApplicationsListProps> = ({
           </Table.Row>
         </Table.Header>
         <Table.Body>
-          {loading ? (
-            <Table.Row colSpan={columns.length} textAlign="center">
-              <Table.Cell colSpan={columns.length} textAlign="center">
-                <Loading />
-              </Table.Cell>
-            </Table.Row>
```
CarlosNZ, line 53: `<Loading>` inside here doesn't look good as it appears in a white box. We weren't seeing this previous cos the loading prop wasn't actually coming through at all (as the whole component didn't render if Loading), but now 

We already have a `<Loading>` component below that shows up now.

src/containers/List/ListWrapper.tsx
```diff
@@ -140,7 +140,7 @@ const ListWrapper: React.FC = () => {
         filterDefinitions={visibleFilters}
         filterListParameters={{ userId: currentUser?.userId || 0, templateCode: type }}
       />
-      {columns && applicationsRows && (
+      {columns && (
```
CarlosNZ, line 143: This was the problem with the "Loading" previously -- the whole "List" wouldn't render until "applicationRows" was defined, by which time it was no longer "Loading". So now we just start with "applicationRows" as an empty array so we still get a List component with a Loading indicator.

src/components/List/ApplicationsList.tsx
nmadruga: Oh good to know Loading on Applications list is fixed 👏

nmadruga: Oh, although the "No applications found" warning still showing on the bottom below the Loading : (

CarlosNZ: OH -- good catch! Will fix...

nmadruga (approved):
Looks great! Only improvement if possible is: Can the "No applications found" not be displayed while loading?

src/components/List/ApplicationsList.tsx
CarlosNZ: Done 59d0e2ff :)

CarlosNZ:
> Looks great! Only improvement if possible is: Can the "No applications found" not be displayed while loading?

Done. Good catch :) 59d0e2ff

PR #1427, opened 2022-11-08
Extra warnings for snapshots

CarlosNZ:
In light of my absent-minded blitz of the demo server earlier today, I thought it would be helpful to have some extra warnings when loading a snapshot. I've made it so the warning only shows when you're on a Production build -- that way it should be really obvious.

Warning for overwriting a snapshot too.

Note: I *haven't* added a warning for deletion, as you normally want to delete a whole bunch at once as part of a "clean-up", and it'll be a real hassle to have an extra click for each one. Plus, it shouldn't be a big deal to delete a snapshot as you should always have a copy elsewhere.

@fergie-nz if you want to test this, just temporarily set `const isProductionBuild = !config.isProductionBuild` to reverse the effect, so the confirmations will show locally.

### DIFF
--- a/src/containers/Dev/Snapshots.tsx
+++ b/src/containers/Dev/Snapshots.tsx
@@ -3,6 +3,8 @@ import { Button, Icon, Input, Label, Loader, Modal, Table, Header } from 'semant
 import config from '../../config'
 import getServerUrl from '../../utils/helpers/endpoints/endpointUrlBuilder'
 import { getRequest, postRequest } from '../../utils/helpers/fetchMethods'
+import useConfirmationModal from '../../utils/hooks/useConfirmationModal'
+import { useToast } from '../../contexts/Toast'
 
 // const diffSnapshotUrl = `${snapshotsBaseUrl}/diff`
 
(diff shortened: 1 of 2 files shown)


src/utils/hooks/useConfirmationModal.tsx
```diff
@@ -36,13 +38,19 @@ const useConfirmationModal = ({
     onConfirm: onConfirm ? onConfirm : () => console.log('Clicked OK'),
     onCancel: onCancel ? onCancel : () => {},
     showCancel,
+    awaitAction,
```
CarlosNZ, line 41: Added this extra `awaitAction` prop so we have the option to close the modal *before* the confirm action runs. Otherwise, since the snapshot operation loads a modal of its own, it looks wrong with a transparent warning modal persisting under the snapshot modal. (But normally you would want this on, so it's the default)

src/containers/Dev/Snapshots.tsx
```diff
@@ -148,39 +154,69 @@ const Snapshots: React.FC = () => {
...
-                    className="clickable"
-                    size="big"
-                    name="random"
-                    onClick={() => setCompareFrom(snapshotName)}
-                  /> */}
+                  className="clickable"
+                  size="big"
+                  name="random"
+                  onClick={() => setCompareFrom(snapshotName)}
+                /> */}
```
CarlosNZ, line 219: Ignore

nmadruga (approved):

PR #1430, opened 2022-11-08
#1429 Template builder fallbacks fixes

### LINKED ISSUE #1429: Template builder bug -- not handling nulls and fallbacks correctly
After fixing the "logo_url" field on company registration for the millionth time, I think I've finally figured out why I've had to do the same fix many times.

Turns out the template builder is not always saving the proper values for its fallback. Like, when you have it set to `null` by default, it's actually just undefined in the expression. But if you change null to something else then back to null, it'll actually be the correct value. 

Similarly there's something fishy going on with the global "fallback" parameter, sometimes it  doesn't get saved, particularly if the value is `null`.

I'm not sure of the exact behaviour that's going on here, but there are switches that if you turn them off then on again (or vice versa) you'll get a different value than what the value is by default. I'll need to investigate further and then make sure the behaviour is consistent with what's expected and what's shown in the GUI.

CarlosNZ:
Fix #1429 

Two main fixes here:

### Global fallback switch
Screenshot 2022-11-09 at 11 50 10 AM
- has a proper `null` value when first switched on
- toggling the switch off now correctly removes the property completely
- values of `null` are preserved in evaluation expression

### Object properties fallback parameter
Screenshot 2022-11-09 at 11 52 23 AM
- can now be disabled (toggle switch) (which you'd want if using global fallback)

Easiest way to test is to just use the "Show as GUI" toggle to compare the state of the UI with what's actually stored in the expression. (And you can see how broken it is when using on develop)

### DIFF
--- a/src/containers/TemplateBuilder/evaluatorGui/EvaluationFallback.tsx
+++ b/src/containers/TemplateBuilder/evaluatorGui/EvaluationFallback.tsx
@@ -12,7 +12,7 @@ type EvaluationOutputTypeProps = {
 const EvaluationFallback: React.FC<EvaluationOutputTypeProps> = ({ evaluation, setEvaluation }) => {
   const { fallback } = evaluation.asOperator
 
-  const [fallbackEnabled, setFallbackEnabled] = useState(!!fallback)
+  const [fallbackEnabled, setFallbackEnabled] = useState(fallback !== undefined)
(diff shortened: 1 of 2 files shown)


nmadruga (approved):
LGTM

Looking forward to have this included in today's build

PR #1434, opened 2022-11-11
#953B Part3 remove final decision reviews in elements

nmadruga:
Final PR of fixes https://github.com/openmsupply/conforma-server/issues/953

Screen Shot 2022-11-11 at 3 49 30 PM

### Changes
- Correct to be copying previous (level 1) reviewResponses from `latestOriginalReviewResponse` (with fall back to thisReviewLatestResponse)
- Removed unused `previousAssignment` that was being passed along even though not in use. 
- Added comment explaining where the reviewResponses pointing to a previous assignment are getting set
- Noticed problem with "Approve All" which needed to remove the `stageNumber` from one mutation

Problem: I'm getting a problem (very difficult to track) which isn't allow the FinalDecision to be self-assigned in one specific type of template. I'll have to investigate further, but for now the Reviews per element are displaying correctly, so thought it should be time for reviews : )


src/containers/Review/assignment/AssignmentTab.tsx
```diff
@@ -118,7 +117,6 @@ const AssignmentTab: React.FC<{
                   sectionId={id}
                   fullStructure={fullStructure}
                   reviewAssignment={assignment}
-                  previousAssignment={assignmentInPreviousStage}
```
CarlosNZ, line 121: Can you just quickly remind me why we're not using this anymore?

src/utils/hooks/useCreateFinalDecisionReview.ts
```diff
@@ -24,48 +18,41 @@ type ConstructReviewPatch = (structure: FullStructure) => ReviewPatch
...
       const { isAssigned, isActiveReviewResponse } = element
-      return (
-        shouldRemakeFinalDecisionReviewResponse(element) && isAssigned && !isActiveReviewResponse
-      )
+      return isAssigned && !isActiveReviewResponse
     })
 
-    // For re-assignment this would be slightly different, we need to consider latest review response of this level
-    // not necessarily this thisReviewLatestResponse (would be just latestReviewResponse, from all reviews at this level)
+    // Generate each reviewResponse for FinalStage review as a copy of previous Lower Level review
```
CarlosNZ, line 38: > Generate each reviewResponse for FinalStage review 

Nit picky, but I think it'd be good to call this a "MakeDecision" review, as that's what its called in the UI. "Final Stage" is very much just an arbitrary name we give to the last stage. And this process you're describing here doesn't happen for every "final stage", only ones that have the "make decision" switch enabled, right?

CarlosNZ (approved):
Looks great, thanks Nicole. Thanks for making sense of all that, I think it'll be tidier going forward.

Just a couple of tiny comments/questions, but let's merge and then we can do some full testing for the "final decision fixes" branch early next week.

src/containers/Review/assignment/AssignmentTab.tsx
nmadruga: It wasn't been used. It was a optional parameter passed along to `ReviewSectionRow` to `useCreateFinalDecisionReview` and `useGetReviewStructureForSectionAsync ` to the helpers which should include that on the `compileVariablesForReviewResponseQuery` that didn't use it, if you check this call:
https://github.com/openmsupply/conforma-web-app/blob/59cf933cb71e72059ab554e0d2c5269c8aa3b31f/src/utils/hooks/useGetReviewStructureForSection/helpers.tsx#L54-L72

Very difficult to track it down, but in the end `previousAssignment` was already being ignored! That's a previous change that missed the tidy up part, the previous Stage responses are fetched in `getReviewResponses` without the need to pass the previousAssignment.

src/utils/hooks/useCreateFinalDecisionReview.ts
nmadruga: Hmmm that's a fair point. I'll rename functions/comments now

PR #1435, opened 2022-11-11
#1431 Reassignment fix

### LINKED ISSUE #1431: Reassign not quite working after back-end update
Pretty sure I know why:
```ts
// From useUpdateAssignment.tsx
assignmentsFiltered
        .filter(({ level: assignmentLevel }) => assignmentLevel === level)
        .forEach((assignment) => {
          if (isChanging(assignment.reviewer.id, assignedSections)) {
            const assignmentPatch = createAssignmentPatch(
              assignment,
              assignedSections,
              currentUser?.userId as number
            )
            results.push(
              updateReviewAssignment({
                variables: {
                  assignmentId: assignment.id,
                  assignmentPatch,
                },
              })
            )
          }
        })
      await Promise.all(results)
      fullStructure.reload()
```

You can see that this runs both mutations (the unassign and the assign one) asynchronously. The problem is, if the "assign" one tries to write first in the database, it won't accept the requested section as it's still assigned elsewhere.

So this block will need to be changed so that the two mutations run in sequence, and the "assign" one starts after the "unassign" one is complete.

CarlosNZ:
Fix #1431 

Requires back-end branch from https://github.com/openmsupply/conforma-server/pull/958#issuecomment-1311189036

Quite simple, just need to make the mutations sequential instead of parallel (and make sure the unassignment one comes first).

Unfortunately, I found a bug in my back-end enforcement function, so I've fixed that too.


fergie-nz (approved):
Nice! Looks great

PR #1436, opened 2022-11-13
956F Review locking updates

CarlosNZ:
Front-end changes required for back-end PR https://github.com/openmsupply/conforma-server/pull/959

We'll need to double check all the cases where locking is relevant before signing off on the feature branch too.

### DIFF
--- a/src/containers/Review/ReviewPage.tsx
+++ b/src/containers/Review/ReviewPage.tsx
@@ -131,14 +131,11 @@ const ReviewPage: React.FC<{
   const isSubmitted =
     thisReview?.current.reviewStatus === ReviewStatus.Submitted ||
     thisReview?.current.reviewStatus === ReviewStatus.ChangesRequested
-  const isLocked =
-    thisReview?.current.reviewStatus == ReviewStatus.Locked ||
-    thisReview?.current.reviewStatus === ReviewStatus.Discontinued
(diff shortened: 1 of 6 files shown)


nmadruga (approved):

src/containers/Review/ReviewPage.tsx
```diff
@@ -131,14 +131,11 @@ const ReviewPage: React.FC<{
   const isSubmitted =
     thisReview?.current.reviewStatus === ReviewStatus.Submitted ||
     thisReview?.current.reviewStatus === ReviewStatus.ChangesRequested
-  const isLocked =
-    thisReview?.current.reviewStatus == ReviewStatus.Locked ||
-    thisReview?.current.reviewStatus === ReviewStatus.Discontinued
+  const isLocked = thisReview?.isLocked
```
nmadruga, line 134: Hey, just wondering if we should keep this check for Discontinued too?

CarlosNZ: I'm not actually sure what "Discontinued" means.

When we review the full branch PR, there's a few places where we've pulled out stuff to do with "ASSIGN.LOCKED" or "is_locked" that we'll need to double-check (like this). Need to make sure that the checks being done are handled in the new case.

PR #1437, opened 2022-11-14
#953 - Fix Self-assign showing after already assigned

nmadruga:
Part of fix for https://github.com/openmsupply/conforma-server/issues/953

~~Pending: Back-end change to correctly set `availableSections` per ReviewAssignment TODO~~ (Should be done now)

### Brief explanation
Before we removed `isLocked` the ReviewAssignment for self-assignments knew it couldn't be assigned when the RA was locked (since already assigned to another), now we had to add this property which basically helps with that - and is clearer.

### Changes
- Added `availableSections` into ReviewAssignment fragment - now this is accessible in query `getReviewInfo`
- Check in `useLoadAssignments` hook to set `isFullyAssigned` - used for AssignAll - Should find if none of ReviewAssignment still have **availableSections** meaning it is Fully assigned...
- Check in `getAssignmentOptions` hook when is Self-assignment should also make sure **availableSections** include the SectionCode to display the option to Self-assign... 

This logic isn't the easiest to follow, but the general idea is that it needs to check the property availableSections of ReviewAssignments to determinate in UI if Self-assignment option should be displayed (since reviewers of Self-assignment only has access to their own ReviewAssignment). The code is different to display for AssignAll dropdown and for each section therefor the change in two places

### Tests

Use the snapshot `Fiji-test4` on our Fiji shared folder in `snapshot-for-testing-PR`.
You can try out assigning/Re-assigning and Self-assigning.
There are 2 templates with Final decision:
PR1 has 3 stages - last one is Final decision (after single-level review)
PR2 has 2 stages - last one is Final decision (after a consolidation)

### DIFF
--- a/src/containers/Review/assignment/AssignmentSectionRow.tsx
+++ b/src/containers/Review/assignment/AssignmentSectionRow.tsx
@@ -64,7 +64,6 @@ const AssignmentSectionRow: React.FC<AssignmentSectionRowProps> = ({
   const assignmentOptions = getAssignmentOptions({
     assignments,
     sectionCode,
-    // elements,
     assignee: assignedSections?.[sectionCode]?.newAssignee,
   })
   if (!assignmentOptions) return null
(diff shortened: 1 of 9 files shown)


CarlosNZ:
Screenshot 2022-11-16 at 10 44 48 AM

Looks really good so far. Just something I wondered -- in the above screenshot, this is what "Basic Reviewer" sees when someone else is assigned (Stage 1, Sponsor registration).

Just wondered if we should have a generic "Already assigned" message that shows instead of the blank space. We don't need to specify *who* is assigned, but might be useful to know that *someone* is, otherwise they'll wonder why they can't assign themselves.

Not a huge deal though, as they can see who is assigned from the timeline on the overview tab, so don't worry about it if it's not quick and simple.

CarlosNZ (approved):
LGTM. Seems to be working well. Just one little suggestion for you to consider, but it's not a biggie, we can do it later if you want.

nmadruga:
> Just wondered if we should have a generic "Already assigned" message that shows instead of the blank space. We don't need to specify _who_ is assigned, but might be useful to know that _someone_ is, otherwise they'll wonder why they can't assign themselves.
> 
> Not a huge deal though, as they can see who is assigned from the timeline on the overview tab, so don't worry about it if it's not quick and simple.

Sounds better to me also. Although, since the user doesn't have access to other reviews. Which means that what we can check is availableSections = [] and in this case we display the message?

nmadruga:
Ok @CarlosNZ it was a very small fix indeed : )

Screen Shot 2022-11-16 at 2 56 30 PM

Issue #961, opened 2021-09-06
DatePicker plugin -- allow SummaryView to handle ISO strings as input


CarlosNZ:
If we want things like expiry dates to show up in the Outcomes display as nicely formatted Dates (as opposed to raw ISO strings), we'll need to use the date picker plugin, but this expects a particular response shape, which is not that useful for actually "using" the date (for expiry checks).

Suggestion: update the date picker plugin so that it can interpret raw ISO strings as valid responses, and display them accordingly

PR #1439, opened 2022-11-15
#961 file upload fixes

CarlosNZ:
Requires back-end branch from https://github.com/openmsupply/conforma-server/pull/963

See back-end PR for details and snapshot to test with.

This just makes a small change to the fileUpload element to allow any number of arbitrary parameters to be passed to the "file-upload" endpoint (such as "template_id", which we need here) without having to manually specify each one. They just get passed directly through untouched.


nmadruga (approved):
Tested! Works very well!

Thanks, Carl

PR #1441, opened 2022-11-16
#1440 Search plugin - "input" mode

### LINKED ISSUE #1440: Update "search" element to have option to display selected value in the search box
As requested [here](https://github.com/openmsupply/conforma-templates/issues/88#issuecomment-1314486759) (amongst other places)

The reason why the default behaviour is to display the selected value *outside* the search box is that we often expect the selected result to be a complex, multi-value item, OR if we are accepting multiple values.

In both of these cases it doesn't make sense to display the selected inside the search, but I guess it seems reasonable when there is only a single, simple value selection.

It seems that we can do this with Semantic-UI search: https://react.semantic-ui.com/modules/search/#types-category

I suggest adding an additional value to the "displayType" parameter -- we currently have "card" and "list", we can add "input".

CarlosNZ:
Fix #1440

This was pretty easy to implement, so we can add this to the Import Permit template for @jagoje if we're happy with it.

Screenshot 2022-11-16 at 2 40 53 PM

Adds an additional displayType option -- "input", which keeps the selected value in the input field (only useful for single selections with plain text display).

To see it, use snapshot `Fiji_search_element_updates` in "Snapshot-for-testing-PR" folder. Log in as "adamdewey" (Jess' usual external password) and go to: http://localhost:3000/application/CPR-0012-22/S1/Page1

You'll see a new "Chemical name" field with an input search box (as shown above). Check it in both application view and summary view.

Docs updated accordingly.

### DIFF
--- a/docs/internal/Element-Type-Specs.md
+++ b/docs/internal/Element-Type-Specs.md
@@ -1,23 +1,28 @@
 _Ongoing authoritative reference of Template Question/Element types, including input parameters and response type (shape). Please ensure this document matches the current implementation at all times._
 
-## Contents
+## Contents <!-- omit in toc -->
 
-- [Template Element fields](#element-fields)
-- [Question/Element types](#types)
-  - [Short Text Input](#short-text)
-  - [Long Text (Multi-line) Input](#long-text)
-  - [Password](#password)
-  - [Text Information](#text)
-  - [Image Display](#image)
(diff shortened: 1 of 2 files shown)


nmadruga (approved):
👍 Looks great

jagoje:
I like it ;)

PR #1442, opened 2022-11-16
#59T Change string to replace List of Questions

nmadruga:
Fixes https://github.com/openmsupply/conforma-templates/issues/59


CarlosNZ (approved):
👍

PR #1447, opened 2022-11-20
#1446 Show manage menu based on contents

### LINKED ISSUE #1446: Allow "Management" menu to appear based on content, not "isAdmin" permission
Currently, any template can be set to appear in the "Manage" menu by setting it's `uiLocation` field. However, the "Manage" menu only appears to users with "isManager" permission.

This is inconsistent, since if it's set to appear in the "Manage" menu and has a different permission, then it should appear to any user with that permission.

It's a pretty small change to fix this -- just check for any items rather than "isManager" to determine whether to show the "Manage" menu at all.

CarlosNZ:
Fix #1446 

Very small change -- just changes the condition for the "Manage" menu appearance to whether there's anything in it or not. Everything else is already working how it should.

This allows us to show the "Remove user..." template in the Manage menu even if the current user doesn't have "isManager" permission.

Screenshot 2022-11-21 at 12 29 54 PM

### DIFF
--- a/src/containers/User/UserArea.tsx
+++ b/src/containers/User/UserArea.tsx
@@ -205,7 +205,7 @@ const MainMenuBar: React.FC<MainMenuBarProps> = ({
             />
           </List.Item>
         )}
-        {isManager && (
+        {managementOptions.length > 0 && (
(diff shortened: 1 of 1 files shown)


nmadruga (approved):
That's a real tiny PR : )

Seems good to me

PR #1448, opened 2022-11-21
#1445 move checks for admin manager inside current user

### LINKED ISSUE #1445: Improve applicationData to allow checking when currentUser is Admin or SystemManager
Something noticed on Template https://github.com/openmsupply/conforma-templates/issues/92
In that template which can be used by either an Applicant (manager of its own company) or a SystemManager/Admin which can select from which company they need to remove a user from. The difference is that, when the user is only an applicant (therefor no permissions related to `admin` or `systemManager` in the system) the template `removeUserOrg` should only remove a user from their own company (pre-selected on the form).

Currently in the front end, although we already fetch from the back-end the information about the the current user checking if any of these permissions apply:
- isAdmin
- isSystemManager

We don't currently pass that along to the object which is accessing in each template used for elements and actions validation. So this issue is basically to expose to the `currentUser` object to also allow this check on the elements/actions validation.

This is how the existing currentUser object is structured:
```json
"currentUser":
{
   "userId": int
   "username": String
   "firstName": String
   "lastName": String
   "email": String
   "dateOfBirth": Date
   "organisation":
   {
      "orgId": int
      "orgName": String
      "userRole": String
      "registration": String
      "address": String
      "logoUrl": String
      "isSystemOrg": bool
   }
"sessionId": String
}
```

After adding isAdmin and isSystemManager to this object checking if the user can select from the list of companies to remove a user would be simplified on the Element/Action to a simple check like this:
```json
{
 "operator": "objectProperties",
 "children": [
  "currentUser.isAdmin"
 ]
}
```

Please note, the permissions is still checked in the Back-end before the user can actually do the removal - so only user with proper `Admin`/`SystemManager` or the Applicant with permission to `manageUsers` would be allowed to remove the users anyway.

nmadruga:
Fixes #1445 

### Changes
- Moves `isAdmin`, `isManager` and `permissionNames` to be in the object for CurrentUser - which can be accessed by the applicationData on Elements
- No need to have preferences to determine the SystemManager in the Front-end anymore.

### Template using new changes
In case someone would like to test it
[removeUserOrg-5 (2).zip](https://github.com/openmsupply/conforma-web-app/files/10051855/removeUserOrg-5.2.zip)

### DIFF
--- a/src/containers/User/UserArea.tsx
+++ b/src/containers/User/UserArea.tsx
@@ -18,14 +18,14 @@ const defaultBrandLogo = require('../../../images/logos/conforma_logo_wide_white
 
 const UserArea: React.FC = () => {
   const {
-    userState: { currentUser, orgList, templatePermissions, isAdmin },
+    userState: { currentUser, orgList, templatePermissions },
     onLogin,
   } = useUserState()
   const {
     templatesData: { templates },
   } = useListTemplates(templatePermissions, false)
   const { dataViewsList } = useDataViewsList()
-  const { intReferenceDocs, extReferenceDocs } = useReferenceDocs(currentUser, isAdmin)
+  const { intReferenceDocs, extReferenceDocs } = useReferenceDocs(currentUser)
 
   if (!currentUser || currentUser?.username === config.nonRegisteredUser) return null
 
@@ -83,7 +83,7 @@ const MainMenuBar: React.FC<MainMenuBarProps> = ({
   })
   const { push, pathname } = useRouter()
   const {
-    userState: { isAdmin, isManager },
+    userState: { currentUser },
   } = useUserState()
 
   // Ensures the "selected" state of other dropdowns gets disabled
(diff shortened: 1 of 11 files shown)


CarlosNZ (commented):
Just wondering -- why are we changing so much here?

I thought we agreed this would just be a small change to what gets passed the form elements evaluation / applicatonData, we didn't want to change the shape of the whole "currentUser" object.

src/containers/User/UserArea.tsx
```diff
@@ -205,7 +205,7 @@ const MainMenuBar: React.FC<MainMenuBarProps> = ({
             />
           </List.Item>
         )}
-        {isManager && (
+        {currentUser?.isManager && (
```
CarlosNZ, line 208: Not this one, I've already changed it: #1447

CarlosNZ (commented):
Okay, I think this is fairly safe and works well. I'll check it again once the back-end is fixed up.

src/containers/User/UserArea.tsx
nmadruga: Done

CarlosNZ (approved):
Okay, looks good thanks.

I've just realised that with the change I made in #1447 that we no longer actually need `isManager` at all.

However, it's fine to leave it for now, as it might come in handy for something else.

PR #1449, opened 2022-11-22
CSS tweaks / Image display improvement

CarlosNZ:
Fix to better handle a wider range of logos in the Header bar.
Screenshot 2022-11-22 at 3 39 27 PM
becomes:

I also took the opportunity to improve the Image Display element to allow custom styles to be passed in. This allows us to prevent this kind of thing:
Screenshot 2022-11-22 at 3 34 07 PM

And restrict it to something like this::
Screenshot 2022-11-22 at 3 33 02 PM

### DIFF
--- a/src/formElementPlugins/imageDisplay/src/SummaryView.tsx
+++ b/src/formElementPlugins/imageDisplay/src/SummaryView.tsx
@@ -1,18 +1,22 @@
 import React from 'react'
 import { Image } from 'semantic-ui-react'
+import { allowedSizeValues } from './ApplicationView'
 import { SummaryViewProps } from '../../types'
 
 const SummaryView: React.FC<SummaryViewProps> = ({ parameters }) => {
-  const { url, size, alignment, altText } = parameters
+  const { url, size, alignment, altText, summarySize = 'small' } = parameters
+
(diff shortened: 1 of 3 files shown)


src/formElementPlugins/imageDisplay/src/SummaryView.tsx
```diff
@@ -1,18 +1,22 @@
...
 import { Image } from 'semantic-ui-react'
+import { allowedSizeValues } from './ApplicationView'
 import { SummaryViewProps } from '../../types'
 
 const SummaryView: React.FC<SummaryViewProps> = ({ parameters }) => {
-  const { url, size, alignment, altText } = parameters
+  const { url, size, alignment, altText, summarySize = 'small', style = {} } = parameters
+
+  // By default, image is displayed "small" in Summary view, but can be
+  // over-ridden by specifying "summarySize"
```
CarlosNZ, line 10: There are cases where we need Summary view to not be restricted to "small" (such as in Data views), so need to allow this to be overridden

(This oversight actually explains a lot of the issues I've had trying to get the Image plugin to display correctly sometimes, so glad to have fixed it)

nmadruga (approved):
LGTM

PR #1452, opened 2022-11-24
#1451 List pagination fix

### LINKED ISSUE #1451: Pagination not being followed on first load in application list
Works okay after selecting a pagination option, but it shows ALL applications at first, which is both (potentially) slow and messy.

CarlosNZ:
Fix #1451 

Very small tweak -- sets proper default page size from prefs (or fallback to 20).

Test -- load a list with more than 20 applications (such as registered companies in current Angola) and check that list is restricted to 20 applications.


nmadruga (approved):
LGTM

PR #1453, opened 2022-11-24
975F Edit serial pattern in Template builder

CarlosNZ:
Requires back-end https://github.com/openmsupply/conforma-server/pull/978

Very simple input field to edit the application serial pattern.


nmadruga (approved):
Testing now - I'll approve Front-end which seems solid.
So just pending a small tweak on Back-end to approve all.

PR #1457, opened 2022-12-02
#1456 Extend deadline fix

### LINKED ISSUE #1456: Extend due date after initial submission
Currently, after application submission, a due date is generated (for bringing in the physical dossier), and reviewers can see the due date on the Overview tab, but they can't yet extend it.

Small change to show the "Extend" button -- it should already be showing up, but I guess the condition is not quite right.

CarlosNZ:
Fix #1456 

I've just relaxed the condition a bit so that any application that is EXPIRED or PENDING (i.e. not APPROVED, REJECTED or WITHDRAWN) will show the "Extend deadline" UI.

I think this is okay, but please let me know if you think of any use cases where this is not appropriate.

Can test with Angola ProdReg --- submit an application and the "extend" option should be available to reviewers (it wasn't previously)

(Sorry, I based this off `feature/core-actions` -- I know it should really be from `develop`, but it's too hard to find and use a snapshot in the old state and switch back and forth consistently without messing something up. Hopefully we'll merge the whole feature branch to develop soon anyway)

### DIFF
--- a/src/containers/Review/overview/Overview.tsx
+++ b/src/containers/Review/overview/Overview.tsx
@@ -6,11 +6,7 @@ import useLocalisedEnums from '../../../utils/hooks/useLocalisedEnums'
 import useConfirmationModal from '../../../utils/hooks/useConfirmationModal'
 import { postRequest } from '../../../utils/helpers/fetchMethods'
 import { FullStructure } from '../../../utils/types'
-import {
-  ActivityLog,
-  ApplicationOutcome,
-  ApplicationStatus,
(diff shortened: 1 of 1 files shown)


fergie-nz (approved):
Looks great. I can't think of any case where allowing any pending application to be extended would be an issue.

Functionality is working fine with my tests.

Only thing I can see is removing obsolete ApplicationStatus import.

CarlosNZ:
> Only thing I can see is removing obsolete ApplicationStatus import.

Good call, thanks 👍

PR #1460, opened 2022-12-08
1459 Fix review history panel

### LINKED ISSUE #1459: ReviewHistory window still failing on Consolidation after a second review
As reported by @nmadruga :

> I took a snapshot with the problem saved to our shared folder Angola Conforma/System Configuration (Templates, Certificates, Lists)/Snapshots/snapshots for bugfixing/Angola_2022_12_08_1028_show_review_history_bug
> Login as Tecnico2/ARMED1234 See review: http://localhost:3000/application/AIM-2022-0122/review/93?activeSections=M1,M2,M3,M4,M5
Check the problem by clicking "Ver histórico" at this element ReviewHistory:
>  Screenshot 2022-12-08 at 4 59 01 PM

CarlosNZ:
Fix #1459 

Turned out to be pretty simple. Seems I overlooked (and didn't check) for stages which only have review responses and not application responses, in which case the `allResponsesByStage[stageNumber]` would be not defined yet. So we just create an empty array, like we already do for  the applicationResponses above this block.

### DIFF
--- a/src/utils/hooks/useGetQuestionReviewHistory.tsx
+++ b/src/utils/hooks/useGetQuestionReviewHistory.tsx
@@ -77,6 +77,7 @@ const useGetQuestionReviewHistory = ({ isApplicant, ...variables }: UseGetQuesti
 
       // Set each entry using HistoryElement values
       if (stageNumber) {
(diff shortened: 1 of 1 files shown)


nmadruga (approved):
Nice! Thanks

PR #1462, opened 2022-12-15
Improve deadline extend button condition

CarlosNZ:
Quick fix for problem @jagoje discovered -- the "Extend deadline" button was still visible even though the deadline had been "Cancelled" due to the application being assigned.

Have added additional condition logic so that "cancelled" (i.e. `is_active = false`) deadlines only allow the button to appear if the application is "Expired". This way we can distinguish between applications that have inactive deadlines due to expiration (i.e. can be "unexpired" using this UI), or inactive due to "cancellation" (i.e. the deadline is effectively "gone" and shouldn't be visible in the UI at all)

### DIFF
--- a/src/containers/Review/overview/Overview.tsx
+++ b/src/containers/Review/overview/Overview.tsx
@@ -84,8 +84,11 @@ export const Overview: React.FC<{
               )}
             </div>
             {applicantDeadline &&
-              (outcome === ApplicationOutcome.Expired ||
-                outcome === ApplicationOutcome.Pending) && (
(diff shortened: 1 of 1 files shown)


nmadruga (approved):
Looks good - I haven't test the actual one (Canceled), but looked that it is still working on Expired and not showing on Rejected. I wasn't sure what **Canceled** meant.

CarlosNZ:
> I wasn't sure what **Canceled** meant.

All the templates that have a deadline set also have a "cancel" event (e.g. when you submit the application) to get rid of the deadline.

PR #1463, opened 2022-12-19
#110 - Fix Other: Help & Reference list menus

nmadruga:
Simple fix to stop showing internal menu to external users

### DIFF
--- a/src/utils/hooks/useReferenceDocs.ts
+++ b/src/utils/hooks/useReferenceDocs.ts
@@ -38,13 +38,13 @@ export const useReferenceDocs = (currentUser: User | null) => {
 }
 
 const shouldSeeInternalDocs = (user: User): boolean => {
-  if (!user?.isAdmin) return false
+  if (user?.isAdmin) return true
   if (!user?.organisation) return false
   return user.organisation.isSystemOrg
 }
 
 const shouldSeeExternalDocs = (user: User): boolean => {
-  if (!user?.isAdmin) return true
+  if (user?.isAdmin) return true
   if (!user?.organisation) return true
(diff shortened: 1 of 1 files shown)


src/utils/hooks/useReferenceDocs.ts
```diff
@@ -38,7 +38,7 @@ export const useReferenceDocs = (currentUser: User | null) => {
 }
 
 const shouldSeeInternalDocs = (user: User): boolean => {
-  if (!user?.isAdmin) return true
+  if (!user?.isAdmin) return false
```
CarlosNZ, line 41: Should be ` if (user?.isAdmin) return true`

What you've got there will prevent all non-admin internal users from viewing them

src/utils/hooks/useReferenceDocs.ts
```diff
@@ -38,7 +38,7 @@ export const useReferenceDocs = (currentUser: User | null) => {
...
   if (!user?.organisation) return false
   return user.organisation.isSystemOrg
 }
```
CarlosNZ, line 44: Needs the same fix on the "shouldSeeExternalDocs" below -- Admin should see all docs (That's how it was previously -- see the git diff I posted)

CarlosNZ (changes requested):

src/utils/hooks/useReferenceDocs.ts
nmadruga: The check is `!` so if NOT admin. I don't like using the not check because it takes longer to see what's the actual check, but I tried not to change this code so much....

CarlosNZ (approved):
LGTM! 👍

PR #1471, opened 2023-01-16
#997B Prevent crash when displaying json data in lookup tables

CarlosNZ:
Front-end fix required for back-end PR https://github.com/openmsupply/conforma-server/pull/1003

Now that lookup tables can have data types other than `varchar`/`string`, the table display component was crashing as it couldn't handle the "object" data.

This just stringifies it for display.


nmadruga (approved):
Looks good

PR #1473, opened 2023-01-18
#1469 #1474 Plugin updates to handle complex pre-filling

### LINKED ISSUE #1469: Update ListBuilder (and maybe Dropdown) to handle default values

CarlosNZ:
Fix #1469 #1474 

~~Bit more than I'd hoped to get this working, and I'm still not 100% sold on it. I'd like to do a more robust overhaul of the plugin default handling (to make them all use a common mechanism), but that'll have to wait for now.~~

Ignore the above, I've managed to squeeze in #1474 which makes this a lot nicer and consistent (and easier to read!) There is a new, very simple hook, `useDefault` that the form elements all share, which handles the logic of checking and updating responses based on changing default values. (I've done for all elements that could possible require a default, except for "Checkboxes", since the checkboxes can be defined with a "selected" state in them already. (Admittedly, this would be a bit tricky to make dynamic, but we can add that when we need it.)

To test it out, use the snapshot `/demo_fiji/AAA_Fiji_Provisional_registration.zip` in the templates repo and look at the "Provisional Medicinal Product Registration" template. Fill in the "Search for existing product" with a valid product ID or name and notice the other fields automatically pre-fill (including the ones on page 2). Note what happens when you change the Product or the Authority (New Zealand / Australia / Other)

~~Please note, however, that they'll only pre-fill on the *first* selection -- if you change the product id again, they won't change again. If this functionality is required, it'll be a decent amount more work, so I'll need to check with @jagoje if it's needed.~~ (Done 🥳)

Currently there's only 5 products in the product lookup table, as I had to create them manually for this purpose, so the valid IDs are:

- 45077
- 278728
- 172845
- 309794
- 14937

**Update**: there's one case I haven't accounted for -- when the "default" source is reset to empty value/null, the dependent form elements *don't* get reset, because we're treating a `falsy` default as not having one, so nothing gets updated. Will have to have a think about how to handle this without messing up other cases (as normally empty default *does* mean there isn't one). Might have to explicitly distinguish between `null` and empty string, perhaps?

**Update 2**: I've kind of handled the above problem in that a default value of `null` will be recognised as a valid default, which can be used to reset certain elements. It won't work for all (as `null` is not a valid default for some elements), and it won't recognise a change if a different condition also outputs `null`, but it'll do for the current use case.

### DIFF
--- a/docs/internal/config/Element-Type-Specs.md
+++ b/docs/internal/config/Element-Type-Specs.md
@@ -72,6 +72,7 @@ _Free-form, single-line text input element_
 - **description**: `string` -- additional explanatory text (usually not required) [Optional]
 - **placeholder**: `string`-- text to display before user input (HTML "placeholder" attribute) [Optional]
 - **default**: `string` -- default response, will change dynamically in response to form changes until the user has edited it [Optional]
+- **replaceResponseOnDefaultChange**: `boolean` -- if `true`, when the `default` value changes, the current response will be replaced by the new default. Otherwise (`false`), the default will only be used if there is no response yet. [Optional; default `true`]
 - **maskedInput**: `boolean` -- if `true`, displays user input as masked (hidden) characters -- i.e. for passwords. [Optional]
 - **maxWidth**: `number` -- the maximum width (in pixels) for the text input box (defaults to fill the width of the container)
 - **maxLength**: `number` -- response must be no longer than this many characters. If the user tries to type more, the response will be truncated to the maximum length.  
@@ -113,6 +114,7 @@ _Free-form, multi-line text input element_
 - **description**: `string` -- additional explanatory text (usually not required) [Optional]
 - **placeholder**: `string`-- text to display before user input (HTML "placeholder" attribute) [Optional]
 - **default**: `string` -- default response, will change dynamically in response to form changes until the user has edited it [Optional]
+- **replaceResponseOnDefaultChange**: `boolean` -- if `true`, when the `default` value changes, the current response will be replaced by the new default. Otherwise (`false`), the default will only be used if there is no response yet. [Optional; default `true`]
 - **lines**: `number` -- height of the TextArea input, in number of lines/rows (default: 5)
 - **maxLength**: `number` -- response must be no longer than this many characters. If the user tries to type more, the response will be truncated to the maximum length. (See Note in ShortText above for how to integrate `maxLength` with validation.)
 
@@ -248,7 +250,8 @@ _Multi-choice question, with one allowed option, displayed as Drop-down list (Co
 - **label**: `string` -- as above
 - **description**: `string` -- as above [Optional]
 - **options**: `array[string | object]` -- array of options for the user to select from. If an array of **strings** is provided, these strings will be displayed to the user. However, if an array of **objects** is provided, you will also need to specify an `optionsDisplayProperty` (see below)
-- **default**: `string`/`number` -- if not provided, defaults to index 0.
+- **default**: `string`/`number` -- can be either a string representing the *value*, or a number representing the index in the `options` array. [Optional]
+- **replaceResponseOnDefaultChange**: `boolean` -- if `true`, when the `default` value changes, the current response will be replaced by the new default. Otherwise (`false`), the default will only be used if there is no response yet. [Optional; default `true`]
 - **search**: `boolean` (default: `false`) -- if `true`, the list of options can be searched and filtered by user
 - **optionsDisplayProperty**: If `options` (above) consists of an array of objects, this parameter specifies the field of each object to be displayed in the options list. For example, if `options` was a list of organisation objects (i.e. `{orgId, name, licenceNumber}`), you'd probably specify `name` as the `optionsDisplayProperty`. Note that even though one field is displayed to the user in the Dropdown list, the _entire_ selected object is saved as the selection. And if `optionsDisplayProperty` refers to a field that doesn't exist on the supplied object, the plugin will fail and show in error in the application.
 - **hasOther**: `boolean` (default `false`) -- if `true`, allows the user to enter a custom "free text" value instead of one of the pre-defined options.
@@ -282,6 +285,7 @@ _Multi-choice question, with one allowed selection, displayed as labelled radio 
 - **description**: `string` -- as above [Optional]
 - **options**: `array[string | object]` -- as above (in [Drop-down](#dropdown))
 - **default**: `string`/`number` -- the value initially selected before user input. If `number`, refers to the index of the options array. If not provided, no options will be pre-selected.
+- **replaceResponseOnDefaultChange**: `boolean` -- if `true`, when the `default` value changes, the current response will be replaced by the new default. Otherwise (`false`), the default will only be used if there is no response yet. [Optional; default `true`]
 - **optionsDisplayProperty**: -- as above (in Drop-down)
 - **layout**: `string` -- if "inline", displays radio buttons horizontally, rather than stacked vertically (default)
 - **hasOther**: `boolean` (default `false`) -- if `true`, displays an additional "Other" option with a free text field for inputting additional user-defined option.
@@ -442,6 +446,8 @@ _Allows user to build a list of items, such as an **Ingredients List**_
 
 - **label**: `string` -- as above
 - **description**: `string` -- as above [Optional]
+- **default** `ResponseList` -- default value for the list. Must be in the format of the `list` parameter from the Response object (i.e. an array of objects)
+- **replaceResponseOnDefaultChange**: `boolean` -- if `true`, when the `default` value changes, the current response will be replaced by the new default. Otherwise (`false`), the default will only be used if there is no response yet. [Optional; default `true`]
 - **createModalButtonText** `string` -- text to display on the button to launch the new item interface (modal) (default: "Add item")
 - **addButtonText** `string` -- text to display on the button to add a new item from the item editing modal (default: "Add")
 - **updateButtonText** `string` -- text to display on the button to update an existing item from the item editing modal (default: "Update")
@@ -555,7 +561,8 @@ Once selected, items are displayed in a "card" view:
     children: [ "search.text" ]
   }
   ```
-
+- **default** `ResponseSelection` / `ResponseSelection[]` -- default value for the selection(s). Must be in the format of the `selection` parameter from the Response object (i.e. an array of objects), or just a single object if not multi-select.
+- **replaceResponseOnDefaultChange**: `boolean` -- if `true`, when the `default` value changes, the current response will be replaced by the new default. Otherwise (`false`), the default will only be used if there is no response yet. [Optional; default `true`]
 - **icon**: `string` -- the name of the icon shown in the search box, from Semantic-UI icons. (default: "search" : 🔍 )
 - **multiSelect**: `boolean` -- whether or not the user can select multiple items for their response (default: `false`)
 - **minCharacters**: `number` -- the minimum number of characters the user must type before the search query executes (default: 1). This is useful in situations where need the user to look up a specific item without being able to freely browse through the entire results list. For example, to look up organisation in our system using "registration" code, we set `minCharacters = 6`, so the user will need to know an exact code rather than being able to try characters one at a time.
@@ -605,6 +612,7 @@ Uses [React Semantic-UI Datepickers](https://www.npmjs.com/package/react-semanti
 - **label**: `string` -- as above
 - **description**: `string` -- as above [Optional]
 - **default**: `(ISO) string` -- a pre-selected date [Optional]
+- **replaceResponseOnDefaultChange**: `boolean` -- if `true`, when the `default` value changes, the current response will be replaced by the new default. Otherwise (`false`), the default will only be used if there is no response yet. [Optional; default `true`]
 - **allowRange**: `boolean` -- if `true`, input is expected to be a date _range_ (a start and end date). Default: `false` (i.e. can only enter single date)
 - **minDate**/**maxDate**: `(ISO) string` -- specifies how far into the future or past the selector can go
 - **minAge**/**maxAge**: `(ISO) string` -- same as above, but a number (in years) relative to the _current_ date. For example, if an applicant was required to be over 18 years old, you'd set `minAge` to `18`.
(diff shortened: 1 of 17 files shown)


src/formElementPlugins/dropdownChoice/src/ApplicationView.tsx
```diff
@@ -123,6 +139,7 @@ const ApplicationView: React.FC<ApplicationViewProps> = ({
         // multiple
         selection
         clearable
+        disabled={!isEditable}
```
CarlosNZ, line 142: This was missing already -- oops!

docs/internal/config/Element-Type-Specs.md
```diff
@@ -72,6 +72,7 @@ _Free-form, single-line text input element_
 - **description**: `string` -- additional explanatory text (usually not required) [Optional]
 - **placeholder**: `string`-- text to display before user input (HTML "placeholder" attribute) [Optional]
 - **default**: `string` -- default response, will change dynamically in response to form changes until the user has edited it [Optional]
+- **replaceResponseOnDefaultChange**: `boolean` -- if `true`, when the `default` value changes, the current response will be replaced by the new default. Otherwise (`false`), the default will only be used if there is no response yet. [Optional; default `true`]
```
CarlosNZ, line 75: There _are_ occasional cases where we *don't* want a value to change after a user has already edited it. In those occasional times, we can use this new parameter. But for the most part, we can ignore it, as it won't apply very often.

src/formElementPlugins/number/src/ApplicationView.tsx
```diff
@@ -124,7 +116,7 @@ const ApplicationView: React.FC<ApplicationViewProps> = ({
     minValue: number = -Infinity,
     maxValue: number = Infinity
   ): { isValid: boolean; validationMessage?: string } => {
-    if (number === NaN || number === null)
+    if (Number.isNaN(number) || number === null)
```
CarlosNZ, line 119: Preferred syntax, according to VSCode :)

src/utils/helpers/utilityFunctions.ts
```diff
@@ -29,7 +29,8 @@ export const substituteValues = (
   // Custom replacement function for regex replace
   const getObjectProperty = (_: string, __: string, property: string) => {
     if (typeof data !== 'object') return data
-    return extractObjectProperty(data, property, `Can't find property: ${property}`)
+    const value = extractObjectProperty(data, property, `Can't find property: ${property}`)
+    return value ?? ''
```
CarlosNZ, line 33: Unrelated, but small tweak so that `null` values don't return "null" string.

nmadruga:
> Fill in the "Search for existing product" with a valid product ID or name and notice the other fields automatically pre-fill 

OK - filling fields in other elements all good

> (including the ones on page 2). 

Nothing on page 2 though, not sure what this was about

> Note what happens when you change the Product or the Authority (New Zealand / Australia / Other)

I noticed this, that perhaps could be improved:
1. I type a product that belongs to NZ, but had Australia selected
Screen Shot 2023-02-17 at 10 54 01 AM
2. Then I change to select New Zealand.
Screen Shot 2023-02-17 at 10 54 12 AM
3. The search area still show "Nothing found"
4. I have to remove one number and type again to see the result

I think it could be better to reset the element after you change the Authority. Ideas?

nmadruga:
Also after I selected the example: 278728	278728_NZ	New Zealand	ADYNOVATE

Not sure why the second combobox is showing this options, but also I imagine with non-editable dropdown we shouldn't show the x button:
Screen Shot 2023-02-17 at 11 03 34 AM

docs/internal/config/Element-Type-Specs.md
nmadruga: I think it would be clearer if you change this property name and use the opposite default value (and swap the logic), just because some property as `persistUserInput` seems a bit clearer than this long property.
So you `persistUserInput` default to `false` we should update the element in case `default` VALUE changes

nmadruga: Another comment: I not add this property explanation under the common properties (line 30) ?

src/formElementPlugins/ApplicationViewWrapper.tsx
```diff
@@ -23,6 +23,8 @@ import getServerUrl from '../utils/helpers/endpoints/endpointUrlBuilder'
 
 const graphQLEndpoint = getServerUrl('graphQL')
 
+export const DEFAULT_LOADING_VALUE = 'Loading...'
```
nmadruga, line 26: Maybe can use localisation here?

CarlosNZ:
> Nothing on page 2 though, not sure what this was about

Yeah, sorry. I think I was intending to test this scenario (pre-filling items on another page) but forgot to. I don't think it'll actually be needed for this form, but it'd be good to check that it works anyway, so i'll try and do that.

src/utils/helpers/utilityFunctions.ts
nmadruga: Ok - nice!

Another string that should possibly be localised....

CarlosNZ:
> Not sure why the second combobox is showing this options, but also I imagine with non-editable dropdown we shouldn't show the x button:

Yeah, I think that would be an easy fix (make the `clearable` prop conditional on `editable`). Will do.

src/formElementPlugins/useDefault.ts
```diff
@@ -0,0 +1,29 @@
...
+  loadingValue = DEFAULT_LOADING_VALUE,
+  replaceResponseOnDefaultChange = true,
+  onChange,
+  additionalDependencies = [],
+}: UseDefaultProps) => {
+  useEffect(() => {
+    if (defaultValue === loadingValue || defaultValue === undefined) return
+
+    if (replaceResponseOnDefaultChange || !currentResponse?.text) onChange(defaultValue)
+  }, [defaultValue, ...additionalDependencies])
```
nmadruga, line 26: I'm curious, what `...additionalDependencies` does in this context? And why we are observing other additionalDependencies changes? Maybe worth to add a comment on the code!

CarlosNZ:
Thanks Nicole, just summarising for myself what needs to be done/checked:

- [ ] Pre-filling not working properly for listBuilder -- it works in Template Builder, but not live form
- [ ] Make sure pre-filling works for items on another page (and check it changes if changing the original)
- [ ] Hide "clearable" icon if element is not enabled
- [ ] (Investigate) How hard is it to make the search re-run in response to other elements changing (and more importantly, make sure that this wouldn't make things *worse* for normal use cases)

src/formElementPlugins/dropdownChoice/src/ApplicationView.tsx
```diff
@@ -30,43 +28,46 @@ const ApplicationView: React.FC<ApplicationViewProps> = ({
...
-      })
-      setSelectedIndex(optionIndex)
-    }
-    if (currentResponse?.text) {
-      const { optionIndex } = currentResponse
-      setSelectedIndex(optionIndex)
+    // This deals with the case when a default response has been externally set
+    // (e.g. by a ListBuilder) and only a selection (with no index) has been
+    // provided.
+    if (options[0] === 'Loading...') return
```
nmadruga, line 54: Should check against same localised string - if that can be changed

src/formElementPlugins/dropdownChoice/src/ApplicationView.tsx
```diff
@@ -30,43 +28,46 @@ const ApplicationView: React.FC<ApplicationViewProps> = ({
...
+    if (currentResponse?.selection || currentResponse?.optionIndex === undefined) {
+      setSelectedIndex(getDefaultIndex(currentResponse?.selection, options))
```
nmadruga, line 56: Hmmm wondering where this case has gone?
```
if (currentResponse?.text) {
      const { optionIndex } = currentResponse
      setSelectedIndex(optionIndex)
}
```

nmadruga (changes requested):
Nice work. I'm happy with the unification of `default` and `defaultValue`!

I have a feeling this is going to need more testing before we can add on a Production build... Please also check my comments, nothing big. But I think it will be difficult to catch any potential bug by reading the code. We should do more tests before we are confident!

CarlosNZ:
> I'm happy with the unification of `default` and `defaultValue`!

Er... I didn't do that. They still exist for different purposes. What were you referring to?

nmadruga:
> Er... I didn't do that. They still exist for different purposes. What were you referring to?

I saw that most plugins now start with having `default: defaultValue` i.e for Datapicker:
Screen Shot 2023-02-17 at 11 45 25 AM

It seems that we are allocating what was set for defaultValue to be the default, am I understanding this wrong?

CarlosNZ:
> It seems that we are allocating what was set for defaultValue to be the default, am I understanding this wrong?

Yes, I'm only calling it `defaultValue` here because `default` is a reserved keyword in javascript so can't use it.

The *actual* `defaulValue` is a common field of `template_element` and never actually reaches the plugin -- it's entirely handled in `ApplicationViewWrapper`. 

I agree that the naming is confusing though... 🙄

docs/internal/config/Element-Type-Specs.md
CarlosNZ: > I think it would be clearer if you change this property name and use the opposite default value (and swap the logic), just because some property as `persistUserInput` seems a bit clearer than this long property.
> So you `persistUserInput` default to `false` we should update the element in case `default` VALUE changes

Yeah, okay good idea. Will do.

> Another comment: I not add this property explanation under the common properties (line 30)

I did think of that, but the "common properties" are ones that have their own database field in the `template_element` table, whereas this one is still a JSON parameter (even though it's available to all ). I did consider making it a "common" field, but I thought that would be a bit unnecessary and involve extra migration etc.. Also, doing that might add to the confusion between the dynamic default parameter and the common "defaultValue" parameter.

And since "default" is a JSON parameter for each plugin, it makes sense to leave another parameter relating to it in the same place.

I do agree though that it's confusing to have a global "defaultValue" field as well as per-element "defaults".

src/utils/helpers/utilityFunctions.ts
CarlosNZ: Probably not -- should only ever be seen in development.

And if it does ever show up in production, then users will be more likely to file a bug report if it's in the wrong language! :)

src/formElementPlugins/useDefault.ts
CarlosNZ: > I'm curious, what `...additionalDependencies` does in this context? And why we are observing other additionalDependencies changes? Maybe worth to add a comment on the code!

For most elements, the response value only needs when the incoming `default` parameter changes. However, for the "Drop-down" plugin, we also need to update it when the `options` parameter changes as well. So you'll note that for that plugin, it also passes `additionalDependencies: [options]` to the `useDefault` hook, so the dependency array for that particular instance will be different to the "standard" one.

src/formElementPlugins/dropdownChoice/src/ApplicationView.tsx
CarlosNZ: I'm a bit reluctant to do that as I think it'll break things. It would be okay for the DEFAULT_LOADING_VALUE, but it won't work easily in this case. That's because the "Loading" values are stored in each plugin's `pluginConfig.json` file, which aren't currently localisable. And we can't use the default in this case, because the `options` default loading value is `[ "Loading..." ]` not `"Loading..." (because it needs to be in an options array). 

So I think safer to leave this for now until we come up with a better overall solution for this (but that'll be quite complicated). (Ideally you don't see the "Loading" strings very often in the UI)

src/formElementPlugins/dropdownChoice/src/ApplicationView.tsx
CarlosNZ: That's now handled by the `useDefault` hook:
```
if (replaceResponseOnDefaultChange || !currentResponse?.text) onChange(defaultValue)
```

:)

CarlosNZ:
> > * The search area still show "Nothing found"
> > * I have to remove one number and type again to see the result
> 
> Yeah, I think this would be pretty tricky, since it only does a new search on keystrokes, not in response to other elements changing. I'll have a look and see how hard it would be to change, but I suspect it would be quite a bit of added complexity for a small benefit.

Have had a look and there's one way to achieve this, but not sure if it's worth it..

The search query is only re-evaluated when the user input changes:
```
  useEffect(() => {
    if (!debounceOutput) return
    evaluateSearchQuery(debounceOutput)
  }, [debounceOutput])
```

I could add `defaultValue` to that dependency array, but in this case it wouldn't help as this element doesn't have a default -- it's the "source" element that triggers default changes in all the others.

It works if I add `allResponses` to that dependency array, but I think this is overkill -- we don't really want the search query to re-run whenever any response changes, it'll be firing all the time in that case!

Options:
1. Just leave it how it is -- it's pretty much fine in most cases.
2. Add a new prop to the Search plugin called `refreshCodes` which is an array of element codes to watch, then add just those to the useEffect dependencies. Have had a quick play and it does work, but I'll need to do a bit more testing with it to make sure I've covered all the use cases.

Let me know what you think @nmadruga

nmadruga:
> 2. Add a new prop to the Search plugin called `refreshCodes` which is an array of element codes to watch, then add just those to the useEffect dependencies. Have had a quick play and it does work, but I'll need to do a bit more testing with it to make sure I've covered all the use cases.

I agree that seems a good option to fix this. Even though, I understand it might not happen very often. I think it's bad to have  logic that when happens creates a problem - that seems like a bug. Even though is just not implemented.
So, I think it would be worth having this flag, and also I liked your idea on how to check for only specific elements

src/formElementPlugins/dropdownChoice/src/ApplicationView.tsx
nmadruga: It might show up in slow connections with bad configured queries (that don't filter the data to show)

CarlosNZ:
> > 2. Add a new prop to the Search plugin called `refreshCodes` which is an array of element codes to watch, then add just those to the useEffect dependencies. Have had a quick play and it does work, but I'll need to do a bit more testing with it to make sure I've covered all the use cases.
> 
> I agree that seems a good option to fix this. Even though, I understand it might not happen very often. I think it's bad to have logic that when happens creates a problem - that seems like a bug. Even though is just not implemented. So, I think it would be worth having this flag, and also I liked your idea on how to check for only specific elements

Turns out my suggested solution was going to be too tricky -- I could add additional dependencies to the `useEffect`, but they couldn't be dependent on other variables within the component (i.e. "refreshCodes"), so it would only work if the watch values were hard-coded.

But I found a simpler way to handle it, which I think works pretty well for most scenarios -- I just added a `onFocus` event to the component which re-runs the search evaluation when coming back to it (as long as there's something to search and there's no selection already made). e3a2494b

CarlosNZ:
- [x] Pre-filling not working properly for listBuilder -- it works in Template Builder, but not live form (Okay, this seems fine now, not sure what was going on earlier for me. 🤷‍♂️)
- [x] Make sure pre-filling works for items on another page (and check it changes if changing the original) (Confirmed: working 👍🏼 )
- [x] Hide "clearable" icon if element is not enabled (Done: 6727d04f - I also hid it if the element is required)
- [x] (Investigate) How hard is it to make the search re-run in response to other elements changing (and more importantly, make sure that this wouldn't make things *worse* for normal use cases) ~~(Comment: https://github.com/openmsupply/conforma-web-app/pull/1473#issuecomment-1433928207)~~ (Better solution: e3a2494b)
- [x] Change `replaceResponseOnDefaultChange` parameter to `persistUserInput` and reverse the logic (Done: e4e90dd1)

Okay @nmadruga have address these now, see above notes/comments. Let me know if it's all working okay for you now.

nmadruga:
I'll have a look on it today : )

nmadruga (approved):
👍 Looks good, thanks Carl!

PR #1475, opened 2023-01-25
#1007B Template export fix

CarlosNZ:
Fix for back-end issue https://github.com/openmsupply/conforma-server/issues/1007 (PR: https://github.com/openmsupply/conforma-server/pull/1008)

Turns out that the snapshot file was sometimes being deleted *before* the download had completed, so I've moved the delete command to be part of the export/download button rather than part of the exportTemplate method.

There's not really much of a difference locally (only sometimes problematic), but it's apparent on the servers. 
I've added a pre-release build to https://conforma-demo.msupply.org:50000 so you can test it. Try exporting a template on that server compared with the other ones and notice how only this server reliably exports valid .zips.

### DIFF
--- a/src/containers/TemplateBuilder/Templates.tsx
+++ b/src/containers/TemplateBuilder/Templates.tsx
@@ -91,18 +91,12 @@ const ViewEditButton: React.FC<CellProps> = ({ template: { id } }) => {
 }
 
 const ExportButton: React.FC<CellProps> = ({ template: { code, version, id } }) => {
-  const downloadLinkRef = useRef<HTMLAnchorElement>(null)
   const { exportTemplate } = useOperationState()
   const snapshotName = `${code}-${version}`
   const JWT = localStorage.getItem(config.localStorageJWTKey)
 
   return (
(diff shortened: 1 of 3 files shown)


src/containers/TemplateBuilder/Templates.tsx
```diff
@@ -91,18 +91,12 @@ const ViewEditButton: React.FC<CellProps> = ({ template: { id } }) => {
...
   const snapshotName = `${code}-${version}`
   const JWT = localStorage.getItem(config.localStorageJWTKey)
 
   return (
     <div key="export">
-      <a
-        ref={downloadLinkRef}
-        href={getServerUrl('snapshot', { action: 'download', name: snapshotName })}
-        target="_blank"
-      ></a>
```
CarlosNZ, line 105: This was doing nothing (an empty link?!), so have removed.

nmadruga (approved):
Cool - indeed it seems like a simple fix. I'll check on instance 50000 and report here if any problems are found.

PR #1477, opened 2023-02-02
#120T Fix for listBuilder LOQ bug

CarlosNZ:
Fix [#120T](https://github.com/openmsupply/conforma-templates/issues/120)

This was quite a subtle bug. The mistake was in "anyIncompleteItems" function, we were iterating over `Object.values(currentInput)` and matching it to `inputFields` with the iteration index. However, this is incorrect as the Object values array doesn't get arranged in the same order as the inputFields (since it's from unordered key-value pairs). So the bug was happening due to the "inputField" not being matched to its correct "currentInput.response"

So now I've changed it to iterate over the `inputFields` array instead and match to the correct "currentInput" by code lookup, which keeps them aligned correctly.

I'm surprised it's taking this long for this bug to become apparent!

Everything else (isRequired, etc) seems to be all working fine.

### DIFF
--- a/src/formElementPlugins/listBuilder/src/helpers.ts
+++ b/src/formElementPlugins/listBuilder/src/helpers.ts
@@ -45,9 +45,7 @@ export const anyInvalidItems = (currentInput: ListItem) =>
   Object.values(currentInput).some((response) => response.isValid === false)
 
 export const anyIncompleteItems = (currentInput: ListItem, inputFields: TemplateElement[]) =>
-  Object.values(currentInput).some(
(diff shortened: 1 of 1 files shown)


CarlosNZ:
As to why it was only showing on LOQ -- well it actually was having the problem whenever the application was reloaded.  So if you filled the form in once, then reloaded the page, you'd have the same problem -- LOQ is a bit of a red herring.

This is because when first filling in the form the Object keys get created in the order of the form elements, so they stay in the correct order with respect to the responses array (JS preserves key order for the most part). However, when saved to the database, the order is lost (via JSON storage) so the *next* time the response is loaded, the object keys are incorrectly sorted.

PR #1478, opened 2023-02-17
Replace "defaultValue" with "initialValue"

CarlosNZ:
Closes back-end issue https://github.com/openmsupply/conforma-server/issues/1010

Required back-end PR: https://github.com/openmsupply/conforma-server/pull/1011

This was mostly just a matter of find-replace on the "defaultValue" usage (the "element"-level one, not within individual plugins), however there was one slight complication: we already had an `initialValue` prop for PluginComponents `<ApplicationView>` components.

However, it turned out that this was exactly the same as the `currentResponse` and we've been passing *both* to plugins all along. We actually noticed that a while ago, but never seemed worth fixing. Until now.

Fortunately, almost all the plugins just use the `currentResponse` prop, except for a couple (see code). But it's quite simple to just switch them over to using `currentResponse` too, and remove `initialValue` (as a plugin prop, not the new one) completely.

So, to be clear: as well as the `default`/`defaultValue` confusion, there was already another `initialValue` prop for ApplicationView components, but it was totally redundant so I've removed it.

See inline comments for more details.


src/formElementPlugins/ApplicationViewWrapper.tsx
```diff
@@ -188,7 +188,6 @@ const ApplicationViewWrapper: React.FC<ApplicationViewWrapperProps> = (props) =>
     <ApplicationView
       onUpdate={onUpdate}
       onSave={onSave}
-      initialValue={currentResponse}
```
CarlosNZ, line 191: You can see here that the existing `initalValue` prop is just the same thing as `currentResponse`, so we can get rid of it.

src/formElementPlugins/checkbox/src/ApplicationView.tsx
```diff
@@ -164,13 +164,13 @@ type KeyMap = {
...
 export const getCheckboxStructure = (
-  initialValue: CheckboxSavedState,
+  initialValue: CheckboxSavedState | null,
   checkboxes: Checkbox[],
   keyMap: KeyMap | undefined,
   isFirstRender: boolean
 ) => {
   // Returns a consistent array of Checkbox objects, regardless of input structure
-  const { values: initValues } = initialValue
+  const { values: initValues } = initialValue ?? {}
```
CarlosNZ, line 173: The `currentReponse` actually had a more accurate typing (`initalValue` was `any`), so I had to add a few catches for the `null` case in a couple of places -- which is better overall.

src/formElementPlugins/types.ts
```diff
@@ -32,7 +32,6 @@ type ValidationState = {
 interface ApplicationViewProps extends ApplicationViewWrapperProps {
   onUpdate: Function
   onSave: Function
-  initialValue: any
```
CarlosNZ, line 35: And removing it completely from the ApplicationView props :)

docs/internal/config/Element-Type-Specs.md
```diff
@@ -42,7 +42,8 @@ These fields are common to all element types and have their own field in the `te
   - default: `{"value": true}`
 - **is_editable**: `JSON` -- dynamic query determining whether can be edited (Would only be false in rare circumstances)
   - default: `{"value": true}`
-- **defaultValue**: `JSON` - the default value for the response. Note: several plugins have their own "default" parameter, which may be easier to use in many situations. What's the difference? In this one, the default response is generated when the application is first created, and from then on it behaves just like a normal question element. The default parameter for individual elements is calculated when the element is rendered, so it can change in response to other changes. If you need a default value to be made from answers the user provided earlier in the form, that would be the one to use. See individual elements for the respective details.  
+- **initialValue**: `JSON` - a dynamic query that will be evaluated to set the initial response value. It's kind of like a "default", but should be thought of as more of a fixed initialisation value (defaults can change in response to other elements). Most plugins have their own "default" parameter, which is probably easier to use in most situations.
```
nmadruga, line 45: Should be improved to explain or give an example for when to use it with just passing the basic element response object

nmadruga (changes requested):
Just a small thing to cover there : )

docs/internal/config/Element-Type-Specs.md
CarlosNZ: Okay, done: 21c0769f

I can't actually think of many realistic use cases for `initialValue` -- the ones I've provided here are pretty rare cases. Can you think of any more common ones?

nmadruga (approved):

nmadruga:
I'm approving, even though I think the explanation on `initialValue` is not really what I wished it was. I think examples in this case are easier. A long explanation like that is difficult to extract something easy to use. And honestly don't think people will read it all and know what to do! But I know we want to merge it now to keep consistent with back-ent

CarlosNZ:
> I'm approving, even though I think the explanation on `initialValue` is not really what I wished it was. I think examples in this case are easier. A long explanation like that is difficult to extract something easy to use. And honestly don't think people will read it all and know what to do! But I know we want to merge it now to keep consistent with back-ent

Thanks. As I mentioned above, I don't actually have any decent use cases. I think we've only used it in one place (for the initially-hidden value in Password reset). Can you think of any others? If so, please feel free to update it, or let me know.

nmadruga:
> Thanks. As I mentioned above, I don't actually have any decent use cases. I think we've only used it in one place (for the initially-hidden value in Password reset). Can you think of any others? If so, please feel free to update it, or let me know.

I think it doesn't need to be show casing a good use of it, just a visual example of how to use, when someone is wondering if that can be useful for something needed in the future. Hope that's helpful, I basically kept most of your text but reshuffled : )
[db21238](https://github.com/openmsupply/conforma-web-app/pull/1478/commits/db21238e515c56c45e32beb407782e60c1b2d203)

CarlosNZ:
> > Thanks. As I mentioned above, I don't actually have any decent use cases. I think we've only used it in one place (for the initially-hidden value in Password reset). Can you think of any others? If so, please feel free to update it, or let me know.
> 
> I think it doesn't need to be show casing a good use of it, just a visual example of how to use, when someone is wondering if that can be useful for something needed in the future. Hope that's helpful, I basically kept most of your text but reshuffled : ) [db21238](https://github.com/openmsupply/conforma-web-app/pull/1478/commits/db21238e515c56c45e32beb407782e60c1b2d203)

Okay, looks good, thanks. I'll merge now :)

PR #1479, opened 2023-02-21
Missing evaluator functions

CarlosNZ:
Quick fix for places where the evaluator functions weren't being passed to the expression evaluator.


nmadruga (approved):
👍

PR #1481, opened 2023-03-10
#1480 Multiple select for Drop-downs

### LINKED ISSUE #1480: Dropdown -- add multi-select option
And make sure the response shape stays compatible with existing instances.

CarlosNZ:
Fix #1480 

To see it implemented use the snapshot `dev/snapshots/core_templates_drop_down_multiselect.zip` from Templates repo and got to Section 1, Page 5 -- the first two elements on that page allow multiple selection -- one using objects in the background and one with strings. The string one (first one) also allows a "custom" option to be selected as well.

Be good to get this tested out in current snapshots too -- in particular, please ensure there are no regressions to existing behaviour!

### DIFF
--- a/docs/internal/Element-Type-Specs.md
+++ b/docs/internal/Element-Type-Specs.md
@@ -266,15 +266,30 @@ _Multi-choice question, with one allowed option, displayed as Drop-down list (Co
 - **search**: `boolean` (default: `false`) -- if `true`, the list of options can be searched and filtered by user
 - **optionsDisplayProperty**: If `options` (above) consists of an array of objects, this parameter specifies the field of each object to be displayed in the options list. For example, if `options` was a list of organisation objects (i.e. `{orgId, name, licenceNumber}`), you'd probably specify `name` as the `optionsDisplayProperty`. Note that even though one field is displayed to the user in the Dropdown list, the _entire_ selected object is saved as the selection. And if `optionsDisplayProperty` refers to a field that doesn't exist on the supplied object, the plugin will fail and show in error in the application.
 - **hasOther**: `boolean` (default `false`) -- if `true`, allows the user to enter a custom "free text" value instead of one of the pre-defined options.
+- **multiSelect**: `boolean` -- whether or not the user can select multiple items for their response (default: `false`)
+
+Note: if including an "Other" options, then options should only be an array of strings, or else you'll be mixing objects with strings, which will cause problems. In other words, you should never have both optionsDisplayProperty and hasOther defined in the same question element.
 
 #### Response type
 
+For single response configurations (`multiSelect = false`):
 ```
 {
   optionIndex: <integer> (index from the options array)
   text: <string> (actual text from options array)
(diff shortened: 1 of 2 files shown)


docs/internal/Element-Type-Specs.md
```diff
@@ -266,9 +266,13 @@ _Multi-choice question, with one allowed option, displayed as Drop-down list (Co
...
 - **optionsDisplayProperty**: If `options` (above) consists of an array of objects, this parameter specifies the field of each object to be displayed in the options list. For example, if `options` was a list of organisation objects (i.e. `{orgId, name, licenceNumber}`), you'd probably specify `name` as the `optionsDisplayProperty`. Note that even though one field is displayed to the user in the Dropdown list, the _entire_ selected object is saved as the selection. And if `optionsDisplayProperty` refers to a field that doesn't exist on the supplied object, the plugin will fail and show in error in the application.
 - **hasOther**: `boolean` (default `false`) -- if `true`, allows the user to enter a custom "free text" value instead of one of the pre-defined options.
+- **multiSelect**: `boolean` -- whether or not the user can select multiple items for their response (default: `false`)
+
+Note: if including an "Other" options, then options should only be an array of strings, or else you'll be mixing objects with strings, which will cause problems. In other words, you should never have both optionsDisplayProperty and hasOther defined in the same question element.
```
fergie-nz, line 271: It seems to work on initial testing with both hasOther and multiSelect for an array of objects.

Maybe it would be nice if it failed more loudly if this would cause issues.

src/formElementPlugins/dropdownChoice/src/ApplicationView.tsx
```diff
@@ -28,19 +28,22 @@ const ApplicationView: React.FC<ApplicationViewProps> = ({
...
     hasOther,
+    multiSelect = false,
     default: defaultValue,
   } = parameters as {
     label?: string
     description?: string
     placeholder?: string
     search?: boolean
     hasOther?: boolean
+    multiSelect?: boolean
```
fergie-nz, line 39: This is a general typescript question - do you need to have the ? for multiSelect if you've defined a default = false?

fergie-nz (approved):
That's great. Working very well.

docs/internal/Element-Type-Specs.md
CarlosNZ: You can use `hasOther` with `multiSelect` just fine -- you just can't use `optionsDisplayProperty` with `hasOther`. This has always been the case, it just wasn't documented.

src/formElementPlugins/dropdownChoice/src/ApplicationView.tsx
CarlosNZ: Good question -- no you don't. It would be fine to make `multiSelect` compulsory in this case (`mutliSelect: boolean`). The only reason I didn't is I didn't want someone casually browsing this to think there was any compulsory parameters for the plugin -- cos in the configs you don't want them to be compulsory.

PR #1484, opened 2023-03-16
#1026B Default filters for Data Views

CarlosNZ:
Front-end part of back-end PR: https://github.com/openmsupply/conforma-server/pull/1027

Just reads the "default_filter_string" provided by the data view API and appends it to the data view url.

To quickly test, load snapshot `/dev/snapshots/Fiji_2023_03_16_1827_default_filters.zip` from templates repo and check out the "Provisional Product Register" data view -- should load filtered to show only *active* products. 

Mess around with the "default_filter_string" value in the data_view definition and see what difference it makes.

### DIFF
--- a/src/containers/List/ListFilters/ListFilters.tsx
+++ b/src/containers/List/ListFilters/ListFilters.tsx
@@ -50,7 +50,7 @@ const ListFilters: React.FC<{
   // Add filters from URL to activeFilters, filters may not have criteria/options yet thus we have to use activeFilter state variable
   useEffect(() => {
     // Reset active filters if coming from Database menu selection
-    if (location?.state?.resetFilters) {
+    if (location?.state?.resetFilters && location.search === '') {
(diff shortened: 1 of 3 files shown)


src/containers/List/ListFilters/ListFilters.tsx
```diff
@@ -50,7 +50,7 @@ const ListFilters: React.FC<{
   // Add filters from URL to activeFilters, filters may not have criteria/options yet thus we have to use activeFilter state variable
   useEffect(() => {
     // Reset active filters if coming from Database menu selection
-    if (location?.state?.resetFilters) {
+    if (location?.state?.resetFilters && location.search === '') {
```
CarlosNZ, line 53: Just means it *doesn't* reset the active filters if there is something in the url query.

fergie-nz (approved):
Looks great!

PR #1487, opened 2023-04-02
Add analytics capability

CarlosNZ:
Was able to add this fairly easily.

Uses the [react-ga4 package](https://www.npmjs.com/package/react-ga4) to implement Google Analytics.

How it works is it looks for a `googleAnalyticsId` and `siteHost` value in the preferences received from the server. If these aren't present, we do nothing. If they are, we use a new `Tracker` component to report the current page via the aforementioned package.

The reason why we require `siteHost` as well as an ID is that we *don't* want tracking enabled on the Demo site or when running locally, so `siteHost` has to match the current url domain (i.e. "fijimedreg.conforma.systems") in order to operate.

Login details for Fiji analytics account in BW, and feature is currently active in the latest build.

Not sure best way to test this PR, you can try enabling the `trackerTestMode` field in `Tracker.tsx`, but not sure how it'll perform locally while it's already running on the server. But you can check out the results in action by looking at the Fiji account.


src/components/Dashboard.tsx
```diff
@@ -156,7 +156,7 @@ const PanelComponent: React.FC<{
     <>
       {filters.map((filter) =>
         totalMatchFilter[filter.id] > 0 ? (
-          <Link to={constructLink(filter)}>{`${filter.title} (${
+          <Link key={filter.id} to={constructLink(filter)}>{`${filter.title} (${
```
CarlosNZ, line 159: Unrelated, but fixes a long-standing key error

src/containers/Main/SiteLayout.tsx
```diff
@@ -12,10 +12,13 @@ import DevRoutes from '../Dev/DevRoutes'
 import DataViews from '../DataDisplay/DataViews'
 import config from '../../config'
 const { isProductionBuild } = config
+import { Tracker } from './Tracker'
 
 const SiteLayout: React.FC = () => {
   return (
     <Router>
+      {/* Google Analytics */}
+      <Tracker />
```
CarlosNZ, line 21: Wouldn't work if `Tracker` was placed outside `Router`. It needs to be as high as possible in the tree to avoid having to repeat it everywhere, and this is about as high as I can insert it. (which means it needs to be added separately to the Login page too)

src/components/Dashboard.tsx
fergie-nz: That's funny my linting/typescript analyzer isn't giving this error - I wonder if my set up isn't the same as yours

fergie-nz (approved):
Looks good!

That analytics info is super cool.

src/components/Dashboard.tsx
CarlosNZ: It's a runtime error (more of a warning really) in the console about a missing key (All iterative components require a unique key so React can keep track of which is which in subsequent re-renders, as the order might change)

PR #1489, opened 2023-04-13
#1486 Don't remove default filters on reset

### LINKED ISSUE #1486: Data Views -- default filters shouldn't be removed when resetting the filters
Image

When you click the "Reset all filters" link (in the Filter dropdown), it removes all filters, including the "default" ones. Might be best if the default filter string was kept, so we don't then start showing cancelled products (for example) in the list. It's still possible to remove the default filter, you just have to do it manually, so you can still filter by "Cancelled" if you want.

CarlosNZ:
Fix #1486 

Requires back-end: https://github.com/openmsupply/conforma-server/pull/1032

To test, go to the Provisional Product Register (Fiji) data view. It should launch with the "isActive=true" filter. Change and update some additional filters, then select "Reset all filters" from Filter button menu. The default `isActive=true` filter should remain (it doesn't in develop). 

Also added a label showing the total result count for all tables (Data Views and Applications List) so user doesn't have to perform mental arithmetic to figure out how many items there are in total for the current table/filter:
Screenshot 2023-04-13 at 12 40 41 PM

See inline comment for more details.

### DIFF
--- a/semantic/src/site/globals/site.overrides
+++ b/semantic/src/site/globals/site.overrides
@@ -1872,9 +1872,10 @@ a:hover {
     }
   }
 
-  .ui.label {
+  .result-count {
     flex-grow: 1;
     text-align: right;
+    font-size: @fontSmall;
   }
 
   // multiple is used to make sure filter menu stays open, but it adds extra
--- a/semantic/src/site/globals/site.overrides
+++ b/semantic/src/site/globals/site.overrides
@@ -320,7 +320,7 @@ a:hover {
   }
   img {
     max-height: @headerHeight;
-    padding-top:8px;
+    padding-top: 8px;
     padding-bottom: 8px;
   }
 }
@@ -551,7 +551,7 @@ a:hover {
     bottom: 0;
     width: 100%;
   }
-  #preview-items{
+  #preview-items {
     display: flex;
     flex-direction: column;
     gap: 10px;
@@ -562,17 +562,19 @@ a:hover {
       background-color: @surfacesLight;
       width: 90%;
       border-radius: @borderRadius;
-      padding: 10px
+      padding: 10px;
     }
     .notification-preview {
       padding-top: 5px;
       padding-bottom: 5px;
-      h1, h2 {
+      h1,
+      h2 {
         font-size: 1.1em;
         font-weight: bold;
       }
-      h3, h4 {
-        font-size: 1.0em;
+      h3,
+      h4 {
+        font-size: 1em;
       }
       p {
         font-size: 0.95em;
@@ -587,7 +589,8 @@ a:hover {
       padding-left: 6px;
       gap: 10px;
     }
-    .error-item, .fallback-preview{
+    .error-item,
+    .fallback-preview {
       p {
         font-size: 1em;
       }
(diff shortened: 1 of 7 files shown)


semantic/src/site/globals/site.overrides
```diff
@@ -551,7 +551,7 @@ a:hover {
     bottom: 0;
     width: 100%;
   }
-  #preview-items{
+  #preview-items {
```
CarlosNZ, line 554: Most changes in here just whitespace due to auto-formatting updates (I guess)

semantic/src/site/globals/site.overrides
```diff
@@ -1867,10 +1872,16 @@ a:hover {
     }
   }
 
+  .result-count {
+    flex-grow: 1;
+    text-align: right;
+    font-size: @fontSmall;
+  }
+
```
CarlosNZ, line 1880: The only *actual* change in this file

src/containers/List/ListFilters/ListFilters.tsx
```diff
@@ -269,6 +272,13 @@ const ListFilters: React.FC<{
             return null
         }
       })}
+      {totalCount ? (
+        <p className="result-count">
+          {totalCount === 1
+            ? strings.APPLICATIONS_LIST_TOTAL_RESULTS_SINGLE
+            : strings.APPLICATIONS_LIST_TOTAL_RESULTS.replace('%1', String(totalCount))}
+        </p>
```
CarlosNZ, line 280: This is a super-clunky way to handle pluralisation. Have made an issue to do it properly: https://github.com/openmsupply/conforma-web-app/issues/1488

src/utils/defaultLanguageStrings.ts
```diff
@@ -29,7 +29,9 @@ export default {
   APPLICATION_DELETION_CONFIRM_MESSAGE:
     'Please confirm you would like to delete a draft application',
   APPLICATIONS_LIST_EMPTY: 'No applications found',
-  APPLICATION_MISSING_TEMPLATE: "The applicantion's template was not found",
+  APPLICATIONS_LIST_TOTAL_RESULTS: '%1 results',
+  APPLICATIONS_LIST_TOTAL_RESULTS_SINGLE: '1 result',
```
CarlosNZ, line 33: 🤮

src/utils/hooks/useRouter.tsx
```diff
@@ -104,3 +106,8 @@ export function useRouter(): RouterResult {
     }
   }, [location])
 }
+
+const parseQueryString = (queryString: string): { [key: string]: string } => {
+  const queries = queryString.split('&').map((query) => query.split('='))
+  return Object.fromEntries(queries)
+}
```
CarlosNZ, line 113: Small enhancement to the "updateQuery" functionality to accept a query string directly (e.g. `isActive=false`) as well as an object of key/values (e.g. `{ isActive: false }` )

src/utils/hooks/useRouter.tsx
```diff
@@ -79,6 +82,20 @@ export function useRouter(): RouterResult {
...
+
+      if (replace)
+        history.replace({
+          search: queryString.stringify(restoreKebabCaseKeys(queryObject), { sort: false }),
+        })
+      else
+        history.push({
+          search: queryString.stringify(restoreKebabCaseKeys(queryObject), { sort: false }),
+        })
+    }
```
CarlosNZ, line 97: Overwrites the current query string rather than updating (merging) it.

fergie-nz (approved):
Looks great and easy to configure

Issue #1006, opened 2021-09-25
Localised Templates (and Filter names? and Outcomes fields?)


CarlosNZ:
I've got an idea for this that I think will work quite well and be fairly easy to implement. I'll whip up a prototype and demonstrate it, see what you think.

But it should work for Templates, Filters, Permission Names -- basically everything should be (optionally) localisable

CarlosNZ:
I will summarise what we discussed.

PR #1490, opened 2023-04-19
#1488 Better localisation substitution

### LINKED ISSUE #1488: Localisation -- better value substitution
Currently the way we substitute variables into localised strings is a bit clunky -- we just put a `.replace()` call in place in every use case.

Short of implement a proper localisation library, we should improve this with our own "translate" function, to avoid implementing in place every time. The "useLanguageProvider" hook should also provide this "translate" function (shorthand "t").

This would allow us to automate pluralisation for number counts, too, which is currently awkward as heck.

Syntax would be:

In languageStrings file, instead of:
```
TIMELINE_WAITING_FOR_REVIEWER: 'Waiting for reviewer %1 to make changes'
```
we have:
```
TIMELINE_WAITING_FOR_REVIEWER: 'Waiting for reviewer {{reviewer}} to make changes',
```

And then when used, instead of:
```
`${strings.TIMELINE_WAITING_FOR_REVIEWER}`.replace('%1', reviewer?.name)`
```
we have:
```
t(TIMELINE_WAITING_FOR_REVIEWER, {reviewer: reviewer.name})
```

The instantiation would change from:
```
```
to:
```
const t = useLanguageProvider("strings") 
```

CarlosNZ:
Fix #1488 

Massive PR, but mostly just find-replace on the translation methods. Main functionality changes are in `Localisation.tsx`

Key points:
- "translate" (and "translatePlugin") methods centralised to "Localisation" context. Instead of each usage extracting a key from "strings", and then doing it's own substitutions, it just passes the key to the translate method (abbreviated `t()`, like in omSupply) along with any properties that need substituting into the string.
- Replacement format uses named properties rather than positional substitution, so instead of "My name is %1", the string would be "My name is {{name}}"
- Replacements can be passed into `translate` as an object with named properties, or, if there's only one replacement, just pass the string. So in the above example, the replacement parameter could be either `{ name: "Carl" }` or just `"Carl"`.
- There is a special parameter for numbers: `{{count}}`. If this is used, then the translation function will look for a different key depending on the number, which is a more elegant way of handling pluralisation (how it's done in localisation libraries like i18n (used in omSupply)). So for example:  
  ```
  REVIEW_OVERVIEW_MODAL_MESSAGE: "This will extend the applicant's deadline by {{count}} days",
  REVIEW_OVERVIEW_MODAL_MESSAGE_1: "This will extend the applicant's deadline by one day",
  ```
  If `count` is anything other than 1, it'll use the plural (first) string, but if `count` is 1, it'll use the singular string
- plugins get a customised "translate" method, with the list of strings customised from their own config (and passed in as a closure on the translate method). I imagine we'll use a similar mechanism for translating templates when we finally get round to doing that (#1006) 

Have made an updated version of the Angola translations, which I'll install once this is in place.


src/contexts/Localisation.tsx
```diff
@@ -44,26 +47,34 @@ interface LanguageState {
...
   selectedLanguage: initSelectedLanguage,
   languageOptions: [],
   languageOptionsFull: [],
   loading: true,
   error: null,
   setLanguage: () => {},
-  getPluginStrings: () => {},
   refetchLanguages: () => {},
+  translate: () => '',
+  t: () => '',
```
CarlosNZ, line 76: This is so we can use either the shortcut name `t`, or the full name `translate` when calling. Mostly we'll use `t`, but there might be the odd occasion where the more verbose name is helpful.

src/formElementPlugins/checkbox/src/ApplicationView.tsx
```diff
@@ -58,8 +58,8 @@ const ApplicationView: React.FC<ApplicationViewProps> = ({
   currentResponse,
   validationState,
 }) => {
-  const { getPluginStrings } = useLanguageProvider()
-  const strings = getPluginStrings('checkbox')
+  const { getPluginTranslator } = useLanguageProvider()
+  const t = getPluginTranslator('checkbox')
```
CarlosNZ, line 62: This is the usage for plugins. It gets a custom version of `t` specific to the plugin, but subsequent usage is exactly the same, i.e. `t('KEY')`

fergie-nz (approved):
Looks great.

I've just tested superficially, but it is failing loudly with everything that I try

CarlosNZ:
> but it is failing loudly with everything that I try

Yikes! What do you mean exactly? (Screenshots?)

PR #1494, opened 2023-04-28
#1493 Dashboard Filters pagination fix

### LINKED ISSUE #1493: Max display range of forms from dashboard showing 20
Possibly pagination issue?

photo_2023-03-23_12-21-19

CarlosNZ:
Fix #1493.

Have kind of reverted this to the way it used to be, before we made it only do a single query for performance reasons. But now I've also made a new query where it only gets the total counts for each filter, which means a lot less data being downloaded for each filter.

I'll make a build of this and we'll see how it performs on the demo server; hopefully won't notice any performance issues, but I won't merge this until we're sure it's working fine.

In order to test, load a recent Fiji snapshot and you should see the new filter counts on the Dashboard. Log in as user "Emi" to see more variety than what "Admin" sees.


src/components/Dashboard.tsx
```diff
@@ -52,6 +57,11 @@ const Dashboard: React.FC = () => {
 
 const TemplateComponent: React.FC<{ template: TemplateInList }> = ({ template }) => {
   const { t } = useLanguageProvider()
+  const {
+    userState: { currentUser },
+  } = useUserState()
+  const [loadedFiltersCount, setLoadedFiltersCount] = useState(0)
```
CarlosNZ, line 63: @fergie-nz This is what you had in here earlier, just putting it back now :)

src/utils/graphql/queries/getApplicationFilterCount.query.ts
```diff
@@ -0,0 +1,9 @@
+import { gql } from '@apollo/client'
+
+export default gql`
+  query getFilteredApplicationCount($filter: ApplicationListShapeFilter, $userId: Int! = 0) {
+    applicationList(filter: $filter, userid: $userId) {
+      totalCount
+    }
```
CarlosNZ, line 7: New query just to get counts. Much more efficient than downloading all the data just to use the count values.

fergie-nz (approved):
Nice nice nice looks good

Issue #1495, opened 2023-05-01
"Default" values not working properly inside listBuilder


CarlosNZ:
I don't think we need this right now, but we should fix it sometime soon.

jagoje:
@CarlosNZ - sorry what is this one?

CarlosNZ:
Defaults values for form fields aren't working within listBuilders. If you try and set one, it'll just show up blank.

Would be good to fix, but I'm not sure how complicated it is.

PR #1496, opened 2023-05-01
#1492 #1495 ListBuilder improvements + misc

### LINKED ISSUE #1492: ListBuilder -- dynamic parameters
Ideally it would work the same as normal form elements, where each element can have dynamic expressions for all their parameters.

I think they currently get evaluated, but don't have access to the other field values within the listBuilder as data.

Should also make sure that all the main fields (visible, editable etc) are dynamically evaluated.

CarlosNZ:
Fixes #1492 and #1495, plus a few other minor tweaks required to make the new Fiji workflows work better.

Fortunately, most of the functionality for dynamic values in listBuilder were already in place, just needed a couple of minor fixes.

However, I needed to re-do the logic in the `useDefault` hook to make it work properly for *all* cases (dynamic values, static values and inside list builder (#1495)). I think it's working well now, but please check.

Test by using snapshot `/conforma-templates/dev/snapshots/Fiji_WIP_2023_05_02_1124.zip` from templates repo.

In particular, the following templates:
- **Test / Demo Feature Showcase** -- the defaults on the first page should behave as described
- **PR2 / Provisional Medicinal Product Registration** -- the pre-filling for products should behave as before (no regressions)
- **PR5 / Import permit for medicines** -- the "Total price" inside the list builder (for registered products) should dynamically update in response to the unit price and number of units inputs.


src/formElementPlugins/listBuilder/src/helpers.ts
```diff
@@ -79,7 +80,7 @@ export const buildElements = async (
     parameters: field.parameters,
     validationExpression: field?.validation || true,
     validationMessage: field?.validationMessage || '',
-    isVisibleExpression: field?.visibilityCondition || true,
+    isVisibleExpression: field?.visibilityCondition ?? true,
```
CarlosNZ, line 83: Makes visibility condition work now.

src/formElementPlugins/number/src/ApplicationView.tsx
```diff
@@ -49,6 +50,8 @@ const ApplicationView: React.FC<ApplicationViewProps> = ({
     style: currency ? 'currency' : undefined,
     currency: currency !== '' ? currency : undefined,
     maximumSignificantDigits: maxSignificantDigits,
+    minimumFractionDigits: decimals,
+    maximumFractionDigits: decimals,
```
CarlosNZ, line 54: Makes displaying dollar values nicer, so we can specify exactly 2 dp.

src/formElementPlugins/useDefault.ts
```diff
@@ -30,11 +30,22 @@ const useDefault = ({
...
 
-    if ((!persistUserInput || !currentResponse?.text) && !isInitialLoad) {
+    if (!currentResponse?.text) {
       onChange(defaultValue)
+      return
     }
+
+    if (persistUserInput) return
+
+    onChange(defaultValue)
```
CarlosNZ, line 48: This should fix the default behaviour, both for existing defaults, and inside listBuilder.

src/contexts/Localisation.tsx
```diff
@@ -88,7 +88,7 @@ export function LanguageProvider({
   const [languageState, setLanguageState] = useState<LanguageState>({
     languageOptions,
     selectedLanguage: initSelectedLanguage,
-    strings: {} as LanguageStrings,
+    strings,
```
CarlosNZ, line 91: Not related, but this just makes sure we start with the default strings available, otherwise we're seeing "LABEL_LOADING" in the UI when first loading, since the language files haven't been fetched yet. So now we'll at least get a sensible default.

fergie-nz (approved):
Looks good and can't see any regressions.

I notice that the multiply field in the list builder is still in the 'QUESTION' category, but also see we can just make these non-editable if we want.

CarlosNZ:
> I notice that the multiply field in the list builder is still in the 'QUESTION' category, but also see we can just make these non-editable if we want.

Did you mean the "Total price" field here?

Screenshot 2023-05-25 at 3 37 52 PM

Looks like I've already made it non-editable in my latest version.

src/containers/TemplateBuilder/evaluatorGui/evaluatorFunctions.ts
```diff
@@ -89,6 +89,15 @@ const extractNumber = (input: string) => {
 // See https://stackoverflow.com/questions/990904/remove-accents-diacritics-in-a-string-in-javascript
 const removeAccents = (input: string) => input.normalize('NFD').replace(/[\u0300-\u036f]/g, '')
 
+// Arithmetic
+const multiply = (num1: number, num2: number, decimals: number) => num1 * num2
```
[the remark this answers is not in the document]
CarlosNZ: Ooh, good catch. I intended to add a rounding option to this function, but then I implemented it in the Number plugin display instead, and then I forgot to remove the parameter here.

Have added in the rounding logic now so we can use it if we ever need it: 691d0172

Issue #1038, opened 2021-11-18
Upload file plugin should be allowed to be answered as shortText alternatively


nmadruga:
[Jess] When not providing a document they need to give an explanation

CarlosNZ:
Wouldn't it be easier to just add an additional shortText element below the file upload in cases when this is an option?

CarlosNZ:
Un-necessary -- there are several possible ways of doing the requested suggestion.

PR #1497, opened 2023-05-04
1028B Preferences UI

CarlosNZ:
Front-end for back-end [PR #1038](https://github.com/openmsupply/conforma-server/pull/1038)

Provides an (Admin) UI to edit the system preferences (so we don't have to re-load a whole snapshot for small tweaks). 
Screenshot 2023-05-04 at 2 01 20 PM

Access it via the "Configuration" menu.

The `react-json-view` package for the tree layout was already installed (used in Template builder), so I just used that to create a simple JSON editor. We can make a more user-friendly UI at a later date, but this should only be used by devs and other people who know what they're doing.


src/components/Admin/AdminLocalisations.tsx
```diff
@@ -115,7 +115,7 @@ export const AdminLocalisations: React.FC = () => {
   return (
     <div id="localisation-panel">
       <WarningModal />
-      <Header as="h1">{t('LOCALISATION_HEADER')}</Header>
+      <Header>{t('LOCALISATION_HEADER')}</Header>
```
CarlosNZ, line 118: Just makes this header consistent with the other Admin page headers

src/contexts/SystemPrefs.tsx
```diff
@@ -11,7 +11,6 @@ interface PrefsState {
     brandLogoFileId?: string
     brandLogoOnDarkFileId?: string
     defaultListFilters?: string[]
-    systemManagerPermissionName?: string
```
CarlosNZ, line 14: This shouldn't be here, it's a server preference.

fergie-nz (approved):

PR #1499, opened 2023-05-04
#1498 Markdown translation method

### LINKED ISSUE #1498: Allow urls to be placed in localised strings
**Edit**: Original suggestion to complicated, better just to use Markdown, since we already have it. Language Provider will just return an additional "translateMarkdown" method to pass the translated string through Markdown processor.

~~Suggested syntax for the string would be something like this:~~

<strike>

```json
"MESSAGE_WITH_LINK": "{{name}}, please click {{link}}this link{{/link}} for more information"
```

</strike>

~~Then, like `{{count}}`, we treat `{{link}}` as a "special" replacement value, and parse it into an HTML link.~~

~~The tricky part is that we don't know exactly what tag should be used (plain HTML uses `<a>`, but we might want React `<Link>`), so I suggest that the implementation can take the tag and the attributes as part of the substitutions object. Something like:~~

<strike>

```ts
<p>
  {t('MESSAGE_WITH_LINK', {
    link: {
      url: 'http://somewhere.com',
      tag: LinkComponent,
      attributes: { target: '_blank' },
    },
  })}
</p>
```

</strike>

~~If no tag is provided, then it would just default to `<a>`.~~

CarlosNZ:
Fix #1498.

Based off #1497 

This is a much more straightforward (and easier to implement) solution than my original suggestion. We just have an additional "translateMarkdown" (shorthand `tFormat`) method returned from the LanguageProvider, and only language strings that have formatting in them need to be passed to it.

(Will need to make sure you have at least https://github.com/openmsupply/conforma-server/pull/1038 branch in back-end so the "Preferences" page works)

Currently only implemented in one place, the System Preferences.

UPDATE: Added a quick fix for #1503 in here as well.

### DIFF
--- a/src/components/Admin/AdminPreferences.tsx
+++ b/src/components/Admin/AdminPreferences.tsx
@@ -10,7 +10,7 @@ import getServerUrl from '../../utils/helpers/endpoints/endpointUrlBuilder'
 import Loading from '../Loading'
 
 export const AdminPreferences: React.FC = () => {
-  const { t } = useLanguageProvider()
+  const { t, tFormat } = useLanguageProvider()
   usePageTitle(t('PAGE_TITLE_PREFS'))
 
   const showToast = useToast({ position: topLeft })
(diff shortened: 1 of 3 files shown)


fergie-nz (approved):
Cool cool cool looks good

PR #1502, opened 2023-05-09
#170T Default application sort by newest

CarlosNZ:
Fix https://github.com/openmsupply/conforma-templates/issues/170

Very simple -- jhust adds the "orderBy" field when none is specified.


fergie-nz (approved):
Looks great! 

Intuitively I would have thought sorting with most recent at the top would make sense - especially when the list is long and the most recent would be on another page.

But I am not sure exactly what Adam was asking for in this case.

CarlosNZ:
> Intuitively I would have thought sorting with most recent at the top would make sense - especially when the list is long and the most recent would be on another page.
> 
> But I am not sure exactly what Adam was asking for in this case.

Newest at the top is what's being asked for, and what this does, right?

Issue #1505, opened 2023-05-17
Timeline: Problem with event


CarlosNZ:
This might be nothing but worth checking into.

Review page url: http://localhost:3000/application/PROV-000156/review?tab=overview (Fiji snapshot)

One of the timeline events shows "Problem parsing this event -- See console for details". Need to look into it.

PR #1507, opened 2023-05-17
#1504 #1505 #1506 Preserve list filters + quick fixes

### LINKED ISSUE #1504: UI quirk -- wrapping in labels
Image

It's only showing up for me locally, not on the live server, so must be some quirk to do with the current column widths. I'd suggested changing the CSS for the labels to not allow any wrapping, and to allow them to overflow their container if needed.

CarlosNZ:
Fix #1504 #1505 #1506

Most of this is for #1506, the other two are one-liners.

To see #1506 in action, go to the application list, apply some filters, then click on one of the links to applications (either Name, Status or Action cells). Then from the application page, click the "back" link up the top of the page. The list page should be restored with the previous filters unchanged.

Try it for both applicant and reviewer.

Also -- if you can think of any other places that need this filter preservation when going back that I might have missed, please let me know.

See inline comments below for more info

### DIFF
--- a/semantic/src/site/globals/site.overrides
+++ b/semantic/src/site/globals/site.overrides
@@ -1555,6 +1555,10 @@ a:hover {
   margin-bottom: 20px;
 }
 
+.stage-label {
+  white-space: nowrap;
+}
+
 .stage-label-spacing {
   margin-bottom: 10px;
 }
--- a/src/components/Application/ApplicationContainer.tsx
+++ b/src/components/Application/ApplicationContainer.tsx
@@ -14,6 +14,8 @@ const ApplicationContainer: React.FC<ApplicationContainerProps> = ({ template, c
   const {
     userState: { currentUser, isNonRegistered },
   } = useUserState()
+  // Need to store in useState, else location.state is lost on subsequent
+  // re-renders
   const [prevQueryString] = useState(location?.state?.prevQuery)
   const { code, name } = template
 
--- a/src/components/Application/ApplicationContainer.tsx
+++ b/src/components/Application/ApplicationContainer.tsx
@@ -1,6 +1,6 @@
 import React from 'react'
 import { Container, Header, Icon, Label } from 'semantic-ui-react'
-import { useRouter } from '../../utils/hooks/useRouter'
+import { useRouter, usePreviousQuery } from '../../utils/hooks/useRouter'
 import { useUserState } from '../../contexts/UserState'
 import { TemplateDetails, User } from '../../utils/types'
 import { useLanguageProvider } from '../../contexts/Localisation'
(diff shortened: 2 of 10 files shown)


semantic/src/site/globals/site.overrides
```diff
@@ -1555,6 +1555,10 @@ a:hover {
   margin-bottom: 20px;
 }
 
+.stage-label {
```
CarlosNZ, line 1558: This is all that's required to fix #1504. Just prevents any wrapping within the label.

src/components/Application/ApplicationContainer.tsx
```diff
@@ -10,12 +10,17 @@ export interface ApplicationContainerProps {
 
 const ApplicationContainer: React.FC<ApplicationContainerProps> = ({ template, children }) => {
   const { t } = useLanguageProvider()
-  const { replace } = useRouter()
+  const { push, location } = useRouter()
   const {
     userState: { currentUser, isNonRegistered },
   } = useUserState()
+  const [prevQueryString] = useState(location?.state?.prevQuery)
```
CarlosNZ, line 19: This shouldn't really need to be a `useState` variable, but the re-renders mess it up. You'll see that the same thing is already done for the Dataview table (going back from the DataView detail page), and we didn't need any extra `useState` there.

One day we can go and hunt down some of these unnecessary re-renders....

src/components/List/Cells/ApplicantActionCell.tsx
```diff
@@ -4,17 +4,22 @@ import { ApplicationStatus } from '../../../utils/generated/graphql'
...
   if (status === ApplicationStatus.ChangesRequired) action = t('ACTION_UPDATE')
   if (status === ApplicationStatus.Draft) action = t('ACTION_CONTINUE')
 
   if (!action)
     return (
-      <Link className="user-action" to={`/application/${serial}`}>
+      <Link
+        className="user-action"
+        to={{ pathname: `/application/${serial}`, state: { prevQuery: location.search } }}
+      >
```
CarlosNZ, line 22: So the basic idea is that we can pass additional state through a link using this syntax. In this case we're passing the filter string so we can use it to generate the "Back" link in the linked component. :)

src/utils/hooks/useTimeline/eventInterpretation.ts
```diff
@@ -228,7 +228,7 @@ const getReviewEvent = (
           ? t('TIMELINE_REVIEW_STARTED', `**${reviewer?.name}**`)
           : t('TIMELINE_CONSOLIDATION_STARTED', `**${reviewer?.name}**`),
       }
-    case value === 'DRAFT' && (prevStatus === 'CHANGES_REQUESTED' || prevStatus === 'PENDING'):
+    case value === 'DRAFT' && ['CHANGES_REQUESTED', 'PENDING', 'DISCONTINUED'].includes(prevStatus):
```
CarlosNZ, line 231: Fix for #1505

fergie-nz (approved):
Nice looks good.

PR #1509, opened 2023-05-18
#1508 listBuilder layout improvements

### LINKED ISSUE #1508: Listbuilder -- parameter to specify all collapsed listboxes should be open initially
As discussed and noted in the "Review of Medicines Import Workflow - 16/05/23" document, Fiji staff would like the collapsed items to start all expanded.

We can do this by adding an additional parameter.

---

While we're at it, at an additional parameter for "table" view, to specify columns that should be hidden, as the tables are often too cluttered.

CarlosNZ:
Fix #1508 

Two new parameters for listBuilder:
- `inlineOpen` -- specify whether the Accordions should appear already open (not collapsed) for "inline" view elements
- `tableExcludeColumns` -- specify columns to hide in Table view (to prevent too much clutter when there are lots of fields)


docs/internal/Element-Type-Specs.md
```diff
@@ -195,35 +195,35 @@ _For displaying blocks of text in the application_
...
 
 #### `style: info`
 
-![style:info](/images/Element-Type-Specs-textInfo-info.png)
+![style:info](images/Element-Type-Specs-textInfo-info.png)
 
 #### `style: warning`
 
-![style:warning](/images/Element-Type-Specs-textInfo-warning.png)
+![style:warning](images/Element-Type-Specs-textInfo-warning.png)
```
CarlosNZ, line 210: This URL syntax was preventing the images showing up on the wiki. Hopefully this fixes it.

src/formElementPlugins/listBuilder/src/types.ts
```diff
@@ -18,13 +19,11 @@ export interface ListLayoutProps {
...
   displayFormat: { title?: string; subtitle?: string; description: string }
   Markdown: any
-  fieldTitles?: string[]
-  codes?: string[]
   editItem?: (index: number) => void
   deleteItem?: (index: number) => void
   isEditable?: boolean
   // These properties required for Inline layout
-  inputFields?: any
+  inputFields: TemplateElement[]
```
CarlosNZ, line 26: Stricter type here means less "as" etc inside. ;)

fergie-nz (changes requested):
I'm getting funny rendering effects when inlineOpen set to true - 

This is an example from the Provisional Medicinal Product Registration from Fiji

CarlosNZ:
> This is an example from the Provisional Medicinal Product Registration from Fiji

That's for the "Search" element *within* the listBuilder, not the listBuilder itself. It shouldn't affect it, but it's weird. I'll take a look.

fergie-nz (approved):
Nice have it working with newer snapshot

PR #1511, opened 2023-05-23
Data view editor

CarlosNZ:
~~Please don't review this until other PRs are merged, @fergie-nz. I based this off a combination of all the other PRs to avoid messy conflicts with the exiting work, but that means that there's no clean diff to any particular branch.~~

~~But once a few more PRs are merged then this will be cleaner. I'll update this a bit more when that happens, just wanted to get this published so we don't lose track of it.~~

Simple "Data view" editor, using the exising `react-json-view` package we already have in the template builder. This means there's now one more thing we can configure without having to upload and install a whole snapshot!

Requires back-end branch from: https://github.com/openmsupply/conforma-server/pull/1041

Screenshot 2023-05-25 at 4 03 58 PM

Should be fairly self-explanatory, see if you can find any bugs ;)

See inline comments below for more detail


src/components/Admin/AdminDataViews/AdminDataViews.tsx
```diff
@@ -0,0 +1,386 @@
...
+  const options = (data.dataTables?.nodes as DataTable[])
+    .filter(({ isLookupTable }) => (includeLookupTables ? true : !isLookupTable))
+    .map(({ id, tableName, isLookupTable }) => {
+      const table = toCamelCase(tableName)
+      return {
+        key: `${table}_${id}`,
+        text: `${table}${isLookupTable ? ` (${t('LOOKUP_TABLE_TITLE')})` : ''}`,
+        value: table,
+      }
+    })
```
CarlosNZ, line 345: We're relying on the "data_tables" being catalogued in the `data_table` table. This should happen automatically if the data table was created as a result of an action (modifyRecord) or imported as a lookup table, but might be missing if it was manually created. 

I also noticed some in there that shouldn't be (such as `userOrganisation`), so we should remove that from snapshots and check to see if it comes back.

src/components/Admin/AdminPreferences.tsx
```diff
@@ -19,131 +18,53 @@ export const AdminPreferences: React.FC = () => {
...
-  const [redoQueue, setRedoQueue] = useState<object[]>([])
 
   useEffect(() => {
-    getRequest(getServerUrl('getAllPrefs')).then((prefs) => setPrefs(prefs))
+    getRequest(getServerUrl('getAllPrefs')).then((prefs) => {
+      setPrefs(prefs)
+    })
   }, [])
 
-  const handleChange = ({
```
CarlosNZ, line 34: Made a common "JsonEditor" component that can be shared by Prefs and Data View editors, so a lot of this logic is moved out of here now. :)

src/config.ts
```diff
@@ -53,7 +53,11 @@ const config = {
   localStorageJWTKey: 'persistJWT',
   applicantDeadlineCode: 'applicantDeadline',
   isProductionBuild,
-  debounceTimeout: 350, // milliseconds
+  debounceTimeout: 350, // milliseconds,
+  // These are the only default tables in the system that we allow to be mutated
+  // directly by modifyRecord or display as data views. All other names must
+  // have "data_table_" prepended.
+  dataViewAllowedTableNames: ['user', 'organisation', 'file'],
```
CarlosNZ, line 58: This was previously only relevant to the back-end, but now we need it so the Data View editor knows what tables to offer. i.e -- all of the tables registered in `data_table` table, plus these three. (The back-end also includes "application" in its allowed tables, but I didn't think we need that here, since we already have a comprehensive application list)

src/utils/defaultLanguageStrings.ts
```diff
@@ -325,10 +360,6 @@ export default {
     'Are you sure?  This will overwrite core settings and can break the system if misconfigured',
   PREFERENCES_SAVED: 'Preferences saved',
   PREFERENCES_SAVE_PROBLEM: 'Problem saving preferences!',
-  PREFERENCES_COPIED: 'Copied to clipboard',
-  PREFERENCES_COPIED_ITEMS: '{{name}} ({{count}} items)',
-  PREFERENCES_COPIED_ITEMS_1: '{{name}} ({{count}} item)',
-  PREFERENCES_COPIED_ITEMS_0: '{{name}}: {{value}}',
```
CarlosNZ, line 331: Just changed the key, as it's no longer just relevant to Preferences.

src/utils/hooks/useUndoQueue.tsx
```diff
@@ -0,0 +1,55 @@
...
+  const setValue = (value: T) => {
+    if (value !== undefined) {
+      if (currentValue !== undefined) {
+        setUndoQueue([...undoQueue, currentValue])
+        setRedoQueue([])
+        setHasChanged(true)
+      }
+      setCurrentValue(value)
+    }
+  }
```
CarlosNZ, line 18: Probably a bit overkill to create a whole Undo queue mechanism, but it was a fun little task 😄

If we find that it's buggy though, I'm happy to replace it with an external package such as https://www.npmjs.com/package/use-undo

CarlosNZ:
Good to go now, @fergie-nz :)

src/utils/hooks/useUndoQueue.tsx
```diff
@@ -0,0 +1,55 @@
+import { useState } from 'react'
+
+export const useUndoQueue = <T,>(initialValue?: T) => {
```
CarlosNZ, line 3: Don't bother reviewing this, as I've replaced it with a 3rd-party package in #1514

CarlosNZ:
@fergie-nz I'll need this one approved before I can merge #1514.

I assume this is okay, as you would have needed to look at the functionality when reviewing #1514 (they're kind of tied together)

CarlosNZ:
Hey @fergie-nz, there's actually a bunch of PRs stacked on top of this one that you've approved (thanks for that, BTW), but can't get them all into develop until this one is merged first, so if you wouldn't mind looking at this one next. Cheers :)

src/components/Admin/JsonEditor.tsx
```diff
@@ -0,0 +1,94 @@
+import React, { useEffect } from 'react'
```
fergie-nz, line 1: Is this json editor different to the Json editor in the other PR?

https://github.com/openmsupply/conforma-web-app/pull/1514/files

fergie-nz (approved):
This update is awesome. Super cool to have this functionality.

No changes, but I was curious as to the json editor here + also another one in the other PR?

src/components/Admin/JsonEditor.tsx
CarlosNZ: They just use a different external package. The one in here is the one we've had all along, but we were only really using it for viewing applicationData in the Template Builder. The newer one is one I made myself due to frustrations at the limitations of this one. The main one being that my new one allows you to edit whole JSON object at once if you want, which means we can copy-paste whole data view configs in one go, rather than adding them one field at a time (which you have to do here).

PR #1514, opened 2023-06-02
Custom JSON editor

CarlosNZ:
The Edit-JSON package we were using is deprecated (abandoned for 2+ years), and had a few things that were quite lacking, so I've replaced it with one I made myself.

It mostly looks and functions the same as what's there,  but with a few improvements that making editing things like Data Views a lot more straightforward:

- you can edit a whole block of JSON at once, which makes it really easy to just copy-paste a whole config from one instance to another
- When you hold "Cmd/Ctrl" while clicking the "Copy" icon, it'll copy the object *path* to the clipboard rather than the value. This is super handy when configuring templates, as you can explore the responses object on the right and just copy/paste the exact path to the deep property you want.
- limit which objects/properties can be edited/deleted or added to on a case by case basis
- Other things which you'll notice as you go along.

I've also replaced the "useUndoQueue" hook with a 3rd party package as it's a bit more robust and still pretty small and efficient.

**Data Views:**
Screenshot 2023-06-02 at 8 38 05 PM

**Application Data sidebar in template builder:**
Screenshot 2023-06-02 at 8 38 38 PM

**Edit raw JSON directly**
Screenshot 2023-06-02 at 8 41 30 PM

Three places it's in use:
- System Preferences editor
- Data Views editor
- Application Data sidebar

I had to upgrade a few dependencies, namely React, so please note if there's any changed behaviour. (This is currently running on 50004, so we'll soon know if there's anything broken)


src/components/PageElements/Elements/ReviewInlineInput.tsx
```diff
@@ -49,7 +49,8 @@ const ReviewInlineInput: React.FC<ReviewInlineInputProps> = ({
   const optionsMap = useOptionsMap()
   const options = isConsolidation ? optionsMap.consolidation : optionsMap.review
 
-  const submit = async () => {
+  const submit = async (e: any) => {
+    e.preventDefault()
```
CarlosNZ, line 53: Required to fix issue caused by React upgrade.

package.json
```diff
@@ -61,7 +62,7 @@
     "@types/papaparse": "^5.3.2",
     "@types/pluralize": "^0.0.29",
     "@types/react": "^16.9.49",
-    "@types/react-dom": "^16.9.8",
+    "@types/react-dom": "^18.2.4",
```
fergie-nz, line 65: I'm generally curious how to manage when a bunch of dependencies are updated. Are there general tests to run? Are there standard ways to figure out if things will clash in unexpected ways? *

src/components/PageElements/Elements/ReviewInlineInput.tsx
fergie-nz: Would there possibly be an issue in reviewInlineNewResponse.tsx which uses the same code?

fergie-nz (approved):
nice looks awesome. 

I don't quite understand how it is all working - but has been well tested on the demo server.

package.json
CarlosNZ: Normally we should do it a bit more consistently as we go along. We've left this way too late. I'd like to update more of the dependencies, but some of them will be quite a bit of work. React Router, for example, is significantly different now, and we'll probably have to re-write our custom "useRouter" hook, which will affect a lot of areas.

src/components/PageElements/Elements/ReviewInlineInput.tsx
CarlosNZ: That's a very good point, thanks for spotting. I'm not totally sure, I'll need to go check this.

CarlosNZ: I've actually just found another problem when I tried to test this -- I can't actually review any questions that the applicant didn't answer, so can't make this method run.

Have made an issue: https://github.com/openmsupply/conforma-web-app/issues/1537

So I'll leave this as-is for now, but I'll make a note to double-check this when doing that one.

PR #1523, opened 2023-06-19
#1450 Checkboxes "Other" option (and improved defaults)

### LINKED ISSUE #1450: "Other" option for checkboxes
Last item shows "Other", and if checked has an input field to enter the custom value.

CarlosNZ:
Fix #1450 

Finally bit the bullet and did this one, as getting the Company Edit template to pre-fill the checkbox values nicely while handling the "Other" case was getting a bit ridiculous. It will be much easier now.

I'll update the company registration and company edit templates to use this properly, and you can use that for testing. (I'll post here when it's available.)

One other thing to check, is please check some of the other templates for potential regressions, in particular the ones that have pre-loaded or dynamic checkbox selections. (e.g. Grant User Permission, Provisional Product Reg)


src/formElementPlugins/checkbox/src/ApplicationView.tsx
```diff
@@ -199,14 +256,54 @@ export const getCheckboxStructure = (
...
+      const newCheckbox = textValues.includes(cb.text)
+        ? { ...cb, selected: true }
+        : { ...cb, selected: false }
+      textValues = textValues.filter((e) => e !== cb.text)
+      return newCheckbox
+    })
+    if (hasOther && textValues.length > 0) {
+      // Leftover textValues must be for the "Other" value
+      newCheckboxes[newCheckboxes.length - 1].text = textValues[0]
+      newCheckboxes[newCheckboxes.length - 1].selected = true
```
CarlosNZ, line 300: I would have like to have used the common `useDefault` hook for default handling, but because this element has its own dynamically handling of changing checkbox values, they wouldn't really play nicely together.

So the default handling is kind of unique to this element, the main feature being that it can uses a comma-delimited string to populate checkboxes (such as is needed for the company "sub-type" field).

src/formElementPlugins/checkbox/src/ApplicationView.tsx
```diff
@@ -199,14 +256,54 @@ export const getCheckboxStructure = (
         selected: cb?.[checkboxPropertyNames.selected] || false,
       }
   })
+
+  if (hasOther)
+    checkboxElements.push({
+      label: otherLabel,
+      text: initValues ? initValues.other?.text ?? '' : '',
```
fergie-nz, line 263: Am I understanding right that a default value for other - or is this just calling "PLACEHOLDER_OTHER" from the strings json?

src/formElementPlugins/checkbox/src/ApplicationView.tsx
fergie-nz: From a configurer's perspective, I would have thought an array of strings would be more intuitive given that is how the options are set - I wonder if this is easy to implement, it might be helpful for setting config?

fergie-nz (approved):
I have a minor preference for fixing the key = 'other' issue and also array of strings vs string split by ',' to make it easier for configurers, but these are subjective so I'm not sure if worth changing and have approved :)

Super awesome to have this feature - I can fix some of the wacky workaround features I've done.

src/formElementPlugins/checkbox/src/ApplicationView.tsx
CarlosNZ: Yeah, good point. The reason I have it this way is because of wanting to prefill values from the database. For example, the "company subtype" output is saved as a comma-seperated string, so in order to pre-fill these checkboxes for company edit, we need it to be able to handle it this way.

You're right though that it makes more sense for the default format to be the same as the actual checkboxes. I might see if I can get it to handle both. I'll let you know if/when I've done it.

src/formElementPlugins/checkbox/src/ApplicationView.tsx
```diff
@@ -44,6 +47,16 @@ export const CheckboxDisplay: React.FC<{
             toggle={type === 'toggle'}
             slider={type === 'slider'}
           />
+          {cb.key === 'other' && (
```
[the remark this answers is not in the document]
CarlosNZ: > This can be manually broken by assigning a key with 'other' as a value.

Hmm, not 100% sure what you mean, but I'll have a play and see if I can reproduce what you're seeing.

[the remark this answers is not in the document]
fergie-nz: You can configure a checkbox option with a simple string as one of the array items, or you can configure it with an object like:

{
 "key": 0,
 "label": "Option 1",
 "selected": true
}

And if the key is set as 'other', it will duplicate the other open text box like above.

Bit of a niche case.

src/formElementPlugins/checkbox/src/ApplicationView.tsx
CarlosNZ: Okay @fergie-nz, I've made a small change to allow an array of strings as default, too. d1be5e46

src/formElementPlugins/checkbox/src/ApplicationView.tsx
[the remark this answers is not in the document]
CarlosNZ: Yeah, so "other" is essentially a "reserved keyword" that you can't use as a checkbox key (can't really see why you'd ever need to -- most of the time we're not even using keys, just leaving it as default `0`, `1`, `2`, etc.)

src/formElementPlugins/checkbox/src/ApplicationView.tsx
CarlosNZ: It's just creating an additional checkbox (on top of the defined ones) if the "hasOther" option parameter is turned on.

Issue #1522, opened 2023-06-16
"Preview" button showing before any decision is available


CarlosNZ:
Pretty sure this wasn't like this previously:

![Image](https://github.com/openmsupply/conforma-web-app/assets/5456533/7c9b81bb-ea65-4ae5-978f-885ce2d0edec)

The "Preview" button shouldn't appear until the "Decision" selection is available

PR #1515, opened 2023-06-05
#1044B Submenus

CarlosNZ:
Front-end for https://github.com/openmsupply/conforma-server/pull/1046

In order to test, you'll need to define some submenus in two places (after the migration runs):
1. The "submenu" field of the "data_view" table
2. Set "is_submenu" to "true" for some categories in "template_category"

As noted in [Semantic UI docs](https://react.semantic-ui.com/modules/dropdown/), nested Dropdowns are not officially supported. However, they can work as demo-ed [here](https://codesandbox.io/s/x7vrmjmyqz?file=/index.js:670-698), albeit slightly clunkily. 

Hopefully we can improve this over time.

Edit: Also added "submenu" option to Template Builder

PR #1524, opened 2023-06-19
1045B Dashboard restrictions

CarlosNZ:
~~**NOTE: Please don't review this until after #1514 and #1515 are merged, as I've based this off *both* of those, and so the diff is not correct for either branch as base. I'll remove/update this notice once it's good to go.**~~

Fix https://github.com/openmsupply/conforma-server/issues/1045

(I also squeezed in a quick fix for #1522)

Back-end PR required: https://github.com/openmsupply/conforma-server/pull/1051 (Migration will need to run)

The rules for Dashboard appearance are:
- If "dashboard_restrictions" is `null` (the current state), then templates will appear on the Dashboard just like they do currently (i.e. as long as you have permission for it)
- However, the "dashboard_restrictions" value can be provided with an array of "filter" codes. In this case, the template *won't* appear on the Dashboard unless at least one of the named filters has a result of at least one application. This way certain templates can be hidden until they have active applications, for example.

I've added some configuration UI to the template builder, so you should be able to add/remove restrictions from the browser.

Screenshot 2023-06-19 at 4 47 17 PM


src/components/Dashboard.tsx
```diff
@@ -33,44 +33,82 @@ const Dashboard: React.FC = () => {
...
-            </div>
-          </div>
+      {dashboardCategories.map(({ templates, templateCategory: { icon, title, code } }) => (
+        <CategoryComponent
+          key={code}
+          templates={templates}
+          icon={icon as SemanticICONS}
+          title={title}
+        />
+      ))}
```
CarlosNZ, line 50: Just moved this mapping to its own component, since it needs a little internal logic/state.

src/components/Dashboard.tsx
```diff
@@ -33,44 +33,82 @@ const Dashboard: React.FC = () => {
...
+
+  return (
+    <div key={title} className={`template-category${readyToDisplay ? '' : ' hidden-element'}`}>
+      <div className="title">
+        {icon && <Icon size="large" color="grey" name={icon} />}
+        <Header as="h4">{title}</Header>
+      </div>
+      <div className="templates">
+        {templates.map((template) => (
+          <TemplateComponent key={template.code} template={template} setReady={setReadyToDisplay} />
```
CarlosNZ, line 72: So the tricky thing here was that we also want to hide categories if there are no templates within it that should appear. However, normally the Category is rendered before all the filters are loaded/checked. Which would have resulted in Categories (and even templates) appearing on the Dashboard for a few seconds until the filters loaded, then disappearing.

So I've added some additional logic here:
- Categories (and templates) that have no restrictions appear as normal (immediately)
- If any of the templates have restrictions, then they remain hidden until at least one filter returns a positive result.

src/components/Dashboard.tsx
```diff
@@ -33,44 +33,82 @@ const Dashboard: React.FC = () => {
...
+const TemplateComponent: React.FC<{
+  template: TemplateInList
+  setReady: Dispatch<SetStateAction<boolean>>
+}> = ({ template, setReady }) => {
   const { t } = useLanguageProvider()
   const {
     userState: { currentUser },
   } = useUserState()
-  const [loadedFiltersCount, setLoadedFiltersCount] = useState(0)
+  const [filterCounts, setFilterCounts] = useState<{ [key: string]: number }>({})
```
CarlosNZ, line 87: Rather than just keeping *count* of the number of loaded filters, we now need to know *which* ones have loaded, so we can set the template/category to "ready" accordingly.

src/containers/Review/ReviewSubmit.tsx
```diff
@@ -51,15 +51,17 @@ const ReviewSubmit: React.FC<ReviewSubmitProps> = (props) => {
...
+        <div className="flex-row-space-between-flex-end">
+          <ReviewDecision
+            decisionOptions={decisionOptions}
+            setDecision={setDecision}
+            isDecisionError={isDecisionError}
+            isEditable={thisReview?.current.reviewStatus === ReviewStatus.Draft}
+          />
+          <ReviewDecisionPreview structure={props.structure} decision={getDecision()} />
+        </div>
+      )}
```
CarlosNZ, line 64: Quick fix for #1522

src/containers/TemplateBuilder/template/General/Filters.tsx
```diff
@@ -112,14 +116,36 @@ const Filters: React.FC = () => {
...
   return (
     <>
       <Header as="h3">Dashboard Filters</Header>
       <div className="flex-row-start-center">
         <DropdownIO
           value={selectedFilter.id}
+          placeholder="Select filter"
           title="Filters"
-          options={allFilters}
+          options={[...allFilters, newFilter].filter((fil) => !filterJoinCodes.includes(fil?.code))}
```
CarlosNZ, line 148: Not strictly part of this issue, but I thought it was better if filters that had already been selected no longer appeared in the menu.

src/containers/TemplateBuilder/template/General/Filters.tsx
```diff
@@ -191,6 +216,50 @@ const Filters: React.FC = () => {
...
+      </div>
+      <div className="spacer-20" />
+      <div className="filter-joins">
+        {template.dashboardRestrictions &&
+          template.dashboardRestrictions.map((filter) => (
+            <Label
+              key={filter}
+              onClick={() => {
+                setSelectedRestrictFilter(filter)
+              }}
```
CarlosNZ, line 256: This is not great UI, but it's on a par with (and uses the same elements) as the existing filter edit UI. One of these days I wanna re-write the whole thing!

CarlosNZ:
> ~**NOTE: Please don't review this until after #1514 and #1515 are merged, as I've based this off _both_ of those, and so the diff is not correct for either branch as base. I'll remove/update this notice once it's good to go.**~

Good to go now 👍🏼

src/containers/TemplateBuilder/template/General/Filters.tsx
fergie-nz: What is bad about it? Is this it this one in particular, or a number of UI components?

fergie-nz (approved):
Nice looks great :)

src/containers/TemplateBuilder/template/General/Filters.tsx
CarlosNZ: The template builder overall, in particular all the  "IO" components with the label/contents pattern. That was never part of the original design and while it looks okay in moderation, a whole page of it (especially when nested) looks pretty bleh (IMHO), not to mention it's definitely not mobile friendly. 
Screenshot 2023-08-04 at 1 55 13 PM

PR #1526, opened 2023-06-21
#1525 Sort and Filter for templates list

### LINKED ISSUE #1525: Add a "Category" filter to the templates page
The "Templates" page is getting a bit unwieldly (lots of templates!) so it would be very handy (and simple to implement) a "Category" drop-down filter at the top.

Maybe add sorting by column headers as well, if it's not too complicated.

CarlosNZ:
Fix #1525 

Basic sorting and filtering for Templates list. Should make it a bit easier to find what we're looking for and organise categories, etc.

No need to make a new database query, since we're already pulling the whole data set -- so this is just an in-place rearrangement of the DB result.

### DIFF
--- a/src/containers/TemplateBuilder/Templates.tsx
+++ b/src/containers/TemplateBuilder/Templates.tsx
@@ -161,8 +161,11 @@ const Templates: React.FC = () => {
   const fileInputRef = useRef<HTMLInputElement>(null)
   const { templates, refetch } = useGetTemplates()
   const { importTemplate } = useOperationState()
-  const [hideInactive, setHideInactive] = useState(false)
-  const [selectedCategories, setSelectedCategories] = useState<string[]>([])
+  const { query, updateQuery } = useRouter()
+  const [hideInactive, setHideInactive] = useState(query.hideInactive === 'true')
+  const [selectedCategories, setSelectedCategories] = useState<string[]>(
+    query.categories ? query.categories.split(',').map((cat) => (cat === 'none' ? '' : cat)) : []
+  )
   const [sortColumn, setSortColumn] = useState<SortColumn>()
   const [sortAsc, setSortAsc] = useState<1 | -1>(1)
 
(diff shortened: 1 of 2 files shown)


src/containers/TemplateBuilder/Templates.tsx
```diff
@@ -20,7 +20,7 @@ type Columns = {
 
 const columns: Columns = [
   {
-    title: '',
+    title: 'code',
```
CarlosNZ, line 23: Not sure why "Code" was never displayed in header

src/containers/TemplateBuilder/Templates.tsx
```diff
@@ -153,20 +153,44 @@ const TemplatesWrapper: React.FC = () => (
...
 const Templates: React.FC = () => {
   const { t } = useLanguageProvider()
   const [selectedRow, setSelectedRow] = useState(-1)
   const fileInputRef = useRef<HTMLInputElement>(null)
   const { templates, refetch } = useGetTemplates()
   const { importTemplate } = useOperationState()
+  const [hideInactive, setHideInactive] = useState(false)
+  const [selectedCategories, setSelectedCategories] = useState<string[]>([])
+  const [sortColumn, setSortColumn] = useState<SortColumn>()
+  const [sortAsc, setSortAsc] = useState<1 | -1>(1)
```
CarlosNZ, line 167: The reason I've used -1 / 1 rather than `true`/`false` is that you can just multiply it by the result of the compare function to switch the direction. (See "sortTemplates" function)

fergie-nz: Nice,

I feel like this is on the boundary for me of easy legibility. If I ever were to come back to this sort of thing with fresh eyes, I'd probably take some time to figure it out. Might just be me, though! Looks good.

fergie-nz (approved):
Nice, looks great!

src/containers/TemplateBuilder/Templates.tsx
CarlosNZ: Yeah, fair enough. It's probably on the borderline. I was just happy with how many less lines of calculation it yielded! :)

PR #1531, opened 2023-06-23
#1529 File upload improvements

### LINKED ISSUE #1529: FileUpload -- show acceptable file formats and max size automatically
As pointed out [here](https://github.com/openmsupply/conforma-templates/issues/188) we often end up with configurations where the stated allowed file formats are not actually the same as the configured allowed file formats.

Thinking it would probably be useful to have this information shown automatically based on the defined formats.

We should make the text smaller, and include it in the upload box itself, rather than wasting the description on it.

It should be optional, so we'd need a parameter to enable or disable it (but on by default). And we'd need a helper function to convert the `fileSizeLimit` in bytes to a readable format: e.g. `5000` => "5MB".

CarlosNZ:
Fix #1529 

Shows the file restrictions (allowed extensions and max size) automatically below the upload box, to save us having to put them in the description (which we've been really bad at keeping consistent):

Screenshot 2023-06-23 at 4 24 45 PM

Unfortunately, we're going to have to go through every template and remove the existing descriptions, but worth it IMHO.

Also, while I was here I improved the error message display, so we get a nice line of text instead of that ugly red box.

Screenshot 2023-06-23 at 5 10 53 PM


docs/internal/Element-Type-Specs.md
```diff
@@ -435,6 +435,7 @@ _Interface for uploading documents or other files_
 - **fileSizeLimit**: `number` -- maximum file size in KB (default: no limit)
 - **subfolder**: `string` -- by default, files are uploaded into a subfolder with the name of the application serial. However, this can be over-ridden by specifying this parameter. This should rarely be required.
 - **showDescription**: `boolean` -- if `true`, an additional text input will be displayed alongside each file to allow the applicant to specify a description for each file. (default `false`)
+- **showFileRestrictions**: `boolean` -- will display the allowed file extensions and maximum size below the upload box (default: `true`)
```
fergie-nz, line 438: I wonder if default would be better at first so we can have a more easy transition with all of the templates?

I haven't fully thought this through, but wondered if this might help.

src/formElementPlugins/fileUpload/localisation/default/strings.json
```diff
@@ -1,10 +1,12 @@
 {
-  "ERROR_FILE_ALREADY_UPLOADED": "File already uploaded",
-  "ERROR_FILE_TYPE_NOT_PERMITTED": "File type not permitted",
-  "ERROR_FILE_TOO_BIG": "File too big",
+  "ERROR_FILE_ALREADY_UPLOADED": "File already uploaded!",
+  "ERROR_FILE_TYPE_NOT_PERMITTED": "File type not permitted!",
+  "ERROR_FILE_TOO_BIG": "File too big!",
   "ERROR_TOO_MANY_FILES": "Too many files!",
-  "ERROR_UPLOAD_PROBLEM": "Problem uploading file",
+  "ERROR_UPLOAD_PROBLEM": "Problem uploading file!",
```
fergie-nz, line 6: I have a personal preference for no exclamation marks - but that is just me! :)

fergie-nz (approved):
This is awesome. Really like the implementation - much smoother and cleaner UI

docs/internal/Element-Type-Specs.md
CarlosNZ: > I wonder if default would be better at first so we can have a more easy transition with all of the templates?

It would in the short term, but in the long run it'd be more annoying I think. 99% of the time we'd want it on (in fact I can't really think of a use case for turning it off).

And it's easy to quickly update templates in place as we see them -- Fiji is already pretty much done.

src/formElementPlugins/fileUpload/localisation/default/strings.json
CarlosNZ: LOL, yeah that's a good point actually. Not sure why I put them in. I can remove: b9ec31de

PR #1532, opened 2023-06-23
#1530 Date picker allowed range fix

### LINKED ISSUE #1530: Bug: Date picker lets you manually enter unallowed dates
Issue 2 from https://github.com/openmsupply/conforma-templates/issues/188

In that case, even though the component has been configured to limit dates to the past, you can circumvent it by manually typing the date.

So need to add an additional check "onBlur" and either display error, or just automatically modify the date to a valid one.

CarlosNZ:
Fix #1530 

Adds an additional check that input date falls within allowed range.

I've also updated the `react-semantic-ui-datepickers` package to the latest version, so don't forget to `yarn`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,7 +42,7 @@
     "react-markdown": "^5.0.3",
     "react-progress-stepper": "^0.2.2",
     "react-router-dom": "^5.3.3",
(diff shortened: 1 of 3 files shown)


fergie-nz (approved):
Nice nice nice looks good

PR #1533, opened 2023-06-28
#1054F Add new data view fields to queries

CarlosNZ:
I just realised when doing https://github.com/openmsupply/conforma-server/pull/1055 that the front-end wasn't querying these new fields in the Data View configuration UI.

So small fix for that.


fergie-nz (approved):
Looks great.

I'm curious as to why so much of the generated file is removed when adding fields?

CarlosNZ:
> I'm curious as to why so much of the generated file is removed when adding fields?

It's just cos the base branch had a bunch of fields from a different snapshot (i.e. the custom data is different). I'll make sure to regen the whole thing from core_templates when I merge back to develop.

PR #1536, opened 2023-06-29
#1534 Form element localisation consistency

### LINKED ISSUE #1534: Extend the localisation module to get plugin translations from the main bundle
We currently keep form element/plugin translations in the respective plugin folders, which means they get neglected when translating and aren't exported/imported with the main localisation configs.

My suggestion is this:

- Keep the "default/strings.json" for each element where it is (in the plugin folder)
- When translating, instead of looking in the plugin folder, get them from the main "strings" json that was fetched from the server. 
- To avoid key collisions with existing strings (and to distinguish them in the localisations files), the plugin keys should be prefixed with the plugin code. e.g. for `BUTTON_RESET_SELECTION: "Reset selections"` in the Checkbox plugin, we'd export it as `checkbox.BUTTON_RESET_SELECTION`.
- Update the Localisation "Export" script to also traverse the plugins folder and pull out the default strings (and add appropriate code prefix as above)
- Update the "translatePlugin" method in Localisation Context to build using these strings rather than the internal ones
- Maybe simplify the folder structure of the form elements since we don't need separate subfolders per language any more. Instead of `localisation/default/strings.json`, we can just have `localisation.json` file in the plugin root.

CarlosNZ:
Fix #1534 

In order to see/test this properly, you'll need to switch to a branch that contains all the latest changes and therefore the new strings that go with them (whatever the latest "build" branch is).

You'll see all the plugin strings have been exported and are included in the master translations list: https://docs.google.com/spreadsheets/d/1YvS76N7wLFdh94kmbCrFGIoSQGF2UUirON7mwmUYl0A/edit#gid=2053384988.

You can check that all the plugin translations get imported and exported correctly on the Admin/Localisations page.


fergie-nz:
I haven't been able to figure out how to change the strings - maybe I can run through this one with you next week?

fergie-nz (approved):
Nice Carl, looks awesome

PR #1540, opened 2023-07-12
#1539 Make Submit button inactive while "loading"

### LINKED ISSUE #1539: Bug: it's possible to click the Review "Submit" button multiple times and cause multiple duplicate triggers
This seems to be the cause of this problem: https://github.com/openmsupply/conforma-templates/issues/194

Can confirm that this occurs for Review submit, but it'd pay to check others as well.

CarlosNZ:
Fix #1539 

Kind of surprising to discover that the Semantic `<Button/>` is still clickable when it's in its "loading" state.

So this just prevents the "onConfirm" method from being called while it's "loading".

Also done a back-end fix to prevent this, too: https://github.com/openmsupply/conforma-server/pull/1058

### DIFF
--- a/src/utils/hooks/useConfirmationModal.tsx
+++ b/src/utils/hooks/useConfirmationModal.tsx
@@ -93,7 +93,11 @@ const useConfirmationModal = ({
           inverted
           icon="checkmark"
           content={modalState.confirmText}
-          onClick={() => handleConfirm(modalState.onConfirm)}
+          // This ensures that the "Confirm" action can only be triggered once.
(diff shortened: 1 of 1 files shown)


fergie-nz:
That is surprising! Does the button loading state not change immediately?

fergie-nz (approved):

CarlosNZ:
> Does the button loading state not change immediately?

The loading state changes, but having the button "loading" doesn't stop it from being clicked again, unless you explicitly set the `disabled` property.

PR #1543, opened 2023-08-06
#1068B File archive and template version improvements

CarlosNZ:
Fix https://github.com/openmsupply/conforma-server/issues/1068

As per original issue, the following fixes/improvements are implemented:

### File Archiving
- [x] Save total size of each archive in each archive's info.json, so this info can be displayed in front-end.
- [x] Bug when creating an archive snapshot when no upper range (happens after already creating one snapshot)
- [x] Group snapshots with the same name in collapsible row (like template versions)
  - [x] Use same technique on Templates page
  - [x] "Cmd" click collapses all
- [x] Display of missing archives not great (should be list):
Screenshot 2023-08-04 at 5 11 13 PM

### Template versioning
- [x] "Commit this version" goes missing after making new version (something to do with "Status", not sure exactly)
- [x] Setting status = "Available" gives warning if version is not committed
- [x] Duplicating template should always set Status = DRAFT (Already done)
- [x] Once committed, shouldn't be able to set it to Draft

#### Collapsing snapshots (with the same name):
Screenshot 2023-08-06 at 12 50 03 PM

#### Improved template version collapsing:
Screenshot 2023-08-06 at 12 49 47 PM

- More consistent with Snapshots
- Easier to tell what's opened
- Can keep multiple expanded
- Doesn't duplicate the "main" version in the expanded list
- Shows combined application count when collapsed and per-version count when expanded


fergie-nz:
Issue with older (30/6/2023) snapshot in shared drive:

to reproduce:
-> Create a new version of an existing template.
-> Create a new version of this newly created and uncommitted template. Commit the previously created template when prompted.
-> Recei
DB437597-012F-4124-994B-6DED984CCE9E
ve duplicate key graphql error:

CarlosNZ:
I'm just going to add a few additional fix/improvement notes for myself here to do in this PR, feel free to add to it:

- [ ] When uploading a new snapshot, it should appear at top of list. Currently it appears in order according to its timestamp. So would probably be better to sort them by file modified time, rather than snapshot timestamp (if it's not too complicated)
- [ ] When selecting "Full archive" in New Snapshot, the Total size is either incorrect, or shows nonsense:
Screenshot 2023-08-08 at 10 57 41 AM

fergie-nz (approved):
Nice looks good!

PR #1546, opened 2023-08-14
Update docs for npm login

fergie-nz:
Not sure if this is in the right place? I wasn't sure where else this should go


CarlosNZ (approved):
That's great, thanks.

PR #1564, opened 2023-10-27
#1559 #1560 #1561 Document Modal part 1

CarlosNZ:
Fix #1559 & #1560 & #1561

Creates "Document Modal" and implements it in the following places:
- **Application view** (in forms)
- **Summary View** (for reviewer/summary page)
- **Database** views
- "**Notes**" tab (review page)
- "**Documents**" tab (review page)
- **Previews** (I know we said we would probably not do this one, but it actually looks pretty good and it was easy, so.._)

Still to do (might update PR, might do in new issues if substantial):

- [x] Downloaded file has wrong name
-  ~~Simple "Zoom" functionality~~ (don't think we need this)
- [x] Cache file so it doesn't take time to re-download a file we just uploaded

Note: you'll need to turn on the preference `showDocumentModal: true` in system prefs (web section) to get the modal functionality.

**Viewing a PDF:**

Screenshot 2023-10-27 at 2 35 38 PM

**Viewing an image file:**

Screenshot 2023-10-27 at 2 35 52 PM

**And when you try to view a file that is not supported (i.e. not pdf or image):**

Screenshot 2023-10-27 at 2 36 08 PM

See inline comments for further detail

PR #1565, opened 2023-10-29
#1561 DocumentModal -- other implementations

### LINKED ISSUE #1561: Document modal -- Implement in other places
Part of Epic #1549 

From the requirements doc, should be implemented in the following places (remembering to follow the global preference, as detailed in #1560)

- Application view / Summary view / Data Views (should already be done in #1559 )
- Documents tab (Review page)
- Notes tab (Review page)
- (Probably not) Preview modal -- it probably won't work to have a modal launched from a modal, but it's worth testing out to see how it works.

CarlosNZ:
Fix #1561 

Added the document modal for:
- "Notes" tab (review page)
- "Documents" tab (review page)
- Previews (I know we said we would probably not do this one, but it actually looks pretty good and it was easy, so.._)

Some other improvements:
- made a generic "SimpleCache" hook which can be used by both the "Notes" implementation and the previous PR's form elements. As explained in comments, this means that a user can view a document they've just added without having to re-download it (which can be quite slow with big docs on slower systems).
- a generic "File handler" for opening files, which is used in each implementation. It handles the logic of whether the modal should be used, considers the file type, and acts accordingly (i.e. either opens the modal, opens in new tab, or downloads immediately)

See inline comment for further detail.

### DIFF
--- a/src/components/common/DocumentModal/DocumentModal.tsx
+++ b/src/components/common/DocumentModal/DocumentModal.tsx
@@ -11,7 +11,7 @@ interface DocumentModalProps {
   filename: string
   url: string
   open: boolean
-  setOpen: (open: boolean) => void
+  onClose: () => void
   cachedFile?: File
 }
 
@@ -21,7 +21,7 @@ export const DocumentModal: React.FC<DocumentModalProps> = ({
   filename,
   url,
   open,
-  setOpen,
+  onClose,
   cachedFile,
 }: DocumentModalProps) => {
   const { t } = useLanguageProvider()
@@ -51,8 +51,7 @@ export const DocumentModal: React.FC<DocumentModalProps> = ({
     <Modal
       closeIcon
       open={open}
-      onOpen={() => setOpen(true)}
-      onClose={() => setOpen(false)}
+      onClose={() => onClose()}
       style={{
         width: fileType === 'pdf' ? contentWidth : 'unset',
         marginLeft: 15,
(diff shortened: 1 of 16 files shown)


src/components/common/DocumentModal/DocumentModal.tsx
```diff
@@ -11,8 +11,9 @@ interface DocumentModalProps {
   filename: string
   url: string
   open: boolean
-  setOpen: (open: boolean) => void
+  onClose: () => void
```
CarlosNZ, line 14: The modal only needs an "onClose" method, as opening is done by changing the state of `open` in the parent component.

src/components/common/DocumentModal/DocumentModal.tsx
```diff
@@ -63,7 +64,7 @@ export const DocumentModal: React.FC<DocumentModalProps> = ({
         <div className="flex-row-space-between-center-wrap" style={{ width: '95%', gap: 10 }}>
           <span style={{ maxWidth: 500 }}>{filename}</span>
           <div className="flex-row-start-center" style={{ gap: 5 }}>
-            {fileType === 'pdf' && (
+            {fileType === 'pdf' && numPages > 1 && (
```
CarlosNZ, line 67: Don't show page jump UI if there's only 1 page :)

src/components/common/DocumentModal/DocumentModal.tsx
```diff
@@ -82,7 +83,7 @@ export const DocumentModal: React.FC<DocumentModalProps> = ({
                 <span className="slightly-smaller-text"> of {numPages}</span>
               </div>
             )}
-            {fileType !== 'other' && (
+            {fileType !== 'other' && !preventDownload && (
```
CarlosNZ, line 86: `preventDownload` hides the option to download or display file in new tab. This is used for Previews, as we don't want preview files stored on users' computers for the most part.

CarlosNZ:
Closing PR and merging into #1564, as I've refactored some of that stuff here.

Issue #1554, opened 2023-10-19
Mobile layout -- Hamburger menu


CarlosNZ:
Part of epic #1548 

- Convert main menu bar to Hamburger menu (maybe [this one](https://github.com/natzar/Semantic-ui-hamburger-menu)?)
- Roll the "User" menu button into this hamburger (but only for mobile)
- Improve layout of company logo/name to fit nicer on mobile

CarlosNZ:
Closed by #1572

Issue #1553, opened 2023-10-19
Mobile layout -- smaller tasks


CarlosNZ:
Part of Epic #1548 

- [x] Login page: reduce padding on outer container so form box fills more of the screen:
![Image](https://github.com/openmsupply/conforma-web-app/assets/5456533/909b5941-6221-4126-a6ac-c082a5a341e7)
- [x] Data View tables -- check filters wrap correctly
- [x] #1581
- [x] ListBuilder modals -- add a close button (for mobile only) 
- [x] Listbuilder "Inline" view (accordion) -- problem with extending out of container (need to be careful with this, as sometimes they contain a table, and we've intentionally let it extend past the container in normal application view, so might need some re-thinking)
- [x] Pagination - check pagination wrap/fit correctly on mobile.

guatemartin:
Added task to check pagination works on mobile.

CarlosNZ:
>  * [x]  Data View tables -- check filters wrap correctly

I checked -- they're all good 👍

>  * [x]  Pagination - check pagination wrap/fit correctly on mobile.

Done in #1577

guatemartin:
Broke out ListBuilder into #1581 as they will need more detailed work.

CarlosNZ:
Closed by #1575

PR #1566, opened 2023-10-30
1555 table layout fix

### LINKED ISSUE #1555: Mobile layout -- table layout fix
Part of Epic #1548 

We need to make the tables collapse nicely on mobile. The original semantic ui components (https://react.semantic-ui.com/collections/table/) mostly collapse fine (except for the header bar), so we should be able to fix this. We've obviously busted something with our custom CSS, so hopefully, we can just tweak that to get it back to something reasonable.

Main table locations:
- Application List (`/applications`)
- Data Views (`/data/...`) (from Database menu)
- Table view of listbuilder elements

guatemartin:
fix #1555 

Basic fix for list table layout in mobile view.  Basically use semantic-ui defaults for mobile ui when it comes to tables.

PaginationBar, ListFilters, and MainMenuBar all need css adjusting to fit mobile.  Used 'hide-on-mobile' class on those components to test. Will need to raise issues separately for Pagination and ListFilters.  The Main User Menu will get fixed as part of #1554 .

Applications View Menu: (notice hidden menus for it to work)
table-layout-fix-mobile

Data view example: 
table-layout-fix-data-view

Normal Tablet view example:
table-layout-fix-tablet


guatemartin:
#1553 already has task to look at the List Filter wrapping.
Added task to above issue to look at pagination issues on mobile.

CarlosNZ:
Thanks @guatemartin, looks good. I'll take a proper look through it this afternoon.

semantic/src/site/globals/site.overrides
```diff
@@ -1728,60 +1726,6 @@ a:hover {
...
-      Data views
----------------------------------*/
-
-#data-view {
-  .flex-column();
-  align-items: center;
-  width: @applicationListWidth;
-
-  .ui.header {
-    padding-top: 30px;
```
CarlosNZ, line 1741: Makes sense to remove this and put it in table over-rides, but can you leave the header here and add a comment to see "Table" overrides file?

semantic/src/site/collections/table.overrides
```diff
@@ -2,89 +2,165 @@
...
+
+    #linked-applications-container {
+      margin-top: 45px;
+    }
   }
 }
+
+// mobile and small tablet overrides
+
+@media only screen and (max-width: @largestMobileScreen - 1) {
```
CarlosNZ, line 162: My linter is complaining about the `-` sign here. Should it be `max-width: calc(@largestMobileScreen - 1)`?

CarlosNZ (approved):
Cool, this is a pretty good start, thanks. (See couple of minor comments on code)

I've changed the base branch to `feature/angola-ui-improvments` cos I think it will be best if we merge quite quickly and then have another thorough look at everything before we merge it into `develop`.

I'm not sure what we need to do with those Table headers. I think we're probably going to need a small mobile-specific header component -- like just a dropdown for chosing the "sortBy:" column or something.

Within each table cell, I think *some* fields will need some kind of "header" label, otherwise it's not clear what they are describing. I have an idea for that, I'll make a new issue to describe what I'm thinking.

The first thing we'll need, though, is a Javascript global value so React knows whether we're in "mobile" mode or not. I'm thinking a simple hook. I made a "useWindowDimensions" hook for my [Modal PR](#1564), so we can probably just extend that to also return a "isMobile" value. I might go and pull that out into its own PR so we can merge it in with this stuff sooner.

We'll also need a bit of extra styling on these collapsed tables, but I think we can leave that till last.

semantic/src/site/collections/table.overrides
guatemartin: That is probably a lot tidier.  Implemented.

semantic/src/site/globals/site.overrides
guatemartin: Moved ui.header back to site.overrides.

guatemartin:
@CarlosNZ - All good to merge?

CarlosNZ:
> @CarlosNZ - All good to merge?

Yes please!

PR #1573, opened 2023-10-31
#1569 Header labels for mobile tables

### LINKED ISSUE #1569: Mobile specific header labels for Tables when in mobile view
Part of Epic #1548

When the tables are collapsed for Mobile, they look okay, but without header labels it's very difficult to make sense of what each bit of information is referring to, e.g.: 

Screenshot 2023-10-31 at 2 28 30 PM

I think we should make a utility component called "TableHeaderLabel", and we wrap each Table.Cell contents in it. Then, in that component we determine if we're in mobile or not (using #1568 ) and if so, pre-pend the "header" label to it. And if not, just return the children unmodified.

We probably don't want this for *all* table cells though, so it should have a "disabled" prop that can be dynamically applied (for Data Views), or we just don't wrap it for static ones (such as the "Status" label on the Application List)

CarlosNZ:
Fix #1569

Adds a wrapper to inject a label for table values when in mobile view. Does nothing when not in mobile. Options to turn off label and hide whole cell.

Changes from this:
Screenshot 2023-11-01 at 10 22 38 AM

to this:

Screenshot 2023-11-01 at 10 21 50 AM

Currently implemented for Application list and Data view tables. We can add elsewhere as required.


semantic/src/site/collections/table.overrides
```diff
@@ -109,7 +109,7 @@
   #data-view {
```
CarlosNZ, line 109: Changes here all just auto-formatting

src/components/List/ApplicationsList.tsx
```diff
@@ -4,9 +4,10 @@ import { useLanguageProvider } from '../../contexts/Localisation'
 import { useDeleteApplicationMutation } from '../../utils/generated/graphql'
 import { ApplicationListRow, ColumnDetails, SortQuery } from '../../utils/types'
 import Loading from '../Loading'
+import { TableCellMobileLabelWrapper } from '../../utils/tables/TableCellMobileLabelWrapper'
 
 interface ApplicationsListProps {
-  columns: Array<ColumnDetails>
+  columns: ColumnDetails[]
```
CarlosNZ, line 10: Same thing, preferred syntax.

src/containers/DataDisplay/DataViewTable.tsx
```diff
@@ -171,15 +172,24 @@ const DataViewTableContent: React.FC<DataViewTableContentProps> = ({
...
             >
               {rowValues.map((value: any, index: number) => (
                 <Table.Cell key={`value_${index}`}>
-                  {getCellComponent(value, headerRow[index], id)}
+                  <TableCellMobileLabelWrapper
+                    label={headerRow[index].title}
+                    // These values are not currently dynamic, hence no rowData
+                    // sent
+                    hideLabel={headerRow[index].formatting?.hideLabelOnMobile}
+                    hideCell={headerRow[index].formatting?.hideCellOnMobile}
```
CarlosNZ, line 188: These values added in Data View configuration in database. Have added additional note to back-end docs.

src/utils/helpers/list/mapColumnsByRole.ts
```diff
@@ -48,16 +50,19 @@ export const useMapColumnsByRole = () => {
       headerDetail: t('COLUMN_DEADLINE_DATE_TOOLTIP'),
       sortName: 'applicant-deadline',
       ColumnComponent: DeadlineCell,
+      hideOnMobileTest: (application) => !application.applicantDeadline,
```
CarlosNZ, line 53: Can be a function so it can be conditional based on the current rowData. e.g. in this case the cell will be hidden if there is no deadline. (Also accepts a boolean to universally hide the cell)

guatemartin (approved):
Looks good.  No comments at this stage..

PR #1576, opened 2023-11-01
#1570 Mobile table header/sort bar

### LINKED ISSUE #1570: Alternate table header element when in mobile
Part of Epic #1548

Currently, once the tables collapse down to "mobile" size, the header row collapses like this:

Screenshot 2023-10-31 at 2 37 38 PM

It's not very useful, as you can't easily see a connection between each header and the value it refers to. The only purposes it serves is to allow us to click on them to sort by that field. And we can add header labels to necessary table values using #1569.

I suggest we replace this header (when in mobile) with just a simple Drop-down where we can select the "Sort by" column.

CarlosNZ:
Fix #1570 

Removes the useless mobile header row and replaces it with a custom component to control sorting.

Changes from this:
Screenshot 2023-11-01 at 2 30 27 PM

to this:
Screenshot 2023-11-01 at 2 30 53 PM

Not sure about the best way to display the ascending/descending state, suggestions welcome.

Currently implemented for Application list and Data view tables.

### DIFF
--- a/src/utils/helpers/list/mapColumnsByRole.ts
+++ b/src/utils/helpers/list/mapColumnsByRole.ts
@@ -100,14 +100,14 @@ export const useMapColumnsByRole = () => {
     },
     REVIEWER_ACTION: {
       headerName: t('COLUMN_REVIEWER_ACTION'),
-      sortName: 'outcome',
+      sortName: '',
       ColumnComponent: ReviewerActionCell,
       hideMobileLabel: true,
       hideOnMobileTest: (application) => !application.reviewerAction,
     },
     APPLICANT_ACTION: {
       headerName: t('COLUMN_APPLICANT_ACTION'),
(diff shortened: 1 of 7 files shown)


src/utils/helpers/list/mapColumnsByRole.ts
guatemartin, line 1: What is the purpose of this change?

guatemartin (approved):
Only wandered what the one change was for in the columns mapping by role.
lgtm though.  Cheers.

src/utils/helpers/list/mapColumnsByRole.ts
CarlosNZ: That was a mistake, those columns had the wrong sort name. Because we don't actually show a header for those columns, I just made the "sortName" empty so that it gets filtered out by the drop-down.

PR #1578, opened 2023-11-01
1554(2) User menu tweaks

CarlosNZ:
Just the last couple of things I mentioned in the last PR -- the User Options menu bits.

I managed to re-use the contents of the original menu and just put them inside a different wrapper in the "UserMenu" component, which made is a bit simpler.

Screenshot 2023-11-01 at 11 51 15 PM

Still a few cosmetic things I'd like to tweak, but I'll make a list of minor tweaks in a later issue.

See inline comments.


src/containers/User/UserArea.tsx
```diff
@@ -312,7 +325,7 @@ const MainMenuBar: React.FC<MainMenuBarProps> = ({
             </Dropdown>
           </List.Item>
         )}
-        {intReferenceDocs.length && (
+        {intReferenceDocs.length > 0 && (
```
CarlosNZ, line 328: I don't really why this worked before, but these were showing up as a "0" in the UI when the arrays were empty. 🤷‍♂️

src/containers/User/UserArea.tsx
```diff
@@ -329,25 +342,13 @@ const MainMenuBar: React.FC<MainMenuBarProps> = ({
...
-                <Dropdown.Item icon="log out" text={t('MENU_LOGOUT')} onClick={() => logout()} />
-              </Dropdown.Menu>
-            </Dropdown>
+            <div className="ui divider menu-divider"></div>
+            <UserMenu
+              user={currentUser as User}
+              templates={templates.filter(({ templateCategory: { uiLocation } }) =>
+                uiLocation.includes(UiLocation.User)
+              )}
+            />
```
CarlosNZ, line 351: Same as in the standard Desktop menu

semantic/src/site/globals/site.overrides
```diff
@@ -2614,70 +2613,71 @@ h6:first-child {
...
-	    opacity: 1;
-	    -ms-filter: progid:DXImageTransform.Microsoft.Alpha(opacity=100);
-	    filter: alpha(opacity=100);
-	    -moz-transition-duration: .2s;
-	    -webkit-transition-duration: .2s;
-	    -o-transition-duration: .2s;
-	    transition-duration: .2s;
-	}
-	.hamburger.active {
-
```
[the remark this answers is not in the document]
guatemartin: Hmmm, I do have it turned on....  I'll see what is going on.

src/containers/User/UserArea.tsx
guatemartin: Yes, was having a bad day... Only had one language so option was not showing up in submenu of user options.

guatemartin (approved):
Thanks for cleaning that up.

PR #1582, opened 2023-11-02
1574 table mobile styles

guatemartin:
- Removed borders
- Less spacing between rows
- Long words will wrap and not break viewport
- Could possibly reduce even more gap between rows... matter of taste really.

Didn't reduce padding inside of rows.  I noticed the padding looks bigger on some applications than others.  I think it has to do with empty row data.
image

wrapping fix:
image

### DIFF
--- a/semantic/src/site/collections/table.overrides
+++ b/semantic/src/site/collections/table.overrides
@@ -1,10 +1,29 @@
 /*******************************
-         Site Overrides
+         General Site Overrides
 *******************************/
 
 .flex-column {
   display: flex;
   flex-direction: column;
+}
+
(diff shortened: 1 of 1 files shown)


CarlosNZ (approved):
Very nice, thanks :)

This made me notice something I didn't quite get right yesterday, so I've added a quick tweak on top of this: https://github.com/openmsupply/conforma-web-app/pull/1584

CarlosNZ:
> I noticed the padding looks bigger on some applications than others. I think it has to do with empty row data.

Yup! https://github.com/openmsupply/conforma-web-app/pull/1584

PR #1583, opened 2023-11-02
#1556 Help text in Application view

### LINKED ISSUE #1556: Mobile layout -- Help text in application view
Part of Epic #1548 

Currently, they're just hidden on mobile. Current plan is to display them *below* each element
- smaller text
- italics
- keep the little blue left border

CarlosNZ:
Fix #1556

Standard Desktop view:
Screenshot 2023-11-02 at 7 37 59 PM

Mobile view:
Screenshot 2023-11-02 at 7 38 23 PM

See inline comments for detail...


semantic/src/site/globals/site.overrides
```diff
@@ -1172,6 +1172,18 @@ a:hover {
...
+  @media only screen and (max-width: @largestMobileScreen) {
+    margin-top: -8px;
+    margin-bottom: 15px;
+    padding-left: 15px;
+    padding-right: 20px;
+    p {
+      font-size: 90%;
+      color: @headersInteractiveLow;
+    }
+  }
```
CarlosNZ, line 1186: Just subtle changes to the dimensions on mobile

semantic/src/site/globals/site.overrides
```diff
@@ -1290,7 +1306,9 @@ a:hover {
 #app-navigation-content {
   .flex-row();
   flex-wrap: wrap;
-  width: @applicationFormContentWidth;
+  max-width: @applicationFormContentWidth;
+  justify-content: space-between;
+  gap: 10px;
```
CarlosNZ, line 1311: Not directly related to issue, but this just makes the Previous/Next and Submit navigation at the bottom wrap a bit nicer when the viewport gets tight.

src/containers/Application/ApplicationPage.tsx
```diff
@@ -69,16 +71,18 @@ const ApplicationPage: React.FC<ApplicationProps> = ({
...
-          <Grid.Column width={4} id="progress-column" className="dev-border">
-            <ProgressArea
-              structure={fullStructure}
-              requestRevalidation={requestRevalidation as MethodRevalidate}
-              strictSectionPage={strictSectionPage as SectionAndPage}
-              context={contextRef}
-            />
-          </Grid.Column>
-        </Ref>
+        {!isMobile && (
```
CarlosNZ, line 74: This was already being hidden using the `hide-on-mobile` class, but I figured it's probably better to prevent rendering of the entire component on mobile, as it's quite a complex one.

(Am I correct in thinking that if it's hidden with a class then React will still render it?)

guatemartin: Yes, it still gets rendered but with display=none.  So that is better really.

guatemartin (approved):
Looks good.

Issue #1574, opened 2023-10-31
Table -- mobile styles more like the Desktop view


CarlosNZ:
Part of Epic #1548

Okay, so we have the tables working okay in a "stacked" view. However, the styles are just bare-bones and not that pretty (black boxes). Would be good if we could add a few of the table styles back to the mobile media query, so they look similar. (Don't have to be exactly the same, and probably should be a bit more "compact")

Current Desktop view:

![Image](https://github.com/openmsupply/conforma-web-app/assets/5456533/1964b6ac-d180-4b1b-b99d-8adc3df134da)

Current Mobile view:

![Image](https://github.com/openmsupply/conforma-web-app/assets/5456533/1a658bca-c27d-44eb-acc6-4abd73da0aa1)

So I'm thinking:

- lose the black border
- slight spacing between rows
- rounded corners
- less padding inside cell
- cell rows closer together

Also...
- The mobile tables still expand past the viewport when there is really long text content in them:
<img width="441" alt="Screenshot 2023-11-02 at 12 16 02 AM" src="https://github.com/openmsupply/conforma-web-app/assets/5456533/a6172d2e-08f8-4852-96d4-6f07addd67b4">

CarlosNZ:
Closed by #1582

PR #1584, opened 2023-11-02
#1574(2) Table mobile extra

CarlosNZ:
This is little something I didn't quite get right yesterday. When a table cell needs to be hidden, we need to hide the whole cell, but the mobile component I made was *inside* the cell, so the cell itself was still there.

So it meant we got an empty "row", but it still took up height, e.g.
Screenshot 2023-11-02 at 9 01 23 PM

So I've removed the "hide" check from the Mobile wrapper component and just done it separately outside the cell. Not quite as "nice", but works as expected:

Screenshot 2023-11-02 at 9 00 12 PM


guatemartin (approved):
Thought it might affect sorting... but of course it doesn't affect the underlying data.
Looks good.

PR #1589, opened 2023-11-03
#1581 ListBuilder improvements

### LINKED ISSUE #1581: ListBuilders -- check the more complex ones and make sure they look okay
ListBuilders have quite a bit going on in them.  None of them scale well into mobile.  Needs more discussion of how mobile layout should look and function.  Should they even be in mobile?
Bringing tasks over from #1553 to complete with the rest of these.

- [x] Check that the list builders scale down for mobile
- [x] ListBuilder modals -- add a close button (for mobile only)
- [x] Listbuilder "Inline" view (accordion) -- problem with extending out of container (need to be careful with this, as sometimes they contain a table, and we've intentionally let it extend past the container in normal application view, so might need some re-thinking)

CarlosNZ:
Fix #1581 

Okay, I think this takes care of most of the list builder tweaks required to get it working nicely. Although there are so many possible cases that I'm sure we'll keep running into things that don't look right -- but they're working okay for the ones I've tried so far.

You can try with any of the list builder elements in the Fiji templates, but a more comprehensive one would be the nested Formulations lists in Angola "P1-ProdRegistrationLong" template. I've put the latest version of the snapshot in the Google drive folder -- not sure if I've given you access to that yet @guatemartin, so let me know if you need it.

The file is: `Angola Conforma/System Configuration (Templates, Certificates, Lists)/Snapshots/Angola_Phase_3_2023-11-03_11-16-31.zip`

If you log in as an external user (try "fergusExt"), you can start of the above applications. The most complex, nested listbuilder is in "Modulo 1" (Section 1), Page 6 -- the "Formulações" element, should look something like this when filled in:

Screenshot 2023-11-03 at 11 32 57 PM

There's several other listbuilder elements in that form (couple more on Page 5), so have a play with them (Desktop and Mobile) and see if anything breaks!

See inline comments for further explanation.


docs/internal/Element-Type-Specs.md
```diff
@@ -534,7 +534,11 @@ _Allows user to build a list of items, such as an **Ingredients List**_
...
+*The following properties only apply to **Table** view and only when viewing on a Mobile display. When displaying on a mobile device, tables get "collapsed" so that each row becomes a "cell", with a stack of labels and values. This makes it possible to view table data on a narrow display without arduous side-scrolling.*
+- **hideFromMobileIfEmpty** `boolean | string[]` -- to save space, if a particular field has no value, we can hide it completely. Put the `code` of any input field that should be treated like this into an array. Or, if the parameter is just set to `true`, the rule will be applied to *all* fields (default `false`)
+- **minMobileLabelWidth** `number | string` -- set a minimum width for the labels within the mobile table. This can help with alignment the values visually. (Can be any valid CSS value, e.g. `150`, `"200px"`, `"50%"`)
+- **maxMobileLabelWidth** `number | string` -- same as above, but a max label width. Labels will wrap onto a new line if they exceed this width.
```
CarlosNZ, line 541: Additional parameters that allow us a bit more control over how the tables look on mobile.

semantic/src/definitions/modules/accordion.less
```diff
@@ -118,9 +118,9 @@
      Styled
 ---------------*/
 
-.ui.styled.accordion {
-  width: @styledWidth;
-}
+// .ui.styled.accordion {
+//   width: @styledWidth;
+// }
```
CarlosNZ, line 123: I really didn't want to have to edit this file (we should be using just the "overrides" file). But I couldn't find a way to get rid of the "width" property (I tried `unset` and `revert` but neither worked properly). 

So we get rid of the fixed width for the accordions and rely on a max-width based on context.

src/formElementPlugins/listBuilder/src/ApplicationView.tsx
```diff
@@ -132,7 +142,7 @@ const ApplicationView: React.FC<ApplicationViewProps> = ({
 
   const editItem = async (index: number, openPanel = true) => {
     setInputState((prev) => ({
-      ...inputState,
+      ...prev,
```
CarlosNZ, line 145: Small unrelated fix.

src/formElementPlugins/listBuilder/src/displayComponents/ListTableLayout.tsx
```diff
@@ -1,35 +1,61 @@
...
   return (
-    <Table celled selectable={isEditable}>
-      <Table.Header>
-        <Table.Row>
-          {displayFields.map(({ title }) => (
-            <TableHeaderCell key={`list-header-field-${title}`}>{title}</TableHeaderCell>
-          ))}
-        </Table.Row>
-      </Table.Header>
+    <Table celled={!isMobile} stackable selectable={isEditable}>
```
CarlosNZ, line 31: Same basic idea as what I did on the other tables -- hide the header, wrap the cells in a "Labeller", and handle optional "hidden" fields.

src/formElementPlugins/listBuilder/src/types.ts
```diff
@@ -8,6 +8,26 @@ export enum DisplayType {
...
+  displayFormat?: { title: string; subtitle: string; description: string }
+  displayType?: DisplayType
+  default?: ResponseFull
+  inlineOpen?: boolean
+  tableExcludeColumns?: string[]
+  // These affect mobile viewing only
+  hideFromMobileTableIfEmpty?: true | string[]
+  minMobileTableLabelWidth?: number | string
+  maxMobileTableLabelWidth?: number | string
+}
```
CarlosNZ, line 29: We were pretty sloppy with our use of Typescript in the early days (lots of `any` around the place), but I'm slowly but surely properly defining types where I can going forward.

semantic/src/definitions/modules/accordion.less
guatemartin: I also found I couldn't 'override' modules the other day...  I had a quick look at how to fix, but decided to move on.

semantic/src/site/collections/table.overrides
```diff
@@ -191,4 +189,40 @@
...
+      border-radius: 0.3em;
+    }
+
+    .ui.table tr td {
+      padding: 1px !important;
+    }
+
+    // .ui.table td p {
+    //   word-wrap: anywhere;
+    // }
```
guatemartin, line 226: Can these just be deleted?

CarlosNZ: Yup, I was wondering if we would need this but didn't have any appropriate data to test it on. Have deleted now, we can add something back in later if we need it. d8ba870f

src/formElementPlugins/listBuilder/src/ApplicationView.tsx
guatemartin, line 1: I can't see exactly where, but there might be a bug here.  It may not be related directly to the code that's changed but...
If I click on "Documentation" in the dropdown here, it throws an error..
Might need to be looked at outside of this PR.

image

Error: 
image

CarlosNZ: > I can't see exactly where, but there might be a bug here. It may not be related directly to the code that's changed but... If I click on "Documentation" in the dropdown here, it throws an error.. Might need to be looked at outside of this PR.

Thanks for spotting this. Can you tell me which template this was for?

CarlosNZ: Okay, I see it. That's a weird one -- there isn't even supposed to be a "Documentation" section for that application type. I'll take a look.

src/formElementPlugins/listBuilder/src/SummaryView.tsx
guatemartin, line 1: We still have a bit of a text wrapping issue on the .ui .card classes

image

guatemartin (commented):
Still struggling to find my way around the app.
There are a few things to look at though.  Comments in the fild changes.

src/formElementPlugins/listBuilder/src/ApplicationView.tsx
CarlosNZ: Okay, I see the problem with this one. There is a "Documentation" section that has been created, but there's nothing in it (no pages). Just need to delete that section and it'll be fine.
Screenshot 2023-11-06 at 11 28 08 AM

guatemartin: Ahh, maybe me messing around with stuff!

src/formElementPlugins/listBuilder/src/SummaryView.tsx
CarlosNZ: Ah right. Yeah, I didn't actually check the cards very well, was mostly focused on the tables.

I've fixed that now, and I've also reduced the padding on most of these layers in this Summary view:

Screenshot 2023-11-06 at 11 50 36 AM

f884b554

guatemartin:
Good to go.

src/formElementPlugins/listBuilder/src/ApplicationView.tsx
CarlosNZ: Nah, I don't think it was you, cos I haven't loaded any snapshots that you've made changes on. Looks like a residual problem.

Issue #1590, opened 2023-11-04
CSS overrides in prefs


CarlosNZ:
I've figured out a way to do this, so anything in the app can be styled in preferences :)

CarlosNZ:
Closed by #1591

PR #1591, opened 2023-11-04
#1590 CSS styles override in preferences

CarlosNZ:
This was quite easy in the end. Just had to figure out how to inject a CSS text string into the head of the document.

You need to run `yarn install` to install one small additional package (see comments).

To get the Angola purple header bar, change the `"style"` value in preferences (Configuration -> System Preferences) to the following actual CSS (instead of our hacky one-off solution)

Screenshot 2023-11-04 at 5 03 29 PM

It'd be good to do something proper with Semantic UI so we could override all the Semantic global variables, but that's quite a bit more complicated, so this will do for now.

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,6 +29,7 @@
     "dotenv": "^16.0.1",
     "graphql": "^15.3.0",
     "json-edit-react": "^0.9.6",
+    "json-to-css": "^0.1.0",
     "lodash": "^4.17.21",
     "luxon": "^1.27.0",
     "nanoid": "^3.1.22",
--- a/semantic/src/site/globals/site.overrides
+++ b/semantic/src/site/globals/site.overrides
@@ -331,6 +331,7 @@ a:hover {
   display: flex;
   flex-direction: column;
   flex-grow: 1;
+  background-color: inherit;
 }
 
 // Menu bar links
@@ -2632,12 +2633,13 @@ h6:first-child {
 
   #menu-bar {
     font-size: inherit;
+    background-color: inherit;
     .list {
       z-index: 2000;
       position: absolute;
       top: @mobileHeaderHeight;
       left: 0px;
-      background: @darkGrey;
+      background-color: inherit;
       width: 95%;
       padding: 5px 30px 20px 25px;
     }
--- a/src/containers/User/UserArea.tsx
+++ b/src/containers/User/UserArea.tsx
@@ -46,12 +46,7 @@ const UserArea: React.FC = () => {
   if (!currentUser || currentUser?.username === config.nonRegisteredUser) return null
 
   return (
-    <Container
-      id="user-area"
-      fluid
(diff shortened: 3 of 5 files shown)


semantic/src/site/globals/site.overrides
```diff
@@ -2632,12 +2633,13 @@ h6:first-child {
...
   #menu-bar {
     font-size: inherit;
+    background-color: inherit;
     .list {
       z-index: 2000;
       position: absolute;
       top: @mobileHeaderHeight;
       left: 0px;
-      background: @darkGrey;
+      background-color: inherit;
```
CarlosNZ, line 2642: Just had to make this `inherit` all the way down from "user-area" to make sure it has the same colour as the main header.

package.json
```diff
@@ -29,6 +29,7 @@
     "dotenv": "^16.0.1",
     "graphql": "^15.3.0",
     "json-edit-react": "^0.9.6",
+    "json-to-css": "^0.1.0",
```
CarlosNZ, line 32: This package just converts CSS defined in an object to a string.
Eg.
```js
{
  "#user-area": {
    "background-color": "#9c85c4"
  }
}
```
becomes `"#user-area{background-color:#9c85c4}"`, which is what gets injected into the `<head>`

It's a pretty tiny package -- only a handful of lines of code.

src/containers/User/UserArea.tsx
```diff
@@ -46,12 +46,7 @@ const UserArea: React.FC = () => {
   if (!currentUser || currentUser?.username === config.nonRegisteredUser) return null
 
   return (
-    <Container
-      id="user-area"
-      fluid
-      // Custom color option (for Angola only currently)
-      style={{ backgroundColor: preferences?.style?.headerBgColor ?? '' }}
```
CarlosNZ, line 53: Gone! 💨

src/contexts/SystemPrefs.tsx
```diff
@@ -46,6 +48,12 @@ export const SystemPrefsProvider = ({ children }: { children: React.ReactNode })
           latestSnapshot,
           loading: false,
         })
+
+        // Install custom CSS into document head
+        if (preferences.style) {
+          const cssString = Css.of(preferences.style)
+          document.head.appendChild(document.createElement('style')).innerHTML = cssString
+        }
```
CarlosNZ, line 56: This is where the magic happens (found after a decent amount of googling) 😉

guatemartin:
LGTM.  We wouldn't be using this all over the place would we?

guatemartin (approved):

CarlosNZ:
> We wouldn't be using this all over the place would we?

No, it's just that Angola have asked for some specific styling options. I've resisted it as much as possible, and we've had that hacky one-off solution for the header bar, but with this we can control anything if we want (but I'd rather avoid it as much as possible).

Issue #1586, opened 2023-11-02
Mobile Layout -- Header bar tweaks


CarlosNZ:
Part of Epic #1548

~~Just need to make the logo and the company name a bit nicer on mobile. Should probably be smaller overall, perhaps just a narrow bar since we have the Hamburger for the menu items now.~~ Fixed by #1592 

Also: the organisation switcher menu goes off the screen:
<img width="541" alt="Screenshot 2023-11-04 at 12 22 22 AM" src="https://github.com/openmsupply/conforma-web-app/assets/5456533/ef940df0-9180-43d5-a1be-449217f5c341">

CarlosNZ:
Closed by #1593

PR #1592, opened 2023-11-05
#1586 Header tweaks

CarlosNZ:
Partly addresses #1586, but doesn't deal with the org dropdown.

Basically, makes things flow better around each other for both mobile and desktop at a wider range of viewport sizes:

### Desktop
- UserMenu button shouldn't wrap, and will expand to fit the name (up to a limit)
- Menu bar shouldn't force org name out of the header (used "position:absolute" for it)
- "Brand Area" logo disappears earlier as the viewport shrinkg

### Mobile
- Header bar size reduced (from 80px to 60)
- Org name fits better into space, and font-size reduced
- Make some margins/padding smaller

### DIFF
--- a/semantic/src/site/globals/site.overrides
+++ b/semantic/src/site/globals/site.overrides
@@ -2654,6 +2654,9 @@ h6:first-child {
       font-size: 100%;
       letter-spacing: unset;
       line-height: 1.3em;
+      #org-label {
+        max-width: 700px;
+      }
     }
     #org-selector .image {
       margin-right: 10px;
--- a/semantic/src/site/globals/site.overrides
+++ b/semantic/src/site/globals/site.overrides
@@ -2644,15 +2644,19 @@ h6:first-child {
 
   #user-area {
     height: @mobileHeaderHeight;
+    padding-left: 10px;
   }
 
   #user-area-left {
     max-width: 100%;
-    padding-right: 20px;
+    padding-right: 25px;
     #org-selector {
-      font-size: 85%;
+      font-size: 100%;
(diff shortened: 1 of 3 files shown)


semantic/src/site/globals/site.overrides
```diff
@@ -325,12 +321,16 @@ a:hover {
     padding-top: 8px;
     padding-bottom: 8px;
   }
+  @media only screen and (max-width: 860px) {
```
CarlosNZ, line 324: The 860 is just an arbitrary amount based on what seemed reasonable on my screen.

src/containers/User/UserArea.tsx
```diff
@@ -490,7 +490,20 @@ const UserMenu: React.FC<{ user: User; templates: TemplateInList[] }> = ({ user,
...
+                    >
+                      <span style={{ fontSize: '150%', marginRight: 5 }}>
+                        {selectedLanguage?.flag}
+                      </span>
+                      <span style={{ overflow: 'hidden', textOverflow: 'ellipsis' }}>{`${
+                        user?.firstName || ''
+                      } ${user?.lastName || ''}`}</span>
+                    </div>
+                  ) as any
+                }
```
CarlosNZ, line 506: This is a bit of a hack -- the `text` prop isn't supposed to take JSX, but it works as long as I force Typescript to ignore the type. 

Lot of hassle just to make the flag icon bigger, but I think it's worth it.

guatemartin (approved):
LGTM.  It tidied up with a really long name as well.
image

Issue #1579, opened 2023-11-01
Hamburger menu -- remaining tweaks


CarlosNZ:
These are all pretty much cosmetic, and are just suggestions -- might need a little discussion.

This is how it currently stands:

<img width="401" alt="Screenshot 2023-11-01 at 11 51 15 PM" src="https://github.com/openmsupply/conforma-web-app/assets/5456533/4fa07f8f-a48b-41da-83c3-c8a35a9e9e62">

- [x] ~~Doesn't need to be as wide, can just contain the list, that should be enough.~~ As discussed in meeting, make it full width (it's currently doesn't quite go all the way over)
- [x] Be good if it animated (slide) in and out. ~~Maybe we should wrap it in a Semantic [Sidebar](https://react.semantic-ui.com/modules/sidebar/) component?~~
- [x] ~~Should it be on the right-hand side, to align with the Hamburger icon? (I realise this might make the sub-menus over-complicated as they would go off screen by default, so might not be worth the hassle)~~
- [x] Slightly wider spacing between items in list
- [x] The drop-downs that pop up seem a bit "big" in this context. Can we lower the font-size perhaps for the drop-downs/sub-menus? The font-size and spacing of the sub-menus should match the font-size and spacing of the main list. (Done in #1593)
- [x] Should we hide the "Configuration" (and maybe the "Manage") menu completely on Mobile?
- [x]  #1590 PR: #1591 Need to change the background colour to use the same constant as the header bar (as it's configurable in prefs):
<img width="329" alt="Screenshot 2023-11-02 at 4 39 42 PM" src="https://github.com/openmsupply/conforma-web-app/assets/5456533/8119f691-fd70-4493-aaf0-25c8bed03f9c">

CarlosNZ:
Closed by #1595

PR #1593, opened 2023-11-06
#1586 Other header tweaks

### LINKED ISSUE #1586: Mobile Layout -- Header bar tweaks

CarlosNZ:
Fix #1586 

Main purpose is to fix the overly wide organisation menu, from this:
Screenshot 2023-11-06 at 3 55 59 PM

to this:
Screenshot 2023-11-06 at 3 56 22 PM

So I also dropped the text and item size down a bit (as listed in #1579)

### DIFF
--- a/src/containers/User/UserArea.tsx
+++ b/src/containers/User/UserArea.tsx
@@ -409,7 +409,13 @@ const OrgSelector: React.FC<{ user: User; orgs: OrganisationSimple[]; onLogin: F
   if (!orgs.some(({ isSystemOrg }) => isSystemOrg))
     dropdownOptions.push({
       key: LOGIN_AS_NO_ORG,
-      text: `> ${t('LABEL_NO_ORG_SELECT')}`,
+      text: (
+        <span>
+          <em>{t('LABEL_NO_ORG')}</em>
+          <br />
+          <em>{t('LABEL_USER_ONLY')}</em>
+        </span>
+      ) as any,
       value: LOGIN_AS_NO_ORG,
     })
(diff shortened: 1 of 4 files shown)


src/containers/User/UserArea.tsx
```diff
@@ -409,7 +409,13 @@ const OrgSelector: React.FC<{ user: User; orgs: OrganisationSimple[]; onLogin: F
...
     dropdownOptions.push({
       key: LOGIN_AS_NO_ORG,
-      text: `> ${t('LABEL_NO_ORG_SELECT')}`,
+      text: (
+        <span>
+          <em>{t('LABEL_NO_ORG')}</em>
+          <br />
+          <em>{t('LABEL_USER_ONLY')}</em>
+        </span>
+      ) as any,
```
CarlosNZ, line 418: Same trick as before -- have to pass HTML to put a line break in. It works but Typescript doesn't think it should.

guatemartin (approved):
All looks good.

CarlosNZ:
Thanks @guatemartin . Having played with this on my phone a little bit, I think I've gone and made the font-size and clickable areas a little small. So I'll quickly fix that before I merge.

PR #1595, opened 2023-11-06
#1579 Hamburger tweaks

### LINKED ISSUE #1579: Hamburger menu -- remaining tweaks

CarlosNZ:
Fix #1579 

Minor tweaks to the hamburger menu

### DIFF
--- a/semantic/src/site/globals/site.overrides
+++ b/semantic/src/site/globals/site.overrides
@@ -2696,7 +2696,9 @@ h6:first-child {
     top: @mobileHeaderHeight;
     left: 0px;
     width: 100%;
+    background-color: inherit;
     .list {
+      background-color: inherit;
       z-index: 2000;
       background: @darkGrey;
       padding: 5px 30px 10px 25px;
--- a/semantic/src/site/globals/site.overrides
+++ b/semantic/src/site/globals/site.overrides
@@ -1835,6 +1835,10 @@ a:hover {
 #data-view-config-panel {
   padding-top: 20px;
   width: 50%;
+  @media only screen and (max-width: @largestMobileScreen) {
+    width: unset;
+    max-width: 100%;
+  }
 
   h2.ui.header {
     padding: 0;
--- a/semantic/src/site/globals/site.overrides
+++ b/semantic/src/site/globals/site.overrides
@@ -2688,14 +2688,18 @@ h6:first-child {
 
   #menu-bar {
     font-size: inherit;
+    position: absolute;
+    top: @mobileHeaderHeight;
+    left: 0px;
+    width: 100%;
     .list {
       z-index: 2000;
-      position: absolute;
-      top: @mobileHeaderHeight;
-      left: 0px;
       background: @darkGrey;
-      width: 95%;
-      padding: 5px 30px 20px 25px;
+      padding: 5px 30px 10px 25px;
+    }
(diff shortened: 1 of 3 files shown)


semantic/src/site/globals/site.overrides
```diff
@@ -1835,6 +1835,10 @@ a:hover {
 #data-view-config-panel {
   padding-top: 20px;
   width: 50%;
+  @media only screen and (max-width: @largestMobileScreen) {
+    width: unset;
+    max-width: 100%;
+  }
```
CarlosNZ, line 1841: This just makes the Data Views and System Preferences config pages appear slightly better on mobile. Not strictly necessary, but was simple enough.

semantic/src/site/globals/site.overrides
```diff
@@ -2688,14 +2692,20 @@ h6:first-child {
 
   #menu-bar {
     font-size: inherit;
+    position: absolute;
+    top: @mobileHeaderHeight;
+    left: 0px;
+    width: 100%;
+    background-color: inherit;
```
CarlosNZ, line 2699: The `inherit` makes the list take the same color as the main header bar (doesn't work properly here, but does in #1591

semantic/src/site/globals/site.overrides
```diff
@@ -2723,6 +2733,7 @@ h6:first-child {
     height: 3px;
     background: @userAreaText;
     margin: 0 0 3px;
+    transition-duration: 0.1s;
```
CarlosNZ, line 2736: Hamburger X animates out too

src/components/Admin/AdminPreferences.tsx
```diff
@@ -66,6 +68,7 @@ export const AdminPreferences: React.FC = () => {
           maxWidth={650}
           restrictDelete={({ level }) => level === 1}
           restrictAdd={({ level }) => level === 0}
+          indent={isMobile ? 2 : 4}
```
CarlosNZ, line 71: Again, not really necessary, but simple enough.

guatemartin (approved):
Looks good.

PR #1600, opened 2023-11-06
1558 review page and tabs

guatemartin:
Have tried to use Semantic as much as possible to do the heavy lifting.  Hopefully I haven't disturbed the desktop portion too much by using Grid columns in some places and leveraging the Stackable property of a grid.

1. Removed ProgressBar for mobile.  Fits well in simple 2 step processes, but in the more complicated it goes awry.
2. Removed Icon in the Review->Overview Tab.   Didn't seem necessary to provide in mobile
3. Tab headers wrap.  Thought about drop down as it might be cleaner looking, but tabs are easier to navigate.
4. Checked assignment tab looks ok with multiple stages.

### DIFF
--- a/semantic/src/site/collections/menu.overrides
+++ b/semantic/src/site/collections/menu.overrides
@@ -6,6 +6,7 @@
   // For review page tabs when mobile, but probably good site wide
   .ui.attached.tabular.menu {
     flex-wrap: wrap;
+    align-items: flex-end;
   }
 
   .ui.tabular.menu .item {
--- a/semantic/src/site/collections/menu.overrides
+++ b/semantic/src/site/collections/menu.overrides
@@ -2,7 +2,7 @@
          Site Overrides
 *******************************/
 
-@media only screen and (min-width: @largestMobileScreen) {
+@media only screen and (max-width: @largestMobileScreen) {
   // For review page tabs when mobile, but probably good site wide
   .ui.attached.tabular.menu {
     flex-wrap: wrap;
--- a/semantic/src/site/collections/menu.overrides
+++ b/semantic/src/site/collections/menu.overrides
@@ -1,3 +1,10 @@
 /*******************************
          Site Overrides
 *******************************/
+
+@media only screen and (min-width: @largestMobileScreen) {
+  // For review page tabs when mobile, but probably good site wide
+  .ui.attached.tabular.menu {
+    flex-wrap: wrap;
+  }
+}
--- a/semantic/src/site/collections/menu.overrides
+++ b/semantic/src/site/collections/menu.overrides
@@ -7,4 +7,14 @@
   .ui.attached.tabular.menu {
     flex-wrap: wrap;
   }
+
+  .ui.tabular.menu .item {
+    background-color: #cacaca24;
+    border-left: 0.5px solid @tabularBorderColor;
+    border-right: 0.5px solid @tabularBorderColor;
+    border-top: 1px solid @tabularBorderColor;
+    border-top-left-radius: @borderRadius;
+    border-top-right-radius: @borderRadius;
+  }
 }
+
--- a/semantic/src/site/collections/menu.overrides
+++ b/semantic/src/site/collections/menu.overrides
@@ -5,8 +5,9 @@
 @media only screen and (max-width: @largestMobileScreen) {
   // For review page tabs when mobile, but probably good site wide
   .ui.attached.tabular.menu {
-    flex-wrap: wrap;
-    align-items: flex-end;
+    flex-wrap: wrap-reverse;
+    flex-direction: row-reverse;
+    justify-content: flex-end;
   }
 
   .ui.tabular.menu .item {
--- a/semantic/src/site/collections/message.overrides
+++ b/semantic/src/site/collections/message.overrides
@@ -43,4 +43,7 @@
   .ui.message {
     padding: 0.5em 0.5em;
   }
+  .ui.info.message {
+    border-left: unset;
+  }
 }
--- a/semantic/src/site/collections/message.overrides
+++ b/semantic/src/site/collections/message.overrides
@@ -38,3 +38,9 @@
 .ui.error.message h6 {
   color: @errorHeaderColor;
 }
+
+@media only screen and (max-width: @largestMobileScreen) {
+  .ui.message {
+    padding: 0.5em 0.5em;
+  }
+}
--- a/semantic/src/site/globals/site.overrides
+++ b/semantic/src/site/globals/site.overrides
@@ -2780,7 +2780,14 @@ h6:first-child {
     }
     .stage-label {
       font-size: 0.57rem;
-    }
+    }    
+  }
+  #overview .left-item {
+    min-width: unset;
+  }
+  #timeline {
+    margin-left: 8px;
+    margin-right: 8px;
   }
   
   #documents-tab {
--- a/semantic/src/site/globals/site.overrides
+++ b/semantic/src/site/globals/site.overrides
@@ -2768,4 +2768,8 @@ h6:first-child {
       font-size: 0.57rem;
     }    
   }
+
+  #notes-tab .options-row {
+    flex-direction: column;    
+  }
 }
--- a/semantic/src/site/globals/site.overrides
+++ b/semantic/src/site/globals/site.overrides
@@ -2779,7 +2779,7 @@ h6:first-child {
       padding: 0.5em 0.5em;
     }
     .stage-label {
-      font-size: 0.57rem;
+      font-size: 0.7rem;
     }    
   }
   #overview .left-item {
@@ -2816,5 +2816,8 @@ h6:first-child {
     .assigning-row > .column {
       padding: 0.5rem 0.5rem !important;
     }
+    .stage-label {
+      margin-left: 3px;
+    } 
   }
 }
--- a/semantic/src/site/globals/site.overrides
+++ b/semantic/src/site/globals/site.overrides
@@ -894,8 +894,7 @@ a:hover {
   }
 
   .ui.image {
-    max-height: 100px;
-    // border: 1px solid @surfacesLight;
+    max-height: 100px;    
   }
 
   .ui.list {
@@ -2781,6 +2780,23 @@ h6:first-child {
     }
     .stage-label {
       font-size: 0.57rem;
+    }
+  }
+  
+  #documents-tab {
+    //smaller icon in Review documents tab
+    .ui.image {
+      max-height: 60px;
+    }
+    .flex-extra {
+      gap: 5px;
+    }
+    .ui.list {
+      margin-top: 0px;
+    }
+    .ui.list > .item {
+      margin-top: 10px;
+      padding: 10px;
     }
   }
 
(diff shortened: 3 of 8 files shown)


semantic/src/site/globals/site.overrides
```diff
@@ -906,8 +906,7 @@ a:hover {
     background-color: @surfacesLight;
     padding: 15px;
     border-radius: @borderRadius;
-    border: 1px solid @surfacesMed;
-    min-width: 400px;
+    border: 1px solid @surfacesMed;    
     h4 {
```
guatemartin, line 910: Removed this min-width site wide.  Couldn't see it being of any benefit anywhere, and it would stuff up the mobile rendering.

semantic/src/site/globals/site.overrides
```diff
@@ -930,6 +929,7 @@ a:hover {
       margin-bottom: 5px;
       .ui.label {
         line-height: 1.5;
+        font-size: 0.57;
       }
       .centered-flex-box-row {
         display: flex;
```
guatemartin, line 935: Why 0.57?  That's just what seemed to look best.

src/containers/Review/assignment/ReviewSectionRow.tsx
```diff
@@ -87,13 +87,19 @@ const ReviewSectionRow: React.FC<ReviewSectionRowProps> = ({
...
             <ReviewSectionRowLastActionDate {...props} />
+          </Grid.Column>
+          <Grid.Column>
             <ReviewSectionRowProgress {...props} />
+          </Grid.Column>
+          <Grid.Column>
             <ReviewSectionRowAction {...props} />
-          </Grid.Row>
+          </Grid.Column>
         </Grid>
```
guatemartin, line 103: I thought I would try to use Semantic's Grid Stackable property if possible.  If it's not what we want, I can find another way.

semantic/src/site/collections/menu.overrides
```diff
@@ -1,3 +1,10 @@
...
-*******************************/
\ No newline at end of file
+*******************************/
+
+@media only screen and (max-width: @largestMobileScreen) {
+  // For review page tabs when mobile, but probably good site wide
+  .ui.attached.tabular.menu {
+    flex-wrap: wrap;
+  }
+}
```
CarlosNZ, line 10: Good to be able to wrap the tabs. However, the upper ones look a little weird just "floating" above the rest of the row and not attached to the container beneath.

Not sure what the solution is. It's a slight improvement if you add the border to the bottom (it only goes on left, right, top normally), but still doesn't look great. Would it be reasonable to *switch* the order so that the "active" one was always on the bottom row, or would that be annoying?

Any other suggestions?

semantic/src/site/globals/site.overrides
CarlosNZ: Screenshot 2023-11-07 at 12 05 40 PM

I reckon we can lose the "STAGE" label, cos it's clear from the coloured bubble that it's referring to stage, which will give more room for the other stuff.

And maybe, let it wrap so the "Stage" bubble can go to next line? (It'd be good if the text/arrow inside the drop-down can stay on a single line)

src/containers/Review/assignment/ReviewSectionRow.tsx
CarlosNZ: Screenshot 2023-11-07 at 12 12 29 PM
Yeah that's great, thanks. Only thing I'd suggest is for the rows within each block to be a bit closer together.

src/containers/Review/overview/Overview.tsx
```diff
@@ -29,12 +30,13 @@ export const Overview: React.FC<{
...
+  const { isMobile } = useViewport()
 
   return (
     <div id="overview">
       <Segment basic>
         <Message info icon>
-          <Icon name="info circle" color="teal" />
+          {!isMobile && <Icon name="info circle" color="teal" />}
           <Message.Content>
             <div className="flex-row-center wrap">
```
CarlosNZ, line 41: Screenshot 2023-11-07 at 12 16 18 PM

Just need to get the "Extend deadline" UI to wrap on mobile. (Hit me up in Telegram if you can't find an application with this set -- or just change the condition to `true` temporarily to make it show up.)

CarlosNZ (changes requested):
Cool, looking much better. Good call to use the Grid properly to stack nicely.

See comments for some small adjustments.

Also, the Notes and Documents tabs could do with a little tidy up -- you want to make a new issue for them?
 
Screenshot 2023-11-07 at 12 21 12 PM
- Left justify sort/files UI
- Input box full width

Screenshot 2023-11-07 at 12 21 54 PM- 
- icon a bit smaller
- reduce spacing

src/containers/Review/ReviewWrapper.tsx
```diff
@@ -105,7 +107,7 @@ const ReviewWrapper: React.FC<ReviewWrapperProps> = ({ structure }) => {
             orgName={org?.name as string}
           />
           <div id="review-home-content">
-            <ReviewProgress structure={structure} />
+            {!isMobile && <ReviewProgress structure={structure} />}
```
CarlosNZ, line 110: Oh one more thing. The progress bar actually looks okay on mobile up to 3 items. Maybe we should keep it unless there are more that two stages (which makes 3 dots in the bar)?

Screenshot 2023-11-07 at 12 28 33 PM

(But just reduce the top/bottom space around it)

CarlosNZ: Actually, don't worry about it. It's still a bit cut off. More trouble than it's worth I think.

CarlosNZ:
Oh, couple of other things I just noticed after posting my review:

There is a min-width of 300px on the `#overview .left-item` class that is causing problems when we get narrow:
Screenshot 2023-11-07 at 12 37 09 PM
It's fine if you turn this off for mobile:

Screenshot 2023-11-07 at 12 37 15 PM


And can we reduce the padding around the Activity region to use more of the screen width (maybe drop the indents down a little too, if it helps?)

Screenshot 2023-11-07 at 12 37 28 PM

I think we can also revert the font-size (0.57 rem) on the Stage bubbles back to what it was, as with the other tweaks I suggested it's probably not necessary.

semantic/src/site/globals/site.overrides
guatemartin: Text Label gone.  Left as much spacing btn dropdown and Stage Label as I could.  It's not much really.
assignment_tab

semantic/src/site/globals/site.overrides
```diff
@@ -2780,16 +2779,47 @@ h6:first-child {
...
+    }
+    .ui.list > .item {
+      margin-top: 10px;
+      padding: 10px;
     }
   }
 
   #notes-tab .options-row {
     flex-direction: column;
+    align-items: flex-start;
```
CarlosNZ, line 2812: Can you just add some left padding to make this row flush with the stuff beneath it?
Screenshot 2023-11-07 at 4 59 25 PM

src/containers/Review/assignment/ReviewSectionRow.tsx
guatemartin: Reduced spacing... we'll see if it's enough, or needs more...

semantic/src/site/collections/message.overrides
```diff
@@ -38,3 +38,9 @@
 .ui.error.message h6 {
   color: @errorHeaderColor;
 }
+
+@media only screen and (max-width: @largestMobileScreen) {
+  .ui.message {
+    padding: 0.5em 0.5em;
```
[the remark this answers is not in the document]
guatemartin: With the removal of other properties that were limiting the space... this is now 1em, and it isn't causing issues.

src/containers/Review/overview/Overview.tsx
guatemartin: Ohh, forgot this one...  Will get it

semantic/src/site/globals/site.overrides
```diff
@@ -2780,16 +2779,47 @@ h6:first-child {
...
   }
 
   #assignment-tab {
     .assigning-row > .column {
+      padding: 0.1rem 0.5rem !important;
+    }
+    .stage-label {
+      margin-left: 3px;
+    }
+    .custom-assignment-grid-column {
```
CarlosNZ, line 2822: Screenshot 2023-11-07 at 5 11 03 PM

I think we need a bit more tidying up of this layout. But lets leave that to another issue.

guatemartin:
> Oh, couple of other things I just noticed after posting my review:
> 

I hadn't seen this one!  Thanks.  It was causing me grief and I couldn't find it!!

> There is a min-width of 300px on the `#overview .left-item` class that is causing problems when we get narrow: Screenshot 2023-11-07 at 12 37 09 PM It's fine if you turn this off for mobile:
> 
> Screenshot 2023-11-07 at 12 37 15 PM
> 

It was actually the margins that were huge.  Have reduced for mobile.

> And can we reduce the padding around the Activity region to use more of the screen width (maybe drop the indents down a little too, if it helps?)
> 
> Screenshot 2023-11-07 at 12 37 28 PM
>

There was a little more space, but I still left it at 0.7rem, otherwise it uses up a lot of space.... I left it at .7 thinking it looked good, but if not, 1rem it is.
 
> I think we can also revert the font-size (0.57 rem) on the Stage bubbles back to what it was, as with the other tweaks I suggested it's probably not necessary.

src/containers/Review/ReviewWrapper.tsx
```diff
@@ -107,7 +107,11 @@ const ReviewWrapper: React.FC<ReviewWrapperProps> = ({ structure }) => {
             orgName={org?.name as string}
           />
           <div id="review-home-content">
-            {!isMobile && <ReviewProgress structure={structure} />}
+            {isMobile && structure.stages.length <= 2 ? (
+              <ReviewProgress structure={structure} />
+            ) : (
+              !isMobile && <ReviewProgress structure={structure} />
+            )}
```
CarlosNZ, line 114: Sorry, I think I put you wrong. Let's just leave it how it was (hide on mobile) for now, and we can look at this a bit later. I think it needs a bit more before consideration it can work on mobile.

(I'm also thinking that it's generally going to be most useful when there are a lot of stages, so it's less useful for only 1-2 stage applications anyway)

src/containers/Review/ReviewWrapper.tsx
guatemartin: Oh, I actually did this.  If more than 2 stages, will not show in mobile.

CarlosNZ (approved):
Cool, working pretty well thanks. There's just a couple of tiny things if you could quickly fix before merging. And I'll make a new issue later on for a few additional tweaks once we've got used to this.

guatemartin:
@CarlosNZ lots of changes here

I just haven't done the "extend date" in overview, as I can't find one that has those properties.

In Overview 
- I removed the left coloured border as without the icon it makes no sense
- reduced the margin of the activities timeline
- removed '.left-area' class that had min-width 300px.  Not needed

Assignment Tab
- Removed the Text Label for the Stage
- Reduced the padding and spacing
- Increased Stage Label size
assignment_tab

Notes Tab
- Left aligned dropdown and toggle button
- Textarea now 'stretches' to use up space
- Reduced spacing
notes_tab_textarea

Documents
- Smaller icon (60 x 60px) could go smaller
- Reduced padding and gap between files
image

PR #1601, opened 2023-11-06
#1587 App intro page and App header fix

### LINKED ISSUE #1587: Mobile -- Application intro page layout
Part of Epic #1548

Image

Could do with a bit of tidying up. Reducing the padding on the outside should mostly do it, then see what else it needs

CarlosNZ:
Fix #1587 

Plus, I fixed up the header at the top of the application page. Turns out we'd been doing something wrong the whole time -- the "organisation" that appears at the top should be the company that made the application we're looking at, not the current logged in company (that's already in the Header bar!), so have fixed that and tidied it up a little.

WrongRight

### DIFF
--- a/semantic/src/site/globals/site.overrides
+++ b/semantic/src/site/globals/site.overrides
@@ -472,6 +472,7 @@ a:hover {
 #content-area .top-container {
   margin-top: 20px;
   max-width: @applicationFormWidth;
+  width: 100%;
 }
 
 // Back to application list label
--- a/semantic/src/site/globals/site.overrides
+++ b/semantic/src/site/globals/site.overrides
@@ -460,6 +460,10 @@ a:hover {
    Application/Review Wrappers
 ------------------------------------*/
 
+#application-area {
+  .flex-column-center();
+}
+
 // Moves back up if UserArea is not displayed
 #application-area.non-registered {
   margin-top: -@headerHeight;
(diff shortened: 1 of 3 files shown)


semantic/src/site/globals/site.overrides
```diff
@@ -461,15 +461,19 @@ a:hover {
...
+
 // Moves back up if UserArea is not displayed
 #application-area.non-registered {
   margin-top: -@headerHeight;
 }
 
 #content-area .top-container {
   margin-top: 20px;
-  flex-wrap: wrap;
-  justify-content: space-between;
```
CarlosNZ, line 472: Flex properties weren't doing anything, as it's not a flex container 🙄

semantic/src/site/globals/site.overrides
```diff
@@ -478,14 +482,7 @@ a:hover {
...
   font-size: @fontSmall;
   font-weight: @regularWeight;
-  padding-left: 10%;
-}
-
-// Fix for company header on mobile to not clash with 'back to applications'
-@media only screen and (min-width: @tabletBreakpoint) {
-  #content-area .heading-alt {
-    margin-top: -25px;
-  }
```
CarlosNZ, line 488: Don't seem to need any more

semantic/src/site/globals/site.overrides
```diff
@@ -498,6 +495,9 @@ a:hover {
   background-color: @surfacesWhite;
   width: 800px;
   padding: 30px 50px 32px;
+  @media only screen and (max-width: @largestMobileScreen) {
+    padding: 20px;
+  }
```
CarlosNZ, line 500: This is all that's need to fix #1587

guatemartin (approved):
All makes sense.

PR #1602, opened 2023-11-07
#1599 Dashboard width fix

### LINKED ISSUE #1599: Mobile Dashboard container size
When first logging in the Dashboard appears a bit too "zoomed in" and you have to pinch to bring it back to correct viewport. Suspect something to do with the Dashboard cell wrapping

CarlosNZ:
Fix #1599 

This is a pretty tiny fix. I won't know for certain that it addresses the exact problem I found until I look at it in a build, but it probably does.


guatemartin (approved):
Done.

PR #1606, opened 2023-11-07
#1605 Mobile application list tweak and linkable logo

### LINKED ISSUE #1605: Mobile -- Action button in list table and Dashboard link on logo
Image
Image

CarlosNZ:
Fix #1605

Getting the "Action" button to appear in the corner was actually harder than I expected. See comments for more detail.

It would be even harder to make it a button -- I tried wrapping in a button, but, because it would contain an existing component, then:
- the styling doesn't work correctly on hover
- we'd have to define a different icon for each type of action, as a generic one wouldn't make sense, and that's adding even more complexity
- it actually doesn't look that great (IMHO)

### DIFF
--- a/semantic/src/site/collections/table.overrides
+++ b/semantic/src/site/collections/table.overrides
@@ -190,6 +190,10 @@
     }
   }
 
+  #applications-list .ui.table tr td:last-child {
+    padding-bottom: 10px !important;
+  }
+
   // For list-builder tables
   .form-element {
     .ui.table {
--- a/semantic/src/site/collections/table.overrides
+++ b/semantic/src/site/collections/table.overrides
@@ -190,8 +190,15 @@
     }
   }
 
-  #applications-list .ui.table tr td:last-child {
-    padding-bottom: 10px !important;
+  #applications-list {
+    .ui.table tr td:nth-last-child(2).has-mobile-action-cell {
+      border-bottom-left-radius: 10px;
+      border-bottom-right-radius: 10px;
+      padding-bottom: 10px !important;
+    }
+    .ui.table tr td:last-child:not(.has-mobile-action-cell) {
+      padding-bottom: 10px !important;
+    }
   }
 
   // For list-builder tables
--- a/src/components/List/ApplicationsList.tsx
+++ b/src/components/List/ApplicationsList.tsx
@@ -6,7 +6,9 @@ import { ApplicationListRow, ColumnDetails, SortQuery } from '../../utils/types'
 import Loading from '../Loading'
 import { TableCellMobileLabelWrapper } from '../../utils/tables/TableCellMobileLabelWrapper'
 import { useViewport } from '../../contexts/ViewportState'
-import { ReviewerActionCell } from './Cells'
+
+// Which column in the column array is the "Action" cell
+const ACTION_CELL_INDEX = 6
 
 interface ApplicationsListProps {
   columns: ColumnDetails[]
@@ -96,6 +98,8 @@ const ApplicationRow: React.FC<ApplicationRowProps> = ({ refetch, columns, appli
 
   if (error) return <Message error header={t('ERROR_APPLICATION_DELETE')} list={[error]} />
 
+  const MobileActionCell = getMobileActionCell(isMobile, columns[ACTION_CELL_INDEX], application)
+
   return (
     <Table.Row
       key={`ApplicationList-application-${application.id}`}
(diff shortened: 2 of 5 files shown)


semantic/src/site/collections/table.overrides
```diff
@@ -190,6 +190,17 @@
...
+  #applications-list {
+    .ui.table tr td:nth-last-child(2).has-mobile-action-cell {
+      border-bottom-left-radius: 10px;
+      border-bottom-right-radius: 10px;
+      padding-bottom: 10px !important;
+    }
+    .ui.table tr td:last-child:not(.has-mobile-action-cell) {
+      padding-bottom: 10px !important;
+    }
+  }
```
CarlosNZ, line 202: So the main problem was that when the "Action" cell was removed from the normal flow (using `position: absolute`), it is still considered the "last child" of the container. So then the rounded bottom would be missing from the the cell. So I had to do some messy business to determine if there is an action cell or not, and style either the last child or the 2nd-to-last accordingly.

src/components/List/ApplicationsList.tsx
```diff
@@ -1,12 +1,15 @@
...
 import { useLanguageProvider } from '../../contexts/Localisation'
-import { useDeleteApplicationMutation } from '../../utils/generated/graphql'
+import { ApplicationStatus, useDeleteApplicationMutation } from '../../utils/generated/graphql'
 import { ApplicationListRow, ColumnDetails, SortQuery } from '../../utils/types'
 import Loading from '../Loading'
 import { TableCellMobileLabelWrapper } from '../../utils/tables/TableCellMobileLabelWrapper'
 import { useViewport } from '../../contexts/ViewportState'
 
+// Which column in the column array is the "Action" cell
+const ACTION_CELL_INDEX = 6
```
CarlosNZ, line 11: This works because we know the structure of the Applications table. It would be a lot more complicated for a generic table (like the Data Views).

src/components/List/ApplicationsList.tsx
```diff
@@ -95,13 +98,34 @@ const ApplicationRow: React.FC<ApplicationRowProps> = ({ refetch, columns, appli
...
+        if (index === ACTION_CELL_INDEX && MobileActionCell)
+          return (
+            <Table.Cell
+              key={`ApplicationList-row-${application.id}-action`}
+              className="mobile-action-cell"
+              style={{ position: 'absolute', top: '1em', right: '1em' }}
+            >
+              <MobileActionCell {...props} />
+            </Table.Cell>
+          )
```
CarlosNZ, line 119: For the reason mentioned above, we have to do some additional logic to even determine if there's going to be an Action cell for this row. And if there is, we render it last (index == 6) as a special case (`MobileActionCell`), but if not, we just render the cells as before -- wrapping in `TableCellMobileLabelWrapper` and hiding/formatting based on definitions in `mapColumns...`

src/containers/User/UserArea.tsx
```diff
@@ -121,7 +120,6 @@ const MainMenuBar: React.FC<MainMenuBarProps> = ({
   const { push, pathname } = useRouter()
   const {
     userState: { currentUser },
-    logout,
```
CarlosNZ, line 124: Removed a couple of unused values

src/containers/User/UserArea.tsx
```diff
@@ -428,10 +428,16 @@ const OrgSelector: React.FC<{ user: User; orgs: OrganisationSimple[]; onLogin: F
       ) as any,
       value: LOGIN_AS_NO_ORG,
     })
+
+  const logoIsLink = viewport.width < 860
```
CarlosNZ, line 432: This is using the same arbitrary "860px" as yesterdays header tweaks. Ideally we'd put this value in a constant somewhere, but not sure how to make a constant that can be available to both the LESS file and React components.

src/utils/helpers/list/mapColumnsByRole.ts
```diff
@@ -106,7 +106,7 @@ export const useMapColumnsByRole = () => {
       sortName: '',
       ColumnComponent: ReviewerActionCell,
       hideMobileLabel: true,
-      hideOnMobileTest: (application) => !application.reviewerAction,
+      hideOnMobileTest: (_) => true,
```
CarlosNZ, line 109: Always hide the action cell in the normal flow now, as we handle it as a special case (as above)

src/containers/User/UserArea.tsx
guatemartin: Not possible I believe.

guatemartin (approved):
Well done.

PR #1610, opened 2023-11-10
#1607 Auto logout functionality

### LINKED ISSUE #1607: Auto-logout after time out
We've already [created an issue](https://github.com/msupply-foundation/conforma-web-app/issues/1200) to do a more comprehensive version of this, but for now we need a more basic solution.

So a first instance, we just need something that measures "inactive" time, and if it exceeds X minutes (defined in prefs, default 60), it runs `logout()`.

Suggest using this resource as a basis for how it should work: https://medium.com/tinyso/how-to-detect-inactive-user-to-auto-logout-by-using-idle-timeout-in-javascript-react-angular-and-b6279663acf2
(in particular, the way to handle multiple tabs is important).

If we have time, we can add a mechanism to also have a time limit in the JWT (`iat` field) and require it be refreshed from the server (run `getUserInfo` method) as well.

CarlosNZ:
Fix #1607 

This was quite complicated once I realised that auto-logout needs to consider the activity of multiple Conforma tabs -- you really don't want to be logged out while you're working if you happen to have another inactive tab open!

I used [this](https://medium.com/tinyso/how-to-detect-inactive-user-to-auto-logout-by-using-idle-timeout-in-javascript-react-angular-and-b6279663acf2) as the basic idea, but then modified it to our requirements/resources.

I tried several different approaches and this was the most efficient way of doing it I could find that covered every possible case (I hope). Could change it if a simpler way exists.

The number of diff files here is quite large, but that's mainly because I made a small change to the Toast provider that meant I had to refactor the syntax everywhere it's used. So you can mostly ignore that.

So there's three main areas here:

1. The auto-logout timer: all the internal logic is in `LoginInactivityTimer.ts`, with a long comment explaining how it works. The implementation is in `UserState.ts`, where the timer is started and stopped upon login/logout.
2. Mechanism to refresh the JWT token -- it runs on a period one minute less than the inactivity logout time (i.e. 59 mins by default) All the implementation of this is in `UserState.ts`, and you'll want https://github.com/msupply-foundation/conforma-server/pull/1076 in the back-end to test it.
3. Miscellaneous small tweaks/fixes -- see inline comments for details

To test the timer, you can enable debug logging by setting `const DEBUG_LOGGING = true` in `LoginInactivityTimer`.  Please try and find edge cases that I might have missed. I'm interested to see how it behaves when devices go to sleep, but I'll wait to test in a build version for that.

Here is the "how it works" copied from the code comment:

```js
/**
 * A module to handle auto-logout after a certain amount of user idle time. The
 * complexity is due to the fact that we need this to work across multiple tabs
 * -- if a user has another Conforma tab open in the background, we don't want
 * that tab to log out if there's another tab still active, as that will clear
 * the auth credentials of the active tab too.
 *
 * To achieve this, we used a shared "expiry" value in localStorage. This is
 * loosely based on an idea presented in https://medium.com/tinyso/b6279663acf2
 *
 * The basic operation is:
 * - The "idleTracker" detects whenever a user goes idle for more than 3
 *   seconds, or returns from idle.
 * - On going idle, we set the expiry time in localStorage (based on input
 *   "idleTimeout") and start a timer that will log out at that time.
 * - If the user becomes active again, the expiry time is deleted from
 *   localStorage and the logout timer is cancelled.
 * - When the logout timer elapses, we read the expiry time from local storage:
 *   - If it's now later than the current time, that means another tab has been
 *     active in the meantime, so we restart our timer to logout at this new
 *     expiry time.
 *   - If the expiry time is no longer present, that means that another tab is
 *     *currently* active, so we restart our timer to logout after our
 *     "idleTimeout" (this cycle will repeat until an expiryTime is detected, or
 *     the current tab becomes active).
 *   - If the expiry time is less than/equal to the current time, we know no
 *     other tabs have been active, so we can log out.
 *
 * There is one more case to consider: if a user is active in one tab and closes
 * that tab before "idle" begins, there won't be any expiry time written to
 * local storage. And then when another (inactive) tab reaches the end of its
 * logout time, it will think another tab is still active and remain in that
 * cycle indefinitely (i.e. never log out. To handle this case, we add an event
 * listener to the window to capture tab/window closing and write the expiry
 * time to local storage on this event.
 */
```


package.json
```diff
@@ -28,6 +28,7 @@
     "deep-equal": "^2.0.5",
     "dotenv": "^16.0.1",
     "graphql": "^15.3.0",
+    "idle-tracker": "^0.1.3",
```
CarlosNZ, line 31: The most light-weight one of these that did what I needed and had event throttling.

src/components/Admin/AdminDataViews/useAdminDataViewConfig.tsx
```diff
@@ -19,7 +19,7 @@ export const useAdminDataViewConfig = (tableName: string) => {
 
   const { query, updateQuery } = useRouter()
 
-  const showToast = useToast({ position: Position.topLeft })
+  const { showToast } = useToast({ position: Position.topLeft })
```
CarlosNZ, line 22: Ignore all these -- just a syntax change

src/components/Dashboard.tsx
```diff
@@ -112,9 +112,9 @@ const TemplateComponent: React.FC<{
       <div className="content">
         <div className="filters">
           <Label className="strong-label clickable">
-            <a href={`/applications?type=${code}&user-role=${userRole}`}>
+            <Link to={`/applications?type=${code}&user-role=${userRole}`}>
```
CarlosNZ, line 115: This was a mistake. Bare `a` links don't work as they bypass React Router and force a full page reload.

src/containers/Dev/Snapshots.tsx
```diff
@@ -244,7 +244,7 @@ const Snapshots: React.FC = () => {
                 showToast({
                   title: 'Download complete',
                   text: `${displayType === 'archives' ? 'ARCHIVE_' : ''}${filename}`,
-                  timeout: 30000,
+                  timeout: 0,
```
CarlosNZ, line 247: Makes it stay permanently until clicked. I thought this was already set on this one.

src/containers/TemplateBuilder/template/Actions/ActionConfig.tsx
```diff
@@ -129,7 +129,7 @@ const ActionConfig: React.FC<ActionConfigProps> = ({ templateAction, onClose })
           style={{ borderTopRightRadius: 8 }}
         >
           <a
-            href="https://github.com/openmsupply/conforma-server/wiki/List-of-Action-plugins"
+            href="https://github.com/msupply-foundation/conforma-server/wiki/List-of-Action-plugins"
```
CarlosNZ, line 132: This was changed company-wide for our organisation. It does forward correctly, but I'm updating them as I find them.

src/containers/User/Login.tsx
```diff
@@ -31,6 +32,8 @@ const Login: React.FC = () => {
   const { t, languageOptions } = useLanguageProvider()
   const { preferences } = usePrefs()
 
+  usePageTitle(t('LABEL_LOG_IN'))
```
CarlosNZ, line 35: Missing page title on this page!

src/contexts/LoginInactivityTimer.ts
```diff
@@ -0,0 +1,143 @@
...
+ * to local storage. And then when another (inactive) tab reaches the end of its
+ * logout time, it will think another tab is still active and remain in that
+ * cycle indefinitely (i.e. never log out. To handle this case, we add an event
+ * listener to the window to capture tab/window closing and write the expiry
+ * time to local storage on this event.
+ */
+
+import IdleTracker from 'idle-tracker'
+
+const DEBUG_LOGGING = false
```
CarlosNZ, line 41: Turn this on to see what's happening in the console.

src/contexts/SystemPrefs.tsx
```diff
@@ -28,6 +29,7 @@ interface PrefsState {
 
 const defaultPrefs: Preferences = {
   showDocumentModal: false,
+  logoutAfterInactivity: 60,
```
CarlosNZ, line 32: Adjust this value in preferences (server object) to test short timeouts.

src/contexts/Toast/ToastDemo.tsx
```diff
@@ -91,6 +94,12 @@ export const ToastDemo: React.FC<any> = () => {
               showToast(toastData)
             }}
           />
+          <Button
+            secondary={toasts.length > 0}
+            content="Clear all toasts"
+            disabled={toasts.length === 0}
+            onClick={clearAllToasts}
+          />
```
CarlosNZ, line 102: Not particularly relevant, but you can play with the Toast notifications by going to `/dev/toast-demo`

src/contexts/Toast/ToastProvider.tsx
```diff
@@ -10,11 +10,17 @@ const setTimeout = window.setTimeout // To ensure return type is number
...
+  toasts: ToastProps[]
 }
 
+type ToastState = ToastProps & { close: () => void }
+
 const ToastProviderContext = createContext<ToastProviderValue>({
   showToast: () => {},
   updateDefaults: () => {},
+  clearAllToasts: () => {},
+  toasts: [],
```
CarlosNZ, line 23: So the reason why I poked around with the Toasts was so I could add a "clear all" method -- this gets invoked on login, so the residual "You have been logged out" notification gets cleared automatically.

src/contexts/Toast/ToastProvider.tsx
```diff
@@ -28,25 +34,32 @@ export function ToastProvider({ children }: { children: React.ReactNode }) {
...
+      ...prevState,
+      { ...toastDefaults, ...newToast, uid: nanoid(8), close: () => {} },
+    ])
   }
 
   const removeToast = (uid: string) => {
     setToasts((prevState) => prevState.filter((toast) => toast.uid !== uid))
   }
 
+  const clearAllToasts = () => toasts.forEach((toast) => toast.close())
```
CarlosNZ, line 54: I could just go `setToasts([])`, but this way we get the nice fade-out on each one as they close :)

src/contexts/UserState.tsx
```diff
@@ -74,38 +81,85 @@ const initialUserContext: {
...
   const client = useApolloClient()
+  const { push } = useRouter()
+  const { preferences } = usePrefs()
+  const { showToast, clearAllToasts } = useToast()
+
+  let refreshTokenTimer = useRef(0)
+
+  const loginTimer = useMemo(
+    () =>
+      new LoginInactivityTimer({
```
CarlosNZ, line 109: Instantiates the timer object, but doesn't start it yet.

The reason I used `useMemo` is to ensure only one instance is created. Without this I was getting up to 4 running simultaneously!

src/contexts/UserState.tsx
```diff
@@ -74,38 +81,85 @@ const initialUserContext: {
...
+    refreshTokenTimer.current = 0
+    clearLocalStorageExcept(['language', LOCAL_STORAGE_EXPIRY_KEY])
     client.clearStore()
-    window.location.href = '/login'
+    setUserState({ type: 'resetCurrentUser' })
+    loginTimer.end()
+    push('/login')
+  }
+
+  const delayedLogout = () => {
```
CarlosNZ, line 135: This is what happens when the front-end tries to run "fetchUserInfo" with an invalid/expired token. We need to implement a more generic overall solution for what happens on various server responses/errors, but that's a bigger job for another day.

src/contexts/UserState.tsx
```diff
@@ -114,6 +168,18 @@ export function UserProvider({ children }: UserProviderProps) {
       })
       dispatch({ type: 'setLoading', isLoading: false })
     }
+
+    loginTimer.start()
```
CarlosNZ, line 172: This is where it gets started -- on login

src/contexts/UserState.tsx
```diff
@@ -74,38 +81,85 @@ const initialUserContext: {
...
+  const { t } = useLanguageProvider()
   const [state, dispatch] = useReducer(reducer, initialState)
   const userState = state
   const setUserState = dispatch
   const client = useApolloClient()
+  const { push } = useRouter()
+  const { preferences } = usePrefs()
+  const { showToast, clearAllToasts } = useToast()
+
+  let refreshTokenTimer = useRef(0)
```
CarlosNZ, line 105: I use `useRef` to get a stable, single mutable value that is unaffected by re-renders and state changes.

src/utils/helpers/fetchMethods.ts
```diff
@@ -27,8 +27,10 @@ export async function postRequest({
       },
       body,
     })
-    return response.json()
+    const responseJSON = await response.json()
+    return responseJSON
```
CarlosNZ, line 31: No actual change here. I wanted to inject a general handler here to notify and logout for any failed request, but it was getting too complicated for right now.

src/utils/hooks/useListTemplates.tsx
```diff
@@ -31,7 +31,7 @@ const useListTemplates = (templatePermissions: TemplatePermissions, isLoading: b
 
   const { data, error: apolloError } = useGetTemplatesQuery({
     skip: isLoading,
-    fetchPolicy: 'network-only',
+    // fetchPolicy: 'network-only',
```
CarlosNZ, line 34: So there was a bug where if you loaded the app in a new tab with valid auth credentials already in storage, then the Dashboard would load but wouldn't fetch any templates. I had a good poke around in here, and it's really weird because the `data` in this hook was showing `undefined`, yet I could clearly see the query in the network tab was returning correct results. 🤷‍♂️

Anyway, the problem goes away by turning this off, no idea why. I'm sure it's fine, there's no critical, updated data here that requires a non-cached response every time.

CarlosNZ (commented):
Sorry, bit of biggie this one.

src/utils/helpers/fetchMethods.ts
guatemartin: Did you want to leave the console logs here?

src/utils/hooks/useListTemplates.tsx
guatemartin: I couldn't replicate this error.  I tried 'no-cache' as well, which I don't think is what you want long term anyway, but I always got data....  So I guess we'll leave it.  It'll be fine....

guatemartin (approved):
I really only have one lame thing that I personally would prefer.  That is, that the Toast be front and center when it logs you out...  But that is just me.  I like 'big' notifications like that to be very very obvious.

I tested it in various ways.  One tab, 4 tabs, Separate windows.....    It seems solid.  It obviously does not close a pdf opened in a separate tab.  Even if that did matter, I don't think there is a way to do that.

What I didn't test though, was with a PDF in modal view....  You pulled this from 'develop' branch which doesn't have that yet it would seem.  It would be interesting to see what that does..

src/utils/helpers/fetchMethods.ts
CarlosNZ: Yeah, cos it's displaying an error for a state that shouldn't really occur. So if it does, it'd be good to see a console error. As I said elsewhere, I'd like to put some proper, generic error handling into these methods, but will make do for now.

CarlosNZ:
Hold the phone, I had a bit more of a think about this and came up with an even simpler process.

Here it is if you're interested: be22bc26

I'll merge it now and we'll see how it goes in testing.

CarlosNZ:
> I really only have one lame thing that I personally would prefer. That is, that the Toast be front and center when it logs you out... But that is just me. I like 'big' notifications like that to be very very obvious.

Yeah, I hear you. I've put it in the top middle now -- I'm not totally sure about it partially blocking the login form, but I think it'll be okay. We'll see how it goes in testing.

Screenshot 2023-11-10 at 10 59 08 PM

> I tested it in various ways. One tab, 4 tabs, Separate windows..... It seems solid. It obviously does not close a pdf opened in a separate tab. Even if that did matter, I don't think there is a way to do that.

Yeah, exactly, it's no longer "our app" :) 

> What I didn't test though, was with a PDF in modal view.... You pulled this from 'develop' branch which doesn't have that yet it would seem. It would be interesting to see what that does..

I actually merged the feature/angola branch into it as I was poking around in a few spots that I knew we'd changed, so didn't want to creat conflicts. Anyway, it's fine in the modal. :)

PR #1618, opened 2023-11-27
BE1077 separate query for count in application list

andreievg:
Part of: https://github.com/msupply-foundation/conforma-server/issues/1077

Remove pageInfo and totalCount from applicationList query, and do a separate query for applicationList count (improves applicationList loading by x3 (ish)

Don't forget to `yarn generate` before running


src/containers/List/ListFilters/ListFilters.tsx
```diff
@@ -271,9 +271,9 @@ const ListFilters: React.FC<{
             return null
         }
       })}
-      {totalCount ? (
+      {totalCount !== 'loading' && (
```
andreievg, line 274: This does mean that total count jumps into view when list is first loaded, what do you think @CarlosNZ ?

src/utils/hooks/useListApplications.tsx
```diff
@@ -56,8 +57,16 @@ const useListApplications = (
     fetchPolicy: 'network-only',
   })
 
+  const [getListCount, { loading: loadingCount, data: countData }] =
```
andreievg, line 60: Wanted to re-fetch it in sequence, after applicationList was loaded

src/utils/hooks/useListApplications.tsx
```diff
@@ -56,8 +57,16 @@ const useListApplications = (
...
+    useGetFilteredApplicationCountLazyQuery({
+      fetchPolicy: 'network-only',
+    })
+
   // Ensures that query doesn't request a page beyond the available total
   useEffect(() => {
+    if (applicationCount == 'loading') {
+      return
```
andreievg, line 68: I wonder if pagination is now broken and query will request beyond pagination ? (if count is only figured out once applicationList loads ?). I wonder how to replicate this use case ?

CarlosNZ:
I don't fully understand why we need to fetch the totalCount in a separate query, but you can explain to me in person. I'd like to find a solution that doesn't require this if possible, because it does seem strange that the count/pagination doesn't get loaded until after the main query.

FYI -- I updated this branch from `develop` with some recent merges in it and had to resolve a couple of minor conflicts, so you probably want to re-pull this. (and re-run `yarn` as there's a few package changes in recent work)

CarlosNZ:
Cool, so summary of what we discussed:

- Just hide the Pagination component completely until count is available
- Don't do this change on the data views, as they work fine with the totalCount included
- When just changing a "sort" value, don't refetch the totalCount

andreievg:
@CarlosNZ in the last commit I've tried to address points we've discussed.
* Reverted 'loading' state (data lists should be the same)
* Hiding pagination until know (also hiding total count while fetching)
* Navigate to first page if not on first page when pagination count is 0
* Reduce flicker of pagination and total count in filter list (this is testable when changing 'page' or when navigating to say last page in unfiltered list, then filtering the by reduce set)

It looked like a lot of work to differentiate between sort and filter (and only update total count when that happens, so didn't want to make it more complex)

CarlosNZ (approved):
Nice -- looks good, Thanks heaps :)

PR #1624, opened 2023-12-13
#1623 Input validation for Search element

### LINKED ISSUE #1623: Input validation for Search element
For the "Search" element, we need to have some validation of the user input and show error when user text doesn't comply.

- Specify a Regex for the pattern to comply with
- Have a way to auto-fix the text case

CarlosNZ:
Fix #1623 

4 new parameters (all optional) for the Search element:

- `inputPattern`
- `inputExample`
- `inputErrorMessage`
- `restrictCase`

See updated docs for specifics.

Also, a bit of improvement to existing functionality.

### DIFF
--- a/src/formElementPlugins/search/src/ApplicationView.tsx
+++ b/src/formElementPlugins/search/src/ApplicationView.tsx
@@ -412,6 +412,9 @@ const getDefaultString = (
   }
 }
 
+// In order to check that user input is consistent with regex pattern, even when
+// it's only partially completed (and therefore won't match the regex), we
+// combine it with a known correct example and then test it against the pattern.
 const partialMatch = (text: string, example: string, pattern?: RegExp) => {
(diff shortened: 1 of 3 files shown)


src/formElementPlugins/search/src/ApplicationView.tsx
```diff
@@ -219,10 +269,10 @@ const ApplicationView: React.FC<ApplicationViewProps> = ({
           disabled={!isEditable}
           input={{ icon, iconPosition: 'left' }}
           noResultsMessage={t('MESSAGE_NO_RESULTS')}
+          onBlur={() => setIsFocused(false)}
+          open={isFocused && !inputError && searchText.length >= minCharacters && !loading}
```
CarlosNZ, line 273: Works a bit better if we manage the open state ourselves.

PR #1636, opened 2024-05-27
#1103B Memoise list actions etc

CarlosNZ:
Front-end changes for back-end https://github.com/msupply-foundation/conforma-server/pull/1106

Main improvement was to get the cache working using [`cache-and-network` policy](https://www.apollographql.com/docs/react/data/queries/#setting-a-fetch-policy). The reason we hadn't been able to get this working in the past is that we were checking for `loading` from Apollo hook and rendering a Loading component. However, when using this policy, Apollo returns `loading:true` while it's doing its background query, so we would be displaying the loader rather than the cached result. Solution is to only show the loader if `loading: true && !data`. `data` is populated immediately with the cached version if it exists, so we can display it immediately.

See review comments for further improvements.

### DIFF
--- a/src/containers/List/ListFilters/OptionFilters.tsx
+++ b/src/containers/List/ListFilters/OptionFilters.tsx
@@ -4,6 +4,7 @@ import { Dropdown, Input } from 'semantic-ui-react'
 import { EnumFilterProps, StaticListFilterProps, SearchableListFilterProps } from './types'
 import { FilterContainer, FilterOptions, FilterTitle } from './common'
 import { useLanguageProvider } from '../../../contexts/Localisation'
+import useDebounce from '../../../formElementPlugins/search/src/useDebounce'
 
 const EnumFilter: React.FC<EnumFilterProps> = ({
   getActiveOptions,
@@ -44,7 +45,7 @@ const StaticListFilter: React.FC<StaticListFilterProps> = ({
   const { query, resultExtractor, variables } = getFilterListQuery({ filterListParameters })
 
   const { data, error } = useQuery(query, {
-    fetchPolicy: 'network-only',
+    fetchPolicy: 'cache-first',
     variables,
   })
 
@@ -75,7 +76,7 @@ const SearchableListFilter: React.FC<SearchableListFilterProps> = ({
   onRemove,
 }) => {
   const { t } = useLanguageProvider()
-  const [searchValue, setSearchValue] = useState('')
+  const [searchValue, setSearchValue] = useDebounce('')
 
   const { query, resultExtractor, variables } = getFilterListQuery({
     searchValue,
@@ -83,7 +84,7 @@ const SearchableListFilter: React.FC<SearchableListFilterProps> = ({
   })
 
   const { data, error } = useQuery(query, {
-    fetchPolicy: 'network-only',
+    fetchPolicy: 'cache-first',
     variables,
   })
   const activeOptions = getActiveOptions()
--- a/src/formElementPlugins/SummaryViewWrapper.tsx
+++ b/src/formElementPlugins/SummaryViewWrapper.tsx
@@ -38,23 +38,27 @@ const SummaryViewWrapper: React.FC<SummaryViewWrapperProps> = ({
   )
 
   useEffect(() => {
-    // Update dynamic parameters when responses change
-    const JWT = localStorage.getItem(globalConfig.localStorageJWTKey)
-    Object.entries(parameterExpressions).forEach(([field, expression]) => {
-      evaluateExpression(expression as EvaluatorNode, {
-        objects: {
-          responses: { ...allResponses, thisResponse: response?.text },
-          currentUser,
-          applicationData,
-          functions,
-        },
-        APIfetch: fetch,
-        graphQLConnection: { fetch: fetch.bind(window), endpoint: graphQLEndpoint },
-        headers: { Authorization: 'Bearer ' + JWT },
-      }).then((result: any) =>
-        setEvaluatedParameters((prevState) => ({ ...prevState, [field]: result }))
-      )
-    })
+    // Update dynamic parameters when responses change, but only if there's no
+    // saved evaluations (usually INFORMATION type), or "showLiveParameters" has
+    // been explicitly set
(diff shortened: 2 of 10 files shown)


src/containers/List/ListFilters/OptionFilters.tsx
```diff
@@ -75,15 +76,15 @@ const SearchableListFilter: React.FC<SearchableListFilterProps> = ({
   onRemove,
 }) => {
   const { t } = useLanguageProvider()
-  const [searchValue, setSearchValue] = useState('')
+  const [searchValue, setSearchValue] = useDebounce('')
```
CarlosNZ, line 79: Text input debouncing for the filter lists reduces unnecessary queries.

src/containers/List/ListFilters/OptionFilters.tsx
```diff
@@ -75,15 +76,15 @@ const SearchableListFilter: React.FC<SearchableListFilterProps> = ({
...
 
   const { query, resultExtractor, variables } = getFilterListQuery({
     searchValue,
     filterListParameters,
   })
 
   const { data, error } = useQuery(query, {
-    fetchPolicy: 'network-only',
+    fetchPolicy: 'cache-first',
```
CarlosNZ, line 87: List filters use `cache-first` so they don't make a query at all (which would compete with the main list query). We may have to keep an eye on this. User/Reviewer lists don't change often, but we might need to explicitly invalidate this cache value at certain times.

src/formElementPlugins/SummaryViewWrapper.tsx
```diff
@@ -38,23 +38,27 @@ const SummaryViewWrapper: React.FC<SummaryViewWrapperProps> = ({
...
-        APIfetch: fetch,
-        graphQLConnection: { fetch: fetch.bind(window), endpoint: graphQLEndpoint },
-        headers: { Authorization: 'Bearer ' + JWT },
-      }).then((result: any) =>
-        setEvaluatedParameters((prevState) => ({ ...prevState, [field]: result }))
-      )
-    })
+    // Update dynamic parameters when responses change, but only if there's no
+    // saved evaluations (usually INFORMATION type), or "showLiveParameters" has
+    // been explicitly set
```
CarlosNZ, line 43: Prevents application evaluation queries from re-running during review/summary. Unfortunately, the "Information" type elements don't store a response, so we can used their previously evaluated values, so they still run in Summary view. (I might change this behaviour at some point)

src/index.tsx
```diff
@@ -57,27 +57,27 @@ const App: React.FC = () => {
...
+    persistCache({
+      cache,
+      storage: window.localStorage,
+      /**
+       * Storage options.
+       */
+      // Maximum size of cache to persist (in bytes).
+      // Defaults to 1048576 (1 MB). For unlimited cache size, provide false.
+      // If exceeded, persistence will pause and app will start up cold on next launch.
+      // maxSize: false,
```
CarlosNZ, line 69: Not really sure if we should set a maximum or not.

src/utils/graphql/queries/getApplicationsList.query.ts
```diff
@@ -35,6 +35,7 @@ export default gql`
         assigners
         reviewers
       }
+      totalCount
```
CarlosNZ, line 38: Only change to the actual list query is to include the totalCount

src/utils/hooks/useListApplications.tsx
```diff
@@ -18,12 +15,6 @@ const useListApplications = (
   graphQLFilterObject?: object
 ) => {
   const FILTER_DEFINITIONS = useGetFilterDefinitions()
-  const [applications, setApplications] = useState<ApplicationListShape[]>([])
```
CarlosNZ, line 21: We had a lot of unnecessary `useState` variables, so have removed quite a few and just returned the state directly from the Apollo hook.

src/utils/hooks/useListApplications.tsx
```diff
@@ -57,55 +44,30 @@ const useListApplications = (
...
+    fetchPolicy: 'cache-and-network',
   })
 
-  const [getListCount, { data: countData }] = useGetFilteredApplicationCountLazyQuery({
-    fetchPolicy: 'network-only',
-    onCompleted: () => setIsLoadingCount(false),
-  })
+  const applications = data?.applicationList?.nodes ?? []
 
-  useEffect(() => {
```
CarlosNZ, line 68: Removing unnecessary state values also means we can get rid of an entire `useEffect` update.

src/utils/hooks/useListTemplates.tsx
```diff
@@ -19,37 +19,25 @@ type TemplatesData = {
...
 const useListTemplates = (templatePermissions: TemplatePermissions, isLoading: boolean) => {
-  const [templatesData, setTemplatesData] = useState<TemplatesData>(emptyTemplateData)
-  const [error, setError] = useState('')
-  const [loading, setLoading] = useState(true)
-
-  const { data, error: apolloError } = useGetTemplatesQuery({
+  const { data, loading, error } = useGetTemplatesQuery({
     skip: isLoading,
-    // fetchPolicy: 'network-only',
+    fetchPolicy: 'cache-and-network',
```
CarlosNZ, line 30: Caching the templates list makes the menus load immediately on subsequent re-loads, which improves user experience a lot.

src/utils/hooks/useListTemplates.tsx
```diff
@@ -19,37 +19,25 @@ type TemplatesData = {
...
-      }
+  const templatesData = emptyTemplateData
 
-      const templates = filteredTemplates
+  if (data?.templates?.nodes) {
+    const filteredTemplates = (data?.templates?.nodes || []).filter(
+      (template) => templatePermissions[String(template?.code)]
+    ) as Template[]
+    if (filteredTemplates.length > 0) {
+      templatesData.templates = filteredTemplates
```
CarlosNZ, line 40: No logic change here, just moved out of unnecessary `useEffect`

CarlosNZ:
Oh, seems this was accidentally closed when I deleted the base branch :(

Have re-created here: https://github.com/msupply-foundation/conforma-web-app/pull/1638

PR #1637, opened 2024-06-10
#1104 review performance - batch mutation for approve all

CarlosNZ:
Uses a batch mutation when clicking "Approve All" which means only one http request is sent rather than one for each response.

The only downside is that it has to use the "updateReview" mutation in order to include a whole array of responses. The consequence is that the whole review section reloads in the UI after pressing "Approve All". Oddly, it only happens the first time, subsequent "Approve All"s within same review session only updates the inline responses.

Couple of other small tweaks too -- see comments.


src/components/PageElements/Elements/ReviewInlineInput.tsx
```diff
@@ -2,7 +2,7 @@ import React, { useState } from 'react'
 import { Button, Form, Radio, TextArea } from 'semantic-ui-react'
 import { useLanguageProvider } from '../../../contexts/Localisation'
 import { ReviewResponse, ReviewResponseDecision } from '../../../utils/generated/graphql'
-import useUpdateReviewResponse from '../../../utils/hooks/useUpdateReviewResponse'
+import useUpdateReviewResponses from '../../../utils/hooks/useUpdateReviewResponses'
```
CarlosNZ, line 5: Just changed name and access path here -- no functionality changes

src/components/Review/ReviewSectionRowAction.tsx
```diff
@@ -154,23 +157,28 @@ const GenerateActionButton: React.FC<ReviewSectionComponentProps> = ({
...
       return setError(true)
     }
   }
 
   if (error) return <Message error title={t('ERROR_GENERIC')} />
 
-  return (
+  return loading ? (
+    <LoadingSmall />
+  ) : (
```
CarlosNZ, line 181: Show a loading spinner when clicking "Start" review -- better than no visual update when there's a long-ish wait.

src/utils/hooks/useGetReviewStructureForSection/useGetReviewStructureForSectionAsync.tsx
```diff
@@ -24,7 +24,7 @@ const useGetReviewStructureForSectionAsync = (props: UseGetReviewStructureForSec
 
   const { data, error } = useGetReviewResponsesQuery({
     variables,
-    fetchPolicy: 'network-only',
+    fetchPolicy: 'cache-and-network',
```
CarlosNZ, line 27: Helps when multiple review response fetches close together.

src/utils/hooks/useUpdateReviewResponses.ts
```diff
@@ -2,31 +2,55 @@ import {
...
+            decision: response.decision,
+            recommendedApplicantVisibility: computeVisibility(response),
+          },
+        })),
+      },
+    }
+    updateMultipleResponsesMutation({ variables: { reviewId, reviewPatch } })
+  }
+
+  return { updateReviewResponse, updateMultipleReviewResponses }
```
CarlosNZ, line 53: Hook now returns two methods instead of just a single one for single response update.

andreievg:
@CarlosNZ, updateReviewResponse mutation returns:

https://github.com/msupply-foundation/conforma-web-app/blob/ceab2ba878a18eed10823d3a5943527d698e9304/src/utils/graphql/mutations/updateReviewResponse.mutation.ts#L20-L22

And the reviewMutation would return:

https://github.com/msupply-foundation/conforma-web-app/blob/ceab2ba878a18eed10823d3a5943527d698e9304/src/utils/graphql/mutations/updateReview.mutation.ts#L9-L15

Not sure what difference it would make overall, and how cache for reviewResponse is updated, probably safer to keep them both the same and create updateReviewResponseViaReview mutation. Btw I think the review refreshes the first time because review.id is returned from updateReview.

src/components/Review/ReviewSectionRowAction.tsx
andreievg: v

src/utils/hooks/useUpdateReviewResponses.ts
```diff
@@ -2,31 +2,55 @@ import {
...
-// also computes and updates recommendedApplicantVisibility
-type UseUpdateReviewResponse = () => (
-  reviewResponse: ReviewResponse,
-) => PromiseReturnType
+// Hook to update review responses. Either updates individual response or a
+// batch via updating review object.
+// Also computes recommendedApplicantVisibility
+type UseUpdateReviewResponse = () => {
+  updateReviewResponse: (reviewResponse: ReviewResponse) => PromiseReturnType
+  updateMultipleReviewResponses: any
```
andreievg, line 18: Hmm, tbh I am not a fan of 'any', unless it's absolutely necessary, in this case input is know and the return is Promise<void>.

andreievg (approved):
Looks good thanks, a couple of minor suggestion, mainly the 'any' and potentially just one method for updatingReviewResponses (through review)

src/utils/hooks/useUpdateReviewResponses.ts
CarlosNZ: Yeah, same. I often put it when I'm trying to test something, and then forget to update it. 🤷‍♂️

CarlosNZ:
> Not sure what difference it would make overall, and how cache for reviewResponse is updated, probably safer to keep them both the same and create updateReviewResponseViaReview mutation. Btw I think the review refreshes the first time because review.id is returned from updateReview.

Yeah, I think that's right, we don't want to risk messing with the existing updateReview process, so I'll create a new one called `updateReviewResponseViaReview` (or something like that) and use it for *both* these (then ditch the old one). 

And yes, if I don't return the reviewId in the return query the page doesn't refresh, so 👍

